PR #308, opened 2014-03-12
Added optional relaxed Liskov Substitution Principle enforcement.

ctabeckey:


vishnuvardhan623:
hi @hferentschik Do we have way to disable constraint validation check on method overrides when running over spring boot? We have controller contacts generated by openapi and for some of the controller implementations we didn't specify not null annotations on parameters for overridden methods and facing ConstraintDeclarationException HV000151(A method overriding another method must not redefine the parameter constraint configuration) . Please let me know if u feel this needs to be discussed on another thread

marko-bekhta:
Hey @vishnuvardhan623, 
changes from this pull request were integrated and you can set the properties to adjust the checks https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#section-method-validation-prerequisite-relaxation 
you can add them through a validation.xml or see how you can do it through Spring.
If you think that there's more to be added to the Validator, it's best to open a new ticket https://hibernate.atlassian.net/jira/software/c/projects/HV/issues

PR #360, opened 2014-12-23
HV-951: Make expression variables added via context.addExpressionVariable(String, Object) accessible from the ConstraintViolation object.

NicolaF:
This commit adds a new getExpressionVariables() getter to ConstraintViolationImpl and minor changes to the internals:
 * add an argument to ConstraintViolationImpl private constructor
 * overload ConstraintViolationImpl.forBeanValidation & co. to handle the new constructor argument, and mark previous, overloaded ones as @Deprecated
 * adjust ValidationContext.createConstraintViolation(ValueContext<?, ?>, ConstraintViolationCreationContext, ConstraintDescriptor<?>) to pass expressionVariables as un unmodifiable Map to the overloaded builders
 * unit test :p

Useful for:
 * Single page web applications, when all i18n/l10n/interpolation stuff is handled client side in JS.
 * Web services, where we want to expose constraint violations in a parsable way. WS client may present the error to users, depending on their prefered locale, and do the interpolation itself.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
@@ -6,28 +6,8 @@
  */
 package org.hibernate.validator.internal.engine;
 
-import java.lang.annotation.ElementType;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Method;
-import java.lang.reflect.Type;
-import java.util.HashSet;
-import java.util.IdentityHashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import javax.validation.ConstraintValidatorFactory;
-import javax.validation.ConstraintViolation;
-import javax.validation.MessageInterpolator;
-import javax.validation.ParameterNameProvider;
-import javax.validation.Path;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -292,10 +292,13 @@ public void addConstraintFailures(Set<ConstraintViolation<T>> failingConstraintV
 		);
 		// at this point we make a copy of the path to avoid side effects
 		Path path = PathImpl.createCopy( constraintViolationCreationContext.getPath() );
-
+		//same for expression variables
+		Map<String, Object> expressionVariables =
```
NicolaF: No, no problem here. It's not supposed to be modified afterwards, immutability is better IMO, especially when it's so easy to achieve. Plus, I'm not omniscent :p

NicolaF: License Agreement signed.

hferentschik: > License Agreement signed.

ok, thanks

hferentschik:
merged :-)

PR #361, opened 2014-12-24
5.1

ningzhaohang:


hferentschik:
Not quite sure what I am looking at here. It does not seem to address an issue. Also I am wondering whether this pull request is opened against the wrong branch. 

I am going to close it. If you think I missed something let me know what you actually try to achieve with this pull request.

PR #363, opened 2015-01-30
adds tests to CPF and CNPJ without mask

otaviojava:
In Brazilian documents, CPF and CNPJ, we may use these information without the mask.
So this test, should work.
So the CPF 134.241.313-00 is valid as 13424131300 is to.
The same thing happend with the CNPJ: 91.509.901/0001-69 and 91509.901000169
I thought in create the other annotations with represents CPF and CNPJ without mask

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/constraints/br/CNPJValidatorTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/constraints/br/CNPJValidatorTest.java
@@ -25,6 +25,11 @@ public class CNPJValidatorTest {
 				new Company( "91.509.901/0001-69" )
 		);
 		assertNumberOfViolations( violations, 0 );
+
+		violations = ValidatorUtil.getValidator().validate(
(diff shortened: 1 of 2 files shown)


hferentschik:
> In Brazilian documents, CPF and CNPJ, we may use these information without the mask.
So this test, should work.

I guess this is just a test case for the issue described here - https://hibernate.atlassian.net/browse/HV-933. As mentioned in the issue, currently `CPF` and `CNPJ` only work when they use the additional punctuation characters. I am not sure whether there is much we can do. From a Hibernate Validator point of view it does not make sense to change the behaviour of `ignoreNonDigitCharacters` again and adding a different set of annotations seems wrong as well. I think the best bet is to create your own custom constraints as described in the issue. (A potential solution is to go away from the pure composing constraint approach)

> I thought in create the other annotations with represents CPF and CNPJ without mask

For your project that's fine.

hferentschik:
I added a link to these tests to HV-933. Closing this for now.

PR #364, opened 2015-02-09
HV-913 Providing new profile for the "tck-runner" module for executing t...

gunnarmorling:
...he TCK with a security manager enabled

The TCK can be executed with a security manager enabled like this:

    mvn clean install -pl tck-runner/ -Dwith-security-manager

If we agree on the overall approach I'll send a PR to update the docs on the web site  as well.

### DIFF
--- a/tck-runner/pom.xml
+++ b/tck-runner/pom.xml
@@ -165,7 +165,8 @@
                 </property>
             </activation>
             <properties>
-                <arquillian.protocol>Local</arquillian.protocol>
+                <!-- The properties "arquillian.protocol" and "surefire.argLine" are set based on this setting via GMaven-->
+                <with-security-manager>true</with-security-manager>
             </properties>
             <dependencies>
                 <dependency>
@@ -176,6 +177,26 @@
             </dependencies>
             <build>
                 <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.gmaven</groupId>
+                        <artifactId>gmaven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>configure-properties</id>
+                                <phase>validate</phase>
+                                <goals>
+                                    <goal>execute</goal>
+                                </goals>
+                                <configuration>
+                                    <source><![CDATA[
+                                        project.properties['arquillian.protocol'] = Boolean.valueOf(pom.properties['with-security-manager']) ? 'LocalSecurityManagerTesting' : 'Local';
+                                        project.properties['surefire.argLine'] = Boolean.valueOf(pom.properties['with-security-manager']) ? '-Djava.security.manager -Djava.security.policy=${project.build.directory}/test-classes/test.policy -Djava.security.debug=access' : '';
+                                        ]]>
+                                    </source>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-surefire-plugin</artifactId>
@@ -190,6 +211,7 @@
                                     <value>true</value>
                                 </property>
                             </systemProperties>
+                            <argLine>${surefire.argLine}</argLine>
                         </configuration>
                     </plugin>
                 </plugins>
@@ -315,48 +337,6 @@
                             <systemPropertyVariables>
                                 <arquillian.launch>incontainer</arquillian.launch>
                             </systemPropertyVariables>
-                        </configuration>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
-        <profile>
-            <id>security-manager</id>
-            <activation>
-                <property>
-                    <name>with-security-manager</name>
-                    <value>true</value>
-                </property>
-            </activation>
-            <properties>
-                <arquillian.protocol>LocalSecurityManagerTesting</arquillian.protocol>
-            </properties>
-            <dependencies>
-                <dependency>
-                    <groupId>org.hibernate.beanvalidation.tck</groupId>
-                    <artifactId>beanvalidation-standalone-container-adapter</artifactId>
-                    <version>${tck.version}</version>
-                </dependency>
-            </dependencies>
-            <build>
-                <plugins>
-                    <plugin>
-                        <groupId>org.apache.maven.plugins</groupId>
-                        <artifactId>maven-surefire-plugin</artifactId>
-                        <configuration>
-                            <systemProperties>
-                                <property>
-                                    <name>validation.provider</name>
-                                    <value>${validation.provider}</value>
-                                </property>
-                                <property>
-                                    <name>excludeIntegrationTests</name>
-                                    <value>true</value>
-                                </property>
-                            </systemProperties>
-                            <argLine>-Djava.security.manager -Djava.security.policy=${project.build.directory}/test-classes/test.policy -Djava.security.debug=access</argLine>
-                            <!-- For debugging -->
-                            <!-- <argLine>-Djava.security.manager -Djava.security.policy=${project.build.directory}/test-classes/test.policy -Djava.security.debug=access -Xdebug -Xrunjdwp:server=y,transport=dt_socket,address=5005,suspend=n</argLine> -->
                         </configuration>
                     </plugin>
                 </plugins>
--- a/tck-runner/pom.xml
+++ b/tck-runner/pom.xml
@@ -22,7 +22,6 @@
     <properties>
         <tck.version>1.1.2.Final</tck.version>
         <tck.suite.file>${project.build.directory}/dependency/beanvalidation-tck-tests-suite.xml</tck.suite.file>
-        <arquillian.version>1.0.0.Final</arquillian.version>
         <wildflyTargetDir>${project.build.directory}/wildfly-${wildfly.version}</wildflyTargetDir>
         <validation.provider>org.hibernate.validator.HibernateValidator</validation.provider>
         <remote.debug />
@@ -66,13 +65,11 @@
         <dependency>
             <groupId>org.jboss.arquillian.testng</groupId>
             <artifactId>arquillian-testng-container</artifactId>
-            <version>${arquillian.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jboss.arquillian.protocol</groupId>
             <artifactId>arquillian-protocol-servlet</artifactId>
-            <version>${arquillian.version}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
@@ -318,6 +315,48 @@
                             <systemPropertyVariables>
                                 <arquillian.launch>incontainer</arquillian.launch>
                             </systemPropertyVariables>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
+            <id>security-manager</id>
+            <activation>
+                <property>
+                    <name>with-security-manager</name>
+                    <value>true</value>
+                </property>
+            </activation>
+            <properties>
+                <arquillian.protocol>LocalSecurityManagerTesting</arquillian.protocol>
+            </properties>
+            <dependencies>
+                <dependency>
+                    <groupId>org.hibernate.beanvalidation.tck</groupId>
+                    <artifactId>beanvalidation-standalone-container-adapter</artifactId>
+                    <version>${tck.version}</version>
+                </dependency>
+            </dependencies>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <systemProperties>
+                                <property>
+                                    <name>validation.provider</name>
+                                    <value>${validation.provider}</value>
+                                </property>
+                                <property>
+                                    <name>excludeIntegrationTests</name>
+                                    <value>true</value>
+                                </property>
+                            </systemProperties>
+                            <argLine>-Djava.security.manager -Djava.security.policy=${project.build.directory}/test-classes/test.policy -Djava.security.debug=access</argLine>
+                            <!-- For debugging -->
+                            <!-- <argLine>-Djava.security.manager -Djava.security.policy=${project.build.directory}/test-classes/test.policy -Djava.security.debug=access -Xdebug -Xrunjdwp:server=y,transport=dt_socket,address=5005,suspend=n</argLine> -->
                         </configuration>
                     </plugin>
                 </plugins>
--- a/tck-runner/src/test/java/org/hibernate/validator/tckrunner/securitymanager/TckRunner.java
+++ b/tck-runner/src/test/java/org/hibernate/validator/tckrunner/securitymanager/TckRunner.java
@@ -1,0 +1,73 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.tckrunner.securitymanager;
+
+import java.util.Collections;
+import java.util.List;
+
+import org.hibernate.beanvalidation.tck.util.IntegrationTestsMethodSelector;
+import org.testng.ITestResult;
+import org.testng.TestListenerAdapter;
+import org.testng.TestNG;
+import org.testng.xml.XmlMethodSelector;
+import org.testng.xml.XmlPackage;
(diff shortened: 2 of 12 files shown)


tck-runner/src/test/resources/test.policy
```diff
@@ -0,0 +1,165 @@
+/**
+ * Policy file for running the Bean Validation TCK within a security manager.
+ *
+ * The goal if this is to make sure that all SM-relevant APIs are invoked by Hibernate Validator using privileged
+ * actions so it can be used within an SM enabled, only giving the required permissions to Hibernate Validator but not
+ * other code which uses it.
+ *
+ * For that purpose this test set up puts one frame onto the call stack (via {@ code DelegatingExecutor}) whose
```
hferentschik: space between '@' and 'code'

tck-runner/src/test/resources/test.policy
```diff
@@ -0,0 +1,165 @@
...
+ * protection domain (target/classes) has no permissions assigned at all. That way any missing privileged actions in
+ * Hibernate Validator would lead to an access control exception. All the other domains involved in tests (e.g.
+ * Arquillian or TestNG) are assigned with "all permissions" to keep this set up manageable. As the privilege-less
+ * DelegatingExecutor frame sits is located in the stack very close to the actual code under test, it is ensured that
```
hferentschik: 'sits'?

tck-runner/pom.xml
```diff
@@ -22,7 +22,6 @@
     <properties>
         <tck.version>1.1.2.Final</tck.version>
         <tck.suite.file>${project.build.directory}/dependency/beanvalidation-tck-tests-suite.xml</tck.suite.file>
-        <arquillian.version>1.0.0.Final</arquillian.version>
```
hferentschik: good catch

tck-runner/pom.xml
```diff
@@ -323,5 +320,47 @@
...
+        <profile>
+            <id>security-manager</id>
+            <activation>
+                <property>
+                    <name>with-security-manager</name>
+                    <value>true</value>
+                </property>
+            </activation>
+            <properties>
+                <arquillian.protocol>LocalSecurityManagerTesting</arquillian.protocol>
```
hferentschik: How does this work? Do you define a new protocol? Is this an Arquillian feature? Wouldn't mind somewhere a link/hint to some more information

tck-runner/pom.xml
```diff
@@ -323,5 +320,47 @@
...
+                                <property>
+                                    <name>validation.provider</name>
+                                    <value>${validation.provider}</value>
+                                </property>
+                                <property>
+                                    <name>excludeIntegrationTests</name>
+                                    <value>true</value>
+                                </property>
+                            </systemProperties>
+                            <argLine>-Djava.security.manager -Djava.security.policy=${project.build.directory}/test-classes/test.policy -Djava.security.debug=access</argLine>
```
hferentschik: So the difference to the 'local' profile is the arg line enabling the security manager and the custom arquillian protocol, right? Did you think about just modifying the local profile to run with security manager and not even provide a switch. Would there be a drawback? On the plus side I can see less lines of config code. 

If we keep the this separate profile a better name might be 'local-with-security-manager' to make clear that is a specialization of 'local'. Also what's about `mvn clean install -Dincontainer -Dwith-security-managet`, that would not make sense, right? If possible we should try to make it impossible ('local' and 'inconainer' are mutual exclusive ) or at least make appropriate notes in the readme.

tck-runner/src/test/java/org/hibernate/validator/tckrunner/securitymanager/TckRunner.java
```diff
@@ -0,0 +1,73 @@
...
+ * -Djava.security.policy=target/test-classes/test.policy
+ * -Djava.security.debug=access
+ * -DexcludeIntegrationTests=true
+ * -Darquillian.protocol=LocalSecurityManagerTesting
+ * -Dorg.jboss.testharness.spi.StandaloneContainers=org.hibernate.jsr303.tck.util.StandaloneContainersImpl
+ * </pre>
+ *
+ * @author Gunnar Morling
+ */
+public class TckRunner {
```
hferentschik: I guess this makes it a bit easier to run the TCK test case. I tend to select the suite from the TCK project and change the classpath.

hferentschik: BTW, trying to use this via the Idea, I get:

    access: access allowed ("java.io.FilePermission" "/Applications/Development/IDE/IntelliJ IDEA 14.app/Contents/lib/idea_rt.jar" "read")
    Exception in thread "main" java.lang.ExceptionInInitializerError
    Caused by: java.security.AccessControlException: access denied ("java.util.PropertyPermission" "idea.launcher.bin.path" "read")
	at java.security.AccessControlContext.checkPermission(AccessControlContext.java:457)
	at java.security.AccessController.checkPermission(AccessController.java:884)
	at java.lang.SecurityManager.checkPermission(SecurityManager.java:549)
	at java.lang.SecurityManager.checkPropertyAccess(SecurityManager.java:1294)
	at java.lang.System.getProperty(System.java:714)
	at com.intellij.rt.execution.application.AppMain.<clinit>(AppMain.java:42)

tck-runner/src/test/java/org/hibernate/validator/tckrunner/securitymanager/arquillian/LocalSecurityManagerTestingExecutionEvent.java
```diff
@@ -0,0 +1,80 @@
...
+public class LocalSecurityManagerTestingExecutionEvent extends LocalExecutionEvent {
+
+	public LocalSecurityManagerTestingExecutionEvent(TestMethodExecutor executor) {
+		super( new DelegatingTestMethodExecutor( executor ) );
+	}
+
+	/**
+	 * Invokes the given test method via {@link DelegatingExecutor} which lives in its own protection domain.
+	 *
+	 * @author Gunnar Morling
```
hferentschik: are you not going a bit overboard with all these author tags?

tck-runner/src/test/resources/test.policy
hferentschik: 'set up' == 'set-up'

hferentschik:
> update the docs on the web site as well

What are you referring to? The `readme.md` needs updating for sure (one way or the other, depending in which form we enable this)

hferentschik:
Gee, I am a bit impressed. You are jumping through quite some hoops here. How did you figure all this out (including the Arquillian extension)? Thanks for the comments/javadoc, w/o I would have had a hard time piecing it together.

tck-runner/pom.xml
gunnarmorling: Yes, it is a custom protocol, essentially it's the same as "Local", only that it puts one frame onto the stack which we can assign with minimum rights in the test.policy so to discover any missing privileged blocks in HV. I don't have any documentation of this at hand unfortunately.

tck-runner/pom.xml
gunnarmorling: > So the difference to the 'local' profile is the arg line enabling the security manager and the custom arquillian protocol, right?

Right.

> Did you think about just modifying the local profile to run with security manager and not even provide a switch. Would there be a drawback? On the plus side I can see less lines of config code.

Interesting idea. Being able to run the tests without the SM may still come in handy when analysing bugs, test failures etc. as there is one less moving part. I've changed it to one profile, running with or without the SM being enabled based on the property. The default is now running with the SM, but it can be disabled like so:

    mvn clean install -pl tck-runner/ -Dwith-security-manager=false

> Also what's about mvn clean install -Dincontainer -Dwith-security-managet, that would not make sense, right?

One actually could think about running the tests on WildFly with the SM enabled. But I won't add it for the time being as it functionally should be the same.

> If possible we should try to make it impossible ('local' and 'inconainer' are mutual exclusive ) or at least make appropriate notes in the readme.

I'll add a note.

tck-runner/src/test/java/org/hibernate/validator/tckrunner/securitymanager/TckRunner.java
gunnarmorling: > I guess this makes it a bit easier to run the TCK test case

Exactly. It's faster than having to fiddle with the XML file, attach a remote debugger etc.

tck-runner/src/test/java/org/hibernate/validator/tckrunner/securitymanager/arquillian/LocalSecurityManagerTestingExecutionEvent.java
gunnarmorling: It's part of my IDE templates for new classes and I didn't really care about it. Feel free to remove it if you like.

gunnarmorling:
> What are you referring to? The readme.md needs updating for sure (one way or the other, depending in which form we enable this)

That's the thing, the readme.md does not describe the build options (which is what I would have expected). Instead, they are described on the [web site](http://hibernate.org/validator/contribute/#build-options).

gunnarmorling:
> Gee, I am a bit impressed. You are jumping through quite some hoops here. How did you figure all this out (including the Arquillian extension)? Thanks for the comments/javadoc, w/o I would have had a hard time piecing it together.

Glad you like it :smile: I was looking for a way to put a frame onto the stack with mininum privileges (Surefire, Arquillian etc. need very many permissions, so it's not feasible to sort these ones out) and the custom protocol was the best thing I could come up with. I'm not sure how I figured it out, I already did it last year when we discussed the SM issues. I guess I debugged a bit through Arquillian and looked for a way to intercept the calls :)

As far as the required policies are concerned, it's basically trial and error while running the test suite with the SM.

tck-runner/src/test/resources/test.policy
gunnarmorling: Fixed all these typos.

gunnarmorling:
@hferentschik, pushed one more commit. It's all part of the "local" profile now, with the SM enabled by default. To disable it, append "-Dwith-security-manager=false". Would be nice if you could play a bit with it (e.g. omit some rights from HV in the policy file etc.) to see how it works for you.

tck-runner/pom.xml
hferentschik: Ok. I wrote this comment before seeing all of the code. Browsing the code in the IDE makes it quite clear how things come together. How did you find this, if there was no documentation?

tck-runner/pom.xml
hferentschik: > I've changed it to one profile, running with or without the SM being enabled based on the property. The default is now running with the SM, but it can be disabled l

+1 Right, that's what I had in mind as well.

hferentschik: > One actually could think about running the tests on WildFly with the SM enabled. But I won't add it for the time being as it functionally should be the same.

Sure, I was more bringing it up, so that we can make sure the docs are reflecting this

tck-runner/src/test/java/org/hibernate/validator/tckrunner/securitymanager/TckRunner.java
hferentschik: > You may want to add an entry for this to the policy file if possible?

I'll try

hferentschik:
> That's the thing, the readme.md does not describe the build options (which is what I would have expected). 

I am referring to the readme in the tck module itself. There is the local vs container mode described. That should be for sure updated. I guess for good measure we can update the website as well.

hferentschik:
> As far as the required policies are concerned, it's basically trial and error while running the test suite with the SM.

Got you

tck-runner/src/test/resources/test.policy
hferentschik: thanks

hferentschik:
> Would be nice if you could play a bit with it (e.g. omit some rights from HV in the policy file etc.) to see how it works for you.

I already did the omitting of permissions to see how it fails. What I am still planning to do is adding the required Idea permission (provided there is a simple way to fix that)

hferentschik:
@gunnarmorling, can you update the readme in the tck module?

gunnarmorling:
> @gunnarmorling, can you update the readme in the tck module?

Ah, wasn't even aware of that one. Yes, I'll update it.

But can we move that stuff to the readme.md in root? It's more the place where people (including me ;) would look for build instructions, as you often don't even navigate to specific sub-modules.

hferentschik:
> Ah, wasn't even aware of that one. Yes, I'll update it.

Cool. Btw, something seems to be broken atm. I pulled your latest changes and got heaps of errors. Also some issues on CI it seems.

> But can we move that stuff to the readme.md in root?

Not sure. If you pull that into the root readme, then you need to discuss other build options as well there. no? So far we just have a single command mentioned (to keep things short) on how to build out of the box. Otherwise we refer  to the website. Maybe a link from the root readme would do?

gunnarmorling:
> Also some issues on CI it seems.

Looking.

> If you pull that into the root readme, then you need to discuss other build options as well there. no? 

Yes, all the build options should be defined there (of course after presenting the simple "one click build"). Spreading that info over several files or even the web site makes it only harder to find. But anyways, it's independent from the issue at hand.

hferentschik:
> But anyways, it's independent from the issue at hand.

Right

gunnarmorling:
Force-pushed to update readme and make the policy file work when running the entire aggregator build.

gunnarmorling:
@hferentschik Are we good to go with this one?

The readme is updated. The issue on CI was caused by the fact that I locally only had built the tckrunner module (thus obtaining the engine JAR from the local repo) whereas the CI build runs the full aggregator build (thus obtaining the engine JAR from engine/target). I've added another entry to the policy file for that.

hferentschik:
> are we good to go with this one?

Not sure :-) I thought I would give the problem of running `TckRunner` in Idea a go and came across a couple of things:

* I am wondering why you are not using the '-' wildcard for the non HV grants? This removes also the use of some hard coded versions which would probably cause problems during a components upgrade. * I could reduce the JAXB permissions to just `permission java.util.PropertyPermission "mapAnyUriToUri", "read";`. Are you sure you need the other ones?

I created #366 which is based on this pull request (commits squashed), adding my changes to the policy file. WDYT?

Btw, I can run `TckRunner` without the security manager (which is probably good enough), but even after quite some fiddling I was not able to fix the permissions for running with the manager enabled.

hferentschik:
Speaking of TckRunner, it might be worth mentioning it in the readme. Some note on how to debug local and in-container. Could be part of another issue though? Wdyt?

gunnarmorling:
> I am wondering why you are not using the '-' wildcard for the non HV grants?

I listed them individually to be extra-safe, but you are right, it should be good enough just with the folder prefixes and wildcards.

> I could reduce the JAXB permissions

Interesting, I saw some access control exceptions around these, but apparently these are handled internally within JAXB. I could swear though I saw them bubble up. But tests pass indeed with your change. A bit odd, but the less permissions the better of course :)

> Btw, I can run TckRunner without the security manager (which is probably good enough), but even after quite some fiddling I was not able to fix the permissions for running with the manager enabled. 

Hum, what permissions is it lacking? Is it due to the IntelliJ JAR?

> Speaking of TckRunner, it might be worth mentioning it in the readme. Some note on how to debug local and in-container. Could be part of another issue though? Wdyt?

For debugging, just run it in debug mode in your IDE. It's a main class, I don't think it can be ran within the container (for that there is the Maven profile). The required args are noted right on top of the class. What else would you document?

Closing this one, as it is superseded by #366.

hferentschik:
> Hum, what permissions is it lacking? Is it due to the IntelliJ JAR?

I add all permissions to the Idea jar, but that does not cut it. I do that and then new failing permissions pop up. It looks like that the origin comes from testng, but that has already all permissions. So I might just not be able to read the security manager debug trace. In the end I started added global grants (one after another) as they were popping up. Again, all indicated testng, eg permission to create the html and xml output files etc, but adding the permissions to testng was no good. Anyways, in the end I had enough global permissions defined to not get any more exceptions and the test "ran", but it apparently did not find any test classes. In the end I gave up.

> It's a main class, I don't think it can be ran within the container (for that there is the Maven profile).

Sure. AFAIU this class is there to have an easy way to debug the TCK test suite from within the tck-runner module. Apart from the security manager problem, I got it to work. I used to do it a bit differently, but this works better.

>  The required args are noted right on top of the class.

The seems not quite correct though. For example the policy file needs to be specified via a URI, aka file:/... Otherwise the policy file cannot even be found and the VM reverts to the other default locations.

 > What else would you document?

That's pretty much it. Nothing out of the ordinary. The thing is, if you are new and need a way to debug TCK test, you don't know where to  start and you won't know that `TckRunner` exists.

PR #365, opened 2015-02-09
HV-953 Changing message interpolation algorithm to allow constraint defi...

hferentschik:
...nition contributors to provide default message bundles

- Bundles must be places into the same package as the custom constraint annotation
- Adding tests for default messages to ConstraintDefinitionContributorTest
- Updating docs
- Removing unused log field in ResourceBundleMessageInterpolator
- Adding comments to org.hibernate.validator.ValidationMessages.java

### DIFF
--- a/documentation/src/main/asciidoc/ch11.asciidoc
+++ b/documentation/src/main/asciidoc/ch11.asciidoc
@@ -694,16 +694,24 @@ mechanism to register additional constraint definitions. All you have to do is t
 _javax.validation.ConstraintValidator_ to _META-INF/services_. In this service file you list the
 fully qualified classnames of your constraint validator classes (one per line). Hibernate Validator
 will automatically infer the constraint types they apply to.
-See <<example-using-service-file-for-constraint-definitions>> for an example.
+See <<example-using-service-file-for-constraint-definitions,Constraint definition via service file>>
+for an example.
 
 [[example-using-service-file-for-constraint-definitions]]
-.Using _META-INF/services/javax.validation.ConstraintValidator_ for constraint definitions
+._META-INF/services/javax.validation.ConstraintValidator_
+
 ====
 [source]
 ----
+# Assuming a custom constraint annotation @org.mycompany.CheckCase
 org.mycompany.CheckCaseValidator
 ----
 ====
+
+To contribute default messages for your custom constraints, place a ContributorValidationMessages.properties_
+and/or its locale-specific specializations into the same package as your custom constraint annotations.
+In the example above, this would mean _org.mycompany_. If you your constraint annotation are spread
+over multiple packages, you need to specify one properties file per package.
 
 [[section-constraint-definition-contributor]]
 ==== Constraint definitions via +ConstraintDefinitionContributor+
--- a/engine/src/main/java/org/hibernate/validator/messageinterpolation/AbstractMessageInterpolator.java
+++ b/engine/src/main/java/org/hibernate/validator/messageinterpolation/AbstractMessageInterpolator.java
@@ -68,6 +68,11 @@ public abstract class AbstractMessageInterpolator implements MessageInterpolator
 	public static final String USER_VALIDATION_MESSAGES = "ValidationMessages";
 
 	/**
+	 * Default name of the message bundle defined by a constraint definition contributor.
+	 */
+	public static final String CONTRIBUTOR_VALIDATION_MESSAGES = "ContributorValidationMessages";
+
+	/**
 	 * The default locale in the current JVM.
 	 */
 	private final Locale defaultLocale;
@@ -213,6 +218,13 @@ public abstract class AbstractMessageInterpolator implements MessageInterpolator
 		if ( resolvedMessage == null ) {
 			ResourceBundle userResourceBundle = userResourceBundleLocator
 					.getResourceBundle( locale );
+
+			ResourceBundleLocator constraintResourceBundleLocator = getResourceBundleLocatorForConstraintContributions(
+					context
+			);
+			ResourceBundle constraintContributorResourceBundle = constraintResourceBundleLocator
+					.getResourceBundle( locale );
+
 			ResourceBundle defaultResourceBundle = defaultResourceBundleLocator
 					.getResourceBundle( locale );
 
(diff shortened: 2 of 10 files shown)


documentation/src/main/asciidoc/ch11.asciidoc
```diff
@@ -694,17 +694,25 @@ mechanism to register additional constraint definitions. All you have to do is t
...
+
 ====
 [source]
 ----
+# Assuming a custom constraint annotation @org.mycompany.CheckCase
 org.mycompany.CheckCaseValidator
 ----
 ====
 
+To contribute default messages for your custom constraints, place a _ValidationMessages.properties_
```
gunnarmorling: How about having one named bundle per archive instead which then is "merged" (e.g. similar to service loader files)?

I think that'd be a bit easier to comprehend/maintain. I'd need to have a different name than `ValidationMesssages` though on order to give the "real" `ValidationMessages` precedence.

gunnarmorling:
Nice! Looks good to me apart from that one comment/question above.

documentation/src/main/asciidoc/ch11.asciidoc
hferentschik: > How about having one named bundle per archive instead which then is "merged" (e.g. similar to service loader files)?

I thought about this, but in the end settled for this. Let me give you some pro and cons for this approach:
Pro:
- Simple, no need for yet another contract/api
- We use `ValidationMessages.properties` consistently. This makes the connection to Bean Validation obvious. It also prevent proliferation of names when everyone tries to name the properties file something else
- It automatically creates a natural namespace. Using a new contract, people could for example put a `ValidationMessages.properties` into the root of their jar. This would then conflict with the users `ValidationMessages.properties`

Con:
- If you have custom constraint spread over multiple packages, you need to provide multiple message bundles. However, I expect this to be the exception. 

> I think that'd be a bit easier to comprehend/maintain

Don't you think this rule is pretty straight forward?

>  It'd need to have a different name than ValidationMesssages though

Kind of what I would like to prevent.

gunnarmorling: > Simple, no need for yet another contract/api

API? It'd be a defined bundle name. I guess you could consider that as a contract.

> This makes the connection to Bean Validation obvious. It also prevent proliferation of names when everyone tries to name the properties file something else

Actually it's re-using the defined name from the spec with slightly other semantics I'm a bit concerned about. A separate name would make more apparent that this is a HV-specific feature. Re everybody using different names, it'd be one single defined name.

> Using a new contract, people could for example put a ValidationMessages.properties into the root of their jar. This would then conflict with the users ValidationMessages.properties

Right, it'd have to be another name for that reason.

> Don't you think this rule is pretty straight forward?

I guess it is. Still I find re-using the BV-defined bundle name not optimal.

hferentschik: > It'd be a defined bundle name. I guess you could consider that as a contract.

Ahh, got you now. You just want to use a different name? What would that be? 

> Actually it's re-using the defined name from the spec with slightly other semantics I'm a bit concerned about

Where is your concern about this?

gunnarmorling: > What would that be?

E.g. `ConstraintValidationMessages`, `ProvidedValidationMessages`, `DefaultValidationMessages`, `HibernateValidatorValidationMessages`.

> Where is your concern about this?

Re-using the BV-defined name as is may lead people to the assumption that this is a portable feature, while it actually is HV-specific.

Also it might conflict in case BV one day actually supports `ValidationMessages` in other places than root. Depending on the then defined semantics, issues may arise upon upgrading. All in all I feel it's safer to stay away from the BV-defined name.

hferentschik: I see where you are coming from. There was a initial misunderstanding on what you meant, but I get your idea of changing the expected name of the properties file.

> ConstraintValidationMessages, ProvidedValidationMessages, DefaultValidationMessages, HibernateValidatorValidationMessages.

None really give me the big aha moment. Maybe the last one, but it does not describe where they originate (constraint definition contribution), just that they are HV specific. 

> Re-using the BV-defined name as is may lead people to the assumption that this is a portable feature, while it actually is HV-specific.

I don't think that is a big risk for a bundle name.

> Also it might conflict in case BV one day actually supports ValidationMessages in other places than root.

Well, one would hope that our way will make into the spec :-)

hferentschik: What's about `ContributorValidationMessages`!?

gunnarmorling: > What's about ContributorValidationMessages!?

Yes, sounds good. But to be sure, what I had in mind is a resource bundle at the root of archives, not one per package. Is that what you think, too?

hferentschik: > But to be sure, what I had in mind is a resource bundle at the root of archives, not one per package. Is that what you think, too?

Hmm, are we not running into a visibility problem again? What if I add more than one contributor jar each providing a `ContributorValidationMessages`? I think we need the namespace one way or the other.

hferentschik:
Pushed an update renaming properties file to `ContributorValidationMessages`

documentation/src/main/asciidoc/ch11.asciidoc
gunnarmorling: > Hmm, are we not running into a visibility problem again? What if I add more than one contributor jar each providing a ContributorValidationMessages?

That's what I meant to address when saying "which then is 'merged' (e.g. similar to service loader files)". I.e. we'd consider all the entries from all the `ContributorValidationMessages` bundles. 

There might be a collission of keys, but best practise is to give them "fully-qualified" so it shouldn't be a problem in practice. I guess the same issue btw. exists with the original proposal when giving a bundle in the same package in two different archives.

hferentschik: > That's what I meant to address when saying "which then is 'merged' (e.g. similar to service loader files)". I.e. we'd consider all the entries from all the ContributorValidationMessages bundles. 

Ahh. You realize that this will require some doing? `ResourceBundle.getBundle` internally just returns any resource bundle of the requested name (uses `Classloader#getResource internally`). There is afaict no way to out of the box get all bundles with the same name. There is a `ResourceBundle.Control` which one apparently can use to customize some of the loading behaviour, but there is of course a risk of getting it wrong. 

The idea is nice, but I am afraid, it will take too much time. I can experiment a bit.

hferentschik:
Closing for now.  I want to push some work in progress to my branch ...

PR #366, opened 2015-02-10
HV-913

hferentschik:


hferentschik:
thanks. I have to say, working with the security manager is a true pita.

PR #367, opened 2015-02-10
HV-915 Update compiler option for documentation module

DavideD:
https://hibernate.atlassian.net/browse/HV-915

For the document module we could actually set the JDK to 1.8, I don't think there is any drawback.
This would avoid some errors when importing the project in eclipse.

I'm going to update the contribution guide on the website to explain how to solve the other errors.

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -63,6 +63,15 @@
     <build>
         <plugins>
             <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <!-- We can keep it 1.8 since we don't have sources in this module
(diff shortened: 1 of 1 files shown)


DavideD:
I have create https://hibernate.atlassian.net/browse/WEBSITE-253 to update the Validator contribute page explaining how to fix the errors due to the import

hferentschik:
Merged. I updated the comment a bit in the pom.xml, since it was not quite accurate. Bottom line, for IDE is the the 'target' and 'source' options of the compiler plugin which determine the language level in the IDE module. 'testSource' and 'testTarget' get ignored in Eclipse as well as in Idea. Since the documentation module does not have any production classes it is indeed ok to just set the 'target' and 'source' level to 1.8 in this case.

PR #368, opened 2015-02-11
HV-915 Easier import in eclipse

DavideD:
This patch will store store the settings for m2e plugin the parent pom and will move all the tests requiring JDK 8 in a separate module.

Now the only things I have to do to import the project without errors are:
1) rename the parent folder to something different from hibernate-validator
2) Run the build
3) Import the project with the m2e plugin
4) Set the JavaSE-1.6 environment runtime to use JDK 1.8

Note that this setup will stay if you refresh the maven configuration.

Hopefully I didn't forget anything when moving the tests.


hferentschik:
Merged thanks. 

The split-out of the tests into its own module helped, but it highlighted another issue again which I am having in Idea as well. When a module depends on another module and an additional qualified artifact created by the same dependent module, Idea screws up the dependencies and only one of them ends up on the classpath. This is the case for the engine 'testing' artifact. On import one needs to manually fiddle with the module setups to make it work. 

Again, it is really an IDE bug, but the solution is similar to the solution here - create a 'test-utils' module and move the affected code there. How far do we go with all this?

hferentschik:
Also following up with #370 which addresses another IDE issue (at least in Idea). Using the maven-processor-plugin for generating the logging classes does not work in the IDE out of the box. On changes to the Log interface, the logging classes need to be updated via the command line build.

DavideD:
> The split-out of the tests into its own module helped, but it highlighted another issue again which I am having in Idea as well. When a module depends on another module and an additional qualified artifact created by the same dependent module, Idea screws up the dependencies and only one of them ends up on the classpath. This is the case for the engine 'testing' artifact. On import one needs to manually fiddle with the module setups to make it work.

Eclipse does not seem to have this issue. It works out of the box.

> Also following up with #370 which addresses another IDE issue (at least in Idea). Using the maven-processor-plugin for generating the logging classes does not work in the IDE out of the box. On changes to the Log interface, the logging classes need to be updated via the command line build.

Again, I didn't experience this problem in eclipse.

The only problem I have at the moment is that you will have some errors if you don't have widlfly in the target folder in some projects. This happens because in some project we copy some artifacts in the server folder and eclipse complains that the folder does not exists. This errors disappear if you run the build.

PR #369, opened 2015-02-11
HV-896 Add "ParenPad" CheckStyle rule 

DavideD:
https://hibernate.atlassian.net/browse/HV-896

### DIFF
--- a/annotation-processor/src/main/java/org/hibernate/validator/ap/checks/GetterCheck.java
+++ b/annotation-processor/src/main/java/org/hibernate/validator/ap/checks/GetterCheck.java
@@ -37,7 +37,7 @@ public class GetterCheck extends AbstractConstraintCheck {
 					)
 			);
 		}
-		else if (!hasReturnValue(element)) {
+		else if ( !hasReturnValue( element ) ) {
 			return CollectionHelper.asSet(
 					new ConstraintCheckError(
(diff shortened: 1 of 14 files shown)


DavideD:
I've added another couple of rules related to whitepspaces that don't seem to have a huge impact on the existing code

DavideD:
I can't figure out why it does not always catch all the violations.

The configuration file seems correct since I've found some violations.
I've also added a couple of rules that don't seem to have a huge impact on the existing code.

hferentschik:
> I've added another couple of rules related to whitepspaces that don't seem to have a huge impact on the existing code

Cool. I think the code is already fairly consistent. Actually quite nice to be in this state.

> I can't figure out why it does not always catch all the violations.

Sure, no problem. I am happy to just down as a checkstyle issues. As long as we agree that in theory with the configuration you added, the examples I gave you should raise checkstyle errors. Funny enough, some violations got caught. Well, well, not worth putting more time into it. 

> The configuration file seems correct since I've found some violations.

+1 That's what I think as well. I vaguely remember once @Sanne mentioning that the `ParenPad` checkstyle rule is not really cutting it. I think that's why he reverted to several regexp rules. But I might just be imagining this.

> I've also added a couple of rules that don't seem to have a huge impact on the existing code.

Ok

hferentschik:
Merged with the addition of the checkstyle plugin upgrade. Thanks.

PR #370, opened 2015-02-11
HV-963 Upgrading JBoss Logging to 3.2.1.Final and removing the use of ma...

hferentschik:
...ven-processor-plugin. Annotation processing occurs now as part of the compilation phase

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -57,6 +57,19 @@
             <artifactId>javax.el</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>jboss-logging-processor</artifactId>
+            <!--
+            HV-963
+            This is actually not a dependency which is needed at runtime, however,
+            Maven does not have a compile time only scope. The dependency is needed to
+            run the JBoss Logging annotation processor as part of the main compilation.
+            Trying different setups via compiler plugin local dependencies or extensions
+            all fail. See also http://stackoverflow.com/questions/14322904/maven-3-how-to-add-annotation-processor-dependency
+            -->
+            <scope>provided</scope>
+        </dependency>
 
         <!--
         Optional dependencies
@@ -164,10 +177,6 @@
                 </dependencies>
             </plugin>
             <plugin>
-                <groupId>org.bsc.maven</groupId>
-                <artifactId>maven-processor-plugin</artifactId>
-            </plugin>
-            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
             </plugin>
@@ -206,7 +215,6 @@
                             <classesDirectory>${project.build.testOutputDirectory}</classesDirectory>
                             <includes>
                                 <include>org/hibernate/validator/testutil/**</include>
-                                <include>org/hibernate/validator/test/internal/metadata/provider/AnnotationMetaDataProviderTestBase*</include>
                             </includes>
                         </configuration>
                     </execution>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -503,37 +508,6 @@
                     <version>1.3</version>
                 </plugin>
                 <plugin>
-                    <groupId>org.bsc.maven</groupId>
```
gunnarmorling: What's the reason for removing this? I liked the fact that this set-up does not add the processor as normal project dependency but only on the separate processor path. There is a separate artifact with only the annotations which can be used as project dependency.

engine/pom.xml
```diff
@@ -206,7 +208,6 @@
                             <classesDirectory>${project.build.testOutputDirectory}</classesDirectory>
                             <includes>
                                 <include>org/hibernate/validator/testutil/**</include>
-                                <include>org/hibernate/validator/test/internal/metadata/provider/AnnotationMetaDataProviderTestBase*</include>
```
DavideD: This class is used in the new engine-jdk8-tests module.
You should have some error if you remove it.

hferentschik: No, I bit of a mistake from my part. I already changed this when merging your pull request. I just forgot to add this as well. I should have just staged this hunk. In the merged code I got rid of the base class and copied the few helpers needed. I think just creating the 'testing' jar from `org/hibernate/validator/testutil` is nicer. In particular if we think about making this its own module.

DavideD: ok

pom.xml
hferentschik: > What's the reason for removing this? 
Several reasons:
* the long outstanding upgrade of JBoss Logging
* letting the annotation processor play "nicer" with the IDE. In its current setup it is not working from Idea and one needs to manually re-create the Log class on changes to the interface
* according to David this is the "recommended" setup

>  I liked the fact that this set-up does not add the processor as normal project dependency but only on the separate processor path

Me too. As you know I was arguing for it for a long time, even refusing an earlier upgrade of JBoss Logging. Since other principles already weakened (see Eclipse pulgin settings), I think this is another logical step. It makes working with the IDE easier and we made that now an important goal.

> There is a separate artifact with only the annotations which can be used as project dependency.

That's not helping the goal of letting the processor run as part of the compile step.

Btw, as a final step I am considering a dedicates testutil module as mentioned in other comments.


I liked the fact that this set-up does not add the processor as normal project dependency but only on the separate processor path. There is a separate artifact with only the annotations which can be used as project dependency.

gunnarmorling: For what's it worth, this works nicely in Eclipse :)

> Since other principles already weakened (see Eclipse pulgin settings)

That's two different things, though. The added plug-in config doesn't do any harm (apart from violating your sense of aesthetics ;) whereas this dependency really is wrong in the sense that it allows to (accidentally) import stuff from the processor JAR.

But in the end I don't care much about it, tests would show up this sort of issues anyways. So if you think that's the lesser evil, let's do it in order to move forward.

hferentschik: > So if you think that's the lesser evil, let's do it in order to move forward.

Make it so! ( I added another comment to be more explicit) 

For now I think it is the lesser evil indeed.

gunnarmorling:
Merged; Just noticing though that still the deprecated annotations from the processor JAR are used rather than the ones from the separate annotation JAR. Will you send another PR for that?

hferentschik:
> Just noticing though that still the deprecated annotations from the processor JAR are used rather than the ones from the separate annotation JAR.

Ok,  I see. I have not even noticed that yet. The primary concern was as said something else. I find this whole JBoss Logging thing a bit of a joke. How it is structured and all. At to top things up, there is no website or other documentation to be found anywhere. Not even a decent readme in the github repo. For a tool which is supposed to be used across so many different projects that is just poor. 

> Will you send another PR for that?

Will do. Keep the Jira issue open and I create another pull request against its key.

Thanks for merging.

PR #372, opened 2015-02-13
HV-963 Switching package names for logging annotations going away from t...

hferentschik:
...he deprecated annotations


hferentschik:
that should complete HV-963 for now

PR #373, opened 2015-02-13
HV-942 Upgrade wildfly

DavideD:
https://hibernate.atlassian.net/browse/HV-942

I had to add a beans.xml in the dummy-bean-validation-provider.jar to make it pass.
Not sure why it was working in the previous version.


hferentschik:
> Not sure why it was working in the previous version.

Hmm, `beans.xml` should really be optional in this version of Weld/CDI. Might be a bug on there side. I guess it does not matter for us.

hferentschik:
Merged. Thanks.

PR #374, opened 2015-02-13
HV-964 Breaking test utils classes out into own module instead of being ...

hferentschik:
...creating in the engine module using a classifier


hferentschik:
Ok, at least from an Idea pov all is working now. Clean import, building, .... No manual re-setting of configuration options :-)

hferentschik:
FYI, three times in a row the same test failure. I am working on this branch rebased onto current master (in case that matters)

hferentschik:
Merged

test-utils/src/main/java/org/hibernate/validator/testutil/ConstraintViolationAssert.java
```diff
@@ -62,8 +63,8 @@ private ConstraintViolationAssert() {
 	 * @param expectedMessages The expected constraint violation messages.
 	 */
 	public static void assertCorrectConstraintViolationMessages(Set<? extends ConstraintViolation<?>> violations,
-																String... expectedMessages) {
-		List<String> actualMessages = newArrayList();
+			String... expectedMessages) {
+		List<String> actualMessages = new ArrayList<String>();
```
gunnarmorling: Interesting, why are you changing this?

test-utils/src/main/java/org/hibernate/validator/testutil/ConstraintViolationAssert.java
```diff
@@ -604,4 +605,215 @@ else if ( !parameterIndex.equals( other.parameterIndex ) ) {
 			return true;
 		}
 	}
+
+	public static class PathImpl implements Path {
```
gunnarmorling: What's that other `PathImpl` good for? Seems like copy & paste?

gunnarmorling:
I'm not quite clear what's the overall motivation for this change? Did you have any problems before? Would be nice in general to have a chance for taking a look before it's getting merged.

test-utils/src/main/java/org/hibernate/validator/testutil/ConstraintViolationAssert.java
hferentschik: Because this used CollectionsHelper which is part of the main engine classes.

test-utils/src/main/java/org/hibernate/validator/testutil/ConstraintViolationAssert.java
hferentschik: Same here, we used the path impl from the engine before

hferentschik:
> I'm not quite clear what's the overall motivation for this change? 

Issues with the IDE. There is a bug in Idea which cannot handle depending on two artifacts from the same module. Also it sticks to the one artifact one module rule. 

> Would be nice in general to have a chance for taking a look before it's getting merged.

Sorry, I thought you where gone for the weekend. I thought this is simple enough to be merged and it was a bit annoying to switch between branches with different modules layout.

test-utils/src/main/java/org/hibernate/validator/testutil/ConstraintViolationAssert.java
gunnarmorling: Ah, makes sense. It'd be nice to use Java 7's diamond in tests. Maven supports it, but the IDEs don't like different target levels for main and test (at least Eclipse doesn't).

gunnarmorling:
> There is a bug in Idea which cannot handle depending on two artifacts from the same module. 

I see. Out of curiosity, is it a new bug? Or have you lived with it for all the time?

> Sorry, I thought you where gone for the weekend. I thought this is simple enough to be merged and it was a bit annoying to switch between branches with different modules layout. 

No problem. I was only wondering what's that change about at all :) Look's good overall, only that duplication of `PathImpl` is a bit annoying.

hferentschik:
> I see. Out of curiosity, is it a new bug? Or have you lived with it for all the time?

It's around for quite some time. It also needs manual tweaking, but in this case the manual changes "stick", so that I was not so bothered. Still, since we are on a little mission here. As it stands now all works as expected in Idea.

PR #375, opened 2015-02-13
HV-950 Fix log message when parameterizedType has more then one type

DavideD:
https://hibernate.atlassian.net/browse/HV-950

Changed the level and the message


hferentschik:
Rebased and merged.

PR #376, opened 2015-02-15
HV-933 Changing @CPF and @CNPJ to validate w/ or w/o separator character...

hferentschik:
...s by introducing actual constraint validator classes.

These constraint are not anymore pure composition constraints.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraints/br/CPF.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/br/CPF.java
@@ -14,8 +14,6 @@ import javax.validation.Payload;
 import javax.validation.ReportAsSingleViolation;
 import javax.validation.constraints.Pattern;
 
-import org.hibernate.validator.constraints.Mod11Check;
-
 import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
 import static java.lang.annotation.ElementType.CONSTRUCTOR;
 import static java.lang.annotation.ElementType.FIELD;
@@ -42,16 +40,6 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 		@Pattern(regexp = "^(?:(?!777\\.?777\\.?777-?77).)*$"),
 		@Pattern(regexp = "^(?:(?!888\\.?888\\.?888-?88).)*$"),
 		@Pattern(regexp = "^(?:(?!999\\.?999\\.?999-?99).)*$")
-})
-@Mod11Check.List({
-		@Mod11Check(checkDigitIndex = 12,
-				endIndex = 10,
-				treatCheck10As = '0',
-				ignoreNonDigitCharacters = true),
-		@Mod11Check(checkDigitIndex = 13,
-				endIndex = 12,
-				treatCheck10As = '0',
-				ignoreNonDigitCharacters = true)
 })
 @ReportAsSingleViolation
 @Documented
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/Mod11CheckValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/Mod11CheckValidator.java
@@ -45,24 +45,44 @@ public class Mod11CheckValidator extends ModCheckBase
 	private char treatCheck11As;
 
 	/**
-	 * @return The threshold for the algorithm multiplier multiplier growth
+	 * The threshold for the algorithm multiplier multiplier growth
 	 */
 	private int threshold;
 
 	@Override
 	public void initialize(Mod11Check constraintAnnotation) {
-		super.initialize(
+		initialize(
 				constraintAnnotation.startIndex(),
 				constraintAnnotation.endIndex(),
 				constraintAnnotation.checkDigitIndex(),
-				constraintAnnotation.ignoreNonDigitCharacters()
+				constraintAnnotation.ignoreNonDigitCharacters(),
+				constraintAnnotation.threshold(),
+				constraintAnnotation.treatCheck10As(),
+				constraintAnnotation.treatCheck11As(),
+				constraintAnnotation.processingDirection()
 		);
-		this.threshold = constraintAnnotation.threshold();
+	}
 
-		this.reverseOrder = constraintAnnotation.processingDirection() == ProcessingDirection.LEFT_TO_RIGHT;
+	public void initialize(int startIndex,
+			int endIndex,
+			int checkDigitIndex,
+			boolean ignoreNonDigitCharacters,
+			int threshold,
+			char treatCheck10As,
+			char treatCheck11As,
+			ProcessingDirection direction
+	) {
+		super.initialize(
+				startIndex,
+				endIndex,
+				checkDigitIndex,
+				ignoreNonDigitCharacters
+		);
+		this.threshold = threshold;
+		this.reverseOrder = direction == ProcessingDirection.LEFT_TO_RIGHT;
 
-		this.treatCheck10As = constraintAnnotation.treatCheck10As();
-		this.treatCheck11As = constraintAnnotation.treatCheck11As();
+		this.treatCheck10As = treatCheck10As;
+		this.treatCheck11As = treatCheck11As;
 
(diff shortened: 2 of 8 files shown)


engine/src/main/java/org/hibernate/validator/constraints/br/CPF.java
```diff
@@ -43,16 +41,6 @@
...
 })
-@Mod11Check.List({
-		@Mod11Check(checkDigitIndex = 12,
-				endIndex = 10,
-				treatCheck10As = '0',
-				ignoreNonDigitCharacters = true),
-		@Mod11Check(checkDigitIndex = 13,
-				endIndex = 12,
-				treatCheck10As = '0',
-				ignoreNonDigitCharacters = true)
```
gunnarmorling: We had the idea of making `ignoreNonDigitCharacters` overridable via `@AnnotationOverrides`. Did this not work out?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
```diff
@@ -0,0 +1,65 @@
...
+		secondMod11CheckWithoutSeparator.initialize( createMod11ProxyAnnotation( 9, 12, 13, true ) );
+	}
+
+	@Override
+	public boolean isValid(CharSequence value, ConstraintValidatorContext context) {
+		if ( value == null ) {
+			return true;
+		}
+
+		if ( value.toString().contains( "." ) ) {
```
gunnarmorling: Does only the dot trigger the special handling? Or any non-digit character?

engine/src/main/java/org/hibernate/validator/constraints/br/CPF.java
hferentschik: AnnotationOverrides won't work in this case. Sure, we can provide a parameter which can change these values, but that required beforehand to know whether the format of the string is w/ or w/o separators. The original behavior (and what imo makes most sense) is to handle both cases transparently.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
```diff
@@ -0,0 +1,65 @@
...
+import org.hibernate.validator.constraints.br.CNPJ;
+import org.hibernate.validator.internal.constraintvalidators.hv.Mod11CheckValidator;
+import org.hibernate.validator.internal.util.annotationfactory.AnnotationDescriptor;
+import org.hibernate.validator.internal.util.annotationfactory.AnnotationFactory;
+
+/**
+ * @author Hardy Ferentschik
+ */
+public class CNPJValidator implements ConstraintValidator<CNPJ, CharSequence> {
+	private final Mod11CheckValidator firstMod11CheckWithSeparator = new Mod11CheckValidator();
```
gunnarmorling: I'd put the "withSeparator" first as it's the more important aspect: `withSeparatorMod11Validator1`, `withSeparatorMod11Validator2`.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
hferentschik: there can be other separators as well in the whole string, but a '.' is a given. One could change this to non digit check I guess. Remember there are still the `@Pattern` constraints of the composing constraints

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
gunnarmorling: I'm not quite sure why there are four validators needed? And were do the numbers used for their initialization come from? Some short explanation of this enitre scheme would be helpful.

hferentschik: might be a bit better. You can change it if you like.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
```diff
@@ -0,0 +1,65 @@
...
+	private final Mod11CheckValidator secondMod11CheckWithSeparator = new Mod11CheckValidator();
+
+	private final Mod11CheckValidator firstMod11CheckWithoutSeparator = new Mod11CheckValidator();
+	private final Mod11CheckValidator secondMod11CheckWithoutSeparator = new Mod11CheckValidator();
+
+	@Override
+	public void initialize(CNPJ constraintAnnotation) {
+		firstMod11CheckWithSeparator.initialize( createMod11ProxyAnnotation( 9, 14, 16, true ) );
```
gunnarmorling: How about adding an `initialize()` method to `Mod11CheckValidator` which takes the actual argument types instead of the annotation. That'd be a bit more straight-forward than creating the annotation proxy just for retrieving the values from it.

hferentschik: > How about adding an initialize() method to Mod11CheckValidator which takes the actual argument types instead of the annotation. That'd be a bit more straight-forward than creating the annotation proxy just for retrieving the values from it.

Interesting idea. I wanted to stay with the constraint validator API and not add new methods, but since this the validator instance is internal anyways, it would not matter. Also this felt like constraint composition, just that I could not to it completely via configuration. I'll look into it.

hferentschik:
@gunnarmorling pushed an update with feedback incorporated

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/Mod11CheckValidator.java
```diff
@@ -45,24 +45,44 @@
...
+				constraintAnnotation.threshold(),
+				constraintAnnotation.treatCheck10As(),
+				constraintAnnotation.treatCheck11As(),
+				constraintAnnotation.processingDirection()
 		);
-		this.threshold = constraintAnnotation.threshold();
+	}
 
-		this.reverseOrder = constraintAnnotation.processingDirection() == ProcessingDirection.LEFT_TO_RIGHT;
+	public void initialize(int startIndex,
```
gunnarmorling: Nice!

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
```diff
@@ -0,0 +1,67 @@
...
+
+	private final Mod11CheckValidator withSeparatorMod11Validator1 = new Mod11CheckValidator();
+	private final Mod11CheckValidator withSeparatorMod11Validator2 = new Mod11CheckValidator();
+
+	private final Mod11CheckValidator withoutSeparatorMod11Validator1 = new Mod11CheckValidator();
+	private final Mod11CheckValidator withoutSeparatorMod11Validator2 = new Mod11CheckValidator();
+
+	@Override
+	public void initialize(CNPJ constraintAnnotation) {
+		// validates CNPJ strings with separator, eg 91.509.901/0001-69
```
gunnarmorling: Cool, thanks for the thorough description!

gunnarmorling:
Merged, thx!

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
hferentschik: no worries. good to have it at hand now

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/Mod11CheckValidator.java
hferentschik: I think it turned out for the better now

hferentschik:
thanks. one down

PR #378, opened 2015-02-16
HV-887 Parameter name of an overridden method is taken from the parent c...

emmanuelbernard:
...lass or interface

Note that I did not follow Gunnar's proposal to favor explicit names over `argN`. We would need to propagate a flag if the `DefaultParameterNameProvider` and it's unclear it is a better and more predictable approach.


hferentschik:
So what do you do to get a deterministic result?

emmanuelbernard:
@hferentschik https://github.com/emmanuelbernard/hibernate-validator/commit/f9750a6ddf87e3fc7293d67e76c66ec30bd66333#diff-ec4255be3f0b5ad505e6f0396cfadc2cR146

I always look for the highest parent class between the various overloaded methods received.

hferentschik:
Looking good.

> I always look for the highest parent class between the various overloaded methods received.

I prefer that as well

hferentschik:
Merged, thanks.

PR #379, opened 2015-02-16
Readme undusting

Sanne:
Some micro-polishing, but including some of these URL's are plain old 404

Does it need a JIRA?

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -23,7 +23,7 @@
     <distributionManagement>
         <site>
             <id>site</id>
-            <url>http://validator.hibernate.org</url>
+            <url>http://hibernate.org/validator</url>
         </site>
     </distributionManagement>
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ If you need help with cloning, compiling or setting the project up in an IDE ple
 If you are just getting started with Git, GitHub and/or contributing to Hibernate Validator there are a
 few pre-requisite steps:
 
-* Make sure you have a [Hibernate JIRA account](https://hibernate.onjira.com)
+* Make sure you have a [Hibernate JIRA account](https://hibernate.atlassian.net)
 * Make sure you have a [GitHub account](https://github.com/signup/free)
 * [Fork](http://help.github.com/fork-a-repo) the Hibernate Validator [repository](https://github.com/hibernate/hibernate-validator).
 As discussed in the linked page, this also includes:
@@ -43,7 +43,7 @@ appreciated btw), please use rebasing rather than merging.  Merging creates
 "merge commits" that really muck up the project timeline._
 
 ## Submit
-* Sign the [Contributor License Agreement](https://cla.jboss.org/index.seam).
+* Sign the [Contributor License Agreement](https://cla.jboss.org/).
 * Push your changes to a topic branch in your fork of the repository.
 * Initiate a [pull request](http://help.github.com/send-pull-requests/)
 * Update the JIRA issue, adding a comment including a link to the created pull request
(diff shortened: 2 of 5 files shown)


engine/pom.xml
```diff
@@ -23,7 +23,7 @@
     <distributionManagement>
         <site>
             <id>site</id>
-            <url>http://validator.hibernate.org</url>
+            <url>http://hibernate.org/validator</url>
```
hferentschik: why do you want to change that?

hferentschik:
@Sanne create an issue for it ;-)

engine/pom.xml
Sanne: because the first one redirects to the second, that means the second one is your canonical representation.
Also, consistency: there is no such thing as "orm.hibernate.org", so essentially those having a subdomain are legacy forwarding to the new order.

Sanne:
re-pushed to add the JIRA code on the commits

engine/pom.xml
hferentschik: >  that means the second one is your canonical representation. 

I am not sure about "canonical representation". Just because one URL is a redirect, does not mean it is less important than the URL it redirects to. The redirect gives Validator a more prominent URL and I think that is the whole purpose. There is no legacy involved here. There is no ORM, because ORM == hibernate.org (of course one could argue that such a redirect would make sense). Last but not least, there is search.hibernate.org and that's what's used in the Search POM.

gunnarmorling: +1 for http://hibernate.org/validator, it's nicer to the user to avoid the re-direct. But guys, let's spend our cycles on more valuable stuff ;)

hferentschik:
Merged, except the validator.hibernate.org change. Good catch with the stale links. OOI, why is is so important to write JIRA? Afaik it is not even an acronym.

Sanne:
> why is is so important to write JIRA? Afaik it is not even an acronym.

Just branding. It really is always and only all capitals. They really care a lot about that (you wouldn't write `RedHat` would you?), but the only official statement is this:

"For those who care - it sounds best if you yell it loudly, as though charging into battle. C'mon - try it!"
-Atlassian

engine/pom.xml
Sanne: >  Last but not least, there is search.hibernate.org and that's what's used in the Search POM.

Sure that's not the best choice either. But here I was just scanning the whole project for links, I wouldn't create a JIRA and a PR just to fix a single link ;)

> I am not sure about "canonical representation". Just because one URL is a redirect, does not mean it is less important than the URL it redirects to. 

It is in context of SEO, "canonical" is a technical term of the field. This is de-valuing the SEO value for your project page, you need to make it clear across all your content which one is the "right" way to spell your homepage.

But +1 to spend cycles on more valuable stuff ;) I was just reviewing all links and went overboard with perfectionism.

Sanne:
> Merged

Whoo I'm a Validator contributor!! :dancer:

engine/pom.xml
Sanne: > There is no ORM, because ORM == hibernate.org (of course one could argue that such a redirect would make sense)

No that's a problem we need to fight.
We should be working hard on making sure that Hibernate.org == Java data platform.
When I asked why we don't have an orm.hibernate.org @emmanuelbernard explained me that it's because the subdomains are legacy.. makes sense to me, but it suggests you should try stopping using them.

hferentschik:
> you wouldn't write RedHat would you

I actually do sometimes, since I never can remember whether it is w/ or w/o space. Maybe now I will remember.

> "For those who care - it sounds best if you yell it loudly, as though charging into battle. C'mon - try it!"

LOL. Thanks for digging this out.

hferentschik:
> Whoo I'm a Validator contributor!! 

Yeah! Keep it coming ;-)

engine/pom.xml
emmanuelbernard: Right don't use subdomains anymore. Prefer the hibernate.org/project/ form.

PR #381, opened 2015-02-17
HV-897 Allowing to customize retrieval of current time for @Future/@Past

gunnarmorling:


hferentschik:
A couple of words about commits. I don't know why, but this is now the second time where the order of the commits on GitHub is not the one of the actual commits. For example #eb41139 is displayed after #a7305fa which is apparently wrong. If i get the branch locally it is correct. I know the multiple commits are supposed to help the review, but if they are not even in the right order, they even do the opposite. There was the same issue with the classloader/osgi pull request. It confused the heck out of you, if you review commit by commit and suddenly something gets used you have not even seen introduced yet. No idea, if this is bug on GitHub or whether you do some git magic which confuses the UI. I don't think I have seen this before (except your other pull request).

Speaking of commits, in my world it would be perfectly ok to have  #eb41139 and  #a7305fa in the same commit, but if you split, I would expect that the first commit just introduces the new API and default impl and the next updates all the validators (past and future). Then I can easily review and digest the actual API ignoring the repeating application of the new contract.

gunnarmorling:
> this is now the second time where the order of the commits on GitHub is not the one of the actual commits.

IMO it's a short-coming of GitHub. They [order the commits](https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/) by time and not as they appear in the branch. If I go back and re-work one earlier commit via interactive rebasing, that commit will appear after other commits as it got modified later.

I also find that confusing and am not sure why it's done that way in the UI. IMO they should order commits as they are on the branch.

> in my world it would be perfectly ok to have #eb41139 and #a7305fa in the same commit, but if you split...

I see. My thinking is that: I aim for a) adding "value" with a commit, while b) not letting grow them too big. 

Just committing the new SPI without actually any usage of it doesn't satisy a). Also I wouldn't know whether the SPI acutally makes sense without any test implementation and usage. Hence I added the update to the future constraints to that commit. For the sake of b), so to keep it digestable, I then let the past stuff for another commit. So that's my reasoning, but surely one can do it differently.

hferentschik:
> IMO it's a short-coming of GitHub. They order the commits by time and not as they appear in the branch. If I go back and re-work one earlier commit via interactive rebasing, that commit will appear after other commits as it got modified later.

How odd. That seems just wrong.

gunnarmorling:
@hferentschik You mentioned you'd still have some concerns with this. What are these exactly? I'd love to get it off my plate before switching to OGM :)

hferentschik:
> You mentioned you'd still have some concerns with this. What are these exactly?

Two things really. 
* Regarding the API, I am not happy with exposing a `Calendar`. I would be much more comfortable with a `long` and say a `TimeZone`. 
* I am also not happy about the validation of `LocalDateTime`. Implementation details aside, I am questioning something like this:
    
    @Past
     LocalDateTime dateTime;

To answer the question whether something is in the future or past you need an instant on the time-line. `LocalDateTime` does not have this information.

>  I'd love to get it off my plate before switching to OGM :)

Sure, I want this to get resolved asap as well. However, I think it will be after the release.

gunnarmorling:
> I am also not happy about the validation of LocalDateTime

I see. IMO it's a valid requirement people may have in practice in their apps.

E.g. assume you are building an HR application and want to manage entry dates of future new hires. It makes sense to store that without TZ information. E.g. someone starts on the literal date "10/2015", this does not shift depending on the TZ one is in. I.e. when looking at a given contract, I always want to read "10/2015", I do not want this to be adjust with respect to any TZ.

Based on that I think the following is a valid use case which we should support:

    class Contract {

        @Future // want to validate it's in the future when creating a new contract
        YearMonth monthOfEntry;
    }

Just assuming we agree that this is a valid requirement, next question is, how to compare it with "now", so we can decide whether it's in the future or not.

You said "you need an instant on the time-line". And this is right. To convert a `YearMonth` (or any other date type without TZ information) into an instant on the time line, we need to have a time zone.

The only reasonable thing to do I can see is to take the TZ from what we consider as "now". So if the HR application's user's TZ is "Australia", it makes sense to use this TZ for checking whether "10/2015" is in the future or not. If the user's TZ is "America", than use that one to find out whether "10/2015" is in the future or not. After all, "10/2015" still may be in the future for the user in America but not for the one in Australia.

So we need to have the instant and the TZ of "now" for validating the constraint.

You suggest to use a long and `TimeZone` to convey that information. Why is it that you prefer that over using `Calendar`? `Calendar` is the canonical type used in Java (prior to 8) to represent an instant on the time line and the time zone to be used to "render" it into something human readable. This is what users will typically have in their apps to represent this data. So why require them to decompose the `Calendar` they likely have themselves instead of us doing this? IMO it'd be an unnecessary usability shortcoming. Also the semantics of `Calendar` is clearly defined, whereas a long by itself can be everything. We'd have to mandate in docs that it is the milli-seconds since epoch.

Therefore it makes most sense to me - semantically and design-wise - to use the JDK type designed for this purpose. But maybe I'm missing a piece; It'd help to know exactly why you'd feel more comfortable with the alternative proposal.

Sanne:
> Just assuming we agree that this is a valid requirement, 

I definitely don't :-)

I do agree that lots of people like doing such things, but it's a very bad practice. Would you want to encourage it rather than showing them how it should be done?

gunnarmorling:
> I do agree that lots of people like doing such things, but it's a very bad practice.

In which way is it bad practice? You cannot arbitrarily call valid use cases "bad practice" ;) There is just a category of cases where a TZ is not needed or even wrong. Or would you e.g. store a birthday with a TZ? This makes no sense.

Sanne:
hum ok but you wouldn't use Date for a birthday, as a Date resolves to a point on the global timeline, while the birthday is a cultural concept connected to "when people normally wake up in your region".

What I mean is that a birthday should be stored either as a `String` or as a `org.joda.time.LocalDate`. But using a `Date` without a timezone will lead any developer into complex timezone related issues, for example tests working on one machine and not the other because they have a different TZ.. never encourage that!

Sanne:
Back on topic:

    class Contract {

        @Future // want to validate it's in the future when creating a new contract
        YearMonth monthOfEntry;
    }

The question: do you need it to be validated in the user's timezone or were the server-side code is running?

gunnarmorling:
> What I mean is that a birthday should be stored either as a String or as a org.joda.time.LocalDate. 

Or `java.time.LocalTime`. Which is exactly my point. And in similar cases its timezone-less siblings `Year`, `YearMonth` etc. So it's exactly the opposite of bad practice, use that date/time time which matches your needs :)

> But using a Date without a timezone will lead any developer into complex timezone related issues, for example tests working on one machine and not the other because they have a different TZ.. never encourage that!

All true. But how is it related? I fully agree and nowhere said anything else. On the contrary, by supporting `@Past`/`@Future` on all these data types we in a way allow people to use those date/time types which are most sensible for a given use case.

> The question: do you need it to be validated in the user's timezone or were the server-side code is running?

That's exactly what this PR is about. So far it used to be the server TZ but obviously that's not always what people need. So this PR seeks to make it customizable what "now" is via an SPI which allows people to plug in a custom "time provider".

In the example it e.g. might be based on the TZ of the application user creating a contract or the country for which the contract applies. This might be a different TZ than on the server. The SPI is also meant for use cases where you want to run with another "logical now" than the real one, e.g. in a batch application which you want to re-run with yesterday's date.

Sanne:
Ok, thanks for the explanations @gunnarmorling I had misunderstood quite some things :)

I don't see how an SPI is helpful though. Let's say you're handling any incoming RPC (HTTP or whatever else), tha requested TZ will be a per-request contextual value. How is the SPI going to associate the validation with the context? Not by using some ThreadLocals I hope ;-)

So @hferentschik 's concerns seem valid to me. In his examples, he seems to require the timezone to be somehow present in the values.

disclaimer: I have no idea what kind of SPIs you might have in this project.. I hope I'm not annoying, I'm trying to figure out how things work.

gunnarmorling:
> I don't see how an SPI is helpful though. Let's say you're handling any incoming RPC (HTTP or whatever else), tha requested TZ will be a per-request contextual value. How is the SPI going to associate the validation with the context?

Yes, that's a good point, but I think it can be solved based on the specific requirements and used technologies. E.g. your `TimeProvider` implementation could be a CDI bean which gets the required request-scoped state injected. Or in a desktop application, there could be a global user context which gets passed to the time provider upon instantiation.

Apart from the TZ stuff, the SPI is also helpful for the batching use case i described above. Or testing, when you also want to run with another "now" than the real one.

> I hope I'm not annoying, I'm trying to figure out how things work.

Not annoying at all. We need to battleproof our APIs/SPIs. In doubt it's better to fail already when questioning ourselves than when it's out in the wild :)

hferentschik:
Ok, you had a quite big discussion. Nice. Here come my 0.02$. I will quote several things now, so I hope it is not to confusing.

> E.g. assume you are building an HR application and want to manage entry dates of future new hires. It makes sense to store that without TZ information.

I disagree. When it comes to contracts be it a hire or something else I would never store it as a MonthYear. What if it becomes important when exactly you joined? Or there is a contract which gets signed by multiple people across the world and it matters who signed first (for example you get a bigger share of something if you sign first). It is one question what gets displayed, but another what is stored in the data store and in the domain of contract I would always store a something which can be placed on the timeline. 

> The only reasonable thing to do I can see is to take the TZ from what we consider as "now".

Who is we? Take an application which is deployed on multiple servers across the world. According to which one your request gets routed (assuming some sort of decoupled client server app) you get different results for your validation. As Sanne is saying, this is asking for trouble. 

> What I mean is that a birthday should be stored either as a String or as a org.joda.time.LocalDate. But using a Date without a timezone will lead any developer into complex timezone related issues, for example tests working on one machine and not the other because they have a different TZ

So you guys talked about birthdays. Let's store my birthday. You are not interested in my actual birthday or how old I am, so you store the information into a MonthDay:

    User {
        MonthDay userBirthDay
    }

How would `@Future` or `@Past` even make sense here. My last birthdays has passed, my next is upcoming.

> On the contrary, by supporting @Past/@Future on all these data types we in a way allow people to use those date/time types which are most sensible for a given use case.

No, we are breaking the semantics of future and past, whether you like it or not. That's why I once floated the idea of a different constraint. Not sure whether this is useful or not, but one could imagine that in the case of date/time instances where there is no time zone information available, you basically set one with the constraint. (just throwing an idea out here).

> So @hferentschik 's concerns seem valid to me. In his examples, he seems to require the timezone to be somehow present in the values.

Exactly my point. And I am skeptical about any solution which the requires then CDI or what not to get the desired behavior.

hferentschik:
Now some words about the API. Why do I prefer `long` + `TimeZone` over `Calendar`. 

* It is overkill for the `Date` based constraints. What is happening here. For validation you create a `Calendar` instance, to then create the `Date` in order to compare. Internally the two `Date` instances then use the time in milli to do the comparison. So, if I can get the time in millis from the time provider directly I can to the comparison much easier without creating all these intermediate objects
* As I keep on saying, with a `long` + `TimeZone` we are already much closer to Java 8 `Clock`. In the future we could add `ZoneId` (potentially removing `TimeZone` in the long run) and basically implement `Clock`. If we are using `Calendar` we are stuck with this quite high level beast
* Or lets consider someone who wants to write a custom date/time constraint using Java 8. He is not interested in a `Calendar`. However, with the time in millis and a TimeZone on which he can call `timeZone.toZoneId()` he has the pieces he needs without the additional indirection via `Calendar`

I don't know, I find this all valid reasons.

gunnarmorling:
>  When it comes to contracts be it a hire or something else I would never store it as a MonthYear. What if it becomes important when exactly you joined?

Ok, I guess it's a bit moot to discuss that example, but that's really how you'd model it? It would make a difference in your business whether someone joined the company at 1:25 a.m. or 8:37 p.m.? I don't know, I have never seen a working contract which had a time in it ;) Have you been asked when getting your working contract at which time of your first day you plan to start working so they can put it into the contract? You started with the company at a specific date, not a time of the day. At least that's how I see it.

> Who is we? Take an application which is deployed on multiple servers across the world. According to which one your request gets routed you get different results for your validation.

That one surprises me. Isn't it the very purpose of this change to make this controllable for the application developer? They can define "now" as it makes sense for their semantics and business needs. So "we" really is the application developer who knows what "now" is in their terms.

> Let's store my birthday. You are not interested in my actual birthday or how old I am, so you store the information into a MonthDay
> No, we are breaking the semantics of future and past, whether you like it or not

Yes, `MonthDay` (as well as `LocalTime`) is a different beast. I agree you cannot say whether "October 12th" or "4:56 p.m." is in the future or not, because it's no unambigious point on the time line. And that's why there are indeed not `@Future`/`@Past` validators for these types. Because they make no sense really.

But `LocalDateTime` (and `YearMonth`, `Year`) is the case we have been talking about. Those are different. I can tell - knowing what "now" is, which I do know thanks to the new SPI - whether these are in the future or not.

> but one could imagine that in the case of date/time instances where there is no time zone information available, you basically set one with the constraint.

But how would that really work? The constraint and its attributes are defined statically in the model, whereas the time zone really depends on the context of given instances.

> Exactly my point. And I am skeptical about any solution which the requires then CDI or what not to get the desired behavior. 

It's not that CDI would be required, but I agree you need some way to obtain the time zone. I don't see it as a big issue, but you may have a point here. Maybe we should allow for some way for pointing the engine to another property of the validated bean which holds the TZ? It must be linked to a specific instance, I don't think it makes sense to have it as constraint attribute. Or maybe some way to pass in a validation context into `Validator` via `unwrap()`?

gunnarmorling:
> It is overkill for the Date based constraints. What is happening here. For validation you create a Calendar instance, to then create the Date in order to compare. Internally the two Date instances then use the time in milli to do the comparison. So, if I can get the time in millis from the time provider directly I can to the comparison much easier without creating all these intermediate objects

Point taken on the `Date` case. For all the others it doesn't make a difference IMO.

> As I keep on saying, with a long + TimeZone we are already much closer to Java 8 Clock. In the future we could add ZoneId (potentially removing TimeZone in the long run) and basically implement Clock. If we are using Calendar we are stuck with this quite high level beast

Either way we are having a legacy type in the SPI. Either `Calendar` or `TimeZone`. Why would we be stuck with `Calendar`? As we can replace `TimeZone` with `ZoneId` we could replace `Calendar` with `Clock` or `ZonedDateTime`.

> Or lets consider someone who wants to write a custom date/time constraint using Java 8. He is not interested in a Calendar. However, with the time in millis and a TimeZone on which he can call timeZone.toZoneId() he has the pieces he needs without the additional indirection via Calendar

I don't think it makes a big difference in the end. Within an application you typically use the date/time types. I've never seen methods taking a long + TZ, but people use the semantic types for representing this data. So at some point the decomposition will have to be done, either in the constraint or within the `TimeProvider` implementation.

All in all using a "proper" date/time type still makes most sense to me. But if you feel strongly about it let's have two separate getters. After all, the `Date`/`Instant` case benefits from it.

gunnarmorling:
Thinking about it, how about using `Date` + `TimeZone` which would be replaced by `Clock` down the road?

A `Date` creation will be required either way; If people don't have it anyways, `Date#after()` expects a `Date`, not a long. Similar with `Instant#isAfter(Instant)` etc.

Sanne:
Maybe you could simply refine the issue by working on the new corrected API for date/time exclusively i.e. don't support such a feature on `Date` but encourage people to want that to either roll their own - and deal with the complexities in a specific responsible way - or move to the better type system.

There is only minor value in supporting such a new feature for the older types.

hferentschik:
> Maybe you could simply refine the issue by working on the new corrected API for date/time exclusively i.e. ...

I am not quite sure what you are saying here. 

> There is only minor value in supporting such a new feature for the older types.

Are you talking about the `TimeProvider` contract? Well, yes there is no real "need" for it for the existing time/date based validators, but it opens up some nice possibilities (eg around testing, this is how this whole contract came about).  The whole discussion around `TimeProvider` contract and which data types to support in general for `@Future` and `@Past` are orthogonal to begin with.

hferentschik:
> A Date creation will be required either way; If people don't have it anyways, Date#after() expects a Date, not a long. Similar with Instant#isAfter(Instant) etc.

Well, you don't really need the `Date`. You could simply get the epoch time from the value date and do a direct comparison of two long values. 

Also talking about `Date` and `Instant`, depending with which API you work with you need to create one or the other. If you work with the Java 8 types you probably want to create an `Instant`. To do so you use the epoch time in milliseconds. If you pass a `Date` you have to unwrap again. So I still vote for `long` + `TimeZone`.

Here is what I think needs to be done:
* Rework the `TimeProvider` contract proposed in this pull request (this is really what HV-897 is about)
* Verify the Java 8 date types we currently allow for `@Future` and `@Past` and disable any validators for types which don't have a time zone (potentially in a separate issue)

Regarding the second point, we then need to have a discussion how `@Future` and `@Past` can be sensibly supported for these types.

hferentschik:
> Maybe we should allow for some way for pointing the engine to another property of the validated bean which holds the TZ?

Hmm, but would such a property exist? If you are using a `LodateDateTime` would you separately keep the timezone? But yes, in general we need some way to remove the ambiguity. 

> It must be linked to a specific instance, I don't think it makes sense to have it as constraint attribute. 

You would "hard-code" the timezone you are going to use. But yes, it might not be so useful either. As said, just throwing out ideas.

> Or maybe some way to pass in a validation context into Validator via unwrap()?

How would that look like? Also does this not break the "programming model" of Bean Validation?

gunnarmorling:
> want to create an Instant. To do so you use the epoch time in milliseconds. If you pass a Date you have to unwrap again.

No, there is `Date.toInstant()`. The JDK provides the required bridges between old and new API. They want to spare users from dealing with epoch milli seconds themselves.

> So I still vote for long + TimeZone.

I see, I only fail to understand why. Would you ever represent time/date properties as milli-seconds in a "normal" application? Personally I wouldn't, the date/time APIs are there for a reason and I would always prefer those. Representing date/time by epoch milliseconds and encouraging to do comparisons based on those rather than working with `before()`/`after()` methods seems just like bad API design to me.

If you are concerned about the required conversion from legacy to new date/time API, I'd even prefer to require Java 8 for the `TimeProvider` contract and use `Clock` or whatever. IMO better to have a proper contract which is based on 8 only rather than one which just seems awkwardly designed.

> Rework the TimeProvider contract proposed in this pull request (this is really what HV-897 is about)

Yes, once we agreed on the direction to go :)

> Verify the Java 8 date types we currently allow for @Future and @Past and disable any validators for types which don't have a time zone (potentially in a separate issue)

Why not. I don't think there are any "wrong" validators (as said we don't support `LocalTime` or similar) but if you find any let's discuss them. As you say let's better do it separately so to move on.

hferentschik:
> Representing date/time by epoch milliseconds and encouraging to do comparisons based on those rather than working with before()/after() methods seems just like bad API design to me.

How you are doing the comparisons is up to you. I was just saying one might as well do the comparison directly. If a user wants to create a `Date` instance or some other `java.time.*` is up to him. He gets the basic value he needs to go any direction he wants.

> IMO better to have a proper contract which is based on 8 only rather than one which just seems awkwardly designed.

I guess we have to agree to disagree on this. IMO there is nothing awkwardly designed in what I am proposing. We just have different opinions on what is good design in this case.

>  I don't think there are any "wrong" validators

There is `FutureValidatorForChronoLocalDate`, `FutureValidatorForYear`, `FutureValidatorForYearMonth`, etc

hferentschik:
@gunnarmorling, where do we stand with this? You want to refactor this or shall I take this over and make final changes?

gunnarmorling:
Hey, as you know this PR is good as is IMO, so I won't do any further changes to it. Feel free to do whatever you want with it. We already agreed to disagree on this one :)

hferentschik:
Ok, then. I close this PR for now.

hferentschik:
Superseded by #402

PR #382, opened 2015-02-17
HV-914 Adding note on required permissions when running with a security ...

gunnarmorling:
...manager to the reference guide


hferentschik:
merged, thanks.

PR #384, opened 2015-02-17
HV-890 Allowing sub-type methods to have different constraints than a private super-type method with the same signature

gunnarmorling:


hferentschik:
> It seems not 100% fixed, though. 

Ok, I'll leave it for now then

gunnarmorling:
@hferentschik Pushed and update, should be good now. I'm excluding meta-data for private super-type methods completely now.

hferentschik:
rebased and merged :-)

PR #386, opened 2015-02-18
HV-968 Avoiding over-sized list allocations, usage of concurrent map and...

gunnarmorling:
... a key look-up in ConstraintHelper


hferentschik:
rebased and merged

PR #387, opened 2015-02-18
HV-953

hferentschik:
Integration test still missing (could potentially be provided later). Ready for review though.


hferentschik:
Added wildfly integration test

engine/src/main/java/org/hibernate/validator/resourceloading/AggregateResourceBundle.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Properties;
+import java.util.ResourceBundle;
+
+/**
+ * Inspired by <a href="http://stackoverflow.com/questions/4614465/is-it-possible-to-include-resource-bundle-files-within-a-resource-bundle">this</a>
+ * Stack Overflow question.
+ *
+ * @author Hardy Ferentschik
+ */
+public class AggregateResourceBundle extends ResourceBundle {
```
gunnarmorling: Is there any difference to the inner class of the same name in `PlatformResourceBundleLocator`? Can this be merged to one class? If not, different names would be nice to make the different semantics apparent.

gunnarmorling: Interesting, this seems to be un-used. Maybe a left-over?

integration/src/test/java/org/hibernate/validator/integration/wildfly/ConstraintDefinitionContributorUnitIT.java
```diff
@@ -0,0 +1,74 @@
...
+	private static final String WAR_FILE_NAME = ConstraintDefinitionContributorUnitIT.class.getSimpleName() + ".war";
+
+	@Inject
+	private Validator validator;
+
+
+	@Deployment
+	public static WebArchive createTestArchive() throws Exception {
+		return ShrinkWrap
+				.create( WebArchive.class, WAR_FILE_NAME )
```
gunnarmorling: Very nice test!

gunnarmorling:
Rebased and applied. Added two more commits to remove that superfluous class and do some other minor fixes.

hferentschik:
Thanks, good catch with the `AggregateResourceBundle`. I meant to delete it, but apparently I forgot.

PR #388, opened 2015-02-18
HV-948 Adding missing return statement

hferentschik:


gunnarmorling:
Rebased and merged. Thx!

PR #390, opened 2015-02-19
HV-915 Excluding transitive dependency to JConsole

gunnarmorling:


pom.xml
```diff
@@ -226,6 +226,12 @@
                 <groupId>org.wildfly</groupId>
                 <artifactId>wildfly-arquillian-container-managed</artifactId>
                 <version>${wildfly.version}</version>
+                <exclusions>
```
hferentschik: fair enough with me, just wondering how this is related to HV-915 and Eclipse setup?

gunnarmorling: It showed up in Eclipse for me. Not in the CLI build. Hence I added it with this issue key.

hferentschik: aha

PR #391, opened 2015-02-19
HV-967 Fixing broken links and pre-resolving some valid but redirecting ...

Sanne:
...links

I wouldn't have spent time on this, but since I had the changeset already around.. do you want it or should I trash it?

(It was part of my previous PR about links)

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ There are more build options available as well. For more information refer to [C
 
 ## Hibernate Validator URLs
 
-* [Home Page](http://validator.hibernate.org)
+* [Home Page](http://hibernate.org/validator/)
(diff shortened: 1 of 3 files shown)


hferentschik:
I rebased and merged it, since it seems the consensus seems to be that this the way to go. Note, that personally I liked 'validator.hibernate.org'.  I think it gives the project more "importance". Also I don't think that in general it is always a good idea to give users a "canonical" url. There are good reasons why redirects exists and make sense. Also the SEO argument does imo not really apply here. Nevertheless, it's merged.

Sanne:
TBH I like the domain more as well. Just applying what had been decided in the past here.. but happy to rethink eventually.

PR #392, opened 2015-02-19
HV-955 Allowing to use constraint declaration API for configuration of default validator factory

gunnarmorling:


hferentschik:
A integration test would be nice here as well, especially since the request comes from this corner.

gunnarmorling:
Good idea with the integration test. Added one.

hferentschik:
Good stuff. One of the true user value features of this sprint. Nice impl.

hferentschik:
rebased and merged

PR #393, opened 2015-02-19
HV-970 Avoiding to access javax.xml.stream.* package if not using XML co...

gunnarmorling:
...nfiguration


hferentschik:
Merged, thanks.

PR #394, opened 2015-02-19
Getting rid of custom Maven mirrors

Sanne:
https://hibernate.atlassian.net/browse/HV-971


Sanne:
@gunnarmorling the build failed because the job should be adjusted to not use the _settings-example.xml_ anymore. I can't fix that myself (and probably shouldn't until this is actually integrated)

gunnarmorling:
Rebased and applied. Thx, @Sanne!

gunnarmorling:
Also changed the job config. Do you have an account on CloudBees? Then I can add you to the admins if you like.

Sanne:
I don't think I have one, np. thanks!

PR #395, opened 2015-02-19
HV-972 Avoid dependency on ASM4 (upgrade groovy-jsr223)

Sanne:
https://hibernate.atlassian.net/browse/HV-972


gunnarmorling:
Thx!

PR #396, opened 2015-02-19
HV-969 Setting scope of test utils dependencies to provided.

hferentschik:
Many modules just want @TestForIssue from the test util package.
If they depend on the assertion or log4j helper they needs to define this
dependency anyways themselves. This avoids also have testng on the classpath
(through transitive dependencies) within a module using JUnit


gunnarmorling:
Rebased and applied.

PR #398, opened 2015-04-13
HV-980. Store java.util.Set element as Node.key in Path

ffeisst:
When a bean property of type java.util.Set is violated, the Path of the ConstraintViolation does contain no information which element of the Set is violated. As a fix for this problem, the violated element is stored as Node.key in the Path.

See https://hibernate.atlassian.net/browse/HV-980 for more details


hferentschik:
As is we cannot merge this, since it is against the spec. If anything we would need to make this a provider specific feature.

hferentschik:
Closing this for now. We will use a different approach where it will be possible to unwrap the `Node` to a Hibernate Validator specific `Node` type which will expose a dedicated method.

PR #399, opened 2015-04-20
HV-975 Adding test case for using Optional in conjunction with private g...

hferentschik:
...etter

I am really unable to re-produce the reported issue. I suggest to add this testcase matching the input from the reporter and mark the issue as non re-producable.


gunnarmorling:
Rebased and applied.

hferentschik:
thanks :-)

PR #401, opened 2015-04-21
HV-982 Downgrading compiler plugin to avoid MCOMPILER-236

hferentschik:


hferentschik:
I think downgrading the plugin is for now the easiest option. There is really no necessity to use version 3.2.

PR #403, opened 2015-04-22
HV-847: Allowing to use @SupportedValidationTarget to resolve target ambiguity of purely composed constraint

gunnarmorling:
For now I just added support for resolving the ambiguity by meta-annotating the constraint definition with `@SupportedValidationTarget`.

Trying to deduct the target from composing constraints led to some chicken/egg issue: to do so, I need to parse the composing constraints first, but for that I need the constraint type of the composed constraint. I guess it could be sorted out somehow, but it didn't seem worth the effort: It's an border case, and explicit resolution is needed anyways as a fallback in case deduction from the composing constraints fails. I'd say we can add that should the subject be defined in the spec.

### DIFF
--- a/documentation/src/main/asciidoc/ch11.asciidoc
+++ b/documentation/src/main/asciidoc/ch11.asciidoc
@@ -350,8 +350,53 @@ public static class MyConstraintMappingContributor implements ConstraintMappingC
 You then need to specify the fully-qualified class name of the contributor implementation in _META-INF/validation.xml_,
 using the property key +hibernate.validator.constraint_mapping_contributor+.
 
+[[section-advanced-constraint-composition]]
+=== Advanced constraint composition features
+
+==== Validation target specification for purely composed constraints
+
+In case you specify a purely composed constraint - i.e. a constraint which has no validator itself but is solely made
+up from other, composing constraints - on a method declaration, the validation engine cannot determine whether that
+constraint is to be applied as a return value constraint or as a cross-parameter constraint.
+
+Hibernate Validator allows to resolve such ambiguities by specifying the +@SupportedValidationTarget+ annotation on the
+declaration of the composed constraint type as shown in <<example-purely-composed-constraint-validation-target>>.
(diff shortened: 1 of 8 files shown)


documentation/src/main/asciidoc/ch11.asciidoc
```diff
@@ -350,8 +350,53 @@ public static class MyConstraintMappingContributor implements ConstraintMappingC
...
+
+==== Validation target specification for purely composed constraints
+
+In case you specify a purely composed constraint - i.e. a constraint which has no validator itself but is solely made
+up from other, composing constraints - on a method declaration, the validation engine cannot determine whether that
+constraint is to be applied as a return value constraint or as a cross-parameter constraint.
+
+Hibernate Validator allows to resolve such ambiguities by specifying the +@SupportedValidationTarget+ annotation on the
+declaration of the composed constraint type as shown in <<example-purely-composed-constraint-validation-target>>.
+The +@ValidInvoiceAmount+ is does not declare any validator, but it is solely composed by the +@Min+ and +@NotNull+
```
hferentschik: no 'is'

engine/src/main/java/org/hibernate/validator/internal/metadata/descriptor/ConstraintDescriptorImpl.java
```diff
@@ -423,13 +433,14 @@ else if ( !isExecutable ) {
...
 					constraintType = ConstraintType.CROSS_PARAMETER;
 				}
-				// Now we are out of luck
-				else {
-					throw log.getImplicitConstraintTargetInAmbiguousConfigurationException( annotationType.getName() );
-				}
 			}
 		}
 
+		// Now we are out of luck
```
hferentschik: lol - so now we are out of luck ;-)

hferentschik:
Looking good, just the typo. I can fix that ...

gunnarmorling:
Cool, thanks!

hferentschik:
Merged.

PR #404, opened 2015-04-22
HV-941 Formatting for chapters 1, 2, 11

gunnarmorling:
This applies the literal formatting for chapter 11.

The second commit is an experiment for including examples directly from sources rather than copying them. I think it's the better approach in general, but converting it now takes quite some time, so I'm not sure whether it's worth the effort atm. Maybe let's better do it when adding/updating given example listings anyways?

Any thoughts?

### DIFF
--- a/documentation/src/main/asciidoc/ch01.asciidoc
+++ b/documentation/src/main/asciidoc/ch01.asciidoc
@@ -158,13 +158,13 @@ public class Car {
 ----
 ====
 
-The@NotNull, @Size and @Min annotations are used to declare the constraints which should be applied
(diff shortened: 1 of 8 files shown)


hferentschik:
Looking good. I want to wait a bit with merging though to see how much more of doc reviewing we get in. We probably won't manage a full review for the CR1, in which case we somehow need to split the JIRA issue.

hferentschik:
Great work, thanks. Some documentation source classes were missing (or better some classes from the engine were included). I added  `Marathon`, `Tournament` and `Runner` to the documentation module. Seems like Eclipse allows for all sorts of cross module imports, ts ts ts ;-)

PR #405, opened 2015-04-23
HV-978

hferentschik:


cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidationExtension.java
```diff
@@ -277,6 +272,7 @@ else if ( bean.getTypes().contains( Validator.class ) || bean instanceof Validat
 				continue;
 			}
 
+			BeanDescriptor beanDescriptor = validator.getConstraintsForClass( method.getDeclaringClass() );
```
gunnarmorling: Hey @hferentschik, I am confused, this is a bit different from what we discussed this morning, right?

I think we really must use the original method for meta-data look-up instead of the overridden one. Otherwise we'd miss the case where the super-type method is unconstrained but the sub-type method adds a return-value constraint (which is legal). We wouldn't detect the need for validation in this case when examining the meta-data for the super-type method.

gunnarmorling: I have created PR https://github.com/hibernate/hibernate-validator/pull/409 for that.

hferentschik:
Superseded by #409

PR #407, opened 2015-04-23
HV-976 Optional and wraper

DavideD:
https://hibernate.atlassian.net/browse/HV-976
https://hibernate.atlassian.net/browse/HV-977

@hferentschik 
this solution will pass all the test but I'm not sure it's the correct one.

In particular, what is the expected behaviour for his scenario (does it make sense)?:

```
@UnwrapValidatedValue(true)
@NotNull(message = "container")
Optional<@NotNullTypeUse(message = "type") String> valueWithNotNull;
```

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -84,6 +84,11 @@ import static org.hibernate.validator.internal.util.logging.Messages.MESSAGES;
  */
 public class ValidatorImpl implements Validator, ExecutableValidator {
 
+	/**
+	 * For compatibility with JDK 7 we cannot use ElementType#TYPE_USE
+	 */
+	private static final String TYPE_USE = "TYPE_USE";
+
 	private static final Log log = LoggerFactory.make();
 
 	/**
@@ -533,13 +538,47 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 			if ( !propertyPathComplete ) {
 				valueContext.appendNode( propertyMetaData );
 			}
-			// set the unwrapping mode for this validation
-			valueContext.setUnwrapMode( propertyMetaData.unwrapMode() );
+
+			if ( TYPE_USE.equals( metaConstraint.getElementType().name() ) ) {
+				// TYPE_USE constraints always require UNWRAP
+				valueContext.setUnwrapMode( UnwrapMode.UNWRAP );
+			}
+			else {
+				// set the unwrapping mode for this validation
+				valueContext.setUnwrapMode( propertyMetaData.unwrapMode() );
+			}
 		}
 		else {
 			valueContext.appendBeanNode();
 		}
 
+		validationSuccessful = validateMetaConstraint( validationContext, valueContext, metaConstraint );
+
+		// reset the unwrapping mode
+		valueContext.setUnwrapMode( UnwrapMode.AUTOMATIC );
+		return validationSuccessful;
+	}
+
+	private boolean validateTypeConstraint(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, boolean propertyPathComplete,
+			MetaConstraint<?> metaConstraint) {
+
+		boolean validationSuccessful = true;
+
+		PropertyMetaData propertyMetaData = beanMetaDataManager.getBeanMetaData( valueContext.getCurrentBeanType() ).getMetaDataFor(
+				ReflectionHelper.getPropertyName( metaConstraint.getLocation().getMember() ) );
+
+			if ( !propertyPathComplete ) {
+				valueContext.appendNode( propertyMetaData );
+			}
+
+			valueContext.setUnwrapMode( UnwrapMode.UNWRAP );
+			validationSuccessful = validateMetaConstraint( validationContext, valueContext, metaConstraint );
+			valueContext.setUnwrapMode( UnwrapMode.AUTOMATIC );
+
+		return validationSuccessful;
+	}
+
+	private boolean validateMetaConstraint(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, MetaConstraint<?> metaConstraint) {
 		if ( isValidationRequired( validationContext, valueContext, metaConstraint ) ) {
 			if ( valueContext.getCurrentBean() != null ) {
 				Object valueToValidate = getValue(
@@ -548,12 +587,9 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 				);
 				valueContext.setCurrentValidatedValue( valueToValidate );
 			}
-			validationSuccessful = metaConstraint.validateConstraint( validationContext, valueContext );
+			return metaConstraint.validateConstraint( validationContext, valueContext );
 		}
-
-		// reset the unwrapping mode
-		valueContext.setUnwrapMode( UnwrapMode.AUTOMATIC );
-		return validationSuccessful;
+		return true;
 	}
 
 	/**
@@ -753,12 +789,14 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 
 	private <T> Set<ConstraintViolation<T>> validatePropertyInContext(ValidationContext<T> context, PathImpl propertyPath, ValidationOrder validationOrder) {
 		List<MetaConstraint<?>> metaConstraints = newArrayList();
+		List<MetaConstraint<?>> typeUseConstraints = newArrayList();
 		Iterator<Path.Node> propertyIter = propertyPath.iterator();
 		ValueContext<?, Object> valueContext = collectMetaConstraintsForPath(
 				context,
 				propertyIter,
 				propertyPath,
-				metaConstraints
+				metaConstraints,
+				typeUseConstraints
 		);
 
 		if ( valueContext.getCurrentBean() == null ) {
@@ -776,7 +814,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		while ( groupIterator.hasNext() ) {
 			Group group = groupIterator.next();
 			valueContext.setCurrentGroup( group.getDefiningClass() );
-			validatePropertyForCurrentGroup( valueContext, context, metaConstraints );
+			validatePropertyForCurrentGroup( valueContext, context, metaConstraints, typeUseConstraints );
 			if ( shouldFailFast( context ) ) {
 				return context.getFailingConstraints();
 			}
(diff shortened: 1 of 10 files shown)


gunnarmorling:
> In particular, what is the expected behaviour for his scenario

On first thought I'd say both constraints should be applied to the String value:

* `@NotNull` as unwrap=true is given
* `@NotNullTypeUse` as it is given on the type use of `String`

Is that how it does behave?

DavideD:
What would you expect instead if UnwrapValidatedValue is set to false (`@UnwrapValidatedValue(false)`)?

DavideD:
I think a test case is missing for that scenario, I will add one.

gunnarmorling:
> What would you expect instead if UnwrapValidatedValue is set to false (@UnwrapValidatedValue(false))? 

Then I'd expect `@NotNull` to be applied to the `Optional` and `@NotNullTypeUse` to be applied to the wrapped String.

engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalTypeAnnotationConstraintTest.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+import static java.lang.annotation.ElementType.*;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import static org.hibernate.validator.testutil.ConstraintViolationAssert.*;
+import static org.hibernate.validator.testutil.ValidatorUtil.getValidator;
+
+/**
+ * @author Lukas Niemeier
```
gunnarmorling: Can we include this? Do we have a CLA from Lukas?

engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalTypeAnnotationConstraintTest.java
```diff
@@ -0,0 +1,216 @@
...
+		model.valueWithoutTypeAnnotation = Optional.empty();
+
+		Set<ConstraintViolation<Model>> constraintViolations = validator.validateProperty(
+				model,
+				"valueWithoutTypeAnnotation"
+		);
+		assertNumberOfViolations( constraintViolations, 0 );
+	}
+
+	/**
```
gunnarmorling: I guess the "Actual..." comments can go away now.

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -262,6 +256,29 @@ private ConstrainedField findPropertyMetaData(Field field) {
 		);
 	}
 
+	private UnwrapMode unwrapMode(Field field, boolean typeArgumentAnnotated) {
```
gunnarmorling: Shouldn't something similar also be done for (getter-defined) properties and method-level constraints?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -530,11 +532,19 @@ private boolean validateConstraint(ValidationContext<?> validationContext,
 				ReflectionHelper.getPropertyName( metaConstraint.getLocation().getMember() )
 			);
 
+			// The user didn't set un
```
gunnarmorling: Seems like something should be following to "un"?

engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalTypeAnnotationConstraintTest.java
DavideD: I don't know, we need to ask him, but these are tests he wrote so I feel bad not to mention him

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
DavideD: Mmmh , make sense, I will add the tests

DavideD:
> What would you expect instead if UnwrapValidatedValue is set to false (@UnwrapValidatedValue(false))?
> 
> Then I'd expect @NotNull to be applied to the Optional and @NotNullTypeUse to be applied to the wrapped String.

Should I validate the wrapped element if the validation of the wrapper fails? I'm prone to say no.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
DavideD: Right, while I was writing this comment I had some doubts about the overall behaviour and I've sent the pull request forgetting to remove it.

I will add commits to clean the various left-overs

engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalTypeAnnotationConstraintTest.java
```diff
@@ -0,0 +1,216 @@
...
+		assertCorrectPropertyPaths( constraintViolations, "valueWithNotNull" );
+		assertConstraintViolation( constraintViolations.iterator().next(), "container" );
+		assertCorrectConstraintTypes( constraintViolations, NotNull.class );
+	}
+
+	/**
+	 * Actual: NotNull violation occurs on container's annotation
+	 */
+	@Test
+	public void not_null_type_on_optional_is_validated_for_empty_value() {
```
gunnarmorling: This one makes me curious.

The property is annotated with `@UnwrapValidatedValue(false)`, so my personal expectation would be that the type parameter constraint is not applied at all (as it can only apply to the wrapped value, not the wrapper). That's different from the assertion in the test. I think I understand the reasoning between the expectation here, but it seems a bit incosistent to me.

@hferentschik, any thoughts?

engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalTypeAnnotationConstraintTest.java
```diff
@@ -0,0 +1,216 @@
...
+				model,
+				"valueWithNullOrNotBlank"
+		);
+		assertNumberOfViolations( constraintViolations, 1 );
+		assertCorrectPropertyPaths( constraintViolations, "valueWithNullOrNotBlank" );
+		assertConstraintViolation( constraintViolations.iterator().next(), "type" );
+		assertCorrectConstraintTypes( constraintViolations, NullOrNotBlank.class );
+	}
+
+	@Test
```
gunnarmorling: Are the last three actually relevant for the issue at hand?

DavideD: as far as I understand the issue is about to test the ability of validator to validate the wrapper object and the wrapped object separately.

This test is checking that the validation fails because of the constraint on the type and not the one on the wrapper.

There is another one on the same field that checks that when the container is null, the test fails only for the validation of the container `null_or_not_blank_type_on_optional_is_validated_for_null_value`.

engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalTypeAnnotationConstraintTest.java
DavideD: I think the idea is to be able to validate the wrapper and the wrapped object separately.

By using `@UnwrapValidatedValue(false)`  it saying that the annotation on the property should be used to validate the wrapper.
The validation on the type are used to validate the wrapped object.
In this context, if you set the constraint on the type you always need to unwrap it.

As far as I can see for JDK 8 you don't need the annotation `@UnwrapValidatedValue` but I guess we need to be consistent with the behaviour for JDK 7.

It seems consistent to me, but I don't have much experience with validator.

hferentschik:
>> In particular, what is the expected behaviour for his scenario

> On first thought I'd say both constraints should be applied to the String value

+1

hferentschik:
>> What would you expect instead if UnwrapValidatedValue is set to false (@UnwrapValidatedValue(false))?

> Then I'd expect @NotNull to be applied to the Optional and @NotNullTypeUse to be applied to the wrapped String.

Initially I was tempted to say the same, but I see a certain conflict here. I guess one could argue that the type level constraint implies unwrapping, so setting UnwrapValidatedValue to false is either illegal configuration or one ignored the type level constraint and just apply `@NotNull`

engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalTypeAnnotationConstraintTest.java
hferentschik: Since he provided the code on JIRA without any authorship I think it would be safer to remove it. In some form he still has credit via the issue itself. This avoids any problems.

DavideD: Ok, no problem

DavideD:
I have updated the tests and the patch should now work.

I had to change `UnwrapModesTestUnwrapModesTest`, I think it was wrong but I'm not sure, basically the `ValueHandlerConstraintValidator` was receiving as value to validate a `ValueHolder` for both the cases when we use an unwrapper and when we don't. I think that when we use the unwrapper we should validate the value unwrapped value and therfore in that case we should receive an Integer. Because the same validator is used in both test I've changed the type to Object.
This required me to add an `UnwrapValidatedValue ` to solve an exception.

I also addded the tests for method and constructurs.

hferentschik:
thanks @DavideD - I am taking a look in a bit ...

DavideD:
@hferentschik I think I fixed all the issue

One note:
at the moment the wrapped value gets validated even if the validation of the container fails.

I couldn't do it for methods and constructor and so, for consistency, I did the same for properties. I guess we can open a separate JIRA for this if it is a requirement.

hferentschik:
@DavideD, looking good. Unfortunately, I noticed a new bug related bug (see comment in `ConstraintValidatorManager`.

hferentschik:
Superseded by #413. 

#413 contains the changes of this pull request + refactoring of `ConstraintValidatorManager` to not modify current `ValueContext`.

PR #408, opened 2015-04-23
HV-980 Exposing value on property nodes of constraint violation paths

gunnarmorling:
@hferentschik If you have a minute, can you take a look at that one? This provides a new HV-specific node interface `PropertyNode` with a method `getValue()` returning the value of the represented property. I.e. it's not only for sets but for all property nodes. If you agree with the general solution, I'll push another commit for the doc update. Thx!

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
@@ -31,7 +31,7 @@ import org.hibernate.validator.internal.util.logging.LoggerFactory;
  * @author Gunnar Morling
  */
 public class NodeImpl
-		implements Path.PropertyNode, Path.MethodNode, Path.ConstructorNode, Path.BeanNode, Path.ParameterNode, Path.ReturnValueNode, Path.CrossParameterNode, Serializable {
+		implements Path.PropertyNode, Path.MethodNode, Path.ConstructorNode, Path.BeanNode, Path.ParameterNode, Path.ReturnValueNode, Path.CrossParameterNode, org.hibernate.validator.path.PropertyNode, Serializable {
 	private static final long serialVersionUID = 2075466571633860499L;
 	private static final Class<?>[] EMPTY_CLASS_ARRAY = new Class<?>[]{};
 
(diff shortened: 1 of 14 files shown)


gunnarmorling:
Pushed one more commit for the doc update.

engine/src/main/java/org/hibernate/validator/path/PropertyNode.java
```diff
@@ -0,0 +1,20 @@
...
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.path;
+
+/**
+ * Node representing a property, providing Hibernate Validator specific functionality.
+ *
+ * @author Gunnar Morling
+ */
+public interface PropertyNode extends javax.validation.Path.PropertyNode {
```
hferentschik: nice to create the public class for this (compared to just exposing something internal). I think we get really good at this.

engine/src/test/java/org/hibernate/validator/test/internal/engine/path/NodeImplTest.java
```diff
@@ -181,6 +189,95 @@ public void testToMapCascadeValidate() {
...
+		assertNumberOfViolations( constraintViolations, 1 );
+		assertCorrectPropertyPaths( constraintViolations, "apartments[].resident.name" );
+
+		Path path = constraintViolations.iterator().next().getPropertyPath();
+		Iterator<Path.Node> nodeIterator = path.iterator();
+
+		Path.Node node = nodeIterator.next();
+		assertEquals( node.getKind(), ElementKind.PROPERTY, "unexpected node kind" );
+		assertEquals( node.getName(), "apartments" );
+		assertEquals( node.as( PropertyNode.class ).getValue(), new Apartment( new Person( "Bob" ) ) );
```
hferentschik: I think I would have created the apartment up at the top and added it to the building and used it for the assertion here. But that's a matter of taste manybe.

engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
```diff
@@ -78,6 +80,7 @@ public static NodeImpl createPropertyNode(String name, NodeImpl parent) {
 				null,
 				ElementKind.PROPERTY,
 				EMPTY_CLASS_ARRAY,
+				null,
```
hferentschik: gets a bit nasty with all this parameters which gets set to 'null' depending on the use-case. Not sure how one could improve that!?

hferentschik:
Nice. This really grows on me. Might be even something for BV 1.2.

hferentschik:
Merged.

PR #409, opened 2015-04-23
HV-978

gunnarmorling:


gunnarmorling:
@hferentschik That's based on #405 and takes the top-most method for retrieval of `@ValidateOnExecution` and the actual method for meta-data retrieval.

hferentschik:
Merged, thanks. I think it is clearer now. See how many more border cased will come up over time. Not the first time we "patch" this class :-)

gunnarmorling:
Yes; It's a bit like good wine, it gets better and better as time passes by :)

PR #410, opened 2015-04-24
HV-846 Adding OSGi metadata to cdi module

hferentschik:
- Upgrading maven bundle plugin
- Adding Imports and Exports for CDI module in maven bundle plugin
- Making CDI bundle a fragment of the main Validator bundle

### DIFF
--- a/cdi/pom.xml
+++ b/cdi/pom.xml
@@ -149,6 +149,32 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>org.apache.felix</groupId>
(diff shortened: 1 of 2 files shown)


hferentschik:
@gunnarmorling wdyt? Not sure whether it really works, but it should not hurt, right?

gunnarmorling:
@hferentschik Yes, it should be fine. Just merged. it. Btw. I noted that the OSGi integration takes quite some time (also before this change). Do you observe the same?

PR #413, opened 2015-04-29
HV-976 & HV-977

hferentschik:


engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalTypeAnnotationConstraintOnConstructorTest.java
```diff
@@ -0,0 +1,255 @@
...
+	public void not_null_type_on_optional_is_validated_for_null_value() throws Exception {
+		Constructor<ModelB> constructor = ModelB.class.getDeclaredConstructor( Optional.class );
+		Object[] values = new Object[] { null };
+		Set<ConstraintViolation<ModelB>> constraintViolations = validator
+				.forExecutables()
+				.validateConstructorParameters(constructor, values);
+
+		assertNumberOfViolations( constraintViolations, 2 );
+		assertCorrectPropertyPaths( constraintViolations, "ModelB.arg0", "ModelB.arg0" );
+		// TODO: Should we validate the type since the container already failed?
```
gunnarmorling: Seems like this TODO is obsolete now (there a similar ones in the other tests)?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -533,13 +538,47 @@ private boolean validateConstraint(ValidationContext<?> validationContext,
...
 		}
 
+		validationSuccessful = validateMetaConstraint( validationContext, valueContext, metaConstraint );
+
+		// reset the unwrapping mode
+		valueContext.setUnwrapMode( UnwrapMode.AUTOMATIC );
+		return validationSuccessful;
+	}
+
+	private boolean validateTypeConstraint(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, boolean propertyPathComplete,
```
gunnarmorling: Why is it that `validateTypeConstraint()` only is invoked for property validation but not for parameter/return value validation?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -956,8 +1006,18 @@ private int validatePropertyForNonDefaultGroup(ValueContext<?, Object> valueCont
 									.size() - numberOfConstraintViolationsBefore;
 						}
 					}
+
+					if ( typeUseConstraints.contains( metaConstraint ) ) {
```
gunnarmorling: "else if" should save a cycle.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -956,8 +1006,18 @@ private int validatePropertyForNonDefaultGroup(ValueContext<?, Object> valueCont
...
+						validationSuccessful = validationSuccessful && tmp;
+						if ( shouldFailFast( validationContext ) ) {
+							return validationContext.getFailingConstraints()
+									.size() - numberOfConstraintViolationsBefore;
+						}
+					}
 				}
-				if ( !validationSuccessful ) {
+
+				if (!validationSuccessful) {
```
gunnarmorling: Interesting, is it even in accordance with the style guide? I assumed there'd always be empty spaces in braces?

engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalTypeAnnotationConstraintOnConstructorTest.java
hferentschik: +1 removing it

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -240,16 +240,10 @@ private ConstrainedField findPropertyMetaData(Field field) {
...
-			unwrapMode = unwrapValidatedValue.value() ? UnwrapMode.UNWRAP : UnwrapMode.SKIP_UNWRAP;
-		}
-
 		Set<MetaConstraint<?>> typeArgumentsConstraints = findTypeAnnotationConstraintsForMember( field );
-		if ( !typeArgumentsConstraints.isEmpty() && !ReflectionHelper.isIterable( ReflectionHelper.typeOf( field ) ) ) {
-			unwrapMode = UnwrapMode.UNWRAP;
-		}
+
+		boolean typeArgumentAnnotated = !typeArgumentsConstraints.isEmpty();
+		UnwrapMode unwrapMode = unwrapMode( field, typeArgumentAnnotated );
```
gunnarmorling: I am not sure I understand what's going on here. What is the meaning of the "unwrap mode" for a field exactly? Is it the unwrap setting for all the non-typeuse constraints? If so, I am wondering why `typeArgumentAnnotated` is part of the calculation.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -144,10 +144,8 @@
 	ValidationException getConstraintFactoryMustNotReturnNullException(String validatorClassName);
 
 	@Message(id = 30,
-			value = "No validator could be found for constraint '%s' validating type '%s'. Check configuration for '%s'")
-	UnexpectedTypeException getNoValidatorFoundForTypeException(String constraintType,
-			String validatedValueType,
-			String path);
+			value = "No validator could be found for constraint '%s' validating type '%s'.")
+	UnexpectedTypeException getNoValidatorFoundForTypeException(String constraintType, String validatedValueType);
```
gunnarmorling: The path seemed a useful bit of information. Can't we provide it any longer?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
hferentschik: Hmm , good question. Obviously the executable validation is not fully unified with the other validate methods. There is also `OptionalTypeAnnotationConstraintOnMethodTest` which tests parameter validation. 

Return value validation could need another test ...

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
hferentschik: Interesting indeed. AFAICT checkstyle did not complain

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
hferentschik: Actually we can. I will push an update

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
hferentschik: `UnwrapMode` tells your whether or not the field should be unwrapped (basically the value of `@UnwrapValidatedValue`). The change will enforce unwrapping in case there are type use constraints. Note, that we basically "roll" type use constraints into the field metadata. The exact unwrapping behavior will be determined at validation time.

gunnarmorling: > UnwrapMode tells your whether or not the field should be unwrapped

But that's my point, this isn't a "per-field" setting, is it? AFAIU, `@UnwrapValidatedValue` only applies to non-typeuse constraints, whereas for typeuse constraints we always unwrap.

hferentschik: true, true - I am not arguing against this. I think we even had the same discussion when Khalid was working on this. In the end we would need to refactor the metadata a bit. AFAICR we are using the field metadata to host this configuration value, because the annotation is hosted on the field (property). 

I am all for re-visiting the metadata. There are several things which could/should be improved. Just not sure whether it is necessary right now. At least not as part of this issue I think.

Does this make sense?

gunnarmorling: No need to change it right now, just wanted to make sure we are on the same page :) Probably a good idea to capture this in a JIRA issue.

hferentschik: We are on the same page. Partly this goes under the long deferred "unification" of the validation process. I'll create an issue.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
hferentschik: you were right, something is missing here. Looking into it. An update will follow ...

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
hferentschik: https://hibernate.atlassian.net/browse/HV-988

hferentschik:
@gunnarmorling, did you notice that I did push an update. Return value validation is considered now as well. Apart from the fact that the metadata design needs an overhaul, I think this issue is at least resolved,

hferentschik:
thanks

PR #419, opened 2015-06-19
HV-999 Upgrading to TCK 1.1.4.Final; Adding tests removed from the TCK to HV's own tests

gunnarmorling:


hferentschik:
merged thanks.

PR #422, opened 2015-07-20
HV-1005 Take recommendend length as per RFC for local and domain part…

hferentschik:
… into consideration

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
@@ -31,9 +31,11 @@ import static java.util.regex.Pattern.CASE_INSENSITIVE;
  * @author Hardy Ferentschik
  */
 public class EmailValidator implements ConstraintValidator<Email, CharSequence> {
-	private static String ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~-]";
-	private static String DOMAIN = ATOM + "+(\\." + ATOM + "+)*";
-	private static String IP_DOMAIN = "\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\]";
+	private static final String ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~-]";
+	private static final String DOMAIN = ATOM + "+(\\." + ATOM + "+)*";
+	private static final String IP_DOMAIN = "\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\]";
+	private static final int MAX_LOCAL_PART_LENGTH = 64;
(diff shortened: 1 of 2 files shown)


hferentschik:
@gunnarmorling, could you have a look at this and let me know what you think?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
```diff
@@ -31,9 +31,11 @@
  * @author Hardy Ferentschik
  */
 public class EmailValidator implements ConstraintValidator<Email, CharSequence> {
-	private static String ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~-]";
-	private static String DOMAIN = ATOM + "+(\\." + ATOM + "+)*";
-	private static String IP_DOMAIN = "\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\]";
+	private static final String ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~-]";
+	private static final String DOMAIN = ATOM + "+(\\." + ATOM + "+)*";
+	private static final String IP_DOMAIN = "\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\]";
```
gunnarmorling: So there are no changes to the actual expressions, right? Only the "final" has been added?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
```diff
@@ -69,29 +71,50 @@ public boolean isValid(CharSequence value, ConstraintValidatorContext context) {
...
+
+		Matcher matcher = pattern.matcher( asciiString );
 		return matcher.matches();
 	}
+
+	private String toAscii(String unicodeString) throws IllegalArgumentException {
+		String asciiString = "";
+		int start = 0;
+		int end = unicodeString.length() <= 63 ? unicodeString.length() : 63;
+		while ( true ) {
```
gunnarmorling: Looks alright, though I find the version from the [second answer here](http://stackoverflow.com/questions/3760152/split-string-to-equal-length-substrings-in-java) slightly more digestable.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
hferentschik: correct

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
hferentschik: Interesting solution indeed. Still a bit "tricky", since it uses advanced regexp features. I'll keep it in mind though. Might come in handy one day.

PR #423, opened 2015-07-24
HV-1007 Using XMLReaderEvent#peek to look ahead for schema version in…

hferentschik:
…stead of mark&reset of stream

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/xml/XmlParserHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/xml/XmlParserHelper.java
@@ -14,7 +14,6 @@ import java.security.PrivilegedAction;
 import java.security.PrivilegedExceptionAction;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLEventReader;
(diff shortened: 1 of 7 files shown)


engine/src/main/java/org/hibernate/validator/internal/xml/XmlParserHelper.java
```diff
@@ -72,21 +71,15 @@
 	 * represented by the "version" attribute of the root element of the stream.
 	 *
 	 * @param resourceName The name of the represented XML resource.
```
gunnarmorling: The `READ_LIMIT` constant is not needed anymore.

gunnarmorling:
Merged; I first was concerned about re-using the same reader for getting the version and subsequent unmarshalling, but it works fine as unmarshalling works fine from the start element retrieved for getting the version. Neat trick :)

Added one more commit for removing the unused constant. I couldn't test on IBM JDK but I assume you did so?

hferentschik:
> Added one more commit for removing the unused constant.

nice

 > I couldn't test on IBM JDK but I assume you did so?

I did

engine/src/main/java/org/hibernate/validator/internal/xml/XmlParserHelper.java
hferentschik: good catch

PR #425, opened 2015-09-03
HV-1011: Allow executable validation by passing overridden super-type method

gunnarmorling:
@emmanuelbernard Seems like you are the only one left to merge it ;) Thx!


gunnarmorling:
Pushed one more commit for the suggested test. The existing overriding check takes care of that already.

emmanuelbernard:
Applied thanks. I attached it to version 5.2.x in JIRA. Not sure that's correct.

PR #429, opened 2015-09-18
HV-1017 javafx detection uses TCCL but JavaFXPropertyValueUnwrapper does not

DavideD:
https://hibernate.atlassian.net/browse/HV-1017

I don't know how to test the issue, so I just check that the TCCL is not used when looking for the JavaFX classes. Let me know if it makes sense.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
@@ -32,7 +32,6 @@ import org.hibernate.validator.cfg.ConstraintMapping;
 import org.hibernate.validator.internal.cfg.context.DefaultConstraintMapping;
 import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorFactoryImpl;
 import org.hibernate.validator.internal.engine.resolver.DefaultTraversableResolver;
-import org.hibernate.validator.internal.engine.valuehandling.JavaFXPropertyValueUnwrapper;
 import org.hibernate.validator.internal.engine.valuehandling.OptionalValueUnwrapper;
 import org.hibernate.validator.internal.util.CollectionHelper;
 import org.hibernate.validator.internal.util.Contracts;
@@ -63,6 +62,8 @@ import static org.hibernate.validator.internal.util.logging.Messages.MESSAGES;
  * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
  */
 public class ConfigurationImpl implements HibernateValidatorConfiguration, ConfigurationState {
+
+	private static final String JFX_UNWRAPPER_CLASS = "org.hibernate.validator.internal.engine.valuehandling.JavaFXPropertyValueUnwrapper";
 
 	static {
 		Version.touch();
@@ -114,7 +115,7 @@ public class ConfigurationImpl implements HibernateValidatorConfiguration, Confi
 		this.validationBootstrapParameters = new ValidationBootstrapParameters();
 		TypeResolutionHelper typeResolutionHelper = new TypeResolutionHelper();
 		if ( isJavaFxInClasspath() ) {
-			validatedValueHandlers.add( new JavaFXPropertyValueUnwrapper( typeResolutionHelper ) );
+			validatedValueHandlers.add( createJavaFXUnwrapperClass( typeResolutionHelper ) );
 		}
 		if ( Version.getJavaRelease() >= 8 ) {
 			validatedValueHandlers.add( new OptionalValueUnwrapper( typeResolutionHelper ) );
@@ -130,6 +131,16 @@ public class ConfigurationImpl implements HibernateValidatorConfiguration, Confi
 				typeResolutionHelper
 		);
 		this.addConstraintDefinitionContributor( defaultConstraintDefinitionContributor );
+	}
+
+	private ValidatedValueUnwrapper<?> createJavaFXUnwrapperClass(TypeResolutionHelper typeResolutionHelper) {
+		try {
+			Class<?> jfxUnwrapperClass = run( LoadClass.action( JFX_UNWRAPPER_CLASS, getClass().getClassLoader() ) );
+			return (ValidatedValueUnwrapper<?>) ( jfxUnwrapperClass.getConstructor( TypeResolutionHelper.class ).newInstance( typeResolutionHelper ) );
+		}
+		catch (Exception e) {
+			throw log.validatedValueUnwrapperCannotBeCreated( JFX_UNWRAPPER_CLASS, e );
+		}
 	}
 
 	@Override
--- a/engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
@@ -635,4 +635,7 @@ public interface Log extends BasicLogger {
 
 	@Message(id = 190, value = "Unable to parse %s.")
 	ValidationException getUnableToCreateXMLEventReader(String file, @Cause Exception e);
+
+	@Message(id = 191, value = "Error creating unwrapper: %s")
+	ValidationException validatedValueUnwrapperCannotBeCreated(String className, @Cause Exception e);
 }


gunnarmorling:
An alternative would be to load the unwrapper class via the same loader
used for checking its presence, i.e. reflectively.

emmanuelbernard:
@DavideD asks for the user feedback (to check that the bug is indeed fixed). Let's give it some time before applying.

DavideD:
@emmanuelbernard Done:

emmanuelbernard:
I know you did, it was more as a marker on the PR. Also I messed up my English tenses ;)

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/LoadClass.java
```diff
@@ -36,13 +36,23 @@
 
 	private final ClassLoader classLoader;
 
+	/**
+	 * when true, it will check the Thread COntext ClassLoader when the class is not found in the provided one
```
hferentschik: _Context_

hferentschik:
> An alternative would be to load the unwrapper class via the same loader used for checking its presence, i.e. reflectively.

I think I like this idea

gunnarmorling:
> I think I like this idea

Ok :) I don't have a strong preference, it would help in some cases, but I suppose the scenario is very exotic. @DavideD?

DavideD:
> ... I suppose the scenario is very exotic. 

I don't know, this kind of errors seem to happen more often lately. In particular, when people start to use OSGI or JBoss Module on WildFly.

Anyway, I can try the classloader approach.

gunnarmorling:
Ok, cool.

> In particular, when people start to use OSGI or JBoss Module on WildFly.

But even then, why should JavaFX classes be visible trough the TCCL but not
the system class loader? It seems like a misconfiguration of that
environment (OSGi runtime with incomplete system package exports). Or
possibly a very old JVM which does not bring JavaFX itself but the user
adds it manually. Anyways, +1 for doing the alternative approach :)

hferentschik:
> But even then, why should JavaFX classes be visible trough the TCCL but not the system class loader?

I guess that's my question here as well. I am not an JAvaFX expert, but I was wondering whether it would be possible to add JavaFX classes to a JVM which otherwise does not provide them. Either way, using the reflection approach seems to be the safer option in this case, covering more potential scenarios.

gunnarmorling:
> whether it would be possible to add JavaFX classes to a JVM which
otherwise does not provide them

Yes, that would have been the case before JFX was added to the JDK (in 1.7
Update 6).

DavideD:
@gunnarmorling 
I've rebased with the change you suggested. although I'm not sure if the way I did it is the one you have suggested. Could you have a look?

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -113,8 +112,10 @@ public ConfigurationImpl(ValidationProvider<?> provider) {
 	private ConfigurationImpl() {
 		this.validationBootstrapParameters = new ValidationBootstrapParameters();
 		TypeResolutionHelper typeResolutionHelper = new TypeResolutionHelper();
-		if ( isJavaFxInClasspath() ) {
-			validatedValueHandlers.add( new JavaFXPropertyValueUnwrapper( typeResolutionHelper ) );
+		ClassLoader classLoader = getClass().getClassLoader();
```
hferentschik: I don't think the intention was to change the `isJavaFxInClasspath` in classpath. In my world it would e kept as is. Then, if it returns `true`, use the `LoadClass` privileged action (which is used under the hood for `isJavaFxInClasspath`) to load `JavaFXPropertyValueUnwrapper`.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -113,8 +112,10 @@ public ConfigurationImpl(ValidationProvider<?> provider) {
...
+		if ( isJavaFxInClasspath( classLoader ) ) {
+			JavaFXPropertyValueUnwrapper jfxUnwrapper = createJavaFXUnwrapperClass( typeResolutionHelper, classLoader );
```
hferentschik: Would we not try to avoid the explicit use of `JavaFXPropertyValueUnwrapper`? Just use `ValidatedValueUnwrapper`!?

DavideD: Right, make sense

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -132,6 +133,16 @@ private ConfigurationImpl() {
 		this.addConstraintDefinitionContributor( defaultConstraintDefinitionContributor );
 	}
 
+	private JavaFXPropertyValueUnwrapper createJavaFXUnwrapperClass(TypeResolutionHelper typeResolutionHelper, ClassLoader classLoader) {
+		try {
+			Class<?> jfxUnwrapperClass = classLoader.loadClass( "org.hibernate.validator.internal.engine.valuehandling.JavaFXPropertyValueUnwrapper" );
```
hferentschik: All classloading has to be done via a `PrivilegedAction`. We cannot call `ClassLoader.loadClass` directly.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -483,16 +494,23 @@ private void applyXmlSettings(ValidationBootstrapParameters xmlParameters) {
...
 	}
 
-	private boolean isJavaFxInClasspath() {
-		return isClassPresent( "javafx.application.Application" );
-	}
-
-	private boolean isClassPresent(String className) {
+	private boolean isJavaFxInClasspath(final ClassLoader classLoader) {
 		try {
-			run( LoadClass.action( className, getClass().getClassLoader() ) );
```
hferentschik: I think you can keep the current `LoadClass` (see comment above), but if you really need a new version which for example does not check the TCCL, then you should either modify `LoadClass` or create a new `LoadClassXYZ`.

DavideD: I will keep  it, I didn't realize that it returns a value

DavideD:
@hferentschik Updated

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -132,6 +133,16 @@ private ConfigurationImpl() {
 		this.addConstraintDefinitionContributor( defaultConstraintDefinitionContributor );
 	}
 
+	private ValidatedValueUnwrapper<?> createJavaFXUnwrapperClass(TypeResolutionHelper typeResolutionHelper) {
```
hferentschik: +1 Right, that's what I had in mind

hferentschik:
I would suggest to merge this in this state and deploy a SNAPSHOT the user can test. If it still fails we can re-iterate on this.

hferentschik:
Rebased and merged

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -132,6 +133,16 @@ private ConfigurationImpl() {
...
+		try {
+			Class<?> jfxUnwrapperClass = run( LoadClass.action( JFX_UNWRAPPER_CLASS, getClass().getClassLoader() ) );
+			return (ValidatedValueUnwrapper<?>) ( jfxUnwrapperClass.getConstructor( TypeResolutionHelper.class ).newInstance( typeResolutionHelper ) );
```
gunnarmorling: Just a quick thought: If `JFX_UNWRAPPER_CLASS` is loaded through the TCCL, will this actually find the right constructor, passing `TypeResolutionHelper` obtained through this's class loader?

hferentschik: `org.hibernate.validator` itself will never be loaded via the TCCL, since it is in the `org.hibernate.validator` namespace. This is handled in `LoadClass`.

hferentschik: hmm, so that might not work!? hard to test really

hferentschik: The current code might just behave like the old code. One would need 
1. something like a `ClassFoundIn` which instead of a `Class` as in `LoadClass` returns a classloader (the one which successfully loaded the class or null)
2. this classloader should then be used to load the unwrapper, but it needs a modified version of `LoadClass` to prevent the special handling of the loading of validator classes

Given 1 +  2, there might then still be the question around the `TypeResoltionHelper` @gunnarmorling mentioned.

gunnarmorling: > org.hibernate.validator itself will never be loaded via the TCCL

`LoadClass#loadClassInValidatorNameSpace()` falls back to the TCCL if a class cannot be loaded through its own loader. Now whether that makes sense...

Tbh. the original fix of only loading the JavaFX unwrapper when its visible to HV's loader seems like the best solution to me. Everything else seems very complex for a tiny potential gain.

DavideD: I'm thinking the same, Let's ue the original fix and hear back from the person with the issue, if he still has problem we will reopen the issue.

I'll send another pull request

hferentschik: > I'll send another pull request

+1

PR #430, opened 2015-09-21
HV-994 Validation of parameter on stateless bean

DavideD:
https://hibernate.atlassian.net/browse/HV-994

It seems the `@WithAnnotation` used with `ValidationExtension#processAnnotatedType` does not check constraint defined on interfaces. If remove it, everything works as expected.

### DIFF
--- a/cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidationExtension.java
+++ b/cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidationExtension.java
@@ -30,12 +30,9 @@ import javax.enterprise.inject.spi.BeforeBeanDiscovery;
 import javax.enterprise.inject.spi.Extension;
 import javax.enterprise.inject.spi.ProcessAnnotatedType;
 import javax.enterprise.inject.spi.ProcessBean;
-import javax.enterprise.inject.spi.WithAnnotations;
 import javax.enterprise.util.AnnotationLiteral;
 import javax.validation.BootstrapConfiguration;
 import javax.validation.Configuration;
-import javax.validation.Constraint;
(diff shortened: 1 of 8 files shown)


cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidationExtension.java
```diff
@@ -212,11 +209,7 @@ else if ( bean.getTypes().contains( Validator.class ) || bean instanceof Validat
 	 * @param processAnnotatedTypeEvent event fired for each annotated type
 	 * @param <T> the annotated type
 	 */
-	public <T> void processAnnotatedType(@Observes @WithAnnotations({
```
hferentschik: I think this might go a bit too far. Using `@WithAnnotations` is intentional here to limit the number of classes to check for automatic method validation. By removing it every annotated type will be processed and checked for constraints. I think we (Bean Validation) even were one of the driving forces on making the behavior of `@WithAnnotations` recursive. 

I think we need to dig deeper here and reach out to the Weld guys as well. It might be a bug in Weld and should be fixed there.

gunnarmorling:
@DavideD Where are we with that? Martin mentions "implicit bean archives" in his answer to your question in the Weld forum. Is the archive at hand such implicit archive, though? Would it help to add beans.xml? I am wondering whether we are on the right track with this one.

DavideD:
I've created a test case but I could not find a solution. It seems related
to weld but I haven't received any answer from their forum.

I've tried different configurations but if you want to test something, it's
easy to change the deployment in the test and see if it works

hferentschik:
Closing for now. Seems to need further investigation.

PR #431, opened 2015-09-21
HV-862 ExecutableHelper#overrides() should take visibilities into acc…

DavideD:
…ount

  This is already working. The commit adds a test.

https://hibernate.atlassian.net/browse/HV-862

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
@@ -103,6 +103,11 @@ public final class ExecutableHelper {
 			return false;
 		}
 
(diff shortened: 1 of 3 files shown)


DavideD:
Is there some other use case to consider?

gunnarmorling:
No, I think that's the one.

gunnarmorling:
Thinking about it, one edge case remains: A super-type declaring a package-private method (default visibility), and a sub-type in another package specifies a method with the identical signature. This is no method overriding, but I suppose the current code would not handle this correctly.

DavideD:
I'll have a look

DavideD:
I added a commit to test the package-private use case

gunnarmorling:
Rebased and applied. Thanks, @DavideD!

PR #432, opened 2015-09-22
HV-807 Allowing to ignore constraint annotations for methods/construc…

gunnarmorling:
…tors via API

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/cfg/context/AnnotationProcessingOptions.java
+++ b/engine/src/main/java/org/hibernate/validator/cfg/context/AnnotationProcessingOptions.java
@@ -13,10 +13,13 @@ package org.hibernate.validator.cfg.context;
  * @author Hardy Ferentschik
  */
 public interface AnnotationProcessingOptions<C extends AnnotationProcessingOptions<C>> {
+
 	/**
 	 * Specifies that annotations specified on the configured type or property should be ignored.
 	 *
 	 * @return Returns itself for method chaining.
+	 * @deprecated Use {@link AnnotationIgnoreOptions#ignoreAnnotations(boolean)} instead.
 	 */
+	@Deprecated
 	C ignoreAnnotations();
 }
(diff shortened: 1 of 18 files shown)


hferentschik:
I'll take a look at this in a bit ...

engine/src/main/java/org/hibernate/validator/cfg/context/AnnotationProcessingOptions.java
```diff
@@ -13,10 +13,13 @@
...
  */
 public interface AnnotationProcessingOptions<C extends AnnotationProcessingOptions<C>> {
+
 	/**
 	 * Specifies that annotations specified on the configured type or property should be ignored.
 	 *
 	 * @return Returns itself for method chaining.
+	 * @deprecated Use {@link AnnotationIgnoreOptions#ignoreAnnotations(boolean)} instead.
 	 */
+	@Deprecated
```
hferentschik: why the deprecation and rename?

gunnarmorling: Because you cannot "turn off" the ignore-annotations setting inherited from elements up in the hierarchy as you could with XML. I created that new interface to leave the current hierarchy unaltered while not forcing all the newly supported elements (methods, parameters etc.) to implement the deprecated method.

hferentschik: Could you explain a bit more? An example?

gunnarmorling: Yes, check out e.g. the new test `testIgnoreAnnotationsOnMethodAndParameter()`. In XML, "ignore-annotations" can be given at several levels in the tree, e.g. being set to true on the entity level, to false for a specific method and to true for a specific parameter of that method. The API seeks to resemble that, so `ignoreAnnotations()` doesn't quite cut it, as you need to be able to set it to false, too.

hferentschik: Ahh, now I get it. I did not notice/reflect on the boolean parameter.

hferentschik:
rebased and applied. thanks

PR #435, opened 2015-09-24
HV-1006 Using correct Arquillian container artifact for in-container …

gunnarmorling:
…TCK profile

Small follow-up.

### DIFF
--- a/tck-runner/pom.xml
+++ b/tck-runner/pom.xml
@@ -246,9 +246,8 @@
             </properties>
             <dependencies>
                 <dependency>
-                    <groupId>org.wildfly</groupId>
+                    <groupId>org.wildfly.arquillian</groupId>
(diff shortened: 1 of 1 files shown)


hferentschik:
good catch

hferentschik:
btw, how did you find out the new GAV coordinates for Arquillian. I find this to be a pain all the time. Things just keep changing.

gunnarmorling:
> btw, how did you find out the new GAV coordinates for Arquillian. I find
this to be a pain all the time. Things just keep changing.

The old artifact id wasn't available in version 10.0.0.CR1. Then I checked
what WF is referencing instead.

Yes, it's a pain, really hard to track. I often think we'd provide from a
BOM provided by WF with all its versions, so we can make sure that we use
the same ones. I manually compared some dependencies, that's how I updated
the Arquillian Core and Weld versions.

PR #436, opened 2015-09-24
HV-1017 Avoid checking the TCCL for JavaFX

DavideD:
https://hibernate.atlassian.net/browse/HV-1017

The first patch I've submitted

### DIFF
--- a/engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/JavaFXClassLoadingTest.java
+++ b/engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/JavaFXClassLoadingTest.java
@@ -1,0 +1,64 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.test.internal.engine.valuehandling;
+
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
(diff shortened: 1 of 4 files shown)


hferentschik:
:-)

engine-jdk8-tests/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/JavaFXClassLoadingTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * @author Davide D'Alto
+ */
+public class JavaFXClassLoadingTest {
+
+	/**
+	 * This class will be present in the TCCL beacuse is part of JDK 8
+	 */
+	private static final String JAVAFX_APPLICATION_CLASS = "javafx.application.Application";
```
hferentschik: That's because if nothing sets explicitly the TCCL, the default is that the TCCL is set to the system classpath, right?

DavideD: That what I expect and it seems consistent with the unit tests. I guess I can set explicitly the TCCL in the the test, but it seems unecessary

hferentschik: I think this is the case - see also http://www.javaworld.com/article/2077344/core-java/find-a-way-out-of-the-classloader-maze.html

"If you don't do anything at all in the entire application, all Threads will end up with the system classloader as their context classloader"

hferentschik:
rebased and merged

PR #437, opened 2015-09-28
HV-1018 Making sure private properties from super-types can be validated

gunnarmorling:


engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
```diff
@@ -568,12 +559,18 @@ public BuilderDelegate(Class<?> beanClass, ConstrainedElement constrainedElement
 				case CONSTRUCTOR:
 				case METHOD:
 					ConstrainedExecutable constrainedExecutable = (ConstrainedExecutable) constrainedElement;
-					methodBuilder = new ExecutableMetaData.Builder(
-							beanClass,
-							constrainedExecutable,
-							constraintHelper,
-							executableHelper
-					);
+					Member member = constrainedExecutable.getExecutable().getMember();
```
hferentschik: :-) LOL

hferentschik:
Merged. Simple enough. Nice solution for now. In the long run one would still wish for a rework of some of the metadata stuff.

PR #439, opened 2015-10-07
HV-1020 support for additional information on constraint violation

julien-may:
This shows some real code of how the feature request of: [BVAL-490] (https://hibernate.atlassian.net/browse/BVAL-490) could look like.

Documentation and probably some more test code is missing. Please let me know if this leads to a direction you could imagine.

Looking forward for your comments.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
@@ -68,4 +68,6 @@ public interface HibernateConstraintValidatorContext extends ConstraintValidator
 	 * @since 5.2
 	 */
 	TimeProvider getTimeProvider();
+
+	HibernateConstraintValidatorContext withInfo(Object info);
 }
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
@@ -69,5 +69,18 @@ public interface HibernateConstraintValidatorContext extends ConstraintValidator
 	 */
 	TimeProvider getTimeProvider();
 
+	/**
+	 *
+	 * Allows to set an object that may further describe the violation.
+	 *
+	 * The user is responsible himself to ensure that this violation context is serializable in case the
+	 * {@code javax.validation.ConstraintViolation} has to be serialized.
+	 *
+	 * @param violationContext an object representing additional information about the violation
+	 *
+	 * @return a reference to itself to allow method chaining
+	 *
+	 * @since 5.3
+	 */
 	HibernateConstraintValidatorContext withViolationContext(Object violationContext);
 }
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
@@ -69,5 +69,5 @@ public interface HibernateConstraintValidatorContext extends ConstraintValidator
 	 */
 	TimeProvider getTimeProvider();
 
-	HibernateConstraintValidatorContext withInfo(Object info);
+	HibernateConstraintValidatorContext withViolationContext(Object violationContext);
 }
--- a/engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
+++ b/engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
@@ -1,0 +1,7 @@
+package org.hibernate.validator.engine;
+
+import javax.validation.ConstraintViolation;
+
+public interface HibernateConstraintViolation<T> extends ConstraintViolation<T> {
+	Object getInfo();
+}
--- a/engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
+++ b/engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
@@ -3,5 +3,5 @@ package org.hibernate.validator.engine;
 import javax.validation.ConstraintViolation;
 
 public interface HibernateConstraintViolation<T> extends ConstraintViolation<T> {
-	Object getInfo();
+	Object getViolationContext();
 }
--- a/engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
+++ b/engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
@@ -1,3 +1,9 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
 package org.hibernate.validator.engine;
 
 import javax.validation.ConstraintViolation;
--- a/engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
+++ b/engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
@@ -8,6 +8,14 @@ package org.hibernate.validator.engine;
 
 import javax.validation.ConstraintViolation;
 
+/**
+ * A custom {@link ConstraintViolation} which allows to get additional information for a constraint violation.
+ *
+ * @since 5.3
+ */
 public interface HibernateConstraintViolation<T> extends ConstraintViolation<T> {
+	/**
+	 * @return an object that further describes the violation or {@code null}.
+	 */
 	Object getViolationContext();
 }
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
@@ -231,12 +231,16 @@ public class ConstraintViolationImpl<T> implements HibernateConstraintViolation<
 		return violationContext;
 	}
 
+	/**
+	 * IMPORTANT - some behaviour of Validator depends on the correct implementation of this equals method! (HF)
+	 *
+	 * {@code expressionVariables} and {@code violationContext} are not taken into account for equality. These
+	 * variables solely enrich the actual Constraint Violation with additional information e.g how we actually
+	 * got to this CV.
+	 *
+	 * @return true if the two ConstraintViolation's are considered equals; false otherwise
+	 */
 	@Override
-	// IMPORTANT - some behaviour of Validator depends on the correct implementation of this equals method! (HF)
-
-	// Do not take expressionVariables into account here. If everything else matches, the two CV should be considered
-	// equals (and because of the scary comment above). After all, expressionVariables is just a hint about how we got
-	// to the actual CV. (NF)
 	public boolean equals(Object o) {
 		if ( this == o ) {
 			return true;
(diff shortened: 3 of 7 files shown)


engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
```diff
@@ -68,4 +68,6 @@
 	 * @since 5.2
 	 */
 	TimeProvider getTimeProvider();
+
+	HibernateConstraintValidatorContext withInfo(Object info);
```
hferentschik: I find _withInfo_ a bit too generic. WDYT about _withViolationContext_ or something like that?

hferentschik:
Hi @julien-may, thanks for the pull request. Except the imo too generic name of _getInfo_, things are looking good. A couple of technicalities: 

* There should be a HV issue for that. I created https://hibernate.atlassian.net/browse/HV-1020 and linked it to BVAL-490. Also the commit message should start with this HV key
* There are a couple of checkstyle violations which make the build fail:

     [INFO] Starting audit...
     engine/src/main/java/org/ hibernate/validator/engine/HibernateConstraintViolation. java:1: Line does not match  expected header line of '/\*'.
     engine/src/test/java/org/hibernate/validator/test/internal/engine/ constraintvalidation/HibernateConstraintValidatorContextTest.java:46:81: '(' is  not followed by whitespace.
     engine/src/test/java/org/hibernate/validator/test/internal/engine/ constraintvalidation/HibernateConstraintValidatorContextTest.java:110:109: '(' is  not followed by whitespace.
     engine/src/test/java/org/hibernate/validator/test/internal/engine/ constraintvalidation/HibernateConstraintValidatorContextTest.java:112:37: '(' is  not followed by whitespace.
     engine/src/test/java/org/hibernate/validator/test/internal/engine/ constraintvalidation/HibernateConstraintValidatorContextTest.java:157:59: '(' is  not followed by whitespace.

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
```diff
@@ -0,0 +1,7 @@
+package org.hibernate.validator.engine;
+
+import javax.validation.ConstraintViolation;
+
+public interface HibernateConstraintViolation<T> extends ConstraintViolation<T> {
+	Object getInfo();
```
hferentschik: Obviously its up to the user to make sure this additional context is serializable in case the _ConstraintViolation_ should be serialized. But serialization is problematic anyways due to the root bean being part of the _ConstraintViolation_

hferentschik:
Ahh, yet another technicality. Could you sign (if you not already have) the JBoss CLA - https://cla.jboss.org/index.seam? All you need is a free JBoss.org user account and a couple of clicks.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
julien-may: withInfo was just something to have a base to discuss on. i pretty much like your proposal. going to change that :+1:

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
julien-may: yes, this was already mentioned by emmanuel bernard. i propose to have this written in the java doc. WDYT?

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -196,7 +206,7 @@ public final Path getPropertyPath() {
 
 	@Override
 	public <C> C unwrap(Class<C> type) {
-		if ( type.isAssignableFrom( ConstraintViolation.class ) ) {
+		if ( type.isAssignableFrom( HibernateConstraintViolation.class ) ) {
```
julien-may: is this ok? as it could in theory break existing code of users.

julien-may:
signed the cla for the hibernate validator project

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
hferentschik: > yes, this was already mentioned by emmanuel bernard. i propose to have this written in the java doc. WDYT?

+1

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
hferentschik: I guess instead of removing the original if statement, we should be able to just add `type.isAssignableFrom( HibernateConstraintViolation.class )` as a second if statement, right?

hferentschik:
> signed the cla for the hibernate validator project

awesome

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
hferentschik: > i pretty much like your proposal. going to change that

ok :-)

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -213,6 +227,11 @@ public Object getExecutableReturnValue() {
 	}
 
 	@Override
+	public Object getViolationContext() {
+		return violationContext;
+	}
+
+	@Override
 	// IMPORTANT - some behaviour of Validator depends on the correct implementation of this equals method! (HF)
```
julien-may: IMHO ```violationContext``` should *not* be taken into account for the equals/hashcode methods the same reason as ```expressionVariables```. I did not document this... should we add that to the doc?

hferentschik: > IMHO violationContext should not be taken into account for the equals/hashcode methods the same reason as

+1

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * A custom {@link ConstraintViolation} which allows to get additional information for a constraint violation.
+ *
+ * @since 5.3
+ */
+public interface HibernateConstraintViolation<T> extends ConstraintViolation<T> {
+	/**
+	 * @return an object that further describes the violation or {@code null}.
+	 */
+	Object getViolationContext();
```
hferentschik: What do you think about:

    public <C> C getViolationContext(Class<C> type) 

This would potentially avoid a cast for the user. The question would be what to do if there is a context object, but of different type. I guess one would need to return `null`, since throwing an exception seems wrong. The user might use multiple custom constraints with different context objects.

hferentschik:
We would also need an update to the online docs, somewhere in chapter 11 - http://docs.jboss.org/hibernate/validator/5.2/reference/en-US/html_single/#validator-specifics. Do you feel up for this?

The file to edit is https://github.com/hibernate/hibernate-validator/blob/master/documentation/src/main/asciidoc/ch11.asciidoc. The documentation is written in Asciidoc, so a plain text editor would do.

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
julien-may: In the beginning I thought about this too and actually even wanted to go one step further so that the user defines the violation context type on the custom constraint annotation.

This type could then have even been used to do some runtime type validation, when it is tried to set a violation context of a different type.

```
@Target( { METHOD, FIELD, ANNOTATION_TYPE, TYPE, PARAMETER })
@Retention(RetentionPolicy.RUNTIME)
@Constraint(validatedBy = { FooValidator.class })
@Documented
public @interface FooConstraint {
	String message() default "";
	Class<?>[] groups() default {};
	Class<? extends Payload>[] payload() default {};
	Class<?>[] violationContext() default {};
}
```

Also from your proposal, it would then make sense to throw an exception when the type parameter given differs from what have been specified in the annotation.

WDYT?

However we can start introducing the definition you suggested and make it return null in case there is a violation context but of a different type.

julien-may:
> We would also need an update to the online docs, somewhere in chapter 11 - http://docs.jboss.org/hibernate/validator/5.2/reference/en-US/html_single/#validator-specifics. Do you feel up for this?

sure... i do something and we can discuss again if things are written bad or unclear.

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
hferentschik: > In the beginning I thought about this too and actually even wanted to go one step further so that the user defines the violation context type on the custom constraint annotation.

The problem is that this would make the use of a custom constraint mandatory and one could use the custom context with a default constraint just using a custom `ConstraintValidator` implementation. I think for now I opt for either plain object or the approach where the expected type is passed to the getter. If no context is set or one of the wrong type `null` is returned. WDYT?

julien-may: > The problem is that this would make the use of a custom constraint mandatory

i go for the one an expected type is passed. :+1:

hferentschik: > i go for the one an expected type is passed.

+1

gunnarmorling:
@julien-may, thanks for giving this a shot! Going to look at this one and add some remarks.

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
gunnarmorling: This is left to be done it seems?

+1 for the `public <C> C getViolationContext(Class<C> type)` signature, it leaves the door open for supporting several context objects, of different types (essentially the "type-safe heterogenous container" pattern from "Effective Java").

I don't like the method name though, how about `getPayload()`. To me, the method returns no context, just some data/payload retrieved from such context.

gunnarmorling:
Looks good overall to me, only see my specific remark above. Also the doc update is still missing. Looking forward to integrate this change, @julien-may :)

hferentschik:
> Looks good overall to me

+1 

>  Also the doc update is still missing. 

Right

> Looking forward to integrate this change

+1

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
hferentschik: > or the public <C> C getViolationContext(Class<C> type) signature, it leaves the door open for supporting several context objects, of different types (essentially the "type-safe heterogenous container" pattern from "Effective Java").

+1

> how about getPayload()

anything with 'payload' is probably a bad name, since it conflicts with the static payload feature which already exists in Bean Validation. Personally I think it is not so bad, since it is contextual data in conjunction of an existing constraint violation. Open for suggestion of course, but as said anything 'payload' is probably a bad idea.

gunnarmorling: I thought about it, but didn't deem it a real issue. "Payload" is rather generic, not necessarily limited to that specific usage. Maybe `getCustomPayload()`, `getDynamicPayload()`? `getExtraData()`?

julien-may: i also think that the term payload would be quiet confusing because of the use in the annotation. but it is true that the word context might be not that accurate as well.

how about ```getViolationDetails()``` or ```getViolationReference()``` ?

julien-may:
sorry for not updating this PR for some time, i was quiet busy. i will do the open points this evening.

hferentschik:
Closing this pull request which is superseded by #453

PR #440, opened 2015-10-14
HV-1021 Cache result of loadClass for el dependencies checking

lucaspouzac:


hferentschik:
Hi @lucaspouzac, in order to accept your patch, could you sign  the JBoss CLA - https://cla.jboss.org/index.seam? All you need is a free JBoss.org user account and a couple of clicks.

lucaspouzac:
It's done

hferentschik:
@lucaspouzac, thanks for the pull request and your interest in Validator. Pull request merged.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -64,6 +64,8 @@
 
 	private static final Log log = LoggerFactory.make();
 
+	private static Boolean missingElDependencies;
```
gunnarmorling: Should this not better be volatile to safely publish the write to other threads?

PR #441, opened 2015-10-14
 HV-1022: Missing check for type argument constraints on validateValue

gastaldi:
Type argument constraints should be checked before returning failing constraints


gastaldi:
Test case included

hferentschik:
Merged. Thanks @gastaldi :-)

PR #443, opened 2015-10-28
HV-1023 Prevent resource aggregation to occur in an Google App Enviro…

hferentschik:
…nment

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/logging/Messages.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/logging/Messages.java
@@ -84,6 +84,14 @@ public interface Messages {
 	@Message(value = "Custom parameterized types with more than one type argument are not supported and " +
 			"will not be checked for type use constraints.")
 	String parameterizedTypesWithMoreThanOneTypeArgument();
+
+	@Message(value = "Hibernate Validator cannot instantiate AggregateResourceBundle.CONTROL. " +
+			"This can happen most notably in a Google App Engine environment. " +
+			"A PlatformResourceBundleLocator without bundle aggregation was created. " +
+			"This only effects you in case you are using multiple ConstraintDefinitionContributor jars. " +
+			"ConstraintDefinitionContributors are a Hibernate Validator specific feature. All Bean Validation " +
+			"features work as expected. See also https://hibernate.atlassian.net/browse/HV-1023.")
+	String unableToUseResourceBundleAggregation();
 }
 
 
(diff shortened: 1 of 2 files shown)


gunnarmorling:
This works, but I am not sure whether I like the GAE-specific coding. I feel I'd prefer just an example in the ref guide showing how to configure a custom locator / interpolator with the aggregation feature disabled. Any thoughts?

hferentschik:
> This works, but I am not sure whether I like the GAE-specific coding. 

I am also a bit torn, but in the end I thought for the ease of use it is better this way. And really we are not importing any App Engine specific classes. I think it makes it works much smoother for people using Validator on App Engine. And we know that it is being used. We had several issues over time related to App Engine. 

> I feel I'd prefer just an example in the ref guide showing how to configure a custom locator / interpolator with the aggregation feature disabled.

The problem is that atm you can  only configure a custom `ResourceBundleLocataor` via the programmatic bootstrap. We would either need to add a custom `MessageInterpolator` for use in App Engine which can already be configured via `validation.xml` or we need some provider specific option to configure `PlatformResourceBundleLocator` accordingly. I would assume that in most cases Validator is used as part of another framework in which case it is easiest/best if things keep working out of the box.

This is maybe not perfect, but in the end I find it the lesser evil. 

Btw, I think it is a oversight that the ResourceBundle.Context class is not white listed. Not quite sure how one would report this or how one would get more detailed information on why it should not be whitelisted.

gunnarmorling:
> We would either need to add a custom MessageInterpolator

I would just add an example showing a custom `MessageInterpolator` using a non-aggregating RBL in the docs and not add any code at all. That way there is no reliance on GAE code on the HV side at all, as a bonus the approach will continue to work also if e.g. that class is renamed in GAE. Sure, we don't import GAE-classes, but we rely on it some specifics of it now.

Anyways, if you feel the usability advantages out-weigh that concern, let's merge it.

hferentschik:
>  if you feel the usability advantages out-weigh that concern, let's merge it

I think so. I think we should merge this. One "improvement" we could make is to not try a Google App Engine specific class, but rather directly `ResourceBundle.Control`. This won't have any effect on normal JVM runtimes and in the case of App Engine we catch the exception and skip aggregation. So it becomes really a generic check with no reliance to App Engine classes at all. The added benefit would be that bundle aggregation would magically work on App Engine in case they decide to whitelist `ResourceBundle.Control`. The only reference to App Engine would be in a comment explaining why we do the class loading check at all. Thoughts?

gunnarmorling:
Yes, that seems a bit better.

hferentschik:
> Yes, that seems a bit better.

Ok then. We go for that. I'll update the pull request.

hferentschik:
pushed an update

gunnarmorling:
Applied, thanks, @hferentschik. Note that I've added one more commit for moving the presence check into a static method invoked once at class load time. The not safely published Boolean made me feel a bit uneasy ;)

hferentschik:
Thanks. 

> The not safely published Boolean made me feel a bit uneasy ;)

Do you think this could really lead to a problem?

gunnarmorling:
The only thing that really could happen is that the check is done several times on different threads. It's more that such not safely published field looks wrong and makes one (well, me) wonder each time whether there is actually a problem or not. Plus, that way the null check during each instantiation is avoided.

hferentschik:
> The only thing that really could happen is that the check is done several times on different threads

Ok, then we agree :-)

PR #445, opened 2015-11-11
HV-1026: provide better context in exceptions

wolfc:
https://hibernate.atlassian.net/browse/HV-1026

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
@@ -233,15 +233,15 @@ public interface Log extends BasicLogger {
 	@Message(id = 64, value = "Unable to instantiate %1$s: %2$s.")
 	ValidationException getUnableToInstantiateException(String message, Class<?> clazz, @Cause Exception e);
 
-    @Deprecated
+	@Deprecated
 	@Message(id = 65, value = "Unable to load class: %s.")
 	ValidationException getUnableToLoadClassException(String className);
 
-    @Deprecated
+	@Deprecated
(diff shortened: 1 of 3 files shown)


hferentschik:
There are some checkstyle violations with the provided patch. I'll take a look.

wolfc:
Hmm, that just begs the question: why is check-style bound to verify phase and not compile?
It should definitely be run before tests are started.

hferentschik:
> why is check-style bound to verify phase and not compile?

because some developers like to work in a "hack mode" where they might create code which is not aligned with the enforced style yet. Of course they still want fast compile/test/debug cycles. To accommodate this type of developers/development the checkstyle task is bound to  verify.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -233,11 +233,16 @@ UnexpectedTypeException getNoValidatorFoundForTypeException(String constraintTyp
...
 	ValidationException getUnableToInstantiateException(String message, Class<?> clazz, @Cause Exception e);
 
+	@Deprecated
 	@Message(id = 65, value = "Unable to load class: %s.")
 	ValidationException getUnableToLoadClassException(String className);
 
+	@Deprecated
 	ValidationException getUnableToLoadClassException(String className, @Cause Exception e);
 
+	@Message(id = 65, value = "Unable to load class: %s from %s.")
```
hferentschik: We normally don't re-use ids, but I guess in this case one could argue that it is still the same exception, just with improved context. We also just remove unused log messages. No need to deprecate. I'll care of these things, since I want to squash the commits in the end anyways.

hferentschik:
thanks for the pull request. Squashed and merged.

PR #446, opened 2015-11-26
There is wrong in the Validation Messages in Korean.

LeeKyoungIl:
There is wrong in the Validation Messages in Korean.
The NotNull and The Null are changed with each other.
Because I fixed it.

Please check.


hferentschik:
@LeeKyoungIl, thanks for pointing the mix up out and providing a fix. Pull request is applied and I also created https://hibernate.atlassian.net/browse/HV-1035 to track the change in our issue tracker.

PR #447, opened 2015-11-27
HV-1034 JDK9 compilation errors

Sanne:
These changes are necessary to *compile* the project successfully on JDK9, but are not enough to have it pass the TCK.

https://hibernate.atlassian.net/browse/HV-1034

### DIFF
--- a/cdi/src/main/java/org/hibernate/validator/internal/cdi/interceptor/ValidationInterceptor.java
+++ b/cdi/src/main/java/org/hibernate/validator/internal/cdi/interceptor/ValidationInterceptor.java
@@ -104,7 +104,7 @@ public class ValidationInterceptor implements Serializable {
 	@AroundConstruct
 	public void validateConstructorInvocation(InvocationContext ctx) throws Exception {
 		ExecutableValidator executableValidator = validator.forExecutables();
-		Set<ConstraintViolation<Object>> violations = executableValidator.validateConstructorParameters(
+		Set<? extends ConstraintViolation<?>> violations = executableValidator.validateConstructorParameters(
 				ctx.getConstructor(),
 				ctx.getParameters()
 		);
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ReturnValueMetaData.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ReturnValueMetaData.java
@@ -59,7 +59,7 @@ public class ReturnValueMetaData extends AbstractConstraintMetaData
 		);
 
 		this.typeArgumentsConstraints = Collections.unmodifiableSet( typeArgumentsConstraints );
-		this.cascadables = Collections.unmodifiableList( isCascading ? Arrays.asList( this ) : Collections.<Cascadable>emptyList() );
+		this.cascadables = Collections.unmodifiableList( isCascading ? Arrays.<Cascadable>asList( this ) : Collections.<Cascadable>emptyList() );
 		this.groupConversionHelper = new GroupConversionHelper( groupConversions );
 		this.groupConversionHelper.validateGroupConversions( isCascading(), this.toString() );
 	}
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ValidatableParametersMetaData.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ValidatableParametersMetaData.java
@@ -9,8 +9,8 @@ package org.hibernate.validator.internal.metadata.aggregated;
 import org.hibernate.validator.internal.metadata.facets.Cascadable;
 import org.hibernate.validator.internal.metadata.facets.Validatable;
 
-import static org.hibernate.validator.internal.util.CollectionHelper.newHashSet;
-
+import java.util.HashSet;
+import java.util.Set;
 /**
  * Represents the constraint related meta data of the arguments of a method or
  * constructor.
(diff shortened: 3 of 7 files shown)


hferentschik:
I am not sure why we need to get into this. Compiling against JDK 9 is imo long distance. If anything I would be more interested to know what happens if I run JDK 8 compiled Validator in a Java 9 JVM. 

Regarding the actual changes, I am with Gunnar on this one. Some of the usages look legit. I think we need to double check with the spec and potentially report back to the JDK guys.

Sanne:
These are the original compile errors, but please try it as well?

    /engine/src/main/java/org/hibernate/validator/internal/util/CollectionHelper.java:[61,44] incompatible types: java.util.HashSet<capture#1 of ? extends T> cannot be converted to java.util.HashSet<T>

    /engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ReturnValueMetaData.java:[62,64] incompatible types: java.util.List<capture#2 of ? extends org.hibernate.validator.internal.metadata.facets.Cascadable> cannot be converted to java.util.List<org.hibernate.validator.internal.metadata.facets.Cascadable>

    /engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ValidatableParametersMetaData.java:[25,46] incompatible types: java.util.HashSet<capture#3 of ? extends org.hibernate.validator.internal.metadata.facets.Cascadable> cannot be converted to java.lang.Iterable<org.hibernate.validator.internal.metadata.facets.Cascadable>

     /engine/src/main/java/org/hibernate/validator/internal/metadata/raw/BeanConfiguration.java:[57,54] incompatible types: java.util.HashSet<capture#4 of ? extends org.hibernate.validator.internal.metadata.raw.ConstrainedElement> cannot be converted to java.util.Set<org.hibernate.validator.internal.metadata.raw.ConstrainedElement>

> Compiling against JDK 9 is imo long distance. If anything I would be more interested to know what happens if I run JDK 8 compiled Validator in a Java 9 JVM.

Sure, my aim was mostly explorative and I was equally surprised that the compiler would refuse it..

As you suggest I was primarily interested in verifying if it would run allright in JDK9; when I hit the compiler errors I thought let's propose a fix to you all so that we have a common starting place to investigate the runtime errors as a second step: because of Maven it's easier to have it compile properly first to then easily run the testsuite with the same JDK version.

> Some of the usages look legit. I think we need to double check with the spec and potentially report back to the JDK guys.

Sure, this might be a JDK regression. Let's talks with them if you're confident of that, this is exactly why I'm testing early.

Still I'm not sure; the compiler is making a good point: unfortunately the Java type system is a bit limited, so taking a method like this:

    public static <T> HashSet<T> newHashSet(Collection<? extends T> c)

As I see it, the client code is interested in the return type to be "of some T" which has to be compatible with the argument being fed to the method; but since there is no such thing in java as covariance of types for generics, your implementation is attempting to game the type system, or suggesting that these rules can be somehow magically relaxed by your helper.. apparently the compiler is stricter with that now?

From an API user perspective, wouldn't it be equally useful to have such a method declare as below ?

    public static <T> HashSet<T> newHashSet(Collection<T> c)

Anyway, that's just my 2 cents about the generic APIs; please throw my patches away if you have a better suggestion, but I'd hope we could move on soonish to have one of you check on the actual TCK failures: this one fixes compilation but I still have some runtime tests failing in the TCK, which would need our attention as well.

Hopefully the test failures aren't caused by my own hacked up changes to have it compile ;-)
Sorry I couldn't look at the test failures yet.

hferentschik:
>  but I'd hope we could move on soonish to have one of you check on the actual TCK failures:

as said, this is not high on my todo list. I see no compelling reason to move the build to JDK 9. Happy to discuss/investigate the issues above, but I don't feel this has high priority. 

Again, the more important question for me would be, is there an issue running Hibernate Validator built with JDK 8 on a Java 9 JVM. That would be a serious issue for me.

hferentschik:
Btw, I recommend we have the discussion around this on HV-1034. This provides better documentation for the future.

gunnarmorling:
> From an API user perspective, wouldn't it be equally useful to have such a method declare as below ?

No, it would be a bit less useful, as you couldn't initialize a collection from a collection of a sub-type. The same is possible using the actual collection type constructors (which are defined exactly like that: `Collection<? extends E> c`):

    List<Long> longs = Arrays.asList( 1L, 2L, 3L );
    Set<Number> numbers = new HashSet<Number>( longs );

As it seems, javac has a bit stricter rules for type deduction now, explicitly requiring the type argument to be specified. Similarly to the case of

    Set<Number> numbers = new HashSet<>( longs );

which also is rejected. I suppose making this behavior consistent is reasonable.

gunnarmorling:
Regarding the TCK, it passes for me when running without the security manager (by default the test runs within the SM). The failures seem to be caused by usage of (certain?) JAXB types which are internal types in JDK 9 apparently. This needs some deeper exploration, but it seems not to be a big issue from first sight.

hferentschik:
Personally I would just opt for closing this pull request for now. I don't see an urgent need to get into this now. We can revisit this once we really switch to Java 9 for compilation at which stage we might address all relevant changes at once.

@gunnarmorling WDYT? Do you want to merge it? 

I would to have a tabula rasa on the HV pull requests as the year turns to an end.

Sanne:
I don't mind closing it, but would like to see Validator to regularly test *running* on Java9.
How can we do that?

BTW I had found an issue in WildFly 10 CR4 which would not have it run on latest Java9 but that's fixed now in CR5.

hferentschik:
> I don't mind closing it

Ok

> but would like to see Validator to regularly test running on Java9. How can we do that?

So what one could do it setting up a Bean Validation TCK run of the latest Validator version running on Java 9. See also https://github.com/hibernate/hibernate-validator/blob/master/tck-runner/pom.xml and https://github.com/beanvalidation/beanvalidation-tck/tree/master/setup-examples. That would be useful indeed. We would verify that a compiled version of Validator works on Java 9 which is really what we want and we get another TCK run setup.

hferentschik:
FYI - https://hibernate.atlassian.net/browse/HV-1042

gunnarmorling:
Yes, good thinking with the TCK run on JDK 9, this should give us what we are after.

Sanne:
+1 ! Thanks @hferentschik

PR #421, opened 2015-07-17
HV-872 optional relaxation of parameter validation

ctabeckey:
Started from version 5.0, rebased to master and merged to there. Built and unit tested.

PR #448, opened 2015-11-28
HV-872 Optional relaxation of parameter validation

hferentschik:
Finally got around taking a closer look at the problems with applying #421.

I cherry-picked the commits introducing the new functionality and fixed merged conflicts. This pull request applies to master and passes all tests.

TODO -  Documentation needs to be written. The pull request had changes to the obsolete docbook documentation which I ignored so far.

### DIFF
--- a/documentation/src/main/asciidoc/ch11.asciidoc
+++ b/documentation/src/main/asciidoc/ch11.asciidoc
@@ -111,7 +111,6 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter11/failfast/F
 Here the validated object actually fails to satisfy both the constraints declared on the `Car` class,
 yet the validation call yields only one `ConstraintViolation` since the fail fast mode is enabled.
 
-
 [NOTE]
 ====
 There is no guarantee in which order the constraints are evaluated, i.e. it is not deterministic
@@ -122,6 +121,68 @@ required, a deterministic evaluation order can be enforced using group sequences
 
 Refer to <<section-provider-specific-settings>> to learn about the different ways of enabling the
 fail fast mode when bootstrapping a validator.
+
+[[section-method-validation-prerequisite-relaxation]]
+=== Relaxation of requirements for method validation in class hierarchies
+
+The Bean Validation specification defines a set of preconditions which apply when defining
+constraints on methods within class hierarchies. These preconditions are defined in
+http://beanvalidation.org/1.1/spec/#constraintdeclarationvalidationprocess-methodlevelconstraints-inheritance[section 4.5.5]
+of the Bean Validation 1.1 specification. See also <<section-method-constraints-inheritance-hierarchies>>
+in this guide.
+
+As per specification a Bean Validation provider is allowed to relax these preconditions.
+With Hibernate Validator you can do this in one of two ways.
+
+First you can use the configuration properties _hibernate.validator.allow_parameter_constraint_override_,
+_hibernate.validator.allow_multiple_cascaded_validation_on_result_ and
+_hibernate.validator.allow_parallel_method_parameter_constraint_ in _validation.xml_. See example
+<<example-relaxing-method-validation-xml>>.
+
+
+[[example-relaxing-method-validation-xml]]
+.Configuring method validation behaviour in class hierarchies via properties
+====
+[source, XML]
+----
+<?xml version="1.0" encoding="UTF-8"?>
+<validation-config
+        xmlns="http://jboss.org/xml/ns/javax/validation/configuration"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://jboss.org/xml/ns/javax/validation/configuration validation-configuration-1.0.xsd">
+    <default-provider>org.hibernate.validator.HibernateValidator</default-provider>
+
+    <property name="hibernate.validator.allow_parameter_constraint_override">true</property>
+    <property name="hibernate.validator.allow_multiple_cascaded_validation_on_result">true</property>
+    <property name="hibernate.validator.allow_parallel_method_parameter_constraint">true</property>
+</validation-config>
+----
+====
+
+Alternatively these settings can be applied during programmatic bootstrapping.
+
+[[example-relaxing-method-validation]]
+.Configuring method validation behaviour in class hierarchies
+====
+[source, JAVA]
+----
+HibernateValidatorConfiguration configuration = Validation.byProvider(HibernateValidator.class).configure();
+
+configuration.allowMultipleCascadedValidationOnReturnValues(true)
+    .allowOverridingMethodAlterParameterConstraint(true)
+    .allowParallelMethodsDefineParameterConstraints(true);
+----
+====
+
+By default, all of these properties are false, implementing the default behavior as defined in the
+Bean Validation specification.
+
+[WARNING]
+====
+Changing the default behaviour for method validation will result in non specification conform and non
+portable application. Make sure to understand what you are doing and that your use case really
+requires changes to the default behaviour.
+====
 
 [[section-programmatic-api]]
 === Programmatic constraint declaration
--- a/engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
+++ b/engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
@@ -22,6 +22,7 @@ import org.hibernate.validator.spi.valuehandling.ValidatedValueUnwrapper;
  * @author Gunnar Morling
  * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
  * @author Hardy Ferentschik
+ * @author Chris Beckey &lt;cbeckey@paypal.com&gt;
  */
 public interface HibernateValidatorConfiguration extends Configuration<HibernateValidatorConfiguration> {
 	/**
@@ -29,6 +30,27 @@ public interface HibernateValidatorConfiguration extends Configuration<Hibernate
 	 * Accepts {@code true} or {@code false}. Defaults to {@code false}.
 	 */
 	String FAIL_FAST = "hibernate.validator.fail_fast";
+
+	/**
+	 * Property corresponding to the {@link #allowOverridingMethodAlterParameterConstraint} method.
+	 * Accepts {@code true} or {@code false}.
+	 * Defaults to {@code false}.
+	 */
+	String ALLOW_PARAMETER_CONSTRAINT_OVERRIDE = "hibernate.validator.allow_parameter_constraint_override";
+
+	/**
+	 * Property corresponding to the {@link #allowMultipleCascadedValidationOnReturnValues} method.
+	 * Accepts {@code true} or {@code false}.
+	 * Defaults to {@code false}.
+	 */
+	String ALLOW_MULTIPLE_CASCADED_VALIDATION_ON_RESULT = "hibernate.validator.allow_multiple_cascaded_validation_on_result";
+
+	/**
+	 * Property corresponding to the {@link #allowParallelMethodsDefineParameterConstraints} method.
+	 * Accepts {@code true} or {@code false}.
+	 * Defaults to {@code false}.
+	 */
+	String ALLOW_PARALLEL_METHODS_DEFINE_PARAMETER_CONSTRAINTS = "hibernate.validator.allow_parallel_method_parameter_constraint";
 
 	/**
 	 * Property corresponding to the {@link #addValidatedValueHandler(ValidatedValueUnwrapper)} method. Accepts a String
@@ -139,6 +161,7 @@ public interface HibernateValidatorConfiguration extends Configuration<Hibernate
 
 	/**
 	 * @return the default {@link org.hibernate.validator.spi.constraintdefinition.ConstraintDefinitionContributor}. Never {@code null}.
+	 *
 	 * @since 5.2
 	 */
 	ConstraintDefinitionContributor getDefaultConstraintDefinitionContributor();
@@ -165,10 +188,10 @@ public interface HibernateValidatorConfiguration extends Configuration<Hibernate
 	 * If no class loader is given, these resources will be obtained through the thread context class loader and as a
 	 * last fallback through Hibernate Validator's own class loader.
 	 *
-	 * @param externalClassLoader
-	 *            The class loader for loading user-provided resources.
+	 * @param externalClassLoader The class loader for loading user-provided resources.
 	 *
 	 * @return {@code this} following the chaining method pattern
+	 *
 	 * @since 5.2
 	 */
 	HibernateValidatorConfiguration externalClassLoader(ClassLoader externalClassLoader);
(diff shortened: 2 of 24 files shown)


hferentschik:
Added a commit w/ some documentation. @gunnarmorling, care for a final review?

hferentschik:
@gunnarmorling, did you get a chance to look at it? I think it is good to go.

gunnarmorling:
Sorry for staying silent, I am lagging behind a bit on other this than
HSEARCH/ES atm. I'll look into this latest tomorrow.

hferentschik:
cool

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -186,4 +209,30 @@
...
+	 * Note that client code relying on such alternative behavior is not portable between Bean
+	 * Validation providers."
+	 * </pre>
+	 *
+	 * @param mvc {@code MethodValidationConfiguration} instance with the method validation behaviour configuration
+	 * settings.
+	 *
+	 * @return {@code this} following the chaining method pattern
+	 */
+	HibernateValidatorConfiguration setMethodValidationConfiguration(MethodValidationConfiguration mvc);
```
gunnarmorling: In which case would one use this setter?

The proposed design seems to break the fluent invocation style; What's the advantage over simply having `allowOverridingMethodAlterParameterConstraint()`, `allowMultipleCascadedValidationOnReturnValues()` and `allowParallelMethodsDefineParameterConstraints()` right here? That way one can invoke them without any intermediary assignments.

gunnarmorling: Regarding `isAllowMultipleCascadedValidationOnReturnValues()` etc., what's the exact semantics of these? Do they return the current value - i.e. including any changes applied by calling the setters - or does it only reflect the default settings? If so - which I think would make sense to be with the existing API - we should somehow encode the "default" world into the names to be inline with other getters and make the meaning apparent.

gunnarmorling: Btw. it'd be nice to add `@since 5.3` tags to new public API members.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -81,16 +81,17 @@
 	private ValidationProviderResolver providerResolver;
 	private final ValidationBootstrapParameters validationBootstrapParameters;
 	private boolean ignoreXmlConfiguration = false;
-	private final Set<InputStream> configurationStreams = CollectionHelper.newHashSet();
```
gunnarmorling: Why is it not final anymore?

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -213,6 +215,25 @@ public final HibernateValidatorConfiguration failFast(boolean failFast) {
 		return this;
 	}
 
+
+	@Override
+	public HibernateValidatorConfiguration setMethodValidationConfiguration(MethodValidationConfiguration mvc) {
+		// the MethodValidationConfiguration cannot be null, it will cause NPE later if it is
```
gunnarmorling: I'd rather fail early with `Contracts#assertNotNull()`, that's what we do in other places, too.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ExecutableMetaData.java
```diff
@@ -263,17 +264,17 @@ public boolean equals(Object obj) {
...
 		 */
-		private static final Set<MethodConfigurationRule> rules = Collections.unmodifiableSet(
-				CollectionHelper.<MethodConfigurationRule>asSet(
-						new OverridingMethodMustNotAlterParameterConstraints(),
-						new ParallelMethodsMustNotDefineParameterConstraints(),
-						new VoidMethodsMustNotBeReturnValueConstrained(),
-						new ReturnValueMayOnlyBeMarkedOnceAsCascadedPerHierarchyLine(),
-						new ParallelMethodsMustNotDefineGroupConversionForCascadedReturnValue()
+		@SuppressWarnings("unchecked")
+		private static final Set<Class<? extends MethodConfigurationRule>> DEFAULT_RULES = Collections.unmodifiableSet(
```
gunnarmorling: This is unused.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ExecutableMetaData.java
```diff
@@ -307,14 +310,35 @@ public boolean equals(Object obj) {
...
 			kind = constrainedExecutable.getKind();
 			executable = constrainedExecutable.getExecutable();
 			add( constrainedExecutable );
+			this.methodValidationConfiguration = methodValidationConfiguration;
+
+			// Build the rules that will be enforced through the metadata created here
+			this.rules = new HashSet<MethodConfigurationRule>();
+			for ( Class<? extends MethodConfigurationRule> ruleClass : this.methodValidationConfiguration.getConfiguredRuleSet() ) {
+				try {
+					this.rules.add( ruleClass.newInstance() );
```
gunnarmorling: Please use `org.hibernate.validator.internal.util.privilegedactions.NewInstance` instead; It may not be strictly needed in this particular case, but I'd prefer it for the sake of consistency throughout the code base (also exception messages will be consistent that way).

Or, couldn't we have the actual rule *instances*  in this set already (as created earlier in `MethodValidationConfigurationImpl()`? That'd avoid the need for reflective instantiation altogether.

engine/src/test/java/org/hibernate/validator/test/internal/engine/ValidatorTest.java
```diff
@@ -272,7 +272,7 @@ public String getE() {
 
 	class Ticket {
 		@NotNull
-		Float €price;
```
gunnarmorling: Please don't remove the "€" character, it was the very basis for this particular test ;)

engine/src/test/java/org/hibernate/validator/test/internal/engine/methodvalidation/ClassLoaderMethodParameterConstraintsTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * Integration test for {@link import org.hibernate.validator.internal.engine.ValidatorImpl} which tests that the ClassLoader of the
+ * base object and of classes in a parameter list do not affect validation.
+ * These tests attempt to simulate a web application that has its own class loader that
+ * does not preferentially delegate to a parent class loader.
+ *
+ * @author Chris Beckey &lt;cbeckey@paypal.com&gt;
+ */
+public class ClassLoaderMethodParameterConstraintsTest {
```
gunnarmorling: I am not quite clear about the rationale behind this test; Did you run into any specific issue which you'd like to cover in the test suite? Is it really related to the rule relaxation?

engine/src/test/suite/cfg-tests.xml
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
+
+<suite name="Hibernate Validator Configuration Tests" verbose="1">
```
gunnarmorling: What's the purpose of this file?

documentation/src/main/asciidoc/ch11.asciidoc
```diff
@@ -123,6 +122,64 @@ required, a deterministic evaluation order can be enforced using group sequences
...
+http://beanvalidation.org/1.1/spec/#constraintdeclarationvalidationprocess-methodlevelconstraints-inheritance[section 4.5.5]
+of the Bean Validation 1.1 specification. See also <<section-method-constraints-inheritance-hierarchies>>
+in this guide.
+
+As per specification a Bean Validation provider is allowed to relax these preconditions.
+With Hibernate Validator you can do this in one of two ways.
+
+First you can use the configuration properties _hibernate.validator.allow_parameter_constraint_override_,
+_hibernate.validator.allow_multiple_cascaded_validation_on_result_ and
+_hibernate.validator.allow_parallel_method_parameter_constraint_ in _validation.xml_. See example
```
gunnarmorling: "See example" = "See the following example" or "See example x.y.z"

documentation/src/main/asciidoc/ch11.asciidoc
```diff
@@ -123,6 +122,64 @@ required, a deterministic evaluation order can be enforced using group sequences
...
+MethodValidationConfiguration mvc = new MethodValidationConfigurationImpl();
+mvc.allowMultipleCascadedValidationOnReturnValues(true);
+mvc.allowOverridingMethodAlterParameterConstraint(true);
+mvc.allowParallelMethodsDefineParameterConstraints(true);
+
+configuration.setMethodValidationConfiguration(mvc);
+----
+====
+
+By default, all of these properties are false, implementing the default behavior as defined in the
```
gunnarmorling: I'd like us to add a warning that the defaults are what they are for a good reason, so one should be very careful and have a good reason when changing them, as this may come to the surprise of other developers.

gunnarmorling:
Hey @hferentschik, finally got to review that one :) Added some remarks inline.

I am no huge fan of the feature in general, but it's ok as the defaults don't change. So if someone reeeally wants to do it, why not... .

engine/src/test/java/org/hibernate/validator/cfgtest/ConfigurationFilePropertiesTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.lang.reflect.Field;
+import java.net.URL;
+import javax.validation.Configuration;
+import javax.validation.Validation;
+import javax.validation.Validator;
+import javax.validation.ValidatorFactory;
+import javax.validation.bootstrap.GenericBootstrap;
+
+import org.junit.After;
```
gunnarmorling: I don't think that's gonna work ;)

engine/src/test/java/org/hibernate/validator/cfgtest/ConfigurationFilePropertiesTest.java
```diff
@@ -0,0 +1,150 @@
...
+		Validator validator = factory.getValidator();
+
+		ValidatorImpl hibernateValidatorImpl = (ValidatorImpl) validator;
+		BeanMetaDataManager bmdm = findPropertyOfType( hibernateValidatorImpl, BeanMetaDataManager.class );
+		MethodValidationConfiguration methodConfig = findPropertyOfType( bmdm, MethodValidationConfiguration.class );
+
+		Assert.assertTrue( methodConfig.isAllowParallelMethodsDefineParameterConstraints() );
+	}
+
+	@After
```
gunnarmorling: Can be removed.

engine/src/test/java/org/hibernate/validator/cfgtest/ConfigurationFilePropertiesTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+	/**
+	 * Reflect into the subject and find the first property of the given type.
+	 *
+	 * @param subject - the instance to reflect on
+	 * @param clazz - exactly the class to match on
+	 *
+	 * @return
+	 */
+	private <T extends Object> T findPropertyOfType(Object subject, Class<T> clazz) {
```
gunnarmorling: Uff, I guess I'd just expose the required member. All these types are internal anyways.

hferentschik:
@gunnarmorling thanks for the thorough review. I will address/comment your comments asap. 

Personally, I think this is a nice feature ;-)

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
hferentschik: > The proposed design seems to break the fluent invocation style

Right. The original pull request has separate setters as well, but then the values which got internally passed around were blowing up the method signatures. For this reason I suggested to introduce `MethodValidationConfiguration` as a parameter object, but I guess this does not have to be exposed externally to the `Configuration`, but can be created internally. I'll update the pull request.

hferentschik: > Btw. it'd be nice to add @since 5.3 tags to new public API members.

+1 I have a todo for this. The reason I have not added it, was that I was not sure which version it would go in. a 5.2.x or a 5.3. I guess new functionality kind of implies a 5.3. Given that we are not going to do a new release this year and we have an early iteration on HV early next year where we most likely will go to 5.3, I think it is safe to add it now.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
hferentschik: slipped through during the merge. There were a couple of things which were "odd", eg some pom changes. I thought I caught most of it. Apparently not.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ExecutableMetaData.java
hferentschik: Good catch.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ExecutableMetaData.java
hferentschik: > It may not be strictly needed in this particular case, but I'd prefer it for the sake of consistency throughout the code base

+1

engine/src/test/java/org/hibernate/validator/test/internal/engine/ValidatorTest.java
hferentschik: I did not notice this on the merge. Good catch.

engine/src/test/suite/cfg-tests.xml
hferentschik: I think it was just used to run the tests he was initially working on. I'll remove it.

engine/src/test/java/org/hibernate/validator/cfgtest/ConfigurationFilePropertiesTest.java
hferentschik: :-)

engine/src/test/java/org/hibernate/validator/cfgtest/ConfigurationFilePropertiesTest.java
hferentschik: Hmm, true. On the other hand, maybe it is good to tests things in slightly different ways sometimes. I would say we just leave it as is.

hferentschik:
@gunnarmorling pushed an update with your review comments addressed.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -186,4 +209,58 @@
...
+	 * This would pose a strengthening of preconditions to be fulfilled by the caller."
+	 * </pre>
+	 *
+	 * @param allow flag determining whether validation will allow overriding to alter parameter constraints.
+	 *
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 5.3
+	 */
+	HibernateValidatorConfiguration allowOverridingMethodAlterParameterConstraint(boolean allow);
```
gunnarmorling: Yes, I think that's much nicer now.

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedExecutable.java
```diff
@@ -257,8 +257,8 @@ public boolean isEquallyParameterConstrained(ConstrainedExecutable other) {
 		int i = 0;
 		for ( ConstrainedParameter parameter : parameterMetaData ) {
 			ConstrainedParameter otherParameter = other.getParameterMetaData( i );
-			if ( parameter.isCascading != otherParameter.isCascading || !getDescriptors( parameter.getConstraints() )
-					.equals( getDescriptors( otherParameter.getConstraints() ) ) ) {
+			if ( parameter.isCascading != otherParameter.isCascading
```
gunnarmorling: It's a pity to have this pure formatting changes in this PR. But I won't go back to sort that out :/

gunnarmorling:
Applied, added one more commit for some clean-up. Thanks Chris and Hardy! Only took us one and half year; Not sure whether I find it more sad or funny ;)

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedExecutable.java
hferentschik: A colleague of mine said something to me the other day - "You get over it" ;-)

hferentschik:
> Applied, added one more commit for some clean-up. 

Thanks

>  Only took us one and half year; Not sure whether I find it more sad or funny ;)

Yeah, it sat there in the shelf waiting for some attention.

PR #452, opened 2015-12-19
HV-1039: make karaf feature work with karaf 4.0.x.

benson-basis:
* update jboss logging
* move paranamer wrap to its own feature
* move to current pax-exam

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/messageinterpolation/ResourceBundleMessageInterpolator.java
+++ b/engine/src/main/java/org/hibernate/validator/messageinterpolation/ResourceBundleMessageInterpolator.java
@@ -11,6 +11,8 @@ import java.util.Locale;
 import org.hibernate.validator.internal.engine.messageinterpolation.InterpolationTerm;
 import org.hibernate.validator.spi.resourceloading.ResourceBundleLocator;
 
+import javax.el.ExpressionFactory;
+
 /**
  * Resource bundle backed message interpolator.
  *
@@ -21,32 +23,53 @@ import org.hibernate.validator.spi.resourceloading.ResourceBundleLocator;
  * @author Adam Stawicki
  */
 public class ResourceBundleMessageInterpolator extends AbstractMessageInterpolator {
+	private final ExpressionFactory expressionFactory;
+
+	// Many constructors here are only called for tests; these just call ExpressionFactory.newInstance().
+	// Some sort of DI would be neater :-)
+
 	public ResourceBundleMessageInterpolator() {
 		super();
+		this.expressionFactory = ExpressionFactory.newInstance();
 	}
 
 	public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundleLocator) {
-		super( userResourceBundleLocator );
+		this( userResourceBundleLocator, (ExpressionFactory) null );
 	}
 
 	public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundleLocator,
-			ResourceBundleLocator contributorResourceBundleLocator) {
+											 ExpressionFactory expressionFactory) {
+		super( userResourceBundleLocator );
+		this.expressionFactory = expressionFactory;
+	}
+
+	public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundleLocator,
+			ResourceBundleLocator contributorResourceBundleLocator,
+			ExpressionFactory expressionFactory) {
 		super( userResourceBundleLocator, contributorResourceBundleLocator );
+		this.expressionFactory = expressionFactory;
+	}
+
+	public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundleLocator,
+											 ResourceBundleLocator contributorResourceBundleLocator) {
+		this( userResourceBundleLocator, contributorResourceBundleLocator, null );
 	}
 
 	public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundleLocator,
 			ResourceBundleLocator contributorResourceBundleLocator,
 			boolean cachingEnabled) {
 		super( userResourceBundleLocator, contributorResourceBundleLocator, cachingEnabled );
+		this.expressionFactory = null;
 	}
 
 	public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundleLocator, boolean cachingEnabled) {
 		super( userResourceBundleLocator, null, cachingEnabled );
+		this.expressionFactory = null;
 	}
 
 	@Override
 	public String interpolate(Context context, Locale locale, String term) {
-		InterpolationTerm expression = new InterpolationTerm( term, locale );
+		InterpolationTerm expression = new InterpolationTerm( term, locale, expressionFactory );
 		return expression.interpolate( context );
 	}
 }
--- a/osgi/integrationtest/src/test/java/mockservice/MockService.java
+++ b/osgi/integrationtest/src/test/java/mockservice/MockService.java
@@ -1,0 +1,80 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+
+package mockservice;
+
+import com.example.CustomerDecimalMin;
+import org.hibernate.validator.HibernateValidator;
+import org.hibernate.validator.HibernateValidatorConfiguration;
+import org.hibernate.validator.messageinterpolation.ResourceBundleMessageInterpolator;
+import org.hibernate.validator.resourceloading.PlatformResourceBundleLocator;
+
+import javax.el.ExpressionFactory;
+import javax.validation.ConstraintViolation;
+import javax.validation.Validation;
+import javax.validation.ValidationProviderResolver;
+import javax.validation.Validator;
+import javax.validation.spi.ValidationProvider;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ *
+ */
+public final class MockService {
+    private MockService() {
+        //
+    }
+
+    public static Set<ConstraintViolation<CustomerDecimalMin>> run( ClassLoader testClassLoader ) {
+        CustomerDecimalMin customer = new CustomerDecimalMin();
+
+        HibernateValidatorConfiguration configuration = Validation.byProvider( HibernateValidator.class )
+                // Cannot use lambda here, as the version of checkstyle in use implodes on it.
+                        .providerResolver( new ValidationProviderResolver() {
+                            @Override
+                            public List<ValidationProvider<?>> getValidationProviders() {
+                                ValidationProvider<HibernateValidatorConfiguration> prov = new HibernateValidator();
+                                List<ValidationProvider<?>> provs = new ArrayList<>();
+                                provs.add( prov );
+                                return provs;
+                            }
+                        }).configure();
+
+        ClassLoader oldTccl = Thread.currentThread().getContextClassLoader();
+        ExpressionFactory expressionFactory;
+        try {
+            // We might have been called under an unfriendly TCCL.
+            // Use _our_ classloader, which has com.sun.el and all that, to obtain
+            // an expression factory, since ExpressionFactory.newInstance() uses the TCCL.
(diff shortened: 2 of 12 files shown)


osgi/karaf-features/src/main/features/features.xml
```diff
@@ -13,17 +13,15 @@
...
-            are overriding the Import-Package header to make this optional
-        -->
-        <bundle>wrap:mvn:org.jboss.logging/jboss-logging/${jboss.logging.version}$overwrite=merge&amp;Bundle-SymbolicName=org.hibernate.org.jboss.logging&amp;Import-Package=org.apache.log4j;resolution:=optional,org.apache.logging.log4j;version="[2.0,3)";resolution:=optional,org.apache.logging.log4j.message;version="[2.0,3)";resolution:=optional,org.apache.logging.log4j.spi;version="[2.0,3)";resolution:=optional,org.jboss.logmanager;version="[1.5,2)";resolution:=optional,org.slf4j;version="[1.7,2)";resolution:=optional,org.slf4j.spi;version="[1.7,2)";resolution:=optional</bundle>
         <bundle>mvn:com.fasterxml/classmate/${classmate.version}</bundle>
         <bundle>mvn:javax.el/javax.el-api/${javax.el.version}</bundle>
         <bundle>mvn:org.glassfish.web/javax.el/${javax.el.version}</bundle>
-        <bundle>wrap:mvn:com.thoughtworks.paranamer:paranamer:${paranamer.version}</bundle>
+        <bundle>mvn:org.jboss.logging/jboss-logging/${jboss.logging.version}</bundle>
         <bundle>org.jsoup:jsoup:${jsoup.version}</bundle>
```
gunnarmorling: While you are at it, could you add the "mvn:" prefix to the last to ones? Apparently Karaf 3.x didn't complain, but in Karaf 4 I am getting an error due to this when trying to install the feature.

benson-basis:
There you go with the mvn:'s. Have I persuaded you at all about the TCCL issue?

gunnarmorling:
>  There you go with the mvn:'s. 

Cool; That needs "/" separators rather than ":", too.

> Have I persuaded you at all about the TCCL issue?

Not quite, because so far there is no test which shows that actually failing *without* setting the TCCL from the outside to something not working ;)

But in fact I agree it may be an issue with the Pax set-up which happens to expose a TCCL with access to the EL packages; I am testing this with Karaf stand-alone as we speak.

Assuming that is the case, the question is how to solve it. I am a bit reluctant to set the TCCL on HV's side because it's required for just a specific EL implementation, for others it may work without that. As a work-around, you always can set the TCCL properly on your side.

benson-basis:
On Mon, Dec 21, 2015 at 4:28 PM, Gunnar Morling <notifications@github.com>
wrote:

> There you go with the mvn:'s.
>
> Cool; That needs "/" separators rather than ":", too.
>

Oh, I am disorganized.

> Have I persuaded you at all about the TCCL issue?
>
> Not quite, because so far there is no test which shows that actually
> failing *without* setting the TCCL from the outside to something not
> working ;)
>
I can make you a test that has a real servlet and all that if you like. I
thought a shrunken simulation was more to the point. I spent many happy
hours finding the smallest possible simulation of the original problem I
hit with pax-web and CXF.


> But in fact I agree it may be an issue with the Pax set-up which happens
> to expose a TCCL with access to the EL packages; I am testing this with
> Karaf stand-alone as we speak.
>
> Assuming that is the case, the question is how to solve it. I am a bit
> reluctant to set the TCCL on HV's side because it's required for just a
> specific EL implementation, for others it may work without that. As a
> work-around, you always can set the TCCL properly on your side.
>
The problem is in the el-api jar, not the implementation jar. I don't know
if that counts as not 'a specific implementation' for you. In any case,
this is why I suggested adding an ExpressionFactory to the Configuration
system. This allows the caller to take responsibility for obtaining it, and
avoids any TCCL manipulation in HV. The alternative is to wrap the call to
ExpressionFactory.newInstance() in a TCCL setting to the
externalClassLoader..

How about this argument: in other SPI apis, there are versions of
newInstance() that take a class loader. If ExpressionFactory.newInstance()
had such an overload, wouldn't you want to pass the externalClassLoader to
it?

benson-basis:
I fixed up the feature descriptor and reordered the commits so that if you
like you can take the feature repairs without the test case.

gunnarmorling:
Ok, I can confirm the issue in plain Karaf now. Pax Exam indeed gives a false positive. Need to think about it for a bit :)

Could you sign the JBoss contributor license agreement in the mean time, which we need to merge your change. It's small web app at https://cla.jboss.org/index.seam. Thanks!

gunnarmorling:
> In any case, this is why I suggested adding an ExpressionFactory to the Configuration system. This allows the caller to take responsibility for obtaining it, and avoids any TCCL manipulation in HV

Ok, this seems like the best solution to me. I prefer it over having HV deal with any TCCL business as it's just something required by the EL reference implementation currently. There may be other EL providers which don't need this set-up. Interestingly, the latest EL 3.0 RI provides a combined JAR with API (javax.el.*) and implementation (com.sun.*) but still uses the TCCL for obtaining the impl class. That way it's not found in OSGi despite being located in the same JAR :(

Would you be willing to implement that addition to `HibernateValidatorConfiguration`? Thanks a lot for your effort, it's good to see this moving forward. I am a bit unhappy about PaxExam not revealing this issue, so even if we had had a test using `@DecimalMin`, we wouldn't have noticed the issue...

benson-basis:
I'm on it.

benson-basis:
@gunnarmorling I need a hint. I need to defer some code (that sets defaultMessageInterpolator) that is currently in the constructor of ConfigurationImpl so that it only runs after there's an opportunity to call the new method. Is there an established scheme for this, or do I need to change the users of that field to do lazy init?

gunnarmorling:
> I need to defer some code (that sets defaultMessageInterpolator) that is currently in the constructor of ConfigurationImpl

Ah, that's an interesting one.

Maybe it's hints us that a slightly different solution would be even better. Specifically, just allowing to pass the EL factory into the constructor of `ResourceBundleMessageInterpolator` (which is a public API type already) and then in OSGi you just pass in an interpolator instance configured with the right EL factory. WDYT?

benson-basis:
Hmm. Well, I have a version that puts it into the configuration. Hopefully it will work well enough to be pushed soon. I added that ctor argument in order to do it, so we could back out the Configuration changes if you don't like what it looks like.

benson-basis:
@gunnarmorling so, here it is.

# you may not like it.
# if you let Maven run both tests, the old one never starts. I may have an idea about that.

But let me know what you think of how I dealt with the insides of ConfigurationImpl.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -186,4 +187,19 @@
...
+
+	/**
+	 * Registers an expression factory for use when constraint violation messages use EL.
+	 * This allows applications to avoid issues with the method {@link ExpressionFactory#newInstance()},
+	 * which uses the TCCL in some implementations.
+	 * @param expressionFactory The expression factory for use.
+	 * @return  {@code this} following the chaining method pattern
+	 * @since 5.3
+     */
+	HibernateValidatorConfiguration expressionFactory(ExpressionFactory expressionFactory);
```
hferentschik: I am also not so happy about exposing `ExpressionFactory` here. It should not be required. If anything there is an issue in the implementation of the EL libs. I'd rather try to get this fixed there and in the meantime have something which "fixes" the issue under the hood. On top of things I needs fiddling with the creation of the message interpolator.

So what are the alternatives here again?

benson-basis: The only alternative I know is to wrap a setting of the TCCL to the 'externalClassLoader'  around the call to `ExpressionFactory.newInstance()` in `ELTermResolver`. The _defined API_ of is to read the TCCL, so it seemed to me that it was not implementation-dependent to do that -- presumably that behavior is sitting in a JSR somewhere.

However, @gunnarmorling was not enamored of that approach, which is why you are reading what you're reading.

I don't think I have any other paths to offer you. Also note that calling setContextClassLoader annoys the security manager in the TCK. I did code it that way at first and hit that.

gunnarmorling: Passing the EL factory in through a new constructor in `ResourceBundleMessageInterpolator`. That'd be my preference.

benson-basis: @gunnarmorling , doesn't that still expose `ExpressionFactory` in just the way that @hferentschik dislikes? If the two of you agree on that, far be it from me to argue.

hferentschik:
> Maybe it's hints us that a slightly different solution would be even better. Specifically, just allowing to pass the EL factory into the constructor of ResourceBundleMessageInterpolator (which is a public API type already) and then in OSGi you just pass in an interpolator instance configured with the right EL factory. WDYT?

I think this sounds like a more promising approach and aligns more with a less "visible" change in order to deal with issue which actually originate somewhere else.

benson-basis:
That looks like agreement to me. OK, stand by for slight rework.

benson-basis:
Gentlemen, I've remove the code in the Configuration department, leaving only the API changes in the ResourceBundleMessageInterpolator. I hope it meets with your satisfaction. Is there any live committee on the API we could harass about {{ExpressionFactory}} that you know of?

engine/src/main/java/org/hibernate/validator/messageinterpolation/ResourceBundleMessageInterpolator.java
```diff
@@ -21,32 +23,54 @@
...
+	// Many constructors here are only called for tests; these just call ExpressionFactory.newInstance().
+	// Some sort of DI would be neater :-)
+
 	public ResourceBundleMessageInterpolator() {
 		super();
+		this.expressionFactory = ExpressionFactory.newInstance();
 	}
 
-	public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundleLocator) {
+	public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundleLocator,
```
gunnarmorling: We must not alter these constructors (it's public facing API). Could you instead add a new one, please? Admittedly, the number of constructors gets unwieldy, probably we should add some kind of builder eventually.

osgi/integrationtest/src/test/java/mockservice/MockService.java
```diff
@@ -0,0 +1,78 @@
...
+                                return provs;
+                            }
+                        }).configure();
+
+        ClassLoader oldTccl = Thread.currentThread().getContextClassLoader();
+        ExpressionFactory expressionFactory;
+        try {
+            // use _our_ classloader, which has com.sun.el and all that, to obtain
+            // an expression factory.
+            Thread.currentThread().setContextClassLoader( MockService.class.getClassLoader() );
```
gunnarmorling: That should not be needed any longer? I think overall it'd be nice to just amend the existing test case to cover `@DecimalMin` (and thus the EL stuff), too.

gunnarmorling:
> Gentlemen, I've remove the code in the Configuration department, leaving only the API changes in the ResourceBundleMessageInterpolator.

Cool; Could you clean up the testing part a bit? Specifically, I don't think we need a new test class, just amending the existing one should be fine. Also could you squash all commits into a single one, that makes for a cleaner history (I can do that, too, before merging if you prefer that).

> Is there any live committee on the API we could harass about {{ExpressionFactory}} that you know of?

It may be worth reaching out to the EL expert group (see [JSR 341](https://jcp.org/en/jsr/detail?id=341)). I don't know whether there are plans for a new revision, but if so, allowing to pass in a classloader for finding the implementation would be a good addition.

benson-basis:
Hmm, I tried to end up by purely adding ctors and delegating with this().
I'll fix it up.

Now that I've eliminated the static I see your point about the test
program, I'll clean it up.

benson-basis:
Squashed to one commit. I claim that the old ctors are all still there, delegating to the new ones. Do we need one more ctor that both allow the expressionFactory and the caching boolean? Do you want me to toss in a builder as a static nested class?

gunnarmorling:
@benson-basis I've reworked your commit a bit to further simplify the test and make the change to the interpolator constructor a bit easier to follow (just adding one new constructor taking all possible arguments now). I've created PR #454 for this, thus I'm going to close this one. Thanks a lot again!

PR #453, opened 2015-12-22
HV-1020 Support for additional payload for HibernateConstraintViolation

hferentschik:
A pull request following up on #439. Switching the new method in `HibernateConstraintViolation` to:

    <C> C getDynamicPayload(Class<C> type);

Also using the term 'dynamicPayload'

### DIFF
--- a/documentation/src/main/asciidoc/ch11.asciidoc
+++ b/documentation/src/main/asciidoc/ch11.asciidoc
@@ -334,6 +334,39 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter11/propertypa
 
 This is specifically useful to obtain the element of `Set` properties on the property path (e.g. `apartments` in the example) which otherwise could not be identified (unlike for `Map` and `List`, there is no key nor index in this case).
 
+[[section-dynamic-payload]]
+=== Dynamic payload as part of ConstraintViolation
+
+In some cases automatic processing of violations can be aided, if the constraint violation provides additional
+data - a so called dynamic payload. This dynamic payload could for example contain hints to the user on how to
+resolve the violation.
+
+Dynamic payloads can be set in <<validator-customconstraints,custom constraints>> using `HibernateConstraintValidatorContext`.
+This is shown in example <<example-constraint-validator-setting-dynamic-payload>> where the
+`javax.validation.ConstraintValidatorContext` is unwrapped to `HibernateConstraintValidatorContext` in order to call
+`withDynamicPayload`.
+
+[[example-constraint-validator-setting-dynamic-payload]]
+.ConstraintValidator implementation setting a dynamic payload
+====
+[source, JAVA, indent=0]
+----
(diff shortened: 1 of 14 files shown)


julien-may:
are you also goig to do the documentation part? otherwise i still like to do that now. (i am in holidays, so i finally have time now)

hferentschik:
> are you also goig to do the documentation part? 

done - 32763598d8b5bf405af9fd9de291e01d52aceaaa. WDYT?

From my perspective we are good to go. @gunnarmorling !?

documentation/src/main/asciidoc/ch11.asciidoc
```diff
@@ -825,7 +860,7 @@ validator factory still is referenced by application code.
 ====
 
 [[section-time-provider]]
-=== Time providers for `@Future` and `@Past`
+=== Time providers for @Future and @Past
```
gunnarmorling: I don't mind, but why did you remove it? Does it look bad with the mark-up in titles?

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
```diff
@@ -0,0 +1,24 @@
...
+ * @since 5.3
+ */
+public interface HibernateConstraintViolation<T> extends ConstraintViolation<T> {
+	/**
+	 * @param type The type of payload to retrieve
+	 *
+	 * @return an instance of the specified type set by the user via
+	 * {@link org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorContext#withDynamicPayload(Object)}.
+	 */
+	<C> C getDynamicPayload(Class<C> type);
```
gunnarmorling: So it doesn't work super-well with parameterized types (e.g. `List`), requiring an unchecked cast as in the test. The alternative would be to use a type literal a la CDI, but I don't think it's worth that or even better from a usability POV.

gunnarmorling:
Rebased and applied; Thanks, @julien-may, @hferentschik, that's a nice one!

documentation/src/main/asciidoc/ch11.asciidoc
hferentschik: > Does it look bad with the mark-up in titles?

+1 Default rendering sucks with back-tics in the title. I hope asciidoctor will give you soon more control over how reference format and numbering looks

engine/src/main/java/org/hibernate/validator/engine/HibernateConstraintViolation.java
hferentschik: > The alternative would be to use a type literal a la CDI, but I don't think it's worth that or even better from a usability POV.

Yeah, not optimal with parameterized types, but as you say, I don't think it is worth getting into more complicated approaches.

hferentschik:
> Rebased and applied

Awesome.

julien-may:
> Rebased and applied

Yay!

PR #455, opened 2016-01-03
HV-864 - Cross-parameter constraint is disallowed by Annotation Processor

nibbio84:
I've analyzed the [Jira issue HV-864](https://hibernate.atlassian.net/browse/HV-864) and found out that cross-parameter constraints were not supported by hibernate annotation processor. In the example provided in Jira, the processor was trying to compare *Object* or *Object[]* to the return type of the method, not to the parameters.

I've implemented the support for cross-parameter constraints, added most of the checks and related test cases.


hferentschik:
Hi @nibbio84, great that you want to help out. We have been talking about getting the annotation processor up to date, but so far we just could not find the time to do so. We were always hoping that someone from the community would step up to the task. Seems it has happened now :-)

I have not looked closer at your changes yet, but will do so asap. A couple of things to start with though.

1. For us to accept your contributions you need to "sign" the JBoss [Contributor License Agreement (CLA)](https://cla.jboss.org/index.seam). All you need it is to sign up for a free JBoss user account and then select Hibernate Validator from the list of JBoss projects in the CLA.
2. As it stands your code does not compile (see also CI job). I had a quick look at the problem and it seems you are now depending on Bean Validation API classes (eg `javax.validation.constraintvalidation.ValidationTarget`), but the `pom.xml` does not include the Bean Validation API. Adding 

        <dependency>
            <groupId>javax.validation</groupId>
            <artifactId>validation-api</artifactId>
        </dependency>

gets the code to compile, but then some checkstyle violations fail the build. 

Regarding the latter, in case you are using Idea, you can find a code formatting template [here](http://hibernate.org/validator/contribute/#idea). If you are using Eclipse or any other IDE, most violations seems to be  braces placement and missing/wrong license headers. It should be quite easy to get your IDE to do apply the right style.

Regarding the missing dependency, I guess it is reasonable to expect that the Bean Validation API is available, but so far the annotation processor is a dependency free jar. There are several advantages to this. The code is not tied to a specific Bean Validation API version and the setup of the annotation processor for the user is simpler, since only a single jar needs to be added to the processor classpath. I think it is possible to add the new functionality without a hard dependency to the BV API. You could just do string based checks for example. However, I am open for a discussion on whether the BV API is really needed.

Hope this make sense. Let us know what you think.

nibbio84:
Thank you for the explanation, I didn't realize that the dependency on the *validation-api* was only with *test* scope (I'm using Eclipse, the problem appears only when compiling off-ide). I completely agree on the advantages of having a dependency-free jar.

Hope that Jenkins will compile the updated version.

annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
```diff
@@ -544,30 +769,22 @@ else if ( assignableTypes.size() == 1 ||
...
-					}
-				}, null
-		);
+	private TypeMirror determineSupportedType(AnnotationValue oneValidatorClassReference) {
 
 		// contains the bindings of the type parameters from the implemented
 		// ConstraintValidator interface, e.g. "ConstraintValidator<CheckCase, String>"
-		TypeMirror constraintValidatorImplementation = getConstraintValidatorSuperType( validatorType );
-
+		TypeMirror constraintValidatorImplementation = getConstraintValidatorSuperType( oneValidatorClassReference );
```
hferentschik: not a big fan of variable names like `oneXYZ`. I think we avoided them as much as possible on the actual engine code base, but in the annotation processor we still might have more of these names left. Just saying ...

annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
```diff
@@ -580,9 +797,78 @@ public TypeMirror visitDeclared(DeclaredType constraintValidatorImplementation,
...
+		);
+		
+		DeclaredType validatorType = validatorClass.accept( new TypeKindVisitor6<DeclaredType, Void>() {
+			@Override
+			public DeclaredType visitDeclared(DeclaredType t, Void p) {
+				return t;
+			}
+		}, null );
+		
+		DeclaredType supportedValidationTargetType = annotationApiHelper.getDeclaredTypeByName( "javax.validation.constraintvalidation.SupportedValidationTarget" );
```
hferentschik: +1, right, that's what I meant ;-)

annotation-processor/src/main/resources/org/hibernate/validator/ap/ValidationProcessorMessages.properties
```diff
@@ -26,3 +26,10 @@ GROUP_SEQUENCE_PROVIDER_ANNOTATION_VALUE_MUST_BE_AN_IMPLEMENTATION_CLASS=The ann
...
 ONLY_NON_VOID_METHODS_MAY_BE_ANNOTATED=Void methods may not be annotated with constraint annotations.
+CROSS_PARAMETER_TARGET_NOT_INFERABLE=The target of the constraint @{0} cannot be inferred.
+CROSS_PARAMETER_CONSTRAINT_VALIDATOR_HAS_INVALID_TYPE=Validator for cross-parameter constraint @{0} does not validate Object nor Object[].
+CROSS_PARAMETER_CONSTRAINT_MULTIPLE_VALIDATORS=The constraint @{0} defines multiple cross parameter validators. Only one is allowed.
+CROSS_PARAMETER_VALIDATION_APPLIES_TO_REQUIRED=Constraints with generic as well as cross-parameter validators must define an attribute validationAppliesTo(), but constraint @{0} does not.
+CROSS_PARAMETER_VALIDATION_APPLIES_TO_MUST_HAVE_IMPLICIT_DEFAULT_VALUE=Default value of the attribute validationAppliesTo() of the constraint @{0} must be ConstraintTarget#IMPLICIT.
+CROSS_PARAMETER_VALIDATION_APPLIES_TO_MUST_HAVE_CONSTRAINT_TARGET_RETURN_TYPE=Return type of the attribute validationAppliesTo() of the constraint @{0} must be javax.validation.ConstraintTarget.
+CROSS_PARAMETER_VALIDATION_ON_PARAMETERLESS_METHOD=Cross parameter constraint @{0} is illegally placed on a parameterless method or constructor.
```
hferentschik: To make git happy we should add a new line here

hferentschik:
> Thank you for the explanation,

no worries 

> I didn't realize that the dependency on the validation-api was only with test scope (I'm using Eclipse, the problem appears only when compiling off-ide).

I guess it depends on the IDE. Idea complained as well. Either way, the reference is always the command line build. I don't care much what IDE one uses or through which hoops one has to jump to set the project up. In the bottom line, the build should be optimized for the command line. I am for example not very fond of all the Eclipse Maven metadata added to the main `pom.xml`. IMO it has nothing to do there. It's really a trade off between clean build and some more comfort for the main stream IDEs Idea and Eclipse. 

> I completely agree on the advantages of having a dependency-free jar.

Cool

> Hope that Jenkins will compile the updated version.

It does and it works also locally for me now. Thanks for changing the style issues, it makes the patch much easier to review as well, since a lot of your previous changes were caused by re-formatting.

Overall this looks quite good to me now, but I let @gunnarmorling be the final judge, since the annotation processor was his initial baby.

hferentschik:
Does this actually also take care of HV-801?

nibbio84:
No, it doesn't.
I can take care of it. Shall I do it in current pull request?

gunnarmorling:
> I can take care of it. Shall I do it in current pull request?

No, if you are interested in tackling that one, too, I recommend to create a follow-up PR. We try to stick to a "one PR per feature" rule which eases reviewing, rebasing due to other concurrent changes etc.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/CrossParameterConstraintCheck.java
```diff
@@ -0,0 +1,192 @@
...
+ * <li>It must have at most one cross-parameter validator.
+ * <li>The cross-parameter validator must resolve to Object or Object[].
+ * <li>If the constraint is both normal and cross-parameter, it must define a 'validationAppliesTo()' attribute.
+ * <li>The 'validationAppliesTo' method, if any, must return a {@code ConstraintTarget}.
+ * <li>The 'validationAppliesTo' method, if any, must declare {@code ConstraintTarget#IMPLICIT} as default return value.
+ * </ul>
+ *
+ * @author Nicola Ferraro
+ */
+public class CrossParameterConstraintCheck extends AbstractConstraintCheck {
```
gunnarmorling: Great commenting!

gunnarmorling: Ah, only the "li" tags should be closed :)

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/CrossParameterConstraintCheck.java
```diff
@@ -0,0 +1,192 @@
...
+			public Boolean visitDeclared(DeclaredType t, Void p) {
+				if ( typeUtils.isSameType( constraintTargetType, t ) ) {
+					return true;
+				}
+
+				return false;
+			}
+		}, null );
+
+		return Boolean.TRUE.equals( res );
```
gunnarmorling: Is there any advantage over simply returning `res`?

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/CrossParameterConstraintCheck.java
```diff
@@ -0,0 +1,192 @@
...
+	private ExecutableElement getValidationAppliesToMethod(Element annotation) {
+		for ( Element e : annotation.getEnclosedElements() ) {
+			ExecutableElement method = e.accept( new ElementKindVisitor6<ExecutableElement, Void>() {
+
+				@Override
+				public ExecutableElement visitExecutableAsMethod(ExecutableElement e, Void p) {
+					if ( e.getSimpleName().contentEquals( "validationAppliesTo" ) ) {
+						return e;
+					}
+					return super.visitExecutableAsMethod( e, p );
```
gunnarmorling: I don't think the call to the super method is needed?

annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
```diff
@@ -361,6 +404,119 @@ public Boolean visitDeclared(DeclaredType constraintValidatorImplementation, Voi
...
+	 * 
+	 * @param element  the method/constructor on which the annotation is applied
+	 * @param annotation  the constraint annotation
+	 * @return the resolved {@code AnnotationProcessorValidationTarget}, null if the target cannot be inferred
+	 */
+	public AnnotationProcessorValidationTarget resolveValidationTarget(ExecutableElement element, AnnotationMirror annotation) {
+		Set<AnnotationProcessorValidationTarget> allowedTargets = getSupportedValidationTargets( annotation.getAnnotationType() );
+		
+		// assume that at least one target (AnnotationProcessorValidationTarget.ANNOTATED_ELEMENT) is present
+		assert allowedTargets.size()>0;
```
gunnarmorling: We usually don't use `assert` statements in the code base, as one cannot really rely on them (it's a user's choice whether to enable them or not). I think you can just remove it, given you obtain the value from another method of the same class so we really know it's never empty.

annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
```diff
@@ -518,7 +674,7 @@ else if ( assignableTypes.size() == 1 ||
 	 * @return A set with the supported types.
 	 */
 	private Set<TypeMirror> getSupportedTypes(DeclaredType constraintAnnotationType) {
-
+
```
gunnarmorling: Better add not the trailing whitespace (also some lines below). Actually I'd expect CS to fail due to that, but maybe I am dreaming. You can add it as a "Save action" in Eclipse to automatically remove any trailing whitespace. That setting works great for me.

annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
```diff
@@ -530,8 +686,77 @@ else if ( assignableTypes.size() == 1 ||
...
+			
+			for(Element e : annotation.getAnnotationType().asElement().getEnclosedElements()) {
+				
+				Boolean isValidationAppliesToMethod = e.accept( new ElementKindVisitor6<Boolean, Void>() {
+					@Override
+					public Boolean visitExecutableAsMethod(ExecutableElement e, Void p) {
+						if(e.getSimpleName().contentEquals( "validationAppliesTo" )) {
+							return true;
+						}
+						return super.visitExecutableAsMethod( e, p );
```
gunnarmorling: Is that one really needed?

annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
gunnarmorling: Yes, to be fair, it's really a left-over from my old style :)

annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
gunnarmorling: I'd just move this (and some other class name references to constants).

annotation-processor/src/main/resources/org/hibernate/validator/ap/ValidationProcessorMessages.properties
```diff
@@ -26,3 +26,10 @@ GROUP_SEQUENCE_PROVIDER_ANNOTATION_VALUE_MUST_BE_AN_IMPLEMENTATION_CLASS=The ann
 GROUP_SEQUENCE_PROVIDER_ANNOTATION_VALUE_DEFINED_PROVIDER_CLASS_WITH_WRONG_TYPE=The annotation @GroupSequenceProvider defines a default group sequence provider for type {0} instead of {1} (super)type.
 GROUP_SEQUENCE_PROVIDER_ANNOTATION_VALUE_CLASS_MUST_HAVE_DEFAULT_CONSTRUCTOR=The annotation @GroupSequenceProvider defines a default group sequence provider class without a public default constructor.
 ONLY_NON_VOID_METHODS_MAY_BE_ANNOTATED=Void methods may not be annotated with constraint annotations.
+CROSS_PARAMETER_TARGET_NOT_INFERABLE=The target of the constraint @{0} cannot be inferred.
+CROSS_PARAMETER_CONSTRAINT_VALIDATOR_HAS_INVALID_TYPE=Validator for cross-parameter constraint @{0} does not validate Object nor Object[].
+CROSS_PARAMETER_CONSTRAINT_MULTIPLE_VALIDATORS=The constraint @{0} defines multiple cross parameter validators. Only one is allowed.
```
gunnarmorling: "cross-parameter" to be consistent with other usages.

annotation-processor/src/main/resources/org/hibernate/validator/ap/ValidationProcessorMessages.properties
```diff
@@ -26,3 +26,10 @@ GROUP_SEQUENCE_PROVIDER_ANNOTATION_VALUE_MUST_BE_AN_IMPLEMENTATION_CLASS=The ann
...
+CROSS_PARAMETER_VALIDATION_APPLIES_TO_REQUIRED=Constraints with generic as well as cross-parameter validators must define an attribute validationAppliesTo(), but constraint @{0} does not.
+CROSS_PARAMETER_VALIDATION_APPLIES_TO_MUST_HAVE_IMPLICIT_DEFAULT_VALUE=Default value of the attribute validationAppliesTo() of the constraint @{0} must be ConstraintTarget#IMPLICIT.
```
gunnarmorling: I suppose everything after "#" will be treated as a comment and not be returned from the properties bundle?

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/crossparameters/MethodLevelValidationUsingCrossParameterConstraints.java
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.ap.testmodel.crossparameters;
+
+import javax.validation.ConstraintTarget;
+import javax.validation.constraints.Size;
+
+public class MethodLevelValidationUsingCrossParameterConstraints {
```
gunnarmorling: Add your name if you like; Credit where credit is due :)

gunnarmorling:
@nibbio84, added some more comments, mostly minor stuff, great work overall! Thanks a lot for this contribution, it's much appreciated!

nibbio84:
Hi, thanks for the suggestion about the trailing whitespace, I didn't know how to solve it automatically :)

The "#" in properties files works on my mac, anyway, I replaced it with the unicode character. I've removed some redundant code and moved some strings to *BeanValidationTypes*.

hferentschik:
I suggest to squash the commits into a single one making sure its message starts with the issue key. Right now some of the commits don't do that and all these follow up commits don't really add new features. It's all about the support of method constraints in the annotation processor.

nibbio84:
Everything squashed.

gunnarmorling:
Rebased and applied. Great work, @nibbio84, thanks a lot!

PR #457, opened 2016-01-20
HV-1048 Fix validator to make to work on JDK9

ctomc:
Fixes project to make it work on JDK9 with JEP 223 
see http://openjdk.java.net/jeps/223 for more about new versioning.

The change to gmaven plugin is to make build work on windows.
policy change is needed to make tck testsuite pass on java 9.

Without this change WildFly testsuite fails spectacularly as pretty much every EE deployment gets validator enabled and it fails on its init.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/Version.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/Version.java
@@ -6,6 +6,9 @@
  */
 package org.hibernate.validator.internal.util;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
 
 /**
@@ -13,6 +16,7 @@ import org.hibernate.validator.internal.util.logging.LoggerFactory;
  * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2012 SERLI
  */
 public final class Version {
+	private static final Pattern JAVA_VERSION_PATTERN = Pattern.compile( "^(?:1\\.)?(\\d+)$" );
 	static {
 		LoggerFactory.make().version( getVersionString() );
 	}
@@ -30,10 +34,18 @@ public final class Version {
 	 * @return the Java release as an integer (e.g. 8 for Java 8)
 	 */
 	public static int getJavaRelease() {
-		// Will return something like 1.8
-		String[] specificationVersion = System.getProperty( "java.specification.version" ).split( "\\." );
+		// Will return something like 1.8 or 9
+		String vmVersionStr = System.getProperty( "java.specification.version" );
 
-		return Integer.parseInt( specificationVersion[1] );
+		Matcher matcher = JAVA_VERSION_PATTERN.matcher( vmVersionStr );  //match 1.<number> or <number>
+
+		if ( matcher.find() ) {
+			return Integer.valueOf( matcher.group( 1 ) );
+		}
+		else {
+			throw new RuntimeException("Unknown version of jvm " + vmVersionStr);
+		}
+
 	}
 
 	// helper class should not have a public constructor
--- a/pom.xml
+++ b/pom.xml
@@ -599,7 +599,7 @@
                 <plugin>
                     <groupId>org.codehaus.gmaven</groupId>
                     <artifactId>gmaven-plugin</artifactId>
-                    <version>1.4</version>
+                    <version>1.5</version>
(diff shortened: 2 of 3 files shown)


hferentschik:
I guess this addresses and fixes https://hibernate.atlassian.net/browse/HV-1048, right? Could you add the issue key to the message?

ctomc:
sure I didn't know jira already existed, let me change commit msg

ctomc:
@hferentschik updated commit message to reflect jira.
Btw, do you think we could have an release with this fix we can include WildFly 10?

gunnarmorling:
@ctomc, do you have any idea why the policy file change is needed? Do you have a stacktrace of the error without the change, i.e. which code path is triggering the policy violation?

hferentschik:
> Btw, do you think we could have an release with this fix we can include WildFly 10?

10 Final? Unlikely. AFAIK there is code freeze for this release. Will be a subsequent maintenance release of Wildfly 10. I would think we try to get an update in, once we have a HV 5.3 final.

ctomc:
as said in the description of PR: "policy change is needed to make tck testsuite pass on java 9."
in any case example of exception:

```
java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "accessClassInPackage.jdk.internal")
        at java.security.AccessControlContext.checkPermission(AccessControlContext.java:468)
        at java.security.AccessController.checkPermission(AccessController.java:894)
        at java.lang.SecurityManager.checkPermission(SecurityManager.java:541)
        at java.lang.SecurityManager.checkPackageAccess(SecurityManager.java:1497)
        at java.lang.Class.checkPackageAccess(Class.java:2405)
        at java.lang.Class.checkMemberAccess(Class.java:2384)
        at java.lang.Class.getMethod(Class.java:1816)
        at org.hibernate.validator.internal.util.privilegedactions.GetMethod.run(GetMethod.java:33)
        at org.hibernate.validator.internal.util.privilegedactions.GetMethod.run(GetMethod.java:17)
        at org.hibernate.validator.internal.metadata.core.ConstraintHelper.run(ConstraintHelper.java:548)
        at org.hibernate.validator.internal.metadata.core.ConstraintHelper.isMultiValueConstraint(ConstraintHelper.java:347)
```
pretty much for all tests.

ctomc:
for release it would be useful anyway, if noting else to get it into EAP7, so another 5.2.x release would be great to have, It would also help us test wildfly/EAP on top of JDK9 even if we manually include it for testing branches.

gunnarmorling:
> as said in the description of PR: "policy change is needed to make tck testsuite pass on java 9."

Got that, I am more after the reason *why* this permission is needed now. It's a usage of the reflection API (`Class#getMethod()`), so it's not quite clear to me why this violates the "accessClassInPackage.jdk.internal" check (which it doesn't on 1.8). For now, we can merge that part as is but we'll need to understand what it exactly is about to see whether a) we need to change something on our side or b) update the required permissions in the docs.

ctomc:
JDK9 is moving to be more strict about using jdk internal apis on many levels  and looks like this is one of them.
looking at http://openjdk.java.net/jeps/220 this looks about right:
> Enable the selective de-privileging of JDK classes that today are granted all security permissions but do not actually require those permissions.

hferentschik:
But where does the code try to access something in the `jdk.internal` package? As @gunnarmorling I would like to understand the why here.

hferentschik:
Thinking about it, we probably have two (it not three separate) issues here. The gmaven upgrade we could have maybe sneaked in here, but at least the security manager changes need probably a dedicated issue and discussion. Potentially we also need to teach out to the JDK guys. BTW, what is the exact error with gmaven and Windows?

gunnarmorling:
Going to close this one, as I created #458, superseding it. Let's continue the discussion over there. 

Regarding the SM change, I agree we need to analyze it, but I don't think this should stop us from merging this change. It just lets run the TCK on JDK 9, we can analyze separately whether that's the right approach or some actual coding change is needed. But being able to pass the TCK as part of the build helps with analyzing other potential issues.

ctomc:
@hferentschik error on windows before upgrading plugin is:
> [ERROR] Failed to execute goal org.codehaus.gmaven:gmaven-plugin:1.4:execute (configure-properties) on project hibernate-validator-tck-runner: startup failed, script1453366734662.groovy: 2: unexpected char: '\' @ line 2, column 191.

hferentschik:
@ctomc - cheers

PR #463, opened 2016-02-04
HV-916 Disabling unstable test

jmartisk:
backport to 4.3


jmartisk:
The checks won't work, apparently they're using a newer JDK 8 build, with which HV 4.3.x is not compilable

gunnarmorling:
Hey @jmartisk, took that commit and applied it to the 4.3 branch.

Note that that branch is some commits ahead of the point where your PR was forked from. You did it from 4.3.1.Final, whereas there later on was a 4.3.2 release and after that also some more commits.

PR #464, opened 2016-02-12
HV-1055, HV-1057 Fixes around redeclared default group sequences

gunnarmorling:
Hey @hferentschik, could you take a look at this PR?

The handling of group sequences is quite painful, but I aimed at doing not more changes than needed to fix the issues at bay, as HV-1055 needs backporting to 4.3. In the long run I think we should do some changes, but the topic is complex and it's not quite clear yet to me how it could look eventually.

One thing I dislike specfically is the iteration in the case where the default group is not redefined. We could skip iteration there and just invoke the validation routine once for the default group. But I refrained from doing it for the sake of backportability for now.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -38,8 +38,8 @@ import javax.validation.metadata.BeanDescriptor;
 
 import org.hibernate.validator.internal.engine.ValidationContext.ValidationContextBuilder;
 import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager;
-import org.hibernate.validator.internal.engine.groups.GroupWithInheritance;
 import org.hibernate.validator.internal.engine.groups.Group;
+import org.hibernate.validator.internal.engine.groups.GroupWithInheritance;
 import org.hibernate.validator.internal.engine.groups.Sequence;
 import org.hibernate.validator.internal.engine.groups.ValidationOrder;
 import org.hibernate.validator.internal.engine.groups.ValidationOrderGenerator;
@@ -468,7 +468,8 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		for ( Class<? super U> clazz : beanMetaData.getClassHierarchy() ) {
 			BeanMetaData<? super U> hostingBeanMetaData = beanMetaDataManager.getBeanMetaData( clazz );
 			boolean defaultGroupSequenceIsRedefined = hostingBeanMetaData.defaultGroupSequenceIsRedefined();
-			List<Class<?>> defaultGroupSequence = hostingBeanMetaData.getDefaultGroupSequence( valueContext.getCurrentBean() );
+			Iterator<Sequence> defaultGroupSequence = hostingBeanMetaData.getDefaultValidationSequence( valueContext.getCurrentBean() );
+
 			Set<MetaConstraint<?>> metaConstraints = hostingBeanMetaData.getDirectMetaConstraints();
 
 			// if the current class redefined the default group sequence, this sequence has to be applied to all the class hierarchy.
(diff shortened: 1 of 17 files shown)


hferentschik:
> The handling of group sequences is quite painful

+1

> In the long run I think we should do some changes, but the topic is complex and it's not quite clear yet to me how it could look eventually.

+1

Could you explain on a higher level on what the issue is and what you've done to fix it. AFAIKT you added one more nested loop by introducing `ExpandedGroup`. Trying to understand why this is necessary.

gunnarmorling:
Yes, it's essentially one more loop. That's needed to properly handle re-defined default group sequences that comprise a) groups which extend other groups or b) other sequences. We missed these before when being part of the default group sequence.

hferentschik:
> Yes, it's essentially one more loop. 

Ok

> That's needed to properly handle re-defined default group sequences that comprise a) groups which extend other groups or b) other sequences. We missed these before when being part of the default group sequence.

Ok, that's what I thought. Gets really fugly though.

engine/src/main/java/org/hibernate/validator/internal/engine/groups/ExpandedGroup.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Set;
+
+/**
+ * Represents a validation group and all the groups it extends ("group inheritance").
+ *
+ * @author Gunnar Morling
+ */
+public class ExpandedGroup implements Iterable<Group> {
```
hferentschik: Cold we not "merge" this functionality into `Group`? I find `ExpandedGroup` a bit misleading. `Group` would be a single group and its inherited groups.

engine/src/main/java/org/hibernate/validator/internal/engine/groups/Sequence.java
```diff
@@ -97,7 +128,7 @@ public String toString() {
 	 * @param group the group for which the inherited groups need to be added to {@code expandedGroups}
 	 * @param expandedGroups The list into which to add all groups
 	 */
-	private void addInheritedGroups(Group group, List<Group> expandedGroups) {
+	private void addInheritedGroups(Group group, Set<Group> expandedGroups) {
```
hferentschik: :-)

engine/src/main/java/org/hibernate/validator/internal/engine/groups/ExpandedGroup.java
gunnarmorling: I tried that first actually, but it didn't work out eventually. In the validator algorithm we still need individual groups to check constraint presence; I still think it'd work in some way, but it'd need more investigation and changes which I tried to minimize as far as reasonably possible for now as the fix needs backporting to 4.3. That said, I am all for revisting the entire implementation at some point. I got some ideas, but it'll be a larger piece of work I am afraid.

hferentschik: > I tried that first actually

Ahh. Good we thought alike.

> but it didn't work out eventually.

Interesting. 

> I still think it'd work in some way, but it'd need more investigation and changes

I see. Personally I think I would have investigate a bit more, but if you think this is the best approach for now, let's go with it.

> That said, I am all for revisting the entire implementation at some point

+1

>  I got some ideas, but it'll be a larger piece of work I am afraid.

Maybe we can talk about it over a chat?

hferentschik: Maybe we can at least find a better name? `GroupWithInheritance`, `GroupInheritance`!?

gunnarmorling: Ok, `GroupWithInheritance` it is. I'll rework it.

gunnarmorling:
Pushed an update with the suggested rename.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -474,31 +478,37 @@ private void validateConstraintsForCurrentGroup(ValidationContext<?> validationC
...
-				for ( MetaConstraint<?> metaConstraint : metaConstraints ) {
-					// HV-466, an interface implemented more than one time in the hierarchy has to be validated only one
-					// time. An interface can define more than one constraint, we have to check the class we are validating.
-					final Class<?> declaringClass = metaConstraint.getLocation().getDeclaringClass();
-					if ( declaringClass.isInterface() ) {
-						Class<?> validatedForClass = validatedInterfaces.get( declaringClass );
-						if ( validatedForClass != null && !validatedForClass.equals( clazz ) ) {
-							continue;
+
+			while ( defaultGroupSequence.hasNext() ) {
```
gunnarmorling: WDYT, should we try and avoid the loop if its the default group? I'd extract the inner contents (the loop over constraints) into a method and straight go there for the single default group in that case. Avoids some iterator allocations, but complicates backporting. Any thoughts, @hferentschik, @emmanuelbernard`

gunnarmorling: Ok, I've pushed one more commit with this change. Avoiding these loops for the default case seems desirable to me.

gunnarmorling:
@hferentschik Can you check it out and let me know whether it's good to go now?

I've done the renaming and pushed one more commit for avoiding some looping for the default case. For backporting, I'll recommend the first three commits (as they fix the reported issue), the forth one would be good to have (the looping avoidance) but may be tougher to back port and the fifth one shouldn't be backported as method validation has changed too much between 4 and 5 and I don't expect it to be used very often in 4.

> but if you think this is the best approach for now, let's go with it.

Yes, IMO it's good enough for now.

>>    That said, I am all for revisting the entire implementation at some point

> +1

 >>   I got some ideas, but it'll be a larger piece of work I am afraid.

> Maybe we can talk about it over a chat?

Yes, absolutely. The entire `ValidatorImpl` class has reached a state that worries me. Of course it works as is, but maintaining it is starting to get nightmare-ish... :/

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
```diff
@@ -465,26 +499,30 @@ public String toString() {
...
 
 		private final MethodValidationConfiguration methodValidationConfiguration;
-		
+
 		private BeanMetaDataBuilder(
-				ConstraintHelper constraintHelper, 
-				ExecutableHelper executableHelper, 
+				ConstraintHelper constraintHelper,
+				ExecutableHelper executableHelper,
+				ValidationOrderGenerator validationOrderGenerator,
```
hferentschik: First I was wondering why you were passing `ValidationOrderGenerator` around, since I thought it was stateless, but it is actually not. It caches the resolved sequences. Wondering about the use of `ConcurrentMap` in `ValidationOrderGenerator` . Maybe it should be a `ConcurrentReferenceHashMap` as well?

hferentschik:
> I've done the renaming and pushed one more commit for avoiding some looping for the default case

Ok

> Yes, absolutely. The entire ValidatorImpl class has reached a state that worries me. 

+1 I think we discussed this already several times. One of the things which always bugged me was that object traversal and validation code is mixed. I had this idea of properly separating these concerns. Handling of groups is another one of the problem areas. In combination we end up with all these nested loops which need breaking.

hferentschik:
Depending on the concrete changes for BV 2.0 one might not get around a bigger change in `ValidatorImpl`!?

hferentschik:
> For backporting, I'll recommend the first three commits (as they fix the reported issue)

+1 I would just fix what's needed. The code base might be quite different anyways.

PR #465, opened 2016-02-18
HV-1058 Some optimizations around BeanMetaDataImpl

gunnarmorling:
Hey @hferentschik, a quick follow-up to HV-1055.

It avoids the repeated default group sequence retrieval when building the metamodel descriptors and also makes `BeanMetaDataImpl` fully immutable. WF 10 Final update as a bonus :)


hferentschik:
> also makes BeanMetaDataImpl fully immutable

+1 i like that

hferentschik:
nice one

PR #467, opened 2016-04-13
HV-1064 Add Travis support

gsmet:


hferentschik:
I added a comment to HV-1064 itself. I am kind of wondering what we gain here.

gunnarmorling:
Anything missing from this? I think it's good to go, right?

hferentschik:
I think we wanted to add some more info into the README

gsmet:
I added something along the lines of what we discussed to the README.md.

gunnarmorling:

PR #468, opened 2016-05-03
HV-501 - Extend programmatic API to allow alteration of ConstraintValidator list for a given constraint

fenrhil:
Here is the patch for HV-501. I split the PR in two commits:
 * the first to introduce the new feature (the programmatic API)
 * the second to replace the older, experimental concept of `ConstraintDefinitionContributor` by this new feature.

Some notes about this PR below.

1. In the constructor of `ValidatorFactoryImpl`, I had to move up the initialization of `this.constraintMappings`. I'd really like if someone could confirm that this has no way of breaking anything... Unit tests do succeed, but, you know...
2. As a result of these commits, `ConstraintMappingContributor` may now be used for attributing validators to constraints, exactly the feature that was previously provided by `ConstraintDefinitionContributor`. Thus the default contributor `ServiceLoaderBasedConstraintDefinitionContributor` is now `ServiceLoaderBasedConstraintMappingContributor`.
3. I replaced `HibernateValidatorConfiguration#getDefaultConstraintDefinitionContributor` with `HibernateValidatorConfiguration#getDefaultConstraintMappingContributor`. Based solely on the javadoc, it's an API break: nothing in this method's javadoc mentioned the fact that it was part of an experimental API. But the reference documentation was, in my opinion, clearly stating that everything related to the `ConstraintDefinitionContributor`s was experimental, so I'd say it's fine.
4. I'm not sure about what the adequate behavior should be when someone tries to define the same constraint multiple times in different constraint mappings (either programmatic or XML). For now, I made sure that an exception is thrown (see `org.hibernate.validator.test.cfg.ValidatorConstraintMappingTest.testMultipleDefinitionForSameConstraintOnDifferentConstraintMappings()`). This seems consistent with the behavior of XML parsing, though it prevents overriding XML configuration with programmatic API calls.
Note that I had to move some error-checking code from the XML parser to `ConstraintHelper`, which resulted mainly in the message of HV000167 changing slightly. Also, error precedence slightly changed, so for instance if an XML file defines a constraint for the second time, and does it with an invalid class in the validators list, then now an exception will be thrown about the validator and not about the duplicate definition. I think it's more of a minor inconvenience than a real API break, but let me know if you disagree.
5. I updated the reference doc and the relevant tests (moved them to `ConstraintApiTest`), but I think it really needs proofreading.
6. Please let me know if anything's wrong/impolite with my use of `@author` in the Javadoc. I'm not quite sure what I did was in line with your policy.


gunnarmorling:
@fenrhil Thanks a lot! Looks great on first sight, going to add some more detailed comments soon if needed.

Could you sign the contributor license agreement at cla.jboss.org for "Hibernate Validator" in between, if you haven't done so yet? Thanks!

fenrhil:
I signed the CLA, it's awaiting validation.

engine/src/main/java/org/hibernate/validator/cfg/ConstraintMapping.java
```diff
@@ -22,9 +26,24 @@
...
+	 *
+	 * @param <A> The annotation type to be configured.
+	 * @param annotationClass The annotation class on which to define the validators. This type must be an
+	 * {@code @interface} annotated with {@code javax.validation.Constraint}. All validators defined after calling
+	 * this method are added to the annotation of the type {@code annotationClass} until the next call
+	 * of {@code type} or {@code annotation}.
+	 *
+	 * @return Instance allowing for defining validators to be executed for the specified constraint.
+	 */
+	<A extends Annotation> ConstraintDefinitionContext<A> constraint(Class<A> annotationClass);
```
gunnarmorling: I'd name it `constraintDefinition()` to differentiate from actually applying constraints via the existing `constraint()` method and be more inline with the XML names.

engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintTarget.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.lang.annotation.Annotation;
+
+/**
+ * Facet of a constraint mapping creational context which allows to the select the constraint (annotation type) to
+ * which the next operations shall apply.
+ *
+ * @author Yoann Rodiere
+ */
+public interface ConstraintTarget {
```
gunnarmorling: `ConstraintDefinitionTarget`?

engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintTarget.java
```diff
@@ -0,0 +1,30 @@
...
+	 * Selects the constraint (i.e. annotation type) to which the next operations shall apply. A given constraint
+	 * may only be configured once.
+	 *
+	 * @param <A> The annotation type to select.
+	 * @param annotationType The annotation type to select. This type must be an {@code @interface} annotated with
+	 * {@code javax.validation.Constraint}.
+	 *
+	 * @return A creational context representing the selected constraint.
+	 */
+	<A extends Annotation> ConstraintDefinitionContext<A> constraint(Class<A> annotationType);
```
gunnarmorling: `constraint()`=`constraintDefinition()`

engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
```diff
@@ -139,6 +139,7 @@
 	private final Map<Class<? extends Annotation>, List<? extends Class<?>>> builtinConstraints;
 
 	private final ValidatorClassMap validatorClasses = new ValidatorClassMap();
+	private ConcurrentMap<Class<? extends Annotation>, Object> overridenAnnotations = newConcurrentHashMap();
```
gunnarmorling: What is the purpose of this map?

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -546,8 +546,8 @@ UnexpectedTypeException getNoValidatorFoundForTypeException(String constraintTyp
 	ValidationException getValidateOnExecutionOnOverriddenOrInterfaceMethodException(Method m);
 
 	@Message(id = 167,
-			value = "A given constraint definition can only be overridden in one mapping file. %1$s is overridden in multiple files")
-	ValidationException getOverridingConstraintDefinitionsInMultipleMappingFilesException(String constraintClass);
+			value = "A given constraint definition can only be overridden once (either through a XML file or the programmatic API). %1$s is overridden in multiple XML files and/or API calls")
```
gunnarmorling: I think that deviates a bit from the behavior of the DSL in other places. There, constraint mappings given via the DSL would be _added_ to the ones given via annotations and/or XML, unless `ignoreAnnotations()` is called (the name seems sub-par, but as per the ref guide it applies to XML, too).

So should we similarly _add_ programmatically given definitions, unless `includeExistingValidators()` is called?

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -640,4 +640,7 @@ UnexpectedTypeException getNoValidatorFoundForTypeException(String constraintTyp
 	@LogMessage(level = WARN)
 	@Message(id = 192, value = "Couldn't determine Java version from value %1s; Not enabling features requiring Java 8")
 	void unknownJvmVersion(String vmVersionStr);
+
+	@Message(id = 193, value = "%s is configured more than once via the programmatic constraint definition API.")
```
gunnarmorling: +1. Within one source (XML, DSL), it should be done only once.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/DefaultConstraintMapping.java
```diff
@@ -83,4 +91,31 @@ public final AnnotationProcessingOptionsImpl getAnnotationProcessingOptions() {
 
 		return configurations;
 	}
+
+	@Override
+	public <A extends Annotation> ConstraintDefinitionContext<A> constraint(Class<A> annotationClass) {
+		Contracts.assertNotNull( annotationClass, MESSAGES.annotationTypeMustNotBeNull() );
+		Contracts.assertTrue( annotationClass.isAnnotationPresent( Constraint.class ),
```
gunnarmorling: Interesting, i'd expect such check (in a more complete fashion) to happen later in the pipeline anyways? It doesn't do harm here, but it seems not strictly needed?

engine/src/test/java/org/hibernate/validator/test/cfg/ValidatorConstraintMappingTest.java
```diff
@@ -0,0 +1,308 @@
...
+		Class<? extends Payload>[] payload() default { };
+	}
+
+	/*
+	 * Implementation note: we are compiling in Java 6 mode, so {@code @SafeVarargs} is of no use here. Thus, we work
+	 * around the compiler warnings by not restraining the type of classes passed as parameters and just casting them.
+	 */
+	@SafeVarargs
+	@SuppressWarnings("unchecked")
+	private static void assertCorrectValidatorTypes(Set<? extends ConstraintViolation<?>> violations,
```
gunnarmorling: Looks good; Couldn't you just assert on specific messages created by the different validators, though? That'd be a bit more in line with what we do in other tests.

engine/src/test/java/org/hibernate/validator/test/cfg/ValidatorConstraintMappingTest.java
```diff
@@ -0,0 +1,308 @@
...
+		config.addMapping( mapping );
+		Validator validator = config.buildValidatorFactory().getValidator();
+
+		Set<? extends ConstraintViolation<?>> violations = validator.validate( new ConstrainedLongFieldBean() );
+		assertNumberOfViolations( violations, 1 );
+		assertCorrectValidatorTypes( violations, NonDefaultLongValidator.class );
+	}
+
+	@Test
+	public void testConstraintMappingDefaultsToIncludingExistingValidators() {
```
gunnarmorling: Is this actually ensuring the right thing? I.e. wouldn't the test still pass if the impl was completely no-op? How about rather expecting an exception that two matching validators are found (the default one and the programmatically configured one) if not `includeExistingValidators()` has been called?

engine/src/test/java/org/hibernate/validator/test/cfg/ValidatorConstraintMappingTest.java
```diff
@@ -0,0 +1,308 @@
...
+
+		Set<? extends ConstraintViolation<?>> violations = validator.validate( new ConstrainedIntegerFieldBean() );
+		assertNumberOfViolations( violations, 1 );
+		assertCorrectValidatorTypes( violations, NonDefaultIntegerValidator.class );
+	}
+
+	@Test
+	public void testConstraintMappingIncludingExistingValidatorsThenExcludingThem() {
+		mapping.constraint( ConstraintAnnotation.class )
+				.includeExistingValidators( true )
```
gunnarmorling: Does this call make a difference? I'd expect only the last value to be given (`false`) to be applied eventually?

documentation/src/main/asciidoc/ch11.asciidoc
```diff
@@ -800,83 +800,40 @@ entries from all the bundles with this name found on the classpath in addition t
 This mechanism is also helpful when creating large multi-module applications: Instead of putting all the constraint
 messages into one single bundle, you can have one resource bundle per module containing only those messages of that module.
 
-[[section-constraint-definition-contributor]]
-==== Constraint definitions via `ConstraintDefinitionContributor`
+[[section-programmatic-constraint-definition]]
+==== Constraint definitions via programmatic constraint declaration API
```
gunnarmorling: "Adding constraint definitions programmatically"

The spec uses "constraint definition" for creating a new constraint type and "constraint declaration" for applying it to a property, method etc. So formally, with your change the API is now for "programmatic constraint definition and declaration". But that's a mouthful, so I wouldn't name it like that everywhere.

documentation/src/main/asciidoc/ch11.asciidoc
```diff
@@ -800,83 +800,40 @@ entries from all the bundles with this name found on the classpath in addition t
...
-<<example-using-constraint-definition-contributor>>.
+definitions. You can use the programmatic constraint declaration API - see
+<<example-using-constraint-definition-api>>.
 
-[[example-using-constraint-definition-contributor]]
-.Using `ConstraintDefinitionContributor` to register constraint definitions
-====
-[source, JAVA]
+[[example-using-constraint-definition-api]]
+.Using the programmatic constraint declaration API to register constraint definitions
```
gunnarmorling: See the comment above.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -160,23 +152,11 @@
...
-	 *
-	 * @return {@code this} following the chaining method pattern
+	 * @return the default {@link org.hibernate.validator.spi.cfg.ConstraintMappingContributor}. Never {@code null}.
 	 *
-	 * @hv.experimental This API is considered experimental and may change in future revisions
-	 * @since 5.2
+	 * @since 5.3
 	 */
-	HibernateValidatorConfiguration addConstraintDefinitionContributor(ConstraintDefinitionContributor contributor);
+	ConstraintMappingContributor getDefaultConstraintMappingContributor();
```
gunnarmorling: What is the default constraint mapping contributor? What does it do?

engine/src/test/java/org/hibernate/validator/test/constraintvalidator/AcmeConstraintDefinitionContributor.java
```diff
@@ -1,36 +0,0 @@
...
- * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
- */
-package org.hibernate.validator.test.constraintvalidator;
-
-import org.hibernate.validator.spi.constraintdefinition.ConstraintDefinitionContributor;
-
-/**
- * @author Hardy Ferentschik
- */
-public class AcmeConstraintDefinitionContributor implements ConstraintDefinitionContributor {
```
gunnarmorling: Instead of deleting the test, could it be adapted and used with the new scheme?

engine/src/test/java/org/hibernate/validator/test/constraintvalidator/ConstraintDefinitionContributorTest.java
```diff
@@ -48,85 +47,19 @@ public void constraint_definitions_can_be_configured_via_service_loader() {
 		assertCorrectConstraintTypes( constraintViolations, MustMatch.class );
 	}
 
-	@Test(expectedExceptions = UnexpectedTypeException.class, expectedExceptionsMessageRegExp = "HV000030.*")
```
gunnarmorling: Why are this and the tests below removed? Couldn't we keep them?

gunnarmorling:
Hey @fenrhil, I've added some comments on the commits. Mostly minor things, it looks great overall. Thanks a lot for taking up this task!

Some more comments to your remarks/questions:

> In the constructor of ValidatorFactoryImpl, I had to move up the initialization of this.constraintMappings. I'd really like if someone could confirm that this has no way of breaking anything... Unit tests do succeed, but, you know...

Looks good to me.

> As a result of these commits, ConstraintMappingContributor may now be used for attributing validators to constraints, exactly the feature that was previously provided by ConstraintDefinitionContributor. Thus the default contributor ServiceLoaderBasedConstraintDefinitionContributor is now ServiceLoaderBasedConstraintMappingContributor.

I had envisioned it a bit differently, but your's seems good, too. What I had in mind originally was to keep `ConstraintDefinitionContributor` as a concept but make it an impl class. It would have been "consumed" as before, also the service loader based impl would have remained as is. Only that users couldn't have registered custom contributors anymore, instead they'd have used constraint mappings. The impl of constraint mappings would have used `ConstraintDefinitionContributor` underneath.

> I replaced HibernateValidatorConfiguration#getDefaultConstraintDefinitionContributor with HibernateValidatorConfiguration#getDefaultConstraintMappingContributor. Based solely on the javadoc, it's an API break: nothing in this method's javadoc mentioned the fact that it was part of an experimental API. But the reference documentation was, in my opinion, clearly stating that everything related to the ConstraintDefinitionContributors was experimental, so I'd say it's fine.

+1 That's ok. I'm wondering though whether `getDefaultConstraintMappingContributor()` is needed at all. I cannot see a case where a user would make use of it? I'd probably omit it.

> I'm not sure about what the adequate behavior should be when someone tries to define the same constraint multiple times in different constraint mappings (either programmatic or XML). For now, I made sure that an exception is thrown (see org.hibernate.validator.test.cfg.ValidatorConstraintMappingTest.testMultipleDefinitionForSameConstraintOnDifferentConstraintMappings()). This seems consistent with the behavior of XML parsing, though it prevents overriding XML configuration with programmatic API calls.

See one of my comments. I think it should be slightly different.

> Note that I had to move some error-checking code from the XML parser to ConstraintHelper, which resulted mainly in the message of HV000167 changing slightly. Also, error precedence slightly changed, so for instance if an XML file defines a constraint for the second time, and does it with an invalid class in the validators list, then now an exception will be thrown about the validator and not about the duplicate definition.

Any chance that we can avoid that? The duplicated definition is on a "higher level", so if possible I'd prefer to report that in this case.

> I updated the reference doc and the relevant tests (moved them to ConstraintApiTest), but I think it really needs proofreading.

See my comments on the commits. I'm wondering why some tests were deleted.

> Please let me know if anything's wrong/impolite with my use of @author in the Javadoc. I'm not quite sure what I did was in line with your policy.

Nothing implite I could spot. On the contrary, the comments look very thorough and descriptive.

engine/src/main/java/org/hibernate/validator/cfg/ConstraintMapping.java
fenrhil: Will do. I just took the name from `ConstraintDefinitionBuilderContext` (IIRC), but indeed you proposal is clearer.

engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintTarget.java
fenrhil: Will do.

engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintTarget.java
fenrhil: Will do.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
fenrhil: The ConstraintHelper is used both for populating the default validators for each annotation and for later override, either through XML configuration or through the programmatic API.
Since the `ConstraintHelper` is the most high-level place where both the XML parser and the programmatic API add "constraint definitions", it seems the best place to add the checks that will throw an exception whenever a user "overrides" the default validators more than once (assuming it's what we want).
And to do that, we must store somewhere the information "this annotation has been overridden", because just checking the existence of an entry in the map is not enough: it may already contain default values. Also, I'm using a map here because it's simpler than finding some thread-safe implementation of Set.
I could have added some complex checks, comparing the content of `validatorClasses` and of `builtinConstraints`, but this does not seem more elegant to me, and it would not handle the case where a user overrides the default with... exactly the default validators.

I will add a comment and fix the typo ("overrid**d**en", missing "d")

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
fenrhil: > I think that deviates a bit from the behavior of the DSL in other places.
I wanted to be consistent with the current behavior of the XMLParser: if two definitions of the same constraint are found, even in different files, an exception is thrown.
Now, if the purpose of the programmatic API is not only to offer an *alternative* to the XML, but also a way to *override* the XML, indeed it makes sense to provide a way to ignore the XML.

> So should we similarly add programmatically given definitions, unless includeExistingValidators() is called?
I guess you mean `includeExistingValidators(false)`. I have mixed feelings about that. That would be the most straightforward solution, but that would also be a move from the current use of `includeExistingValidators` in the XML configuration: there, multiple constraint definitions for the same annotation are forbidden, so "existing validators" strongly implies "default validators".
On the other hand, as you suggest, we already have bad names taken from XML configuration that have a broader meaning in the programmatic API.
Also, allowing the user to pick what validators it want to preserve (default, XML, default and XML, or none) would result in major changes in the way we handle validator registration (currently the source of a particular validator registration is not recorded).

So... well, I guess you're right. I can't think of any solution that would be both clearer to the user, consistent with the current programmatic API, consistent with the XML configuration, and would not result in an oversized implementation (compared to what it provides).

I will do as you proposed. On the bright side, I guess the whole thing with changing the `ConstraintHelper` and how I did it will not matter anymore, since I will have to revert it.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
fenrhil: Currently that message is only used when it's done multiple time on the **same** `ConstraintMapping` object. When done on multiple `ConstraintMapping` objects, that's the code I added on `ConstraintHelper` that takes over. I'm not sure that the programmatic API handles this kind of misuse for constraint mappings; does it? Should it?

engine/src/main/java/org/hibernate/validator/internal/cfg/context/DefaultConstraintMapping.java
fenrhil: Well, I don't see why we would check for null values and stop there. I guess there are null checks later in the pipeline too, but if we removed this null check here, the stacktrace would only be less clear.
That being said, indeed, I also stopped to what's simple to implement: I did not check the presence of the mandatory attributes on the annotation. I just wanted to make sure that the argument is what we expect (a constraint annotation), as opposed to validating the correctness of this annotation.

Thinking about it a little, if Java had allowed inheritance in annotations, there would probably have been a `ConstraintAnnotation` supertype, instead of a `@Constraint` annotation, so that you could have enforced the presence of the various annotation attributes. Thus I think the presence of the `@Constraint` annotation is almost some kind "type requirement", a contract between this API and its users.

engine/src/test/java/org/hibernate/validator/test/cfg/ValidatorConstraintMappingTest.java
fenrhil: Well, I could, but when someone reads the tests, that's not the meaning I want to convey: I just want to know which validators were executed, no matter their messages. Note that it's exactly how it's implemented, though.

Now, if it's bothering you, indeed I can just put a static final string in each validator type and go with `assertCorrectConstraintViolationMessages` inside each test.

engine/src/test/java/org/hibernate/validator/test/cfg/ValidatorConstraintMappingTest.java
fenrhil: Well, this test mainly ensures that `validatedBy` does not **override** the existing validators, but just add them. So I just assert later that the default validator is still registered after a call to `validatedBy`. This test might pass if the impl was a no-op, but right there that's exactly what we're testing: the absence of side-effect.

As for testing when multiple validators are registered for the same type and the same annotations, `testMultipleValidatorsForSameType` does exactly that. I can add a test without a call to `includeExistingValidators(true)`, if you wish, but since it's the default (and since `testConstraintMappingDefaultsToIncludingExistingValidators` ensures it's the default), I didn't think it would be necessary. Will do anyway, there's no harm in that.

engine/src/test/java/org/hibernate/validator/test/cfg/ValidatorConstraintMappingTest.java
fenrhil: That's exactly what I'm testing :) If it did not, the last validation would fail because there would be multiple validators for the type "Integer". Maybe I should add some comments?

documentation/src/main/asciidoc/ch11.asciidoc
fenrhil: Okay, will change that. Agreed, there's no simple name for this API anymore.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
fenrhil: I'd say it contributes defaults to constraint mappings :D 

More seriously, that's the equivalent of a method I removed, `getDefaultConstraintDefinitionContributor`. I had to, because `ConstraintDefinitionContributors` have been replaced with `ConstraintMappingContributors`. This method (and now its replacement) was used by `ValidatorFactoryImpl` to retrieve default constraint definitions. Currently, the default implementation is `ServiceLoaderBasedConstraintMappingContributor`, which performs automatic discovery of `ConstraintValidator`s. 

But, as you say below, this method is indeed useless in the public API. It could just exist in `ConfigurationImpl`. I will remove it from `HibernateValidatorConfiguration`.

engine/src/test/java/org/hibernate/validator/test/constraintvalidator/AcmeConstraintDefinitionContributor.java
fenrhil: See below (on the actual test class).

engine/src/test/java/org/hibernate/validator/test/constraintvalidator/ConstraintDefinitionContributorTest.java
fenrhil: Because the concept of `ConstraintDefinitionContributor` does not exist anymore. Which was asked explicitely of me: "Let's not provide two different APIs to achieve the same thing, it's against the spirit of design we agreed long time ago.". So : 

 * `addConstraintDefinitionContributor` disappeared completely, so the tests around this method are not relevant anymore
 * `HibernateValidatorConfiguration.CONSTRAINT_DEFINITION_CONTRIBUTORS` disappeared too (`ConstraintMappingContributor`s may be used instead), so the tests around this property are not relevant anymore
 * I removed `validating_constraint_without_registered_constraint_validator_throws_exception` because it was already tested there: `org.hibernate.validator.test.internal.engine.valuehandling.UnwrapModesTest.no_constraint_validator_for_wrapped_and_unwrapped_value_throws_exception()`

I checked that `ValidatorConstraintMappingTest` covered the same features in the new API. I didn't see the point of having two places where I test the same thing.

fenrhil:
Hi Gunnar, 

> > Note that I had to move some error-checking code from the XML parser to ConstraintHelper, which resulted mainly in the message of HV000167 changing slightly. Also, error precedence slightly changed, so for instance if an XML file defines a constraint for the second time, and does it with an invalid class in the validators list, then now an exception will be thrown about the validator and not about the duplicate definition.
>
> Any chance that we can avoid that? The duplicated definition is on a "higher level", so if possible I'd prefer to report that in this case.

I don't think we can, or at least it's not obvious. The duplicated definition is not duplicated when looking at the definitions on a "higher level": `ConstraintMapping`s have no view over the other (already-existing) definitions, and `org.hibernate.validator.internal.engine.ValidatorFactoryImpl.processConstraintDefinitionContribution(ConstraintDefinitionContribution<A>, ConstraintHelper)` only transmits information without knowing what it means (and to me it's better that way).
Besides, I don't really see the problem with making `ConstraintHelper` throw exceptions to the end user, since it already does (in `isConstraintAnnotation`).

Regarding the other issues, I replied to you messages directly.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gunnarmorling: Sounds good :)

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gunnarmorling: It should be illegal do define one constraint more than once across all `ConstraintMapping`s (as it's illegal to do in one more more XML file(s)).

engine/src/main/java/org/hibernate/validator/internal/cfg/context/DefaultConstraintMapping.java
gunnarmorling: Ok. I suppose ideally we'd have a simple to use method `assertIsConstraintType()` which accepts an annotation type and does the validation wherever it's needed.

engine/src/test/java/org/hibernate/validator/test/cfg/ValidatorConstraintMappingTest.java
gunnarmorling: Makes sense. We can leave it as is, was just interested in the reasoning :)

engine/src/test/java/org/hibernate/validator/test/cfg/ValidatorConstraintMappingTest.java
gunnarmorling: I'm not sure I'm following. Are you saying that `includeExistingValidators(false)` is going to ignore validators previously set in the same `ConstraintMapping`? What I had in mind is that it only should consider validators given via annotations or XML. Then only one call to `includeExistingValidators()` would make sense.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
gunnarmorling: +1.

gunnarmorling:
@fenrhil Replied to some of your replies :)

engine/src/test/java/org/hibernate/validator/test/cfg/ValidatorConstraintMappingTest.java
fenrhil: Agreed, only one call to `includeExistingValidators()` would make sense. But still, it's possible to call it multiple times, so the behavior should be defined and tested.
I am not considering the (rather complex) behavior you just described.
In this test, I'm simply ensuring that calling `.includeExistingValidators( true ).includeExistingValidators( false )` produces the exact same effect than just calling `.includeExistingValidators( false )`. So I just test that 
What I tried to say in my previous message is that, if there was a bug in the tested code, and somehow the first value (true) was taken into account instead of the second (false), then the test would throw an unexpected exception because we would have two validators for the same type. So the test would fail.

fenrhil:
Most issues have been tackled in the last few commits. I tried to keep the commits atomic to make it easier to review, but I can squash everything whenever you want.

Most of the work was in https://github.com/hibernate/hibernate-validator/pull/468/commits/14c557b3abdb2b4f95ad252b43aecb3d820c3a11 and tackled those two issues (which were quite linked):

 * Making it impossible to add constraints more than once through the  programmatic API:https://github.com/hibernate/hibernate-validator/pull/468#discussion-diff-62456987
 * Making it possible to override XML constraint definitions through the programmatic API: https://github.com/hibernate/hibernate-validator/pull/468#discussion-diff-62456955

Here are the subjects still pending (no agreement was reached, or so it seems):

 * A discussion about some tests I wrote: https://github.com/hibernate/hibernate-validator/pull/468/files/c55308612b2c3d995a9aba5470fab427f249cc11#r62457852
 * A discussion about some tests I removed: https://github.com/hibernate/hibernate-validator/pull/468#discussion-diff-62465276

engine/src/main/java/org/hibernate/validator/constraints/URL.java
```diff
@@ -40,22 +40,18 @@
...
  * <p>
  * In case URLs with non default protocol handlers need to be validated, Hibernate Validator can be configured to use
- * a regular expression based URL validator only. This can be done programmatically via a {@code ConstraintDefinitionContributor}:
+ * a regular expression based URL validator only. This can be done programmatically via a {@code org.hibernate.validator.cfg.ConstraintMapping}:
  * <pre>
  * {@code
- * HibernateValidatorConfiguration configuration = Validation
- *         .byProvider( HibernateValidator.class )
- *         .configure();
+ * HibernateValidatorConfiguration config = ValidatorUtil.getConfiguration( HibernateValidator.class );
```
gunnarmorling: I'll revert this line as `ValidatorUtil` is a test-only utility.

gunnarmorling:
Rebased and applied. Many thanks, @fenrhil, great contribution!

PR #469, opened 2016-06-02
HV-1068 Deriving MessageDescriptorFormatException from correct Valida…

gunnarmorling:
…tionException type

@gsmet A quick one :)


gsmet:
Rebased and applied! Thanks!

PR #470, opened 2016-06-10
HV-1070

gunnarmorling:
Hey @gsmet, feedback welcome! The WF integration tests and the docs still don't build on JDK 9. The other stuff should do. Note you need set this before running `mvn clean install` on 9:

    export MAVEN_OPTS="-addmods java.annotations.common,java.xml.bind"

I think we still could merge it, also if some issues remain. But engine and the TCK runner pass which are the most important ones.


gsmet:
> HV-1070 Avoiding split package "testutil"

Cannot say I'm very fond of this change. Would probably have been nicer to have "engine" integrated in the package name from the start but too late for that.

gsmet:
OK, I fixed the case of ${surefire.argline.extension} and merged it.

Thanks!

gsmet:
@gunnarmorling do you want me to close https://hibernate.atlassian.net/browse/HV-1070 or do you have more work planned on it? As you haven't attached the PR to the issue, I'm not sure what you have in mind. Thanks!

PR #475, opened 2016-06-17
HV-1076 Fix Checkstyle configuration and all the violations.

gsmet:
https://hibernate.atlassian.net/browse/HV-1076

Better merge this fast to avoid merge conflicts.

The checkstyle.xml is now mostly similar to the Search one. We also use the check goal instead of checkstyle as it finds more violations (similar to what we did in Search recently).


gunnarmorling:
Weird that we not did that before.

gunnarmorling:
Rebased and applied. Thanks, @gsmet, for cleaning-up everything!

PR #476, opened 2016-06-20
HV-1070 Making build more friendly towards Java 9

gunnarmorling:


gunnarmorling:
@gsmet Another quick one :)

distribution/pom.xml
```diff
@@ -95,6 +95,7 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <configuration>
+                    <additionalJOption>-addmods java.xml.bind</additionalJOption>
```
gsmet: This breaks the build on Java 8. It should be added in a profile (or removed as you claim we do not support building the distribution with Java 9).

Thanks!

gunnarmorling: Doh, yes. Pushed an update.

gsmet:
Rebased and merged! Thanks!

PR #477, opened 2016-06-20
HV-1062 Support property rule and type use rule together on a Map

gsmet:
https://hibernate.atlassian.net/browse/HV-1062

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
@@ -258,15 +258,15 @@ public class AnnotationMetaDataProvider implements MetaDataProvider {
 	}
 
 	private UnwrapMode unwrapMode(Field field, boolean typeArgumentAnnotated) {
-		boolean notIterable = !ReflectionHelper.isIterable( ReflectionHelper.typeOf( field ) );
+		boolean indexable = ReflectionHelper.isIndexable( ReflectionHelper.typeOf( field ) );
 		UnwrapValidatedValue unwrapValidatedValue = field.getAnnotation( UnwrapValidatedValue.class );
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -278,8 +278,8 @@ private UnwrapMode unwrapMode(ExecutableElement executable, boolean typeArgument
 		return constraints;
 	}
 
-	private UnwrapMode unwrapMode(boolean typeArgumentAnnotated, boolean notIterable, UnwrapValidatedValue unwrapValidatedValue) {
-		if ( unwrapValidatedValue == null && typeArgumentAnnotated && notIterable ) {
+	private UnwrapMode unwrapMode(boolean typeArgumentAnnotated, boolean notIndexable, UnwrapValidatedValue unwrapValidatedValue) {
+		if ( unwrapValidatedValue == null && typeArgumentAnnotated && notIndexable ) {
```
gunnarmorling: Can we make it a "positive" variable name? It's easier to grok that way, and more idiomatic: `!indexable`.

engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
```diff
@@ -253,6 +253,28 @@ public static Object getValue(Method method, Object object) {
 	}
 
 	/**
+	 * Indicates if the type is considered indexable (ie is an <code>Iterable</code>, an array or a <code>Map</code>).
```
gunnarmorling: Could you use `{@code ...}` over `<code>...</code>` (as per http://hibernate.org/validator/contribute/)? If it's the other style still in some places, that's left-overs from earlier times (fix stuff you see as you go, if you like).

gunnarmorling:
@gsmet Nice fix. Added two minor comments inline.

gsmet:
@gunnarmorling I addressed your comments.

gunnarmorling:

PR #481, opened 2016-08-11
HV-1082 Upgrade maven-jdocbook-plugin and pressgang

gsmet:
It allows to generate valid HTML5 output and fixes a bug where the user
was not brought to the right place when clicking on an anchor.


gunnarmorling:
Merged; thanks, @gsmet!

PR #482, opened 2016-08-12
HV-1083 Properly ignore hibernate-noorm-release-scripts directory

gsmet:
* https://hibernate.atlassian.net/browse/HV-1083

### DIFF
--- a/distribution/src/main/assembly/dist.xml
+++ b/distribution/src/main/assembly/dist.xml
@@ -55,6 +55,10 @@
             <outputDirectory>/</outputDirectory>
         </file>
         <file>
+            <source>../CONTRIBUTING.md</source>
+            <outputDirectory>/</outputDirectory>
+        </file>
+        <file>
             <source>../changelog.txt</source>
             <outputDirectory>/</outputDirectory>
         </file>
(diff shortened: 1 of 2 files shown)


distribution/src/main/assembly/dist.xml
```diff
@@ -85,11 +90,17 @@
...
                 <exclude>**/zanata.xml</exclude>
                 <exclude>src/main/scripts/**</exclude>
+                <exclude>hibernate-noorm-release-scripts/**</exclude>
 
                 <!-- actual files which should be ignored -->
+                <exclude>.travis.yml</exclude>
+                <exclude>travis/**</exclude>
                 <exclude>.git</exclude>
                 <exclude>.gitignore</exclude>
                 <exclude>**/target/**</exclude>
```
yrodiere: Might be a bit agressive, isn't it? I hope we don't have a package named "target" somewhere...

gunnarmorling: Good thinking; we had the same concern in MapStruct and added a [CS rule](https://github.com/mapstruct/mapstruct/blob/master/build-config/src/main/resources/build-config/checkstyle.xml#L55) to avoid such package names. I'll do the same here and merge.

gunnarmorling:
Rebased, applied and added one more commit as discussed above. Thanks, @gsmet!

PR #487, opened 2016-08-24
HV-1013 Avoiding call of hashCode() of validated objects in caching t…

gunnarmorling:
…raversable resolver

@hferentschik You may be interested. Actually it's doing the same now as `ValidationContext.BeanAndPath`.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/resolver/CachingTraversableResolverForSingleValidation.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/resolver/CachingTraversableResolverForSingleValidation.java
@@ -9,6 +9,7 @@ package org.hibernate.validator.internal.engine.resolver;
 import java.lang.annotation.ElementType;
 import java.util.HashMap;
 import java.util.Map;
+
 import javax.validation.Path;
 import javax.validation.TraversableResolver;
 
@@ -20,8 +21,8 @@ import javax.validation.TraversableResolver;
  * @author Emmanuel Bernard
  */
 public class CachingTraversableResolverForSingleValidation implements TraversableResolver {
-	private TraversableResolver delegate;
-	private Map<TraversableHolder, TraversableHolder> traversables = new HashMap<TraversableHolder, TraversableHolder>();
+	private final TraversableResolver delegate;
+	private final Map<TraversableHolder, TraversableHolder> traversables = new HashMap<TraversableHolder, TraversableHolder>();
 
 	public CachingTraversableResolverForSingleValidation(TraversableResolver delegate) {
 		this.delegate = delegate;
@@ -142,7 +143,9 @@ public class CachingTraversableResolverForSingleValidation implements Traversabl
 		}
 
 		public int buildHashCode() {
-			int result = traversableObject != null ? traversableObject.hashCode() : 0;
+			// HV-1013 Using identity hash code in order to avoid calling hashCode() of objects which may
+			// be handling null properties not correctly
+			int result = traversableObject != null ? System.identityHashCode( traversableObject ) : 0;
 			result = 31 * result + traversableProperty.hashCode();
 			result = 31 * result + rootBeanType.hashCode();
 			result = 31 * result + pathToTraversableObject.hashCode();


gunnarmorling:
This one should be merged into 5.3, too.

gsmet:
Mmmh, there are 2 different changes here.

I agree with the first one (the System.identityHashCode change).

But I'm not sure I follow you on the second one. Are you sure using just the object and the property node as key for the cache is sufficient? It makes me a little nervous for 2 reasons:
- the values cached (reachable and cascadable) depend on all the other parameters as we call the methods with all the parameters,
- I would have expected the full property path to be meaningful and the leaf property node does not look sufficient to me.

gunnarmorling:
About your concern, I see where you are coming from.

I can't see a case really where the outcome for the same object and property would be different, so I believe the other params are passed as some means of context (either a given reference is loaded or not, no matter the element type of the association or what's the root of the current traversal). But I'll revert that part of the change to be on the safe side.

gunnarmorling:
@gsmet Force pushed an update.

gsmet:
Merged! Thanks!

gsmet:
And cherry-picked to 5.3.

fcamblor:
I'm still getting an error (but with different stacktrace) after an upgrade to `5.4.1.Final` :
```
	foo.bar.web.assets.AssetsController$DynamicSVGParams.hashCode(AssetsController.java:134)
	org.hibernate.validator.internal.engine.ConstraintViolationImpl.createHashCode(ConstraintViolationImpl.java:328)
	org.hibernate.validator.internal.engine.ConstraintViolationImpl.<init>(ConstraintViolationImpl.java:165)
	org.hibernate.validator.internal.engine.ConstraintViolationImpl.forBeanValidation(ConstraintViolationImpl.java:56)
	org.hibernate.validator.internal.engine.ValidationContext.createConstraintViolation(ValidationContext.java:344)
	org.hibernate.validator.internal.engine.ValidationContext.createConstraintViolations(ValidationContext.java:260)
	org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateSingleConstraint(ConstraintTree.java:454)
	org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:127)
	org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:87)
	org.hibernate.validator.internal.metadata.core.MetaConstraint.validateConstraint(MetaConstraint.java:73)
	org.hibernate.validator.internal.engine.ValidatorImpl.validateMetaConstraint(ValidatorImpl.java:616)
	org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraint(ValidatorImpl.java:581)
	org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForSingleDefaultGroupElement(ValidatorImpl.java:527)
	org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForDefaultGroup(ValidatorImpl.java:495)
	org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForCurrentGroup(ValidatorImpl.java:460)
	org.hibernate.validator.internal.engine.ValidatorImpl.validateInContext(ValidatorImpl.java:410)
	org.hibernate.validator.internal.engine.ValidatorImpl.validate(ValidatorImpl.java:207)
	org.springframework.validation.beanvalidation.SpringValidatorAdapter.validate(SpringValidatorAdapter.java:253)
```

My `DynamicSVGParams` class looks like this :
```
    public static class DynamicSVGParams {
        @NotNull
        String path;
        @NotNull
        String color;

        @AssertTrue
        @JsonIgnore
        public boolean isSvgPathValid() {
            // That's an important security constraint, otherwise, we're going to be able to serve any file in the WEB-INF
            // folder (like configuration files for instance...)
            return path.endsWith(".svg");
        }

        @AssertTrue
        @JsonIgnore
        public boolean isSvgColorValid() {
            return Colors.isValidHexColor(color);
        }

        public String getPath() {
            return path;
        }

        public void setPath(String path) {
            this.path = path;
        }

        public String getColor() {
            return color;
        }

        public void setColor(String color) {
            this.color = color;
        }

        @Override
            if (this == o) return true;
            if (o == null || getClass() != o.getClass()) return false;

            DynamicSVGParams that = (DynamicSVGParams) o;

            if (!path.equals(that.path)) return false;
            return color.equals(that.color);

        }

        @Override
        public int hashCode() {
            int result = path.hashCode();
            result = 31 * result + color.hashCode();
            return result;
        }
    }
```

gsmet:
It is a different issue: https://hibernate.atlassian.net/browse/HV-1373 .

It has been fixed in 6.0.4.Final. Could you consider an upgrade to the latest 6.0.x? We are at 6.0.16.Final right now.

fcamblor:
Wow thanks for the quick response and the pointer (I tried some search but didn't found it, sorry for the useless noise)

Will try to upgrade to BV 2.0 and HV 6.x thanks :-)

PR #491, opened 2016-08-28
HV-1066 email domains with a 63 character length fails EmailValidator

marko-bekhta:
changed the way unicode string is split in  cases when it's length > 63 (the limit for IDN.toASCII). Also updated unit test cases to check for strings of different lengths

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
@@ -36,7 +36,7 @@ public class EmailValidator implements ConstraintValidator<Email, CharSequence> 
 	private static final String IP_DOMAIN = "\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\]";
 	private static final int MAX_LOCAL_PART_LENGTH = 64;
 	private static final int MAX_DOMAIN_PART_LENGTH = 255;
-	private static final int ASCII_MAX_LABEL_LENGTH = 63;
+	private static final int IDN_TO_ASCII_MAX_LABEL_LENGTH = 63;
 
 	/**
 	 * Regular expression for the local part of an email address (everything before '@')
@@ -105,8 +105,7 @@ public class EmailValidator implements ConstraintValidator<Email, CharSequence> 
 	private String toAscii(String unicodeString) throws IllegalArgumentException {
 		String asciiString = "";
 		int start = 0;
-		int index = getBreakingIndex( unicodeString, ASCII_MAX_LABEL_LENGTH );
-		int end = unicodeString.length() <=  index ? unicodeString.length() : index;
+		int end = getIDNToASCIIPartEnd( unicodeString, IDN_TO_ASCII_MAX_LABEL_LENGTH);
 		while ( true ) {
 			// IDN.toASCII only supports a max "label" length of 63 characters. Need to chunk the input in these sizes
 			asciiString += IDN.toASCII( unicodeString.substring( start, end ) );
@@ -114,16 +113,15 @@ public class EmailValidator implements ConstraintValidator<Email, CharSequence> 
 				break;
 			}
 			start = end;
-			index = getBreakingIndex( unicodeString, start + ASCII_MAX_LABEL_LENGTH );
-			end = index > unicodeString.length() ? unicodeString.length() : index;
+			end = getIDNToASCIIPartEnd( unicodeString, start + IDN_TO_ASCII_MAX_LABEL_LENGTH);
 		}
 
 		return asciiString;
 	}
 
-	private int getBreakingIndex(String unicodeString, int index) {
+	private int getIDNToASCIIPartEnd(String unicodeString, int index) {
 		if (unicodeString.length() <= index) {
-			return index;
+			return unicodeString.length();
 		}
 		return unicodeString.charAt( index ) == '.' ? Math.max( 0, index - 1 ) : index;
 	}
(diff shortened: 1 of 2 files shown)


gsmet:
ok to test

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
```diff
@@ -36,6 +36,7 @@
 	private static final String IP_DOMAIN = "\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\]";
 	private static final int MAX_LOCAL_PART_LENGTH = 64;
 	private static final int MAX_DOMAIN_PART_LENGTH = 255;
+	private static final int ASCII_MAX_LABEL_LENGTH = 63;
```
gsmet: Could you rename it IDN_TO_ASCII_MAX_LABEL_LENGTH?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
```diff
@@ -104,17 +105,26 @@ private boolean matchPart(String part, Pattern pattern, int maxLength) {
 	private String toAscii(String unicodeString) throws IllegalArgumentException {
 		String asciiString = "";
 		int start = 0;
-		int end = unicodeString.length() <= 63 ? unicodeString.length() : 63;
+		int index = getBreakingIndex( unicodeString, ASCII_MAX_LABEL_LENGTH );
+		int end = unicodeString.length() <=  index ? unicodeString.length() : index;
```
gsmet: Couldn't you make the 2 lines above a single method instead of having 2 parts and repeating it below? Something like `getIDNToASCIIPartEnd(String unicodeString, int start)`.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
```diff
@@ -104,17 +105,26 @@ private boolean matchPart(String part, Pattern pattern, int maxLength) {
...
 		}
 
 		return asciiString;
 	}
+
+	private int getBreakingIndex(String unicodeString, int index) {
+		if (unicodeString.length() <= index) {
+			return index;
+		}
+		return unicodeString.charAt( index ) == '.' ? Math.max( 0, index - 1 ) : index;
```
gsmet: So, as I understand it, there is an issue if the string ends with a dot, but not if the string starts with a dot? It would be nice to check this case too.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
```diff
@@ -154,10 +154,12 @@ public void testEMailWithTrailingAt() throws Exception {
 	}
 
 	@Test
-	@TestForIssue(jiraKey = "HV-1005")
+	@TestForIssue(jiraKey = {"HV-1005", "HV-1066"})
 	public void testEmailWith64CharacterLocalPartIsValid() {
 		// Local part should allow up to 64 octets: https://tools.ietf.org/html/rfc5321#section-4.5.3.1.1
-		isValidEmail( stringOfLength( 64 ) + "@foo.com" );
+		for (int num = 1; num < 64; num++) {
```
gsmet: You're missing 64 here. `<=` would be better.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
```diff
@@ -167,10 +169,12 @@ public void testEmailWith65CharacterLocalPartIsInvalid() {
 	}
 
 	@Test
-	@TestForIssue(jiraKey = "HV-1005")
+	@TestForIssue(jiraKey = {"HV-1005", "HV-1066"})
 	public void testEmailWith255CharacterDomainPartIsValid() {
 		// Domain part should allow up to 255
-		isValidEmail( "foo@" + stringOfLength( 251 ) + ".com" );
+		for (int num = 1; num < 251; num++) {
```
gsmet: Same here, you're missing 251.

gsmet:
@marko-bekhta Thanks for your patch. See my comments above.

Don't worry about the CI, we are transitioning to a new CI, that's why you have build failures.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
marko-bekhta: @gsmet 
actually the issue is when the string passed starts with the dot. As in the test case for this issue the domain part of the email was split in two parts - first part contained a series of 'a' chars and second part was just '.com' and when this '.com' was passed to `IDN.toASCII` it was throwing an exception. 
So what I've done to overcome the problem - is to check if we are not trying to split  a string in such "wrong" way and at least leave one character before the '.' That's why I return `index - 1` in such case (as substring second parameter index is exclusive we will leave at least one character before '.')

hope I've explained it well :)

gsmet:
OK, I've put some more thinking into it and I think the approach of splitting the domain in chunks was a bad idea in the first place.

Typically, your approach will fail if there are 2 dots in a row just placed where they shouldn't be.

Apart from that, using IDN.toASCII on chunks is not equivalent to using IDN.toASCII on the whole string.

I'm preparing a PR along those lines. I'm closing this one for now.

marko-bekhta:
Yes, I thought about the two dots in row as well - but in that case the domain will be invalid (containing two dots) and the validation will fail which should be expected. But if you found a better solution it's fine :) 
I was happy to provide some input to the issue :)

gsmet:
@marko-bekhta input welcome on the PR I just opened. Thanks for your input!

gsmet:
I think the thing that was really bothering me is that splitting arbitrarily the domain name to run IDN.toASCII on each part might lead to an invalid domain (for instance because of the length of a label or the length of the whole domain name) while the domain is perfectly valid. I'm especially worried about the fact that IDN.toASCII adds a `xn--` prefix at the beginning of the domain. If working on n chunks containing non ASCII characters, you add it n times instead of only one time at the beginning of the whole domain name.

PR #492, opened 2016-08-29
HV-1066 Don't use IDN.toASCII on the local part of the email

gsmet:
* https://hibernate.atlassian.net/browse/HV-1066

Using IDN.toASCII on the local part of the email was a bad idea to begin
with, forcing us to introduce a splitting logic on top of IDN.toASCII.

Executing IDN.toASCII on each chunk of a split string is not equivalent
to executing it on the whole string, potentially opening the way for
other bugs.

Thus, we ended up doing the following:
- allow the \u0080-\uFFFF character range in the local part of the email
  thus allowing local parts containing UTF-8 characters;
- only use IDN.toASCII on the domain name as it should be.

This simplifies the logic a lot and should limit the number of bugs in
this area.

In passing:
- improve the feedback in email validation tests;
- be more cautious about how we generate the domain names in tests: each
  label of a domain name must be at most 63 characters long.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
@@ -6,15 +6,16 @@
  */
 package org.hibernate.validator.internal.constraintvalidators.hv;
 
+import static java.util.regex.Pattern.CASE_INSENSITIVE;
+
 import java.net.IDN;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
 
 import org.hibernate.validator.constraints.Email;
-
-import static java.util.regex.Pattern.CASE_INSENSITIVE;
 
 /**
  * Checks that a given character sequence (e.g. string) is a well-formed email address.
@@ -29,25 +30,31 @@ import static java.util.regex.Pattern.CASE_INSENSITIVE;
  *
  * @author Emmanuel Bernard
  * @author Hardy Ferentschik
+ * @author Guillaume Smet
  */
 public class EmailValidator implements ConstraintValidator<Email, CharSequence> {
-	private static final String ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~-]";
-	private static final String DOMAIN = ATOM + "+(\\." + ATOM + "+)*";
+	private static final String LOCAL_PART_ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~\u0080-\uFFFF-]";
+	private static final String DOMAIN_LABEL = "[a-z0-9!#$%&'*+/=?^_`{|}~-]";
+	private static final String DOMAIN = DOMAIN_LABEL + "+(\\." + DOMAIN_LABEL + "+)*";
 	private static final String IP_DOMAIN = "\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\]";
 	private static final int MAX_LOCAL_PART_LENGTH = 64;
+	/**
+	 * This is the maximum length of a domain name. But be aware that each label (parts separated by a dot) of the
+	 * domain name must be at most 63 characters long. This is verified by {@link IDN#toASCII(String)}.
+	 */
 	private static final int MAX_DOMAIN_PART_LENGTH = 255;
 
 	/**
 	 * Regular expression for the local part of an email address (everything before '@')
 	 */
-	private final Pattern localPattern = java.util.regex.Pattern.compile(
-			ATOM + "+(\\." + ATOM + "+)*", CASE_INSENSITIVE
+	private static final Pattern LOCAL_PART_PATTERN = Pattern.compile(
+			LOCAL_PART_ATOM + "+(\\." + LOCAL_PART_ATOM + "+)*", CASE_INSENSITIVE
 	);
 
 	/**
 	 * Regular expression for the domain part of an email address (everything after '@')
 	 */
-	private final Pattern domainPattern = java.util.regex.Pattern.compile(
+	private static final Pattern DOMAIN_PATTERN = Pattern.compile(
 			DOMAIN + "|" + IP_DOMAIN, CASE_INSENSITIVE
 	);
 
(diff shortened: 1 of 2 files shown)


gsmet:
@marko-bekhta thoughts?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
```diff
@@ -29,25 +29,31 @@
...
+	 * domain name must be at most 63 characters long. This is verified by {@link IDN#toASCII(String)}.
+	 */
 	private static final int MAX_DOMAIN_PART_LENGTH = 255;
 
 	/**
 	 * Regular expression for the local part of an email address (everything before '@')
 	 */
-	private final Pattern localPattern = java.util.regex.Pattern.compile(
-			ATOM + "+(\\." + ATOM + "+)*", CASE_INSENSITIVE
+	private static final Pattern LOCAL_PART_PATTERN = java.util.regex.Pattern.compile(
```
gunnarmorling: Not strictly related, but there should be no need for the FQN?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
```diff
@@ -29,25 +29,31 @@
  *
  * @author Emmanuel Bernard
  * @author Hardy Ferentschik
+ * @author Guillaume Smet
  */
 public class EmailValidator implements ConstraintValidator<Email, CharSequence> {
-	private static final String ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~-]";
-	private static final String DOMAIN = ATOM + "+(\\." + ATOM + "+)*";
+	private static final String LOCAL_PART_ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~\u0080-\uFFFF-]";
```
gunnarmorling: So there is a wider range of characters allowed in the local part than in the domain part?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
```diff
@@ -70,51 +76,42 @@ public boolean isValid(CharSequence value, ConstraintValidatorContext context) {
...
+		Matcher matcher = DOMAIN_PATTERN.matcher( asciiString );
 		return matcher.matches();
 	}
 
-	private String toAscii(String unicodeString) throws IllegalArgumentException {
-		String asciiString = "";
-		int start = 0;
-		int end = unicodeString.length() <= 63 ? unicodeString.length() : 63;
-		while ( true ) {
-			// IDN.toASCII only supports a max "label" length of 63 characters. Need to chunk the input in these sizes
```
gunnarmorling: How comes that this chunking isn't needed any longer?

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
```diff
@@ -190,25 +202,41 @@ private String stringOfLength(int length) {
...
 	}
 
 	private void isValidEmail(CharSequence email, String message) {
-		assertTrue( validator.isValid( email, null ), message );
+		assertTrue( validator.isValid( email, null ), String.format( message, email ) );
 	}
 
 	private void isValidEmail(CharSequence email) {
-		isValidEmail( email, "Expected a valid email." );
+		isValidEmail( email, "Expected a valid email for %1$s." );
```
gunnarmorling: Sounds a bit odd, better use "Expected %1$s to be a valid email".

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
```diff
@@ -190,25 +202,41 @@ private String stringOfLength(int length) {
...
 	}
 
 	private void isInvalidEmail(CharSequence email, String message) {
-		assertFalse( validator.isValid( email, null ), message );
+		assertFalse( validator.isValid( email, null ), String.format( message, email ) );
 	}
 
 	private void isInvalidEmail(CharSequence email) {
-		isInvalidEmail( email, "Expected a invalid email." );
+		isInvalidEmail( email, "Expected an invalid email for %1$s." );
```
gunnarmorling: Same here.

gunnarmorling:
Added some comments/questions.

Would also be great to hear back from @marko-bekhta whether this solution works for you. Thanks!

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
gsmet: Nope, removed them.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
gsmet: Not really.

The local part accepts this new range of character as valid.

The domain is first converted to ASCII using IDN.toASCII (which verifies that everything is legit, UTF-8, label length and so on), then is validated with the regexp.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
gsmet: The chunking was introduced because the local part was mistakenly passed to IDN.toASCII to be converted from UTF-8 to ASCII.

The local part max length is 64 characters while each label of a domain name (parts separated by dots) must be at most 63 characters long. The chunking was introduced to solve this issue. See https://hibernate.atlassian.net/browse/HV-1005 and https://github.com/hibernate/hibernate-validator/pull/422/files .

The fundamental error was to apply IDN.toASCII to the local part in the first place, it's not supposed to be used on other things than domain names where the 63 characters limit on each label is perfectly legit.

So I did the following:
- allow the UTF-8 range that is now allowed in the local part regexp
- only use IDN.toASCII to validate domain names - and this part does not need chunking anymore

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
gsmet: Right, fixed.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
gsmet: Yup!

gsmet:
Force pushed an update.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
gunnarmorling: Ah, got it. `toAscii()` can deal with individual labels or entire domains.

gunnarmorling:
Rebased and pushed to 5.3 as well as master (6.0).

@marko-bekhta it'd be great if you could check out out, build (just run `mvn clean install` or refer to the readme.md) and give it a test run. Thx!

marko-bekhta:
@gunnarmorling done that, all works great 👍 also looked through the code and - now I think the problem is solved correctly, thanks @gsmet

gunnarmorling:
Cool, thanks for reporting back! We'll do a 5.3.0.CR1 release with this fix
within the next days.

PR #494, opened 2016-08-29
HV-1093 Update the CI links

gsmet:
* https://hibernate.atlassian.net/browse/HV-1093

When I preview the committed README.md, I got an image not found on the Build passing image. The link is OK though. Any idea?

If we can't make it work, we might as well remove it, it's not that useful.


gunnarmorling:
Rebased, applied andd added one more commit to use the "protected" link for the build status. It still is visible to everyone. The other link wouldn't work for me either.

PR #500, opened 2016-08-31
HV-1014 Clarify what @ValidPart is in example 2.3 of the reference guide

gsmet:
* https://hibernate.atlassian.net/projects/HV/issues/HV-1014

-> Should be backported to 5.3.


gunnarmorling:
Applied to master and 5.3. Thanks, @gsmet!

PR #501, opened 2016-09-01
HV-763 Use @FormatWith consistently in logging API

gsmet:
* https://hibernate.atlassian.net/browse/HV-763

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
@@ -5,11 +5,18 @@
  * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
  */
 package org.hibernate.validator.internal.util.logging;
+import static org.jboss.logging.Logger.Level.DEBUG;
+import static org.jboss.logging.Logger.Level.INFO;
+import static org.jboss.logging.Logger.Level.WARN;
+
+import java.lang.annotation.Annotation;
 import java.lang.annotation.ElementType;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Member;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
+import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 import java.util.regex.PatternSyntaxException;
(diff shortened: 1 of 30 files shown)


gsmet:
Jenkins, please retest this

gsmet:
Jenkins, retest this please

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -228,10 +246,10 @@ UnexpectedTypeException getNoValidatorFoundForTypeException(String constraintTyp
 	@Message(id = 63, value = "Unable to instantiate %s.")
 	ValidationException getUnableToInstantiateException(String className, @Cause Exception e);
```
gunnarmorling: What about this one?

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -228,10 +246,10 @@ UnexpectedTypeException getNoValidatorFoundForTypeException(String constraintTyp
...
 
-	ValidationException getUnableToInstantiateException(Class<?> clazz, @Cause Exception e);
```
gunnarmorling: Hum, no `@Message` annotation? Something is odd here.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -387,11 +411,13 @@ UnexpectedTypeException getNoValidatorFoundForTypeException(String constraintTyp
...
 	@Message(id = 118, value = "Unable to cast %s (with element kind %s) to %s")
-	ClassCastException getUnableToNarrowNodeTypeException(String actualDescriptorType, ElementKind kind, String expectedDescriptorType);
+	// TODO
+	ClassCastException getUnableToNarrowNodeTypeException(@FormatWith(ClassObjectFormatter.class) Class<?> actualDescriptorType, ElementKind kind,
+			@FormatWith(ClassObjectFormatter.class) Class<?> expectedDescriptorType);
 
 	@LogMessage(level = INFO)
 	@Message(id = 119, value = "Using %s as parameter name provider.")
-	void usingParameterNameProvider(String parameterNameProviderClassName);
+	void usingParameterNameProvider(@FormatWith(ClassObjectFormatter.class) Class<ParameterNameProvider> parameterNameProviderClass);
```
gunnarmorling: Wondering, wouldn't it technically be `Class<? extends ParameterNameProvider>`?

engine/src/main/java/org/hibernate/validator/internal/util/logging/formatter/CollectionOfClassesObjectFormatter.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Used with JBoss Logging to display collection of class names in log messages.
+ *
+ * @author Guillaume Smet
+ */
+public class CollectionOfClassesObjectFormatter {
+
+	private final String stringRepresentation;
+
+	public CollectionOfClassesObjectFormatter(Collection<Class<?>> classes) {
```
gunnarmorling: Or  this:

    stringRepresentation = classes.stream()
        .map( c -> c.getName() )
        .collect( Collectors.joining( ", " ) );

Sorry, couldn't resist :)

gunnarmorling:
@gsmet Looks good, added some minor notes. I reckon you don't want to backport it given it's using Java 7 syntax already?

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gsmet: Sure. There are several like that. But the caller is casting the argument to Class<ParameterNameProvider>, that's why I did it like that. Maybe we should fix the caller?

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gsmet: IIRC, the error is called when it's unable to instantiate the class. Thus we can't have a class parameter :).

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gsmet: Yup, I was wondering if there was some magic to get the message from the previous one (as I would expect them to be the same)?

gsmet:
Yes, I don't think we should backport it. It does not bring anything to the user.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gsmet: I confirm there's some magic involved: the second method reuses the message of the first method. So it's weird but OK.

engine/src/main/java/org/hibernate/validator/internal/util/logging/formatter/CollectionOfClassesObjectFormatter.java
gsmet: Gunnar happy camper, it is!

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gsmet: I fixed the casts in the callers and the Log methods.

gunnarmorling:
Rebased and applied. Added one more commit to get rid of that duplicated method (it was about instantiating not loading a class by name, so we had the class object already).

PR #502, opened 2016-09-01
HV-1031 Creating copy of PathImpl during marking processed paths

gunnarmorling:


gsmet:
Merged and backported to 5.3. Thanks.

PR #503, opened 2016-09-01
HV-1065 Add hibernate.validator.constraint_mapping_contributors option

gsmet:
* https://hibernate.atlassian.net/browse/HV-1065

The current hibernate.validator.constraint_mapping_contributor option
does not accept a comma separated list of contributors which is not in
line with the other options.

Deprecate hibernate.validator.constraint_mapping_contributor. Will be
removed in 6.

### DIFF
--- a/engine/src/test/resources/org/hibernate/validator/test/cfg/constraint-mapping-contributor-validation.xml
+++ b/engine/src/test/resources/org/hibernate/validator/test/cfg/constraint-mapping-contributor-validation.xml
@@ -11,5 +11,6 @@
     xsi:schemaLocation="http://jboss.org/xml/ns/javax/validation/configuration validation-configuration-1.1.xsd"
     version="1.1">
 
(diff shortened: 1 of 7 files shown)


engine/src/test/resources/org/hibernate/validator/test/cfg/constraint-mapping-contributor-validation.xml
```diff
@@ -11,5 +11,5 @@
     xsi:schemaLocation="http://jboss.org/xml/ns/javax/validation/configuration validation-configuration-1.1.xsd"
     version="1.1">
 
-    <property name="hibernate.validator.constraint_mapping_contributor">org.hibernate.validator.test.cfg.ConstraintMappingContributorConfiguredInValidationXmlTest$MyConstraintMappingContributor</property>
```
gunnarmorling: Is there still a test for using the existing option?

gunnarmorling:
@gsmet Looks good but could you restore the test for using the old prop for the time being (if it has vanished) and then merge yourself? Thx!

PR #504, opened 2016-09-01
HV-1097 Remove hibernate.validator.constraint_mapping_contributor option

gsmet:
* https://hibernate.atlassian.net/browse/HV-1097

Also contains HV-1065 which has been applied to 5.3 in another PR.


gunnarmorling:
Rebased and applied.

PR #509, opened 2016-09-02
HV-763 Removing redundant logging method

gunnarmorling:
Small left-over from https://hibernate.atlassian.net/browse/HV-763.


gsmet:
Merged! Thanks!

(still not convinced your automatic replacement of diamonds is a good idea)

gunnarmorling:
> (still not convinced your automatic replacement of diamonds is a good
idea)

But that one wasn't meant to be backported, so it should be fine?

gsmet:
Sure. But it makes each patch less readable.

gunnarmorling:
Ok, now I see that this is your concern. I feel it's digestable if it's
just a small number (similar to import re-org), but I could do separate
commits if you prefer it?

PR #510, opened 2016-09-05
HV-1098 Adapting test.policy file

gunnarmorling:


gunnarmorling:
@gsmet as stated in the issue, it's not totally clear to me why this only now became a problem. Does the TCK pass (with SM enabled) for you without this fix? If so, could you debug into {{org.hibernate.beanvalidation.tck.tests.bootstrap.ValidationProviderTest.testFirstMatchingValidationProviderResolverIsReturned()}} and see how the permissions check for accessing the system property in {{org.hibernate.beanvalidation.tck.util.TestUtil.instantiateValidationProviderUnderTest()}} looks like? If you can't reproduce the issue, I wouldn't bother, the fix should do no harm either way. Thx!

gsmet:
Merged! Thanks!

PR #513, opened 2016-09-15
HV-1001 @NotEmpty and @NotBlank definition is not consistent with other constraints

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1001

Updated @NotEmpty and @NotBlank annotations to not use @NotNull and added a property canBeNull to allow null values to be valid.


gunnarmorling:
Hi @marko-bekhta, first off, many thanks for this contribution and your effort.

I'm having some general doubts about the usefulness of HV-1001, though (it's only now that I see it, otherwise I think we'd have it closed already). I see these constraints as some specialization of `@NotNull`, hence the general recommendation to consider null as valid value does not really apply to these two constraints. They are meant as shortcuts really. If one want to consider null valid, one always can use `@Size(min=1)` (admittedly, it's not that easy for `@NotBlank`).

All in all I'd rather have us reject this issue and leave things as are. That said, your contribution is appreciated very much and I can dig out some other issues to work on if you like (though I think @gsmet gave you some already?). WDYT?

marko-bekhta:
Hi @gunnarmorling, it's up to you to decide what to do with the issue and this pull request right :) 
Even if it will be rejected, for me it's ok, as I've learned a thing or two while working on it :)

And yes, @gsmet provided me with the list of items (filtered with 'Suitable for new contributors'), so I'll look through those and try find something more to work on. Or if you have any other issues or some issue in particular for me to work on I'd be glad to try.

gunnarmorling:
Ok, then I'd say let's close this issue and the PR. Glad to hear that you are up for other stuff nevertheless, that's great!

In terms of issues to work on, @gsmet surely has compiled a nice list. I've also tagged some issues with the "hackergarten" label which are easy starters: https://hibernate.atlassian.net/issues/?jql=project%20%3D%20HV%20AND%20labels%20%3D%20hackergarten

Feel free to tackle any issue you like. I'd only recommend you leave a short comment on the issue when you get started, so others are aware of you working on it. Looking forward to your further PRs!

PR #512, opened 2016-09-14
HV-1094 Mark constraints to support TYPE_USE

gsmet:
* https://hibernate.atlassian.net/browse/HV-1094

Note: the build will fail until the PR for beanvalidation-api has been merged and a new snapshot pushed.

PR #514, opened 2016-09-15
HV-1100 Convert the reference guide to use source includes

gsmet:
This one is on top of https://github.com/hibernate/hibernate-validator/pull/512 .

 * https://hibernate.atlassian.net/browse/HV-1100


gunnarmorling:
@gsmet do you have any idea why that removes so many more lines than it adds? On first thought I'd have said that the extraction should be neutral in terms of lines (give or take).

gsmet:
Well, not in this case, as most of the sources were already extracted. I just had to point to them (well and check they were effectively in sync).

I extracted a few classes but not much.

gunnarmorling:
Ok, makes sense then :)

gunnarmorling:
Rebased and applied. It's very hard to compare the actual files, I trust you did the right thing, @gsmet :)

PR #515, opened 2016-09-16
HV-1099 Extend tests around @Email constraint

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1099

updated regex to match quotation inside local parts and added a few more test cases for email validation

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
@@ -55,67 +55,63 @@ public class EmailValidatorTest {
 
 	@Test
 	public void testValidEmail() throws Exception {
-		Arrays.asList(
-				"emmanuel@hibernate.org",
-				"emmanuel@hibernate",
-				"emma-n_uel@hibernate",
-				"emma+nuel@hibernate.org",
-				"emma=nuel@hibernate.org",
-				"emmanuel@[123.12.2.11]",
-				"*@example.net",
-				"fred&barny@example.com",
-				"---@example.com",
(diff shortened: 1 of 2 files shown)


engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
```diff
@@ -54,36 +55,67 @@ public void testNullAndEmptyString() throws Exception {
...
+				"admin@mailserver1",
+				"#!$%&'*+-/=?^_`{}|~@example.org",
+				"\"()<>[]:,;@\\\"!#$%&'-/=?^_`{}| ~.a\"@example.org",
+				"\" \"@example.org",
+				"example@localhost",
+				"example@s.solutions",
+				"user@localserver",
+				"user@tt",
+				"user@[IPv6:2001:DB8::1]" )
+				.forEach( this::isValidEmail );
```
gsmet: I'm not convinced by this change. I agree it looks better but when you have one address failing, it's far harder to find it because you can't count on your IDE leading you to the right line.

Would you mind going back to the previous pattern?

gsmet:
Nice work!

Apart from my minor comment on the test pattern, it looks good.

I hope we don't enter a rabbit hole with the IPv6 regexp and the quoted local part :).

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/EmailValidatorTest.java
marko-bekhta: No not at all :)
done in the next commit

gsmet:
Squashed and merged! Thanks!

PR #516, opened 2016-09-16
HV-1056 Avoid usage of ReflectionHelper#getPropertyName() in validation code path

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1056

added cached property name to ConstraintLocation and replaced calls to ReflectionHelper#getPropertyName() with a getter from ConstraintLocation

@gunnarmorling is that what you had in mind ? Should it be lazy initialised ?


gsmet:
Looks good to me.

Merged! Thanks!

gunnarmorling:
Nice one, @marko-bekhta!

PR #517, opened 2016-09-16
HV-1101 Fix a few infelicities in the XML parsing of the mapping configuration

gsmet:
* https://hibernate.atlassian.net/browse/HV-1101

Should be backported to 5.3.

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/internal/xml/XmlParsingTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/internal/xml/XmlParsingTest.java
@@ -1,0 +1,71 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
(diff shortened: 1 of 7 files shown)


engine/src/test/java/org/hibernate/validator/test/internal/xml/XmlParsingTest.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Set;
+
+import javax.validation.Validator;
+import javax.validation.metadata.BeanDescriptor;
+import javax.validation.metadata.ConstraintDescriptor;
+
+import org.assertj.core.api.Assertions;
+import org.hibernate.validator.internal.metadata.descriptor.ConstraintDescriptorImpl;
+import org.hibernate.validator.testutil.TestForIssue;
+import org.junit.Test;
```
gunnarmorling: HV uses TestNG. I'll fix it. Should we add a CheckStyle check of sorts to prevent importing the JUnit annotations?

gunnarmorling:
Rebased, applied and added one more commit with some clean-ups. Thanks, @gsmet!

PR #518, opened 2016-09-19
HV-1102 Add org.junit to the Checkstyle IllegalImport rule

gsmet:
* https://hibernate.atlassian.net/browse/HV-1102
 * https://hibernate.atlassian.net/browse/HV-1107

I'm not convinced switching to TestNG and assertj is a good choice for the integration tests as we need to deploy them but, at the time, I hoped we might get everything consistent so the work was already done when I had the issue with the OSGi integration tests.

Consistency is IMHO a good thing to avoid bad surprises.

As for the OSGi integration tests, I didn't get TestNG working in Karaf (well I did get it working but it didn't find any runnable tests). Adding TestNG at Karaf deployment didn't help so I decided to let it go, especially because the tests are very concise and we probably won't add anything there.

### DIFF
--- a/build-config/src/main/resources/checkstyle-documentation.xml
+++ b/build-config/src/main/resources/checkstyle-documentation.xml
@@ -125,11 +125,9 @@
                         STATIC_INIT, INSTANCE_INIT" />
         </module>
 
-        <!-- METHOD_CALL breaks inline casts
-            (i.e. Iterable<?> collection = ((Map<?,?>) value).values();)
-            in 6.15-->
         <module name="ParenPad">
-            <property name="tokens" value="CTOR_CALL, SUPER_CTOR_CALL"/>
+            <property name="tokens" value="CTOR_CALL, METHOD_CALL, SUPER_CTOR_CALL, LITERAL_FOR, LITERAL_IF,
+                        LITERAL_WHILE, LITERAL_SWITCH, LITERAL_NEW"/>
             <property name="option" value="space"/>
         </module>
 
--- a/build-config/src/main/resources/checkstyle-documentation.xml
+++ b/build-config/src/main/resources/checkstyle-documentation.xml
@@ -1,0 +1,174 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Hibernate Validator, declare and validate application constraints
+  ~
+  ~ License: Apache License, Version 2.0
+  ~ See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+  -->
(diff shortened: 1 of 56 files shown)


build-config/src/main/resources/checkstyle-documentation.xml
```diff
@@ -0,0 +1,174 @@
...
+<!--
+  ~ Hibernate Validator, declare and validate application constraints
+  ~
+  ~ License: Apache License, Version 2.0
+  ~ See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+  -->
+<!DOCTYPE module PUBLIC "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
+        "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
+
+<module name="Checker">
```
gunnarmorling: Why is it that we need another one here?

engine/src/test/java/org/hibernate/validator/bugs/TooBigMessageTest.java
```diff
@@ -17,7 +17,7 @@
 
 import org.hibernate.validator.testutil.TestForIssue;
 import org.hibernate.validator.testutils.ValidatorUtil;
-import org.junit.Test;
+import org.testng.annotations.Test;
```
gunnarmorling: Good!

cdi/src/test/java/org/hibernate/validator/test/internal/cdi/injection/InjectionOfCustomProviderTest.java
```diff
@@ -78,43 +74,43 @@ public static JavaArchive createDeployment() {
 
 	@Test
 	public void testInjectionOfDefaultFactory() throws Exception {
-		assertNotNull( defaultValidatorFactory );
-		assertTrue( defaultValidatorFactory instanceof MyValidatorFactory );
-		assertTrue( defaultValidatorFactory.unwrap( MyValidatorFactory.class ) instanceof MyValidatorFactory );
+		assertThat( defaultValidatorFactory ).isNotNull();
```
gunnarmorling: Cool; I think we still can use the simpler form (`org.testng.Assert.assertNotNull()`) in general, i.e. no need to use the fluent form everywhere. Traditionally we tended to use the simple one for not-null and equals and the fluent ones for more complex or type-specific assertions. But I don't mind either way, just meaning to say that from my perspective there is no need to change it wherever we see it.

gunnarmorling:
Merged. Thanks, @gsmet! I've added two more commits to avoid the dependency versions in the base test class and have deployment units not being written out by default to save some build time.

PR #520, opened 2016-09-19
HV-1103 Removing remaining switches on Java version

gunnarmorling:


engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -705,7 +726,119 @@ public ConstraintType getPartition(ConstraintDescriptorImpl<?> v) {
 	 * @return a set of type arguments constraints, or an empty set if no constrained type arguments are found
 	 */
 	protected Set<MetaConstraint<?>> findTypeAnnotationConstraintsForExecutableParameter(Member member, int i) {
-		// To be extended by subclasses
-		return Collections.emptySet();
+		Parameter parameter = ( (Executable) member ).getParameters()[i];
```
gunnarmorling: That's not new code, it just has been inlined from the separate sub-class we had before.

engine/src/main/java/org/hibernate/validator/internal/util/Version.java
```diff
@@ -17,15 +14,9 @@
  * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2012 SERLI
  */
 public final class Version {
-	private static final Pattern JAVA_VERSION_PATTERN = Pattern.compile( "^(?:1\\.)?(\\d+)$" );
```
gsmet: Mmmmh, why do we remove all this? I think we should keep it. We probably will have version specific features in the future and I don't see the point in removing this?

gunnarmorling: I don't like keeping code which is unused. It's still in the repo should there be the need to resurrect it.

gsmet: I usually agree with this motto but not in this case. Anyway, I'll merge it and keep my "I'll tell you so" for later :).

gsmet:
Rebased and merged! Thanks, nice cleanup!

PR #521, opened 2016-09-20
HV-1104 Upgrade Karaf to 4.0.7 in our OSGi integration test

gsmet:
- https://hibernate.atlassian.net/browse/HV-1104

Minor but I want to see if it helps mitigate our CI issues.

### DIFF
--- a/osgi/karaf-features/src/main/features/features.xml
+++ b/osgi/karaf-features/src/main/features/features.xml
@@ -5,10 +5,10 @@
   ~ License: Apache License, Version 2.0
   ~ See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
   -->
-<features xmlns="http://karaf.apache.org/xmlns/features/v1.2.0"
+<features xmlns="http://karaf.apache.org/xmlns/features/v1.4.0"
     name="hibernate-validator-osgi-features"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://karaf.apache.org/xmlns/features/v1.2.0">
+    xsi:schemaLocation="http://karaf.apache.org/xmlns/features/v1.4.0">
 
     <feature name="hibernate-validator" version="${project.version}">
         <bundle>mvn:org.hibernate/hibernate-validator/${project.version}</bundle>
@@ -21,6 +21,7 @@
         <bundle>mvn:joda-time/joda-time/${joda-time.version}</bundle>
     </feature>
     <feature name="hibernate-validator-paranamer" version="${project.version}">
+        <feature prerequisite="true">wrap</feature>
         <feature>hibernate-validator</feature>
         <bundle>wrap:mvn:com.thoughtworks.paranamer:paranamer:${paranamer.version}</bundle>
     </feature>
--- a/pom.xml
+++ b/pom.xml
@@ -137,7 +137,7 @@
         <!-- OSGi dependencies -->
         <pax.exam.version>4.9.2</pax.exam.version>
         <pax.url.version>2.4.7</pax.url.version>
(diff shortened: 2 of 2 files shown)


gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

gsmet:
@gunnarmorling it does not solve the test issue but I think we should commit this anyway as we will need to upgrade Karaf in the future for JDK 9 support (see the tests made by Freeman Fang in https://hibernate.atlassian.net/browse/HV-1051) and we need to fix the prerequisite thing to make it work.

gunnarmorling:
Hum, this one gives me a JVM crash when running the OSGi tests:

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project hibernate-validator-osgi-integrationtest: ExecutionException The forked VM terminated without properly saying goodbye. VM crash or System.exit called?

gsmet:
Maybe it triggers the same bug we have on CI? Did you try with a more recent JVM?

gunnarmorling:
@gsmet Tried with JDK 1.8b101, same exception.

gsmet:
This is very bad news indeed.

gsmet:
@gunnarmorling could you share the error file dumped by the JVM?

gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

gunnarmorling:
> @gunnarmorling could you share the error file dumped by the JVM?

@gsmet Where would I find that file or what options would I need to create this file?

gsmet:
@gunnarmorling last time my JVM got killed like that, I had an error file in my workspace. If it did create it, a `git status` should point it to you.

gsmet:
@gunnarmorling the file is named hs_err_pidXXXX.log (with XXXX, the pid number)

gsmet:
@gunnarmorling the file is named hs_err_pidXXXX.log (with XXXX, the pid number)

gsmet:
Jenkins, retest this please

gunnarmorling:
Hi @gsmet, so downgrading to Surefire 2.18.1 as you suggested makes things work indeed. Any idea what's the issue in 2.19? I'd add a commit for this and merge this PR then. Note that I'm still seeing this slowdown in OSGi tests (it seems to wait for a 30 sec timeout), but it still passes. It's not so nice but still better than getting this exception.

gunnarmorling:
Hi @gsmet, what should we do with this one? Seems the problems on CI are fixed now, also without it? If so, I'd prefer to not merge it as it gives me that 30 sec penalty in the build.

gsmet:
@gunnarmorling We can either merge it as a first step or we can wait a version compatible with JDK9 before upgrading. I think it would be better to upgrade but I'm still concerned about your timeout issue. It would be a good idea to `kill -3` the process to understand what's going on.

gsmet:
@gunnarmorling any news on the `kill -3` front?

gsmet:
I upgraded to 4.0.7. Interested to see if you can reproduce your issue and get a `kill -3` output of the stalled process if so. I would like us to be ready when Karaf releases a JDK9 ready version.

gunnarmorling:
Rebased, applied and added one more commit to address that annoying timeout issue. The problem was that Karaf 4.0.4 and beyond [pull in](http://repo1.maven.org/maven2/org/apache/karaf/features/standard/4.0.4/standard-4.0.4-features.xml) the "management" feature with the "standard" feature we had enabled as boot feature in the test. This causes problems with firewalls potentially which prevent the exposure of the JMX server. Changed it to only use "system" now which seems enough for our purposes. If needed, we can re-add further features from the "standard" or "minimum" feature later on.

PR #522, opened 2016-09-20
HV-1095 - Provide module ZIP file for easily upgrading Hibernate Validator in a WildFly instance

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1095

+added new maven module for creating a module zip file for WildFly
+configured assembly plugin in new module to create a zip with validator and validator-cdi libs.


gunnarmorling:
Nice one! 

Could you also update the existing integration test module so that it uses this new module ZIP?Currently we just unzip WildFly, replace the existing JARs within the ones from our own build and that's it. Instead, we should unzip WildFly, unzip our new module ZIP and then let the test depend on the specific "slot" of the new modules (similar to what's done in the ORM module integration test for instance).

This also will require to disable the implicitly added HV module dependencies (see [here](https://docs.jboss.org/author/display/WFLY10/Class+Loading+in+WildFly) and [here](https://docs.jboss.org/author/display/WFLY8/Implicit+module+dependencies+for+deployments), so that users's don't have the HV coming with WF and the one from the module ZIP added to their deployment.

modules/src/main/modules/org/hibernate/validator/module.xml
```diff
@@ -0,0 +1,45 @@
...
+~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+~ Lesser General Public License for more details.
+~
+~ You should have received a copy of the GNU Lesser General Public
+~ License along with this software; if not, write to the Free
+~ Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+<module xmlns="urn:jboss:module:1.3" name="org.hibernate.validator.cdi">
```
gunnarmorling: name=org.hibernate.validator

modules/src/main/modules/org/hibernate/validator/module.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
```
gunnarmorling: Could you add the header we use in other XML files instead.

modules/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>hibernate-validator-modules</artifactId>
+
+    <name>Hibernate Validator WildFly Module</name>
+    <description>Hibernate Validator modules creation to use in WildFly</description>
+
+    <properties>
+        <module.xml.basedir>${basedir}/src/main/modules</module.xml.basedir>
+        <hibernate.validator.module.slot>${project.version}</hibernate.validator.module.slot>
```
gunnarmorling: That looks good. For ORM and HSEARCH we are using the alias mechanism in addition, allowing to use "6" or "6.1" when referencing the module, making it a bit simpler for the user. Would be nice to have the same here (you can find examples for how it's done in these other projects).

marko-bekhta: Thanks for the comments. I'll make the changes today.

gunnarmorling:
Going to close this one, as #523 is based on top of it. Thanks, @marko-bekhta!

marko-bekhta:
great !

PR #523, opened 2016-09-22
HV-1095

gunnarmorling:


gunnarmorling:
Based on top of @marko-bekhta's work. The missing piece was the _services=import_ declaration in _jboss-deployment-structure.xml_ (see the JIRA issue for more context).

@gsmet The tests make sure that any injected validator (factory) and the one used by JPA are the ones from the module ZIP. The fact remains though that some of the container modules (e.g. JCA, RESTEasy) refer to the "main" slot, so I suppose these functionalities may end up using the wrong one (depending on how they actually obtain validators; relying on injection/JNDI should yield the right one, bootstrapping themselves likely not). Not sure what to do, we could document it as unsupported for now or go all the way and provide tests plus a script for updating descriptors in-place if needed (at least for RESTEasy). WDYT?

In a next step we'll also need to add javax.validation to the module ZIP (so WF users can play with BV 2.0). And an update to the docs.

.gitignore
```diff
@@ -32,6 +32,7 @@ atlassian-ide-plugin.xml
 .project
 .settings
 .factorypath
+.checkstyle
```
gsmet: Where does this file come from?

integration/src/test/java/org/hibernate/validator/integration/wildfly/jpa/CustomValidatorFactoryInPersistenceUnitIT.java
```diff
@@ -86,6 +89,31 @@ public void testValidatorFactoryPassedToPersistenceUnitIsCorrectlyConfigured() t
 		log.debug( "testValidatorFactoryPassedToPersistenceUnitIsCorrectlyConfigured completed" );
 	}
 
+	/**
+	 * Makes sure the HV added via our module ZIP is used instead the one coming with WF by relying on functionality not
+	 * present in the WF-provided HV
+	 */
+	// TODO How to make that work reliably also after a HV upgrade within WF?
```
gsmet: So I suppose we will need a new feature unsupported in WildFly on each Validator upgrade in WildFly?

There's no way to get and check the module version used by WildFly?

gsmet:
Nice to see us being consistent with the other NoORM projects!

As for RESTEasy and others, I think we should document the limitation for now and wait for user feedback on this front.

I think we can commit this as is. Will you create JIRA issues for the next steps you were writing about (validation-api and documentation)?

.gitignore
gunnarmorling: It's from the CS plug-in for Eclipse.

integration/src/test/java/org/hibernate/validator/integration/wildfly/jpa/CustomValidatorFactoryInPersistenceUnitIT.java
gunnarmorling: Yes, I hope we can find a better way. Maybe checking on the protection domain of the HV JAR or so.

gsmet:
Merged! Thanks to both of you!

PR #524, opened 2016-09-23
HV-840 Annotation Processor doesn't detect errors in parameter constraints in inheritance hierarchies

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-840
  
  added test-utils to pom file
  added basic check for method parameters in overriding methods
  added simple test case for the overriding check


marko-bekhta:
Hey, @gsmet, @gunnarmorling wanted to ask if I'm on right track with this one, before completing everything (need to add some code to go through class hierarchy and probably add more tests to cover different scenarios). 
Also looking at https://hibernate.atlassian.net/browse/HV-801 I can see that in HV-801 it's said that the constrains should be defined in the class where a method is defined rather than in any that override them and in HV-840 it is said that the constrains on overridden method shouldn't be stricter than on the parent one. So what should be implemented in the end ?
Thanks !

gunnarmorling:
Hi @marko-bekhta, overriding methods may not alter the parameter constraints from the overridden method (i.e. no parameter constraints may be added in overriding methods). You can find the exact set of rules in [section 4.5.5](http://beanvalidation.org/1.1/spec/#constraintdeclarationvalidationprocess-methodlevelconstraints-inheritance) of the spec, i.e. these are the rules to implement.

In _engine_, these rules are implemented in the `org.hibernate.validator.internal.metadata.aggregated.rule` package. It's a bit sad that we need to re-implement them in the annotation processor module, but there is no good way of achieving code re-use (apart from creating our own abstraction of the two APIs which doesn't seem attractive, either).

marko-bekhta:
Parameter Checks are implemented except some cases for validation of parallel implementation. 

As a processor runs on annotated items there can be a case like this (from documentation):
```
		public interface OrderService {

			void placeOrder(String customerCode, String item, int quantity);
		}

		public interface OrderPlacementService {

			void placeOrder(
					@NotNull @Size( min = 3, max = 20 ) String customerCode,
					@NotNull String item,
					@Min( 1 ) int quantity);
		}

		public static class SimpleOrderService implements OrderService, OrderPlacementService {

			@Override
			public void placeOrder(String customerCode, String item, int quantity) {
			}
		}
```
In such case annotated parameters are only on interface and I cannot see how to get from there to the classes that implement the interface and to other interfaces.

But in such case:
```
		public interface OrderService {

			void placeOrder(String customerCode, String item, int quantity);
		}

		public interface OrderPlacementService {

			void placeOrder(
					@NotNull @Size( min = 3, max = 20 ) String customerCode,
					@NotNull String item,
					@Min( 1 ) int quantity);
		}

		public static class SimpleOrderService implements OrderService, OrderPlacementService {

			@Override
			public void placeOrder(String customerCode, String item, @Min( 1 ) int quantity) {
			}
		}
```
when there's an annotated parameter on class method that implements different interfaces - all works as you can get from this class method to super class methods and methods from implemented interfaces.

gunnarmorling:
@marko-bekhta, ah yes, that's an interesting case.

I think it's ok if we ignore that one for now, any new checks are better than none :) Could you push a commit for a documentation update, too (see 12.2, "Features", in the ref guide), mentioning the new checks and maybe also this one shortcoming?

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/MethodParameterOverrideCheck.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Checks, that the given parameter has a correct set of constraints if it's overriding a method form other class.
+ * Applies to methods parameters.
+ */
+public class MethodParameterOverrideCheck extends AbstractConstraintCheck {
```
gunnarmorling: Could you add an `@author` tag?

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/MethodParameterOverrideCheck.java
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	/**
+	 * Determine if one method 'correctly' overrides another one
+	 *
+	 * @param current    method from a current subclass
+	 * @param overridden method from a super class
+	 * @return {@code true} if method is overridden 'correctly', {@code false} otherwise
+	 */
+	private boolean correctlyOverrides(ExecutableElement current, ExecutableElement overridden) {
```
gunnarmorling: Have you seen `Elements#overrides()`? Let's better use that one instead if implementing it ourselves.

annotation-processor/src/main/resources/org/hibernate/validator/ap/ValidationProcessorMessages.properties
```diff
@@ -33,3 +33,4 @@ CROSS_PARAMETER_VALIDATION_APPLIES_TO_REQUIRED=Constraints with generic as well
 CROSS_PARAMETER_VALIDATION_APPLIES_TO_MUST_HAVE_IMPLICIT_DEFAULT_VALUE=Default value of the attribute validationAppliesTo() of the constraint @{0} must be ConstraintTarget\u0023IMPLICIT.
 CROSS_PARAMETER_VALIDATION_APPLIES_TO_MUST_HAVE_CONSTRAINT_TARGET_RETURN_TYPE=Return type of the attribute validationAppliesTo() of the constraint @{0} must be javax.validation.ConstraintTarget.
 CROSS_PARAMETER_VALIDATION_ON_PARAMETERLESS_METHOD=Cross-parameter constraint @{0} is illegally placed on a parameterless method or constructor.
+INCORRECT_METHOD_OVERRIDING=Method is incorrectly overridden. Method parameters cannot have a wide range of constraints than the one from a parent class.
```
gunnarmorling: wide=wider

annotation-processor/src/test/java/org/hibernate/validator/ap/ConstraintValidationProcessorTest.java
```diff
@@ -96,6 +98,35 @@ public void fieldLevelValidationUsingBuiltInConstraints() {
...
+
+		assertFalse( compilationResult );
+		assertThatDiagnosticsMatch(
+				diagnostics,
+				new DiagnosticExpectation( Kind.ERROR, 29 ),
+				new DiagnosticExpectation( Kind.ERROR, 47 ),
+				new DiagnosticExpectation( Kind.ERROR, 63 ),
+				new DiagnosticExpectation( Kind.ERROR, 82 ),
+				// commented out as method return values annotations are not processed yet
+				// new DiagnosticExpectation( Kind.ERROR, 117 ),
```
gunnarmorling: Is that for a follow-up PR?

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/MethodParameterOverrideCheck.java
```diff
@@ -0,0 +1,182 @@
...
+	private Types typeUtils;
+
+	public MethodParameterOverrideCheck(Elements elementUtils, Types typeUtils) {
+		this.elementUtils = elementUtils;
+		this.typeUtils = typeUtils;
+	}
+
+	@Override
+	public Set<ConstraintCheckError> checkField(VariableElement element, AnnotationMirror annotation) {
+		// get a current class in which a method is located.
```
gunnarmorling: I'm confused that this is implemented in `checkField()`. Shouldn't it rather be `checkMethod()`?

Thinking of it, do you support validation of parameter/return value constraints on constructors, too (I guess that's more for your other PR on constraint sanity, though).

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/MethodParameterOverrideCheck.java
marko-bekhta: yes, I've used this method in `getOverriddenElement` at line 178 to filter out methods that are not overriding a method under validation. And as for this private method `correctlyOverrides` it's verifying that the annotations are correctly used on current method and the one that was overridden, as `Elements#overrides()` seems to ignore annotations. So this `correctlyOverrides` consumes two methods - overridden and overriding.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/MethodParameterOverrideCheck.java
marko-bekhta: Yes, that's a good question. I've started to implement this as checkMethod. But actually the annotations are placed not on the method but on the method parameters. So that's why I've added `ElementKind.PARAMETER` to `SingleValuedChecks` class. Because otherwise those methods with annotations on parameters were ignored.

Also those constraint sanity checks should be added to `parameterChecks` map in `ConstraintCheckFactory` when they are merged in to master. Or should I just merge that branch (HV-270) into this one so those checks are present in here as well ? 

Return value constraints should be already supported in HV-270, but parameter constraints - no. I didn't want to add the same code to support parameters in two different  branches, so I was thinking that when one of the PR will be merged, I'll update the other.

annotation-processor/src/test/java/org/hibernate/validator/ap/ConstraintValidationProcessorTest.java
marko-bekhta: In the original task only constraints on parameters  were mentioned, but in the documentation which you pointed out in your first comment I saw also constraints on method return values. So I've added the test case for it. 

I suppose we can add this functionality in this PR. Because to implement it we will need a lot of functionality present in current `MethodParameterOverrideCheck` (all what is related to finding overridden methods). So I can see two way we can go to implement it:

- first one will be to add a methodCheck implementation to `MethodParameterOverrideCheck`, which will validate annotations on return values of overridden methods. And probably rename `MethodParameterOverrideCheck` to something more suitable for validating both parameters and return values.
- second one - move all common functionality (related to finding overridden methods) to some abstract class and provide to sub classes one for method parameter checks and another for method return value checks.

I didn't want to add it all until we make sure that what is already implemented is correct :)

gsmet:
Jenkins, retest this please

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/ConstraintCheckFactory.java
```diff
@@ -209,6 +254,8 @@ public ConstraintChecks getConstraintChecks(Element annotatedElement, Annotation
 		AnnotationType annotationType = constraintHelper.getAnnotationType( annotation );
 
 		switch ( annotatedElement.getKind() ) {
+			case PARAMETER:
```
gunnarmorling: `CONSTRUCTOR` seems to be missing still.

gunnarmorling:
@marko-bekhta Could you rebase this PR onto the latest master?

In terms of the general implementation, I feel there should be a new method `ConstraintCheck#checkParameter() which would trigger checking of parameter constraints (IIUC, you currently do this via `checkField()`). And `checkMethod()` should be used for the overriding checks as well as the return value checks (the latter could be done separately if you prefer).

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/ConstraintCheckFactory.java
```diff
@@ -115,13 +154,19 @@ public ConstraintCheckFactory(Types typeUtils, ConstraintHelper constraintHelper
...
 				)
 		);
 		methodChecks.put(
 				AnnotationType.GRAPH_VALIDATION_ANNOTATION,
-				new SingleValuedChecks( new GetterCheck( methodConstraintsSupported ), new StaticCheck(), new MethodAnnotationCheck( constraintHelper ),
-						new PrimitiveCheck() )
+				new SingleValuedChecks(
+						new GetterCheck( methodConstraintsSupported ),
+						new StaticCheck(),
+						new MethodAnnotationCheck( constraintHelper ),
```
gunnarmorling: I'm wondering why we apply `MethodAnnotationCheck` to `@Valid` (I know, it has been here before). This seems not correct.

gunnarmorling:
@marko-bekhta Added some more comments. That one is a tricky beast :) Maybe you split the PR and first send one for the parameter checks and then a follow-up for the override checks?

marko-bekhta:
@gunnarmorling yes, I think that's a good idea. I'll create another PR and for parameter checks and link it to new issue in JIRA (correct ?).

marko-bekhta:
@gunnarmorling what do you think of such approach to implementing this method overriding ? I've used the idea that you provided in last comment and here's what I came up with:

I've removed the Checks driven from annotations and implemented a new Visitor - `ClassVisitor`. It starts by walking through all the root elements and if it visits a type - it starts visiting all of it's members and if it visits a method than it performs two checks on the method (if parameters are annotated  correctly and if return value is annotated correctly).

annotation-processor/src/main/java/org/hibernate/validator/ap/ClassVisitor.java
```diff
@@ -0,0 +1,137 @@
...
+		this.elementUtils = processingEnvironment.getElementUtils();
+
+		this.factory = ClassCheckFactory.getInstance(
+				processingEnvironment.getTypeUtils(),
+				processingEnvironment.getElementUtils()
+		);
+	}
+
+	/**
+	 * Doesn't perform any checks at the moment but calls a visit methods on it's own elements.
```
gsmet: s/it's/its/

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ClassCheck.java
```diff
@@ -0,0 +1,50 @@
...
+ * supported by the check.
+ * </p>
+ *
+ *
+ * @author Marko Bekhta
+ */
+public interface ClassCheck {
+
+	/**
+	 * Checks, whether the given method is written correctly.
```
gsmet: You really don't want a comma here :)

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ClassCheck.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.Collection;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
+
+import org.hibernate.validator.ap.checks.ConstraintCheckError;
+
+/**
+ * <p>
+ * Implementations represent checks, which determine whether a given class
```
gsmet: s/, which/ that/ (no comma!)

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ClassCheck.java
```diff
@@ -0,0 +1,50 @@
...
+ * is implemented in correct way.
```
gsmet: s/in correct way/correctly/?

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ClassCheck.java
```diff
@@ -0,0 +1,50 @@
...
+	 *         method is written correctly), an empty set must be returned.
+	 */
+	Collection<ConstraintCheckError> checkMethod(ExecutableElement element);
+
+
+	/**
+	 * run all checks on the element
+	 *
+	 * @param element an element under investigation;
+	 * @return
```
gsmet: Javadoc for @return.

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ClassCheck.java
```diff
@@ -0,0 +1,50 @@
...
+	 *
+	 * @return A collection with errors, that describe, why the given method
+	 *         is not correctly implemented. In case no errors occur (the
+	 *         method is written correctly), an empty set must be returned.
+	 */
+	Collection<ConstraintCheckError> checkMethod(ExecutableElement element);
+
+
+	/**
+	 * run all checks on the element
```
gsmet: s/run/Run/ and a dot at the end.

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ClassCheck.java
```diff
@@ -0,0 +1,50 @@
...
+	 *
+	 * @param element an element under investigation;
```
gsmet: No ; at the end (and according to http://hibernate.org/validator/contribute/#javadoc, no dot either for @param and @return - note that the existing code is not very consistent about that).

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ParametersMethodOverrideCheck.java
```diff
@@ -0,0 +1,79 @@
...
+package org.hibernate.validator.ap.classchecks;
+
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.util.Elements;
+import javax.lang.model.util.Types;
+
+/**
+ * Checks if the parameters of overridden and overriding methods have correctly placed annotations.
+ * Overriding method cannot have more wider range of constraints on parameters the overridden one.
```
gsmet: -more

on parameters the overridden one -> on parameters than the overridden one

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ClassCheckFactory.java
```diff
@@ -0,0 +1,55 @@
...
+	 * Holds the checks to be executed for method elements.
+	 */
+	private final Collection<ClassCheck> methodChecks;
+
+	public ClassCheckFactory(Types typeUtils, Elements elementUtils) {
+		methodChecks = CollectionHelper.newArrayList();
+
+		methodChecks.add( new ReturnValueMethodOverrideCheck( elementUtils, typeUtils ) );
+		methodChecks.add( new ParametersMethodOverrideCheck( elementUtils, typeUtils ) );
+
```
gsmet: Unnecessary whitespace.

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ClassCheckFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Collections;
+
+import javax.lang.model.element.Element;
+import javax.lang.model.util.Elements;
+import javax.lang.model.util.Types;
+
+import org.hibernate.validator.ap.util.CollectionHelper;
+
+/**
+ * @author Marko Bekhta
```
gsmet: Javadoc? You added one everywhere but there so I suppose you forgot.

annotation-processor/src/main/resources/org/hibernate/validator/ap/ValidationProcessorMessages.properties
```diff
@@ -38,3 +38,5 @@ INVALID_PATTERN_ANNOTATION_PARAMETERS=Invalid annotation parameters. Regular exp
 INVALID_SCRIPT_ASSERT_ANNOTATION_PARAMETERS=Invalid annotation parameters. Script, lang and alias parameters cannot be blank.
 INVALID_DIGITS_ANNOTATION_PARAMETERS=Invalid annotation parameters. Fraction and integer should be greater than or equal to 0.
 INVALID_DECIMAL_MIN_MAX_ANNOTATION_PARAMETERS=Invalid annotation parameters. Value should be a valid BigDecimal string representation.
+INCORRECT_METHOD_PARAMETERS_OVERRIDING=Method is incorrectly overridden. Method parameters cannot have a wider range of constraints than the one from a parent class.
```
gsmet: Maybe we could include the name of the parent class in the message? It would be helpful.

documentation/src/main/asciidoc/ch12.asciidoc
```diff
@@ -40,6 +40,8 @@ getter methods (optionally, see below)
 annotations themselves
 * definition of dynamic default group sequence with @GroupSequenceProvider is valid
 * annotation parameter values are meaningful and valid
+* method parameter constraints in inheritance hierarchies are applied correctly, except some cases of parallel implementation.
```
gsmet: Is the "except some cases of parallel implementation" still true with the new implementation?

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ParametersMethodOverrideCheck.java
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * Determine if one method 'correctly' overrides another one in terms of annotated parameters
+	 *
+	 * @param currentMethod method from a current subclass
+	 * @param otherMethod method from a super class
+	 *
+	 * @return {@code true} if method is overridden 'correctly', {@code false} otherwise
+	 */
+	@Override
+	protected boolean checkOverriddenMethod(ExecutableElement currentMethod, ExecutableElement otherMethod) {
```
gsmet: Maybe you could call `otherMethod` `overriddenMethod`? It would be more clear.

annotation-processor/src/main/resources/org/hibernate/validator/ap/ValidationProcessorMessages.properties
```diff
@@ -38,3 +38,5 @@ INVALID_PATTERN_ANNOTATION_PARAMETERS=Invalid annotation parameters. Regular exp
...
+INCORRECT_METHOD_RETURN_OVERRIDING=Method is incorrectly overridden. Overridden method return value cannot have wider range of constraints than the overriding one.
```
gsmet: Same as above with "Return value constraints of {name of parent class} must not be weakened in sub type."

gsmet:
@marko-bekhta I added a couple more comments. I think you're on the right track.

It would be nice if you could rebase your patch (instead of merging from master) and flatten it. It's kinda hard to follow now :).

Thanks!

documentation/src/main/asciidoc/ch12.asciidoc
marko-bekhta: No :) not anymore. I'll remove that.

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ClassCheck.java
marko-bekhta: Sorry :)
My english and punctuation is not as good as I would like them to be :)
Thanks for the patience and all those comments on "wording" and grammar :)

marko-bekhta:
I've applied all the changes in the las commit and after this I'll do the rebasing :)

annotation-processor/src/main/java/org/hibernate/validator/ap/ConstraintAnnotationVisitor.java
```diff
@@ -112,6 +113,23 @@ public Void visitVariableAsField(VariableElement annotatedField, List<Annotation
...
+	 * method parameter. The following checks are performed:
+	 * </p>
+	 * <ul>
+	 * <li>
+	 * Overridden method parameters cannot have wider range of constraints.
+	 * </li>
+	 * </ul>
+	 */
+	@Override
+	public Void visitVariableAsParameter(VariableElement annotatedField, List<AnnotationMirror> mirrors) {
```
gsmet: These changes have already been merged, haven't they? If so, if you rebase your PR, they should be included. You might have to solve some conflicts though. When you rebase, git status gives you useful hints on what you should fix and what you should do.

marko-bekhta:
rebased and cleaned up the commits :)

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/MethodOverrideCheck.java
```diff
@@ -0,0 +1,274 @@
...
+
+		// find if there's a method that was overridden by a current one.
+		List<ExecutableElement> overridden = findAllOverriddenElements( currentTypeElement, currentMethod );
+		if ( overridden.isEmpty() ) {
+			return Collections.emptySet();
+		}
+
+		// if there's more than one overridden method we need to make sure that all of them match
+		Set<ConstraintCheckError> errors = CollectionHelper.newHashSet();
+		for ( ExecutableElement firstExecutable : overridden ) {
```
gsmet: (Sorry for the delay on this one, it took me some time to find some spare cycles to fully dedicate to this review)

I might be wrong but I don't think this is right. You're comparing each method with the others while the order is important and you should only compare an overriding method to the ones it overrides not the other way around.

I think what would be in order is to build several separate inheritance trees for each path and then runs the check on each inheritance tree in the right order. You might have several inheritance trees because of the interfaces. Doing it like that will also make the detection of parallel declarations robust.

I think you could trigger the issue by creating an inheritance with several super classes.

Did I miss something? Does what I say make sense to you?

marko-bekhta: @gsmet thanks for the review. And no worries about the delay, I completely understand it - this one is not as easy as adding translation for messages :) 

As for what you wrote it seems to make sense to me... But I need to think about it a bit before I get back to you with a good answer/solution. 

Also just as information about what I was trying to do by this part of code when I wrote it - it supposed to be only responsible for finding issues in parallel declarations. As the check in hierarchy tree is done in the loop on line 77. And also reading the comment that I wrote - `we need to make sure that all of them match` - it seems that it's doing exactly that (when I was implementing this I thought that if there are two "parallel methods" then their constraints should match, but obviously there can be methods in this list that can have different constraints) - but actually it might not be the right thing to check....

So this is what I'll do next - I'll prepare a few more different test cases and then I'll see how this parallel check can be corrected.

gunnarmorling:
Where are we with this one? Is there anything left to be done (apart from rebasing it)?

gsmet:
@gunnarmorling it would have been merged otherwise :). GitHub is hiding our last discussion with Marko. See https://github.com/hibernate/hibernate-validator/pull/524#pullrequestreview-5125189 . So this PR still needs work.

marko-bekhta:
Thanks @gsmet for answering :) I'll have some time to work on it today and hope to have some results. 

Even thought this is not relevant to this task, but I saw your comment @gunnarmorling in JIRA on Jandex - and here what I was thinking. So I remember that you said that you would want to do a release of 5.3 with those improvements to annotation processor, and I found all of those opened issues related to it. Atm there are two left - current one and also this - https://hibernate.atlassian.net/browse/HV-451. So I was thinking that I might look into it and then all annotation processor related items will be ready (well just only those that are opened in JIRA:) ) And after that I'll be happy to work on that other task with Jandex. But if you have other priorities for these tasks - just let me know :)

gsmet:
@marko-bekhta btw one nice small addition to the annotation processor would be to make the error message more exploitable when there is a mismatch between the error lines. I think just having the items sorted by line number would be sufficient. As I forgot to update the line numbers in my patch, you have a good example to start with :).

gsmet:
Forgot to answer about this:
> So I was thinking that I might look into it and then all annotation processor related items will be ready (well just only those that are opened in JIRA:) ) And after that I'll be happy to work on that other task with Jandex.

Sounds like a good plan. @gunnarmorling was just talking to me about the Jandex stuff this morning.

Btw the idea is to release a **5.4** with your annotation processor enhancements and javax.money stuff. We won't add any new features to 5.3.

marko-bekhta:
@gsmet I've made a few changes. The idea behind them is next: 

- for each method all overridden ones are found and are compared to current one to see if inheritance rules are not broken
- all found overridden methods are grouped into two groups:
-- one group contains methods that themselves override some other method
-- another group contains methods that are originally declared in class/interface and are not overriding any other method.
-- for those methods that are in the second group (the "original" ones) we need to check if they are completely the same to make sure that parallel implementation is not braking any rules

Not sure if I've explained it good enough :)

marko-bekhta:
@gsmet 

>  I think just having the items sorted by line number would be sufficient.

you mean in tests ? when one fails and two sets are printed and it's hard to find which item is wrong ? or something else ? :)

gsmet:
> you mean in tests ? when one fails and two sets are printed and it's hard to find which item is wrong ?

Yes, exactly. Looks like you already had the issue :).

marko-bekhta:
It's ready for review again :)

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/AbstractMethodOverrideCheck.java
```diff
@@ -0,0 +1,338 @@
...
+
+	private final Types typeUtils;
+
+	public AbstractMethodOverrideCheck(Elements elementUtils, Types typeUtils) {
+		this.elementUtils = elementUtils;
+		this.typeUtils = typeUtils;
+	}
+
+	@Override
+	public Collection<ConstraintCheckIssue> checkMethod(ExecutableElement currentMethod) {
```
gsmet: I'm not sure this can be implemented in the abstract method. I suspect the logic will be totally different between the parameters and the return values.

marko-bekhta: you mean in abstract class ? if so - than the idea here was that for both cases for parameters and for return values one should still:

- check if any checks should be performed at all (which is addressed by abstract method `needToPerformAnyChecks` plus checking if the given class is not java.lang.Object) 
- find all those overridden methods and compare current one to the overridden ones (which is addressed by another abstract method that takes two methods as parameters, and then in implementation the specific is present - do we need to check return value or do we need to check parameters)

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/AbstractMethodOverrideCheck.java
```diff
@@ -0,0 +1,338 @@
...
+		}
+		// find if there's a method that was overridden by the current one.
+		Map<Boolean, List<ExecutableElement>> overriddenMethods = findAllOverriddenElements( currentTypeElement, currentMethod );
+		if ( overriddenMethods.get( Boolean.TRUE ).isEmpty() && overriddenMethods.get( Boolean.FALSE ).isEmpty() ) {
+			return Collections.emptySet();
+		}
+
+		// if there's more than one overridden originally declared method we need to make sure that all of them match
+		Set<ConstraintCheckIssue> errors = CollectionHelper.newHashSet();
+		for ( ExecutableElement firstExecutable : overriddenMethods.get( Boolean.TRUE ) ) {
```
gsmet: OK, so I did my homework (finally, you might say) and the rules for parameters are much stricter than what you implemented. I'm a bit surprised you followed this road, we probably didn't guide you very well.

So the rules for parameters constraints are pretty clear:
* you can't define a constraint on a parameter of an overriding/implementing method - so it's not a matter of the constraints match or whatever: you simply can't do it. So you need to check that methods having constraint annotations in the hierarchy does not override another method. I think it's far more simple that what you did.
* the second rule is that if you have 2 parallel hierarchies both of which implementing the same method, you can't define a parameter constraint at all for this method (anywhere in the hierarchy, not even once). I think this check should be done first.

Tell me if I'm wrong but I think we were a bit off before.

You also have the rules for the `@Valid` annotation that might be interesting to implement (referred as cascaded in the doc links below).

See https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#section-method-constraints-inheritance-hierarchies for reference on the constraint parameters. http://beanvalidation.org/latest-draft/spec/#constraintdeclarationvalidationprocess-methodlevelconstraints-inheritance is also interesting to understand the whole thing.

Another thing: we need to be very careful on the error message we provide: the user should have all the required information to fix the issue. This is especially true when the errors are not about the class itself but its hierarchy (we need both classes/interfaces in the error message).

gsmet:
@marko-bekhta about this one, I think we need to be on the same page as for the checks we enforce. Can you confirm me that what I wrote makes sense and you have the same understanding of the spec + HV documentation.

Thanks!

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/ReturnValueMethodOverrideCheck.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public class ReturnValueMethodOverrideCheck extends AbstractMethodOverrideCheck {
+
+	public ReturnValueMethodOverrideCheck(Elements elementUtils, Types typeUtils) {
+		super( elementUtils, typeUtils );
+	}
+
+	@Override
+	protected boolean checkOverriddenMethod(ExecutableElement currentMethod, ExecutableElement otherMethod) {
+		return annotationMirrorContainsAll( currentMethod.getAnnotationMirrors(), listOnlyConstraintAnnotations( otherMethod.getAnnotationMirrors() ) );
```
marko-bekhta: for this one it's even easier :) just check if there is <= 1 `@Valid` annotation present in the hierarchy for a particular method, right ?

marko-bekhta:
@gsmet yes ! All seems clear now. The only question is about `@Valid` for parameters. You were saying about rules for it and the only thing that I've found is this:

> null arguments and null return values are ignored
> 
> The validation is recursive; that is, if validated parameter or return value objects have references marked with @Valid themselves, these references will also be validated
> 
> Bean Validation providers must guarantee the prevention of infinite loops during cascaded validation

But this seems to be unrelated to inheritance rules and more of specific rules for `@Valid` please correct me if I'm missing something.

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/AbstractMethodOverrideCheck.java
marko-bekhta: after thinking about it I'd say even though it might be better to have different checks for each rule from a logical point of view but it will save some time (during execution) if we keep both rules for parameter checks in one check class (as for both of the rules inheritance hierarchy is needed)

gsmet: As for `@Valid`, the rule is that you cannot have `@Valid` in the situation where you can't define a parameter constraint. So it should be simple to implement.

Look for "marked for cascaded validation" in the paragraph I pointed here: http://beanvalidation.org/latest-draft/spec/#constraintdeclarationvalidationprocess-methodlevelconstraints-inheritance

As for having 2 different checks, I don't think it's necessary, just make sure the parallel check is the first one as it seems more logical to me.

marko-bekhta: if there's noting special for `@Valid` then everything else implemented. I've removed all the check logic from the `checkMethod` in abstract class. What I left there is a check to see if any further checks need to be made (call to abstract `AbstractMethodOverrideCheck#needToPerformAnyChecks`) and getting the overridden methods (if there's no overridden methods then no need to check anything else) as these two things are common. After that a call to abstract `AbstractMethodOverrideCheck#checkMethodInternal` is made that do the real check of parameters/return value.

I've also cleaned up the code and used `ConstraintHelper` methods to determine if it's a constraint annotation or `@Valid` (previously there was a set of private methods in `AbstractMethodOverrideCheck` to do this kind of filtering of annotations)

Please let me know if anything else is needed. And thanks again for your patience and time spent on this ticket !

gsmet: OK, thanks! I'll take a look later today!

gsmet:
OK, I think I'll need to take a shot at it later this week. I'm not very happy with the `Map<Boolean, XXX>` and I think we need to do it differently but I need to experiment myself.

marko-bekhta:
@gsmet after looking at it once more - I've started to dislike the `Map` approach as well. So I've came up with a different one.  You can check it in the latest commit when you have time :)

marko-bekhta:
Hi @gsmet have you had a chance to take a look at what I've done with the "`InheritanceTree` approach" ? I understand that all of you are busy with your own work :) I'm just checking in :) Also I was thinking to do something similar for the GroupSequences if the answer is positive for the test case that I've put in the comment of that pull requests. So I just wanted to know if that'll be a good idea before trying to reuse this `InheritanceTree` thing :) Thanks !

gsmet:
Hi @marko-bekhta , sorry about the delay. I didn't forget about this PR, it's the next thing on my TODO list. I'm currently working on polishing the java.time support, it shouldn't be too long now. I'll get back at you.

marko-bekhta:
@gsmet no worries :) as I've said - I understand that you are busy and I just wanted to make sure that you'll notice the latest changes before you experiment:
> but I need to experiment myself

As maybe the latest changes is what you were looking for :)

gsmet:
@marko-bekhta FYI, I'm working on this, so please do not push any more updates! Thanks!

gsmet:
Superseded by https://github.com/hibernate/hibernate-validator/pull/584 . Closing.

PR #525, opened 2016-09-28
HV-1108 Add @Repeatable to NON_COMPOSING_CONSTRAINT_ANNOTATIONS

gsmet:
* https://hibernate.atlassian.net/browse/HV-1108


engine/src/main/java/org/hibernate/validator/internal/metadata/descriptor/ConstraintDescriptorImpl.java
```diff
@@ -75,7 +76,8 @@
 			Retention.class.getName(),
 			Target.class.getName(),
 			Constraint.class.getName(),
-			ReportAsSingleViolation.class.getName()
+			ReportAsSingleViolation.class.getName(),
```
gunnarmorling: Cool. Wondering a bit why we use the class names instead of the classes themselves here. But that's for another time.

gunnarmorling:
Merged. Thanks, @gsmet.

PR #532, opened 2016-09-30
HV-1113-1115 Assorted fixes to the XML parsing

gsmet:
* https://hibernate.atlassian.net/browse/HV-1113
 * https://hibernate.atlassian.net/browse/HV-1114

@gunnar wondering if we should have a more comprehensive form of `LocalNamespace` in validation-api? It could include the `MappingXmlParser.SCHEMAS_BY_VERSION` information too.

Something similar to https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/boot/jaxb/internal/stax/LocalSchema.java without the 3 static methods.

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -267,7 +267,7 @@
                             </execution>
                         </executions>
                         <configuration>
-                            <packageName>org.hibernate.validator.internal.xml</packageName>
+                            <packageName>org.hibernate.validator.internal.xml.binding</packageName>
                             <extension>true</extension>
                             <!-- Generate correct getters for Boolean properties -->
                             <enableIntrospection>true</enableIntrospection>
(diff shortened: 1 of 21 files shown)


gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

gunnarmorling:
> @gunnar wondering if we should have a more comprehensive form of LocalNamespace in validation-api? It could include the MappingXmlParser.SCHEMAS_BY_VERSION information too.

Dunno, would we gain that much from it? Actually I'm not even fully convinced of the enum itself, seems we only use it once for doing the mapping, i.e. simple constants would be as good.

gunnarmorling:

gsmet:
@gunnarmorling well, it would be more useful if we add the schema paths and versions and if we include it in validation-api, it would provide an unified way to access the schemas and the namespace.

PR #533, opened 2016-10-01
HV-962 Remove Chronos/JMeter performance test harness and replace with JMH based tests

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-962

migrated performance tests to JMH
added default profile in pom for current HV version


gunnarmorling:
Hey @marko-bekhta, awesome that you tackled this one! Has been on the agenda for a long time. Could you update the readme file under _performance_ so it describes the new set-up? Seems we'd also need some new "profiles" to compare with 5.2 and 5.3.

performance/pom.xml
```diff
@@ -118,6 +150,7 @@
                 <dependency>
                     <groupId>org.hibernate</groupId>
                     <artifactId>hibernate-validator</artifactId>
+                    <version>5.1.3.Final</version>
```
gunnarmorling: Can't we refer the `${validator.version}` property here?

performance/pom.xml
```diff
@@ -58,45 +70,42 @@
...
-                            <value>${validator.version}</value>
-                        </property>
-                    </jmeterVariables>
-                    <options>
-                        <option>${profilingOptions}</option>
-                    </options>
-                </configuration>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>2.2</version>
```
gunnarmorling: Could you add this version to the `pluginManagement` block in the parent POM?

performance/pom.xml
```diff
@@ -58,45 +70,42 @@
...
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
```
gunnarmorling: Why is it that we need the shading plug-in here?

performance/pom.xml
```diff
@@ -107,6 +116,29 @@
...
+            <dependencies>
+                <dependency>
+                    <groupId>org.hibernate</groupId>
+                    <artifactId>hibernate-validator</artifactId>
+                </dependency>
+                <dependency>
+                    <groupId>org.glassfish</groupId>
+                    <artifactId>javax.el</artifactId>
+                </dependency>
+                <dependency>
```
gunnarmorling: Could you add a log4j config which only emits WARNING or ERROR? Currently, I see lots of DEBUG messages on the console which for sure will take an effect on the benchmark.

performance/src/main/java/org/hibernate/validator/performance/simple/SimpleValidation.java
```diff
@@ -38,24 +43,31 @@
 			"William"
 	};
 
-	private static Validator validator;
-	private static Random random;
+	private static final Random random = new Random();
```
gunnarmorling: Should the `Random` bee part of the benchmark state, too?

performance/pom.xml
gsmet: Don't think so as it's a profile used to run 5.1.x.

That being said, maybe we would need a 5.3 target to compare 6 to 5.3? And maybe a 5.2 target to compare 5.3 to 5.2 as 5.3 is not released yet and we would like to ensure we don't have any performance regression when we release it.

marko-bekhta:
Made suggested changes, added profiles and updated a readme.md file. Also I'm looking that there was a maven profile to run profiling. I see that JMH has profilers as well. So I plan to investigate how can we include profiling as well :)

marko-bekhta:
Also added profilers now. Should anything else be done here ? Also which report format is preferred if any :) ? There're such options:

- TEXT,
- CSV,
- SCSV,
- JSON,
- LATEX,

performance/pom.xml
marko-bekhta: it is used to create an executable jar file

performance/readme.md
```diff
@@ -20,74 +23,69 @@ Choosing a profile executes the tests against the specified Hibernate Validator
...
 
-The following command line will run all performance tests found under _src/jmx/_  (_*.jmx_ files):
+The following command line will run all performance tests listed in the main method of TestRunner class:
 
-    > mvn clean package chronos-jmeter:jmeter -P hv-5.1
+    > mvn clean package -P hv-current
+    > java -jar target/hibernate-validator-performance.jar
 
-To generate a report from the test execution run:
+It will generate a set of reports from the each test execution. Currently it'll be next three files:
```
gsmet: s/from the each/from each/

performance/pom.xml
gunnarmorling: Ok, I see. I thought there'd be a dedicated JMH plug-in, but they use the shade plug-in in their archetype, too.

gunnarmorling:
I've no preference on the report format I think.

Speaking of reports, did you do a run for let's say compare master to 5.3 to earlier versions? Any indications on perf regressions?

marko-bekhta:
@gunnarmorling here's a part of generated report files:

```
org.hibernate.validator.performance.cascaded.CascadedValidation.testCascadedValidation",
        "mode" : "thrpt",    
        6:  
            "score" : 100.33959245564917,
            "scoreError" : 5.412813694666855,
            "scoreConfidence" : [
                94.92677876098232,
                105.75240615031602
            ],
        5.3:
            "score" : 102.98369569669524,
            "scoreError" : 12.436551997747943,
            "scoreConfidence" : [
                90.5471436989473,
                115.42024769444318
            ],

org.hibernate.validator.performance.simple.SimpleValidation.testSimpleBeanValidation",
        "mode" : "thrpt",
        6:
            "score" : 265.82260489992564,
            "scoreError" : 5.947972550120035,
            "scoreConfidence" : [
                259.8746323498056,
                271.77057745004566
            ],
        5.3:
            "score" : 185.4581175043451,
            "scoreError" : 2.2775284177362547,
            "scoreConfidence" : [
                183.18058908660885,
                187.73564592208135
            ],

org.hibernate.validator.performance.simple.SimpleValidation.testSimpleBeanValidationRecreatingValidatorFactory
        "mode" : "thrpt",
        6:
            "score" : 2.0487057105068747,
            "scoreError" : 0.22190676864418532,
            "scoreConfidence" : [
                1.8267989418626893,
                2.27061247915106
            ],
        5.3:
            "score" : 2.185760526884566,
            "scoreError" : 0.17732633783429103,
            "scoreConfidence" : [
                2.008434189050275,
                2.363086864718857
            ],
...
```

So based on this I would say that 6 is not worse than 5.3 :)
Also I've made a few changes to benchmarks. And I've commented out one of them. I'll put a comment on that benchmark.

performance/src/main/java/org/hibernate/validator/performance/cascaded/CascadedValidation.java
```diff
@@ -50,19 +67,22 @@ public void testCascadedValidation() {
...
-	@Test
-	public void testCascadedValidationMultiThreaded() throws Exception {
+//	@Benchmark
+//	@BenchmarkMode(Mode.All)
+//	@OutputTimeUnit(TimeUnit.MILLISECONDS)
+//	@Fork(value = 1)
+//	@Threads(2)
+//	@Warmup(iterations = 10)
+//	@Measurement(iterations = 50)
+	public void testCascadedValidationMultiThreaded(CascadedValidationState state) throws Exception {
```
marko-bekhta: Is this one a good benchmark test ? as you can see previously it was marked as one to be executed manually. And based on what it's doing what would be a difference between this and just running a benchmark on testCascadedValidation (which is called in the end of current test anyway) ? 
I've commented this one out as it is taking to much time to execute and with previous configuration of 100/50 treads it even hasn't been able to run as it was creating too much treads and VM just wouldn't allow it.

gunnarmorling:
@marko-bekhta Rebased and applied. Thanks a lot! I've added one commit with some minor clean-ups.

Regarding the reports, how should I read this. E.g. for testSimpleBeanValidation() the "score" in 6 is much higher than in 5.3. Does it mean 6 has a higher through-put here?

Now that the conversion is done, we could look into analyzing and fixing some actual bottlenecks. Probably not the right time to do it on 6 (master) though, given it's under heavy development. But if there were very apparent issues on 5.3 they may be interesting to fix.

PR #534, opened 2016-10-04
HV-1116 Added support for javax.money validation

whiskeysierra:
This includes validators for `@DecimalMax`/`@DecimalMin` and `@Max`/`@Min`

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -89,6 +89,11 @@
             <artifactId>paranamer</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>javax.money</groupId>
+            <artifactId>money-api</artifactId>
+            <optional>true</optional>
+        </dependency>
 
         <!--
         Test dependencies
@@ -126,6 +131,11 @@
         <dependency>
             <groupId>org.jboss.shrinkwrap</groupId>
             <artifactId>shrinkwrap-impl-base</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.javamoney</groupId>
+            <artifactId>moneta</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/DecimalMinValidatorForNumber.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/DecimalMinValidatorForNumber.java
@@ -16,7 +16,7 @@ import org.hibernate.validator.internal.util.logging.Log;
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
 
 /**
- * Check that the number being validated is less than or equal to the maximum
+ * Check that the number being validated is greater than or equal to the minimum
  * value specified.
  *
  * @author Hardy Ferentschik
--- a/documentation/src/main/asciidoc/ch02.asciidoc
+++ b/documentation/src/main/asciidoc/ch02.asciidoc
@@ -482,7 +482,7 @@ impact portability of your application between Bean Validation providers.
               `int`, `long` and the
               respective wrappers of the primitive types; Additionally
               supported by HV: any sub-type of
-              `Number`|Checks whether the annotated value is less than the
+              `Number` and `javax.money.MonetaryAmount`|Checks whether the annotated value is less than the
               specified maximum, when inclusive=false.
               Otherwise whether the value is less than or equal to the
               specified maximum. The parameter value is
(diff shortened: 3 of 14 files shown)


engine/src/test/java/org/hibernate/validator/test/internal/xml/XmlMappingParserTest.java
```diff
@@ -50,21 +51,23 @@ public void testAdditionalConstraintValidatorsGetAddedAndAreLastInList() {
 				DecimalMin.class
 		);
 
-		assertEquals( validators.size(), 2, "Wrong number of default validators" );
+		assertEquals( validators.size(), 3, "Wrong number of default validators" );
```
whiskeysierra: I feel like this test could be done in a less fragile way, but I didn't want to rewrite it without consent. My idea would be to just check whether the list is non-empty and the custom validator is the last, without also verifying the default validators explicitly.

gunnarmorling:
Jenkins, add to whitelist.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/DecimalMinValidatorForCharSequence.java
```diff
@@ -15,6 +15,9 @@
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
 
 /**
+ * Check that the character sequence (e.g. string) being validated represents a number, and has a value
```
whiskeysierra: This class was missing a javadoc altogether, so I created one based on the `DecimalMaxValidatorForCharSequence`.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/DecimalMinValidatorForNumber.java
```diff
@@ -16,7 +16,7 @@
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
 
 /**
- * Check that the number being validated is less than or equal to the maximum
+ * Check that the number being validated is greater than or equal to the minimum
```
whiskeysierra: While looking at other validators and docs I realized that this was most likely a copy&paste error.

AlexanderYastrebov:
Feels like it is only constrains the amount but not the currency. How one would validate amount boundaries and allowed currencies together?

whiskeysierra:
> Feels like it is only constrains the amount but not the currency.

True. This limits the use of `@Max`/`@Min` on MonetaryAmount fields to either single currency situations and validation *around zero*, i.e. checking if a value is positive or negative.

> How one would validate amount boundaries and allowed currencies together?

For now, a custom constraint, I'd say.

AlexanderYastrebov:
@whiskeysierra Got it - the idea is to add support for Money into Min/Max and DecimalMin/Max and not to implement full currency validation

whiskeysierra:
> not to implement full currency validation

Exactly. If you have a good idea how that could look like, I'd be interested though.

gunnarmorling:
> Exactly. If you have a good idea how that could look like, I'd be
interested though.

+1. What would be the requirements and how could a solution look like?

engine/pom.xml
```diff
@@ -89,6 +89,12 @@
             <artifactId>paranamer</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>javax.money</groupId>
+            <artifactId>money-api</artifactId>
+            <version>1.0.1</version>
```
gunnarmorling: Could you add properties for these versions in the parent POM and add them to the `dependencyManagement` block in there?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/DecimalMinValidatorForCharSequence.java
gunnarmorling: Nice!

gunnarmorling:
Ah, and could you open an issue in the HV issue tracker (https://hibernate.atlassian.net/projects/HV/summary) and reword the commit message so it starts with the issue id? Thanks!

engine/src/test/java/org/hibernate/validator/test/internal/xml/XmlMappingParserTest.java
gunnarmorling: Yes, that sounds good.

AlexanderYastrebov:
> What would be the requirements and how could a solution look like?

I can imagine this additional constraints:

1. @MoneyMin/Max which constraints on amount and currency at the same time
2. @Currency which constraints only on set of currencies and could be mixed with @Min/Max/DecimalMin/DecimalMax

But this feels like a separate effort and goal of negative/positive check is achieved already with this one

whiskeysierra:
@gunnarmorling I made the requested changes:

- version properties
- dependency management
- new issue
- commit message prefix
- mentioned supported data types in docs

Can you take a look?

gunnarmorling:
Merged and rebased. Many thanks to you, @whiskeysierra and Lukas! I've added two more commits to add you two to copyright.txt and add a link to JSR 354 to the docs.

gsmet:
Hmmm, the validators should have been added to the `hv` package, not the `bv` one, as they are not defined in the Bean Validation specification.

Could you create a PR to move them appropriately? (except if @gunnarmorling already moved them?)

gsmet:
Oh and nice work, btw :).

gsmet:
Oh, sorry I don't why I was redirected to this PR somehow. I thought it was a recent one.

Please ignore :).

PR #535, opened 2016-10-04
Add a ValidationSchema enum centralizing the BV schema information

gsmet:
**This shouldn't be committed as is**

@gunnarmorling this is what I had in mind with the enum to gather the schema information. I don't know if `ValidationSchema` is something we could move to validation-api. If we don't, we can even move the calls to the logger to `ValidationSchema.get(Configuration|Mapping)Schema` and simplify a bit more.

What I like about this is that we have all the schema information in the same place and it's easier to add a new one without touching unrelated code. And it might serve as a reference for implementors if we move it to validation-api.


gunnarmorling:
Regarding moving it to _validation-api_, I'm not convinced. It seems to be an implementation detail really. In terms of simplification I'm not sure either. Seems it adds way more code lines than it removes ;) All in all I'm not sure about the problem we want to solve here I think. Happy to discuss though if you feel strongly about it.

gsmet:
@gunnarmorling as mentioned, IMHO the main advantage is that it centralizes the information about the schemas in one and only place. No strong opinion, I just think it's nicer.

gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

gsmet:
Jenkins, retest this please

gsmet:
Closing.

PR #536, opened 2016-10-05
HV-1117 Downgrade surefire and failsafe to 2.18.1

gsmet:
Newer versions cause random failures in our OSGi integration tests.


gunnarmorling:
Merged, thanks, @gsmet. Added one more commit to extract a version for the property.

PR #540, opened 2016-10-06
HV-822 Add check on default message to recognize missing {} for keys

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-822

introduced functionality to report warnings in annotation processing. added a check for probably incorrect messages. added a couple of tests for new functionality and new check.

I would say that this one would be more valuable as adding functionality to produce warnings from checks rather than the check of messages.


annotation-processor/src/main/java/org/hibernate/validator/ap/ConstraintAnnotationVisitor.java
```diff
@@ -6,9 +6,11 @@
  */
 package org.hibernate.validator.ap;
 
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
-
+import java.util.stream.Collectors;
```
gunnarmorling: With @gsmet we've been contemplating to get out all your nice improvements around the annotation processor as part of a 5.4 release (the reason being that 6 is the reference implementation for BV 2.0, so it will take quite a while until it will be delivered in a Final version). The usage of Java 8 prohibits that though. So if you like the idea of releasing these changes in 5.4, could you change that to use traditional loops etc.?

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/ConstraintCheckIssue.java
```diff
@@ -0,0 +1,132 @@
...
+		return messageKey;
+	}
+
+	public Object[] getMessageParameters() {
+		return messageParameters;
+	}
+
+	@Override
+	public String toString() {
+		return "ConstraintCheckError [annotationMirror=" + annotationMirror
```
gunnarmorling: ConstraintCheckError=ConstraintCheckIssue

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationMessageCheck.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Checks that the message provided as annotation parameter is valid and gives a warning otherwise.
+ *
+ * @author Marko Bekhta
+ */
+public class AnnotationMessageCheck extends AnnotationParametersAbstractCheck {
+
+	// for "long ish without space"
+	private static final int LONG_MESSAGE_SIZE = 35;
```
gunnarmorling: I'm not convinced of this one really. Whatever threshold we take, it seems arbitrary. I'd limit it to the next one.

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/ValidMessageParameters.java
```diff
@@ -0,0 +1,89 @@
...
+		Class<? extends Payload>[] payload() default {};
+	}
+
+
+	public static class Case1 {
+
+		@CancellationCodeInvalid
+		private String string;
+
+		@CancellationCodeInvalid(message = "some overridden message")
```
gunnarmorling: This should not be flagged as invalid. The idea would be to only check the default value in the constraint definition. Giving a plain message when _using_ a constraint is perfectly fine.

annotation-processor/src/main/java/org/hibernate/validator/ap/ConstraintAnnotationVisitor.java
```diff
@@ -191,16 +195,19 @@ public Void visitTypeAsInterface(TypeElement e, List<AnnotationMirror> p) {
...
 			try {
 
 				ConstraintChecks constraintChecks = constraintCheckFactory.getConstraintChecks(
 						annotatedElement, oneAnnotationMirror
 				);
-				Set<ConstraintCheckError> errors = constraintChecks.execute( annotatedElement, oneAnnotationMirror );
-				messager.reportErrors( errors );
+				Set<ConstraintCheckIssue> errors = constraintChecks.execute( annotatedElement, oneAnnotationMirror );
+
+				Map<Class, Set<ConstraintCheckIssue>> issues = errors.stream().collect( Collectors.groupingBy( issue -> issue.getClass(), Collectors.toSet() ) );
```
gunnarmorling: Instead of the two sub-types, couldn't you just add something like `getKind()` to `ConstraintCheckError` (renaming it is still a good idea though) an then just use this kind when reporting the diagnostic?

marko-bekhta: So leave the name as `ConstraintCheckIssue` and create an enum of different issue types ERROR, WARNING ? Right ?

annotation-processor/src/main/java/org/hibernate/validator/ap/ConstraintAnnotationVisitor.java
marko-bekhta: Well I kind of like this stream API and lambdas, but i suppose I can change it all to regular loops. And maybe later, when the time for 6.0 comes) create a task to update it all back to Java 8 stuff :)

As a follow up question - in a meanwhile I'm working on HV-840 - it's a part of annotation processor as well, so I'd better not use lambdas/streams there as well right ?

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/ValidMessageParameters.java
gunnarmorling: Ok, cool.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/ConstraintCheckIssue.java
```diff
@@ -0,0 +1,205 @@
...
+	 * @param annotationMirror The annotation that causes the error.
+	 * @param messageKey A key for retrieving an error message template from the bundle
+	 * <p>
+	 * <code>org.hibernate.validator.ap.ValidationProcessorMessages.</code>
+	 * </p>
+	 * @param messageParameters An array with values to put into the error message template
+	 * using {@link java.text.MessageFormat}. The number of elements must match
+	 * the number of place holders in the message template.
+	 */
+	public static ConstraintCheckIssue getConstraintCheckError(Element element, AnnotationMirror annotationMirror, String messageKey, Object... messageParameters) {
```
gsmet: Maybe we could shorten that to `error` and `warning`? WDYT?

annotation-processor/src/main/java/org/hibernate/validator/ap/util/MessagerAdapter.java
```diff
@@ -63,10 +63,8 @@ public Messager getDelegate() {
 	 * @param errors A set with errors to report. May be empty but must not be
 	 * null.
 	 */
-	public void reportErrors(Set<ConstraintCheckError> errors) {
-		for ( ConstraintCheckError oneError : errors ) {
-			reportError( oneError );
-		}
+	public void reportErrors(Collection<ConstraintCheckIssue> errors) {
+		errors.forEach( this::reportError );
```
gsmet: Looks like there is a Java 8 construct left.

gsmet:
@marko-bekhta it would be nice if you could separate this in 2 commits:
- the one introducing ConstraintCheckIssue
- the one checking the message

You should probably split the grouped commit in 2:
https://emmanuelbernard.com/blog/2014/04/14/split-a-commit-in-two-with-git/

Then apply the fixes in 2 separate commits, one for the general one and one for the check.

Then rebase -i and apply the fixes to the respective commit (using fixup in rebase -i).

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/ConstraintCheckIssue.java
```diff
@@ -0,0 +1,205 @@
...
+	 * <p>
+	 * <code>org.hibernate.validator.ap.ValidationProcessorMessages.</code>
+	 * </p>
+	 *
+	 * @param messageParameters An array with values to put into the error message template
+	 * using {@link java.text.MessageFormat}. The number of elements must match
+	 * the number of place holders in the message template.
+	 */
+	public ConstraintCheckIssue(Element element, AnnotationMirror annotationMirror, IssueKind kind, String messageKey, Object... messageParameters) {
+
```
marko-bekhta: wasn't me :) I just renamed the class 
I'll clean it up. But it got me thinking - maybe the style check can be added for such thing ? I remember that you've said that there's older code that uses such style. But maybe when the check is added all the code can be cleaned up ?
Or is that a bad thing to do ?

marko-bekhta:
@gsmet I've made the changes that you asked for, please have a look if all is ok. Thanks !

marko-bekhta:
I was thinking more about this task. And I think that when https://github.com/hibernate/hibernate-validator/pull/524 is merged in, we'd be able to add a check on the annotations themselves (meaning on the files where annotation are declared), to check if that message key is ok or not. would that be of any use ?

gunnarmorling:
That PR provides a commit with some clean-up: https://github.com/marko-bekhta/hibernate-validator/pull/1. Could you incorporate it?

gunnarmorling:
> we'd be able to add a check on the annotations themselves (meaning on the files where annotation are declared), to check if that message key is ok or not. would that be of any use ?

Yes, that was my thinking as well (see my other comment above, wrote it in parallel to yours). That message should be raised at the declaration of the `message()` method, not upon usage of this constraint.

marko-bekhta:
@gunnarmorling  What if we do it like this:

(1) update the current check to look only for annotated elements where message was provided by the user. Meaning for the cases like this:
```
@SomeConstraintAnnotation(message = "{a.bad.message.example") // there's a closing '}' missing
public SomeClass someField;
```
(2) add a new Class check based on https://github.com/hibernate/hibernate-validator/pull/524 or define a check subclassed from AbstractConstraintCheck which will look for default values of annotation messages where they are declared (look for types annotated with @Constraint, than loop through inner elements to find message and validate it). 

which option will be preferable to implement for the second case ?

gunnarmorling:
Yes, that seems very reasonable, @marko-bekhta.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationMessageCheck.java
```diff
@@ -0,0 +1,79 @@
...
+	@Override
+	protected boolean canCheckThisAnnotation(AnnotationMirror annotation) {
+		return true;
+	}
+
+	@Override
+	protected Set<ConstraintCheckIssue> doCheck(Element element, AnnotationMirror annotation) {
+
+		//check if default message on annotation is correct or not:
+		if ( ElementKind.ANNOTATION_TYPE.equals( element.getKind() ) ) {
```
gunnarmorling: Nice. I'd only suggest to put have two separate check classes: one for checking the default message given at a constraint type definition and another one for checking the messages when using them.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/ConstraintCheckFactory.java
```diff
@@ -197,7 +202,8 @@ public ConstraintCheckFactory(Types typeUtils, Elements elementUtils, Constraint
 						new ConstraintValidatorCheck( constraintHelper, annotationApiHelper ),
 						new AnnotationTypeMemberCheck( annotationApiHelper, typeUtils ),
 						new CrossParameterConstraintCheck( annotationApiHelper, constraintHelper, typeUtils ),
-						new AnnotationMessageCheck( annotationApiHelper, elementUtils )
+						new AnnotationUserMessageCheck( annotationApiHelper ),
+						new AnnotationDefaultMessageCheck( annotationApiHelper, elementUtils )
```
gunnarmorling: How comes the default message check is specified more than once? I'd have expected it only once for `@Constraint` on an annotation type.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/ConstraintCheckFactory.java
```diff
@@ -189,7 +201,9 @@ public ConstraintCheckFactory(Types typeUtils, ConstraintHelper constraintHelper
 						new TargetCheck( annotationApiHelper ),
 						new ConstraintValidatorCheck( constraintHelper, annotationApiHelper ),
 						new AnnotationTypeMemberCheck( annotationApiHelper, typeUtils ),
-						new CrossParameterConstraintCheck( annotationApiHelper, constraintHelper, typeUtils )
+						new CrossParameterConstraintCheck( annotationApiHelper, constraintHelper, typeUtils ),
+						new AnnotationUserMessageCheck( annotationApiHelper ),
+						new AnnotationDefaultMessageCheck( annotationApiHelper, elementUtils )
```
gunnarmorling: How comes the default message check is given more than once? I'd have only expected it for `@Constraint` on an annotation type.

marko-bekhta: Yes ... you are correct here - I've just separated the check into two as you said, but didn't give much thought about it and just replaced the previous check with two new ones...
But `AnnotationDefaultMessageCheck` should only be present in `annotationTypeChecks` right ? 

I'll change the previous commit to have default check only in annotation checks.

gunnarmorling:
Yes, exactly. And only if the annotation type is CONSTRAINT_META_ANNOTATION.

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/ValidMessageParameters.java
```diff
@@ -0,0 +1,117 @@
...
+		private String string3;
+
+	}
+
+	public static class Case2 {
+
+		@Size(message = "just some custom message which is completely fine")
+		private String string1;
+
+		@Size(message = "now.this.one.is.probably.not.what.user.wanted.to.get")
```
gunnarmorling: @marko-bekhta, what is the issue with this one?

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/ConstraintCheckFactory.java
gunnarmorling: Ok, that looks better now. Only I am wondering why `AnnotationUserMessageCheck` is given here? My thinking is that there should be one check for the default message and another one for messages given when applying a constraint. I don't think the latter should also be applied for the default value, or should it?

marko-bekhta: It was supposed to work for a case that I've just added with the new commit - `ValidMessageParameters.Case3`. But I only thought that this would work for it and didn't actually write the test case. So sorry for that. Now with the test case added I removed `AnnotationUserMessageCheck` from list of `CONSTRAINT_META_ANNOTATION` and added it to CONSTRAINT_ANNOTATION (in `annotationTypeChecks` map)

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/ValidMessageParameters.java
marko-bekhta: it matches the `long word without spaces with dots and without {}` 
So it looks like the user wanted to give a key but forgot '{}'
and we should produce a warning in such case right ?

gunnarmorling: Ah, so it checks for "without spaces", too. So the message "Something is wrong. Really wrong." would not be flagged?

marko-bekhta: Yes correct. I've added this case as well to the tests and rebased this branch to master

gunnarmorling:
Rebased and applied. Thanks, @marko-bekhta!

PR #541, opened 2016-10-06
5.3 backports proposal

gsmet:


gunnarmorling:
Merged; thanks, @gsmet! I've assigned all backported issues to version 5.3.0.Final in JIRA with the exception of HV-1102 where only one commit was backported. I've left that for 6.0 as that's where this "feature" is really delivered (ideally there would be a separate issue just for fixing that wrong import but no need to go overboard ;).

PR #542, opened 2016-10-07
Preview: Lambda-based constraints

gunnarmorling:
Hey @gsmet, @emmanuelbernard That's a preview how Lambda-based constraints could look like as proposed in [BVAL-515](https://hibernate.atlassian.net/browse/BVAL-515). Check out the test for the actual usage. Some thoughts from my side:

* I kinda like how it optimizes for the simplest case (no initialize() needed as the constraint has no parameters; no access to constraint validator context needed)
* Should we provide an alternative for where these things are needed (or leave that to implementing an actual class)?
* It doesn't pose a compatability issues as originally was David's concern, as we just add new types but don't alter `ConstraintValidator`

On the implementation, I've added it to our DSL to get quickly going, in BV it'd likely be on `Configuration` or some other defined place in the bootstrap API. And don't get hung up on the usage of ByteBuddy, it was just the quickest thing for me to use it, I think it can be implemented without bytecode generation (needs some changes in HV though which expects actual class types for validators)

I think it'd make sense to have this as some sort of prototype API first in HV 6 (Alpha/Beta) and then add it to the spec. Thoughts?

### DIFF
--- a/documentation/src/main/asciidoc/ch11.asciidoc
+++ b/documentation/src/main/asciidoc/ch11.asciidoc
@@ -735,6 +735,18 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter11/constraint
 ----
 ====
 
+If your validator implementation is rather simple (i.e. no initialization from the annotation is needed,
+and `ConstraintValidatorContext` is not used), you also can use this alternative API to specify the constraint logic using a Lambda expression or method reference:
+
+[[example-using-constraint-definition-api]]
+.Adding constraint definition with a Lambda expression
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter11/constraintapi/ConstraintApiTest.java[tags=constraintDefinitionUsingLambda]
+----
+====
+
 Instead of directly adding a constraint mapping to the configuration object, you may use a `ConstraintMappingContributor`
 as detailed in <<section-programmatic-api-contributor>>. This can be useful when
 configuring the default validator factory using _META-INF/validation.xml_ (see
--- a/documentation/src/main/asciidoc/ch11.asciidoc
+++ b/documentation/src/main/asciidoc/ch11.asciidoc
@@ -6,6 +6,13 @@ in addition to the functionality defined by the Bean Validation specification. T
 fail fast mode, the API for programmatic constraint configuration and the boolean composition of
 constraints.
 
+New APIs or SPIs are tagged with the `org.hibernate.validator.Incubating` annotation as long as they
+are under development. This means that such elements (e.g. packages, types, methods, constants etc.)
+may be incompatible altered - or removed - in subsequent releases. Usage of incubating API/SPI members
+is encouraged (so the development team can get feedback on these new features) but you should be
+prepared for updating code which is using them as needed when upgrading to a new version of Hibernate
+Validator.
+
 [NOTE]
 ====
 Using the features described in the following sections may result in application code which is not
--- a/engine/src/main/java/org/hibernate/validator/Incubating.java
+++ b/engine/src/main/java/org/hibernate/validator/Incubating.java
@@ -1,0 +1,23 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * Marks the annotated element as incubating. The contract of incubating elements (e.g. packages, types, methods,
+ * constants etc.) is under active development and may be incompatible altered - or removed - in subsequent releases.
+ * <p>
+ * Usage of incubating API/SPI members is encouraged (so the development team can get feedback on these new features)
+ * but you should be prepared for updating code which is using them as needed.
+ *
+ * @author Gunnar Morling
+ */
+@Retention(RetentionPolicy.CLASS)
+public @interface Incubating {
+}
--- a/engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintDefinitionContext.java
+++ b/engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintDefinitionContext.java
@@ -39,4 +39,27 @@ public interface ConstraintDefinitionContext<A extends Annotation> extends Const
 	 * @return This context for method chaining.
 	 */
 	ConstraintDefinitionContext<A> validatedBy(Class<? extends ConstraintValidator<A, ?>> validator);
+
+	/**
+	 * Allows to configure a validation implementation using a Lambda expression or method reference. Useful for simple
+	 * validations without the need for accessing constraint properties or customization of error messages etc.
+	 *
+	 * @param type The type of the value to validate
+	 * @return This context for method chaining
+	 */
+	<T> ConstraintValidatorDefinitionContext<A, T> validateType(Class<T> type);
+
+	interface ConstraintValidatorDefinitionContext<A extends Annotation, T> {
+
+		/**
+		 * Applies the given Lambda expression or referenced method to values to be validated. It is guaranteed that
+		 * never {@code null} is passed to these expressions or methods.
+		 */
+		ConstraintDefinitionContext<A> with(ValidationCallable<T> vc);
+	}
+
+	@FunctionalInterface
+	interface ValidationCallable<T> {
+		boolean isValid(T object);
+	}
 }
--- a/engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintDefinitionContext.java
+++ b/engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintDefinitionContext.java
@@ -10,6 +10,8 @@ import java.lang.annotation.Annotation;
 
 import javax.validation.ConstraintValidator;
 
+import org.hibernate.validator.Incubating;
+
 
 /**
  * Constraint mapping creational context representing a constraint (i.e. annotation type). Allows to define which
@@ -43,12 +45,19 @@ public interface ConstraintDefinitionContext<A extends Annotation> extends Const
 	/**
 	 * Allows to configure a validation implementation using a Lambda expression or method reference. Useful for simple
 	 * validations without the need for accessing constraint properties or customization of error messages etc.
+	 * <p>
 	 *
 	 * @param type The type of the value to validate
 	 * @return This context for method chaining
 	 */
+	@Incubating
 	<T> ConstraintValidatorDefinitionContext<A, T> validateType(Class<T> type);
 
+	/**
+	 * Allows to specify a validation implementation for the given constraint and data type using a Lambda expression or
+	 * method reference.
+	 */
+	@Incubating
 	interface ConstraintValidatorDefinitionContext<A extends Annotation, T> {
 
 		/**
@@ -62,6 +71,7 @@ public interface ConstraintDefinitionContext<A extends Annotation> extends Const
 	 * Callable implementing a validation routine. Usually given as method reference or Lambda expression.
 	 */
 	@FunctionalInterface
+	@Incubating
 	interface ValidationCallable<T> {
 		boolean isValid(T object);
 	}
--- a/engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintDefinitionContext.java
+++ b/engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintDefinitionContext.java
@@ -53,11 +53,14 @@ public interface ConstraintDefinitionContext<A extends Annotation> extends Const
 
 		/**
 		 * Applies the given Lambda expression or referenced method to values to be validated. It is guaranteed that
-		 * never {@code null} is passed to these expressions or methods.
+		 * {@code null} is never passed to these expressions or methods.
 		 */
 		ConstraintDefinitionContext<A> with(ValidationCallable<T> vc);
 	}
 
+	/**
+	 * Callable implementing a validation routine. Usually given as method reference or Lambda expression.
+	 */
 	@FunctionalInterface
 	interface ValidationCallable<T> {
 		boolean isValid(T object);
--- a/engine/src/main/java/org/hibernate/validator/internal/cfg/context/ConstraintDefinitionContextImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/cfg/context/ConstraintDefinitionContextImpl.java
@@ -51,11 +51,32 @@ class ConstraintDefinitionContextImpl<A extends Annotation>
 		return this;
 	}
 
+	@Override
+	public <T> ConstraintDefinitionContext.ConstraintValidatorDefinitionContext<A, T> validateType(Class<T> type) {
+		return new ConstraintValidatorDefinitionContextImpl<>( type );
+	}
+
 	@SuppressWarnings("unchecked")
 	ConstraintDefinitionContribution<A> build() {
 		return new ConstraintDefinitionContribution<>(
 				annotationType,
 				CollectionHelper.newArrayList( validatorTypes ),
 				includeExistingValidators );
+	}
+
+	private class ConstraintValidatorDefinitionContextImpl<T> implements ConstraintDefinitionContext.ConstraintValidatorDefinitionContext<A, T> {
+
+		private final Class<T> type;
+
+		public ConstraintValidatorDefinitionContextImpl(Class<T> type) {
+			this.type = type;
+		}
+
+		@Override
+		public ConstraintDefinitionContext<A> with(ConstraintDefinitionContext.ValidationCallable<T> vc) {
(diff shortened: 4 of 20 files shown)


engine/src/test/java/org/hibernate/validator/test/cfg/ConstraintDefinitionTest.java
```diff
@@ -269,6 +271,32 @@ public void testMultipleValidatorsForSameTypeWithNoCallToIncludeExistingValidato
...
+
+		Set<? extends ConstraintViolation<?>> violations = validator.validate( new MyBean() );
+		assertNumberOfViolations( violations, 1 );
+		assertCorrectConstraintTypes( violations, Directory.class );
+	}
+
+	@Test
+	public void testConstraintBasedOnLambdaExpression() {
+		mapping.constraintDefinition( Directory.class )
+				.validateType( File.class ).with( b -> b.exists() );
```
gunnarmorling: ...or a Lambda.

gsmet:
> I kinda like how it optimizes for the simplest case (no initialize() needed as the constraint has no parameters; no access to constraint validator context needed)

+1

> Should we provide an alternative for where these things are needed (or leave that to implementing an actual class)?

I don't think so. Let's keep it simple for the simple case and implement a class for further needs.

> I think it can be implemented without bytecode generation

Yes, it would be nice to avoid it if we can.

> I think it'd make sense to have this as some sort of prototype API first in HV 6 (Alpha/Beta) and then add it to the spec.

+1, I think people needs to play with it in real life cases so that we can have useful feedback.

hferentschik:
So what's the idea really, provide an alternative/new constraint mapping API? I guess from that perspective it might be ok, but it in the end BV has always been about declarative constraint mapping. Also, I find it odd that this is just working for simple cases. I would expect something which works for all cases. If not, I'd rather stick to what we have. Lambadas are the new generics and will get overused as generics did.

gunnarmorling:
Hey Hardy,

> So what's the idea really, provide an alternative/new constraint mapping API?

No, not a mapping API really. Rather a simplified way to implement a constraint's logic. It means to address user requests about making constraint implementation more light-weight.

> BV has always been about declarative constraint mapping. 

That wouldn't change, you'd still put your constraints in a declarative fashion.

> I would expect something which works for all cases

Yes, that's a question I have, too. I can see how we could have two overloaded methods or so. But it'd violate the "provide one way to do a thing" paradigm. In that light I kinda like to provide just a more light-weight approach for the most common case.

> Lambadas are the new generics and will get overused as generics did.

Hah, don't be that pessimistic. I believe in the good in people :)

I think we should add it in one way or another as experimental in HV 6 Alpha1 and look for feedback.

hferentschik:
> That wouldn't change, you'd still put your constraints in a declarative fashion.

Ok

> Yes, that's a question I have, too. I can see how we could have two overloaded methods or so. But it'd violate the "provide one way to do a thing" paradigm. In that light I kinda like to provide just a more light-weight approach for the most common case.

As said, I am not a big fan, especially if it just addresses one case.

> I think we should add it in one way or another as experimental in HV 6 Alpha1 and look for feedback.

I guess I could also see it as some experimental feature to explore it, but I don't think it is ready for BV.

gunnarmorling:
Hey @gsmet I've pushed an update, it's not using ByteBuddy anymore now. Let me know what you think. From my side it's good (apart from doc update).

gunnarmorling:
@gsmet Can we merge it?

engine/src/main/java/org/hibernate/validator/cfg/context/ConstraintDefinitionContext.java
```diff
@@ -39,4 +39,27 @@
...
+	 * @param type The type of the value to validate
+	 * @return This context for method chaining
+	 */
+	<T> ConstraintValidatorDefinitionContext<A, T> validateType(Class<T> type);
+
+	interface ConstraintValidatorDefinitionContext<A extends Annotation, T> {
+
+		/**
+		 * Applies the given Lambda expression or referenced method to values to be validated. It is guaranteed that
+		 * never {@code null} is passed to these expressions or methods.
```
gsmet: "{@code null} is never passed" would be better?

engine/src/main/java/org/hibernate/validator/internal/cfg/context/ConstraintDefinitionContextImpl.java
```diff
@@ -32,7 +32,7 @@
 
 	private boolean includeExistingValidators = true;
 
-	private final Set<Class<? extends ConstraintValidator<A, ?>>> validatorTypes = newHashSet();
+	private final Set<ConstraintValidatorDescriptor<A>> validatorTypes = new HashSet<>();
```
gsmet: I don't think `validatorTypes` is accurate as the descriptor also contains the lambda in the lambda case. Why not `validatorDescriptors`?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintdefinition/ConstraintDefinitionContribution.java
```diff
@@ -18,14 +19,14 @@
  */
 public class ConstraintDefinitionContribution<A extends Annotation> {
 	private final Class<A> constraintType;
-	private final List<Class<? extends ConstraintValidator<A, ?>>> constraintValidators = new ArrayList<Class<? extends ConstraintValidator<A, ?>>>();
+	private final List<ConstraintValidatorDescriptor<A>> validatorTypes;
```
gsmet: Same here.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ClassBasedValidatorDescriptor.java
```diff
@@ -0,0 +1,74 @@
...
+		if ( constraintValidator == null ) {
+			throw LOG.getConstraintFactoryMustNotReturnNullException( validatorClass );
+		}
+
+		return constraintValidator;
+	}
+
+	@Override
+	public Type getValidatedType() {
+		return TypeHelper.extractType( validatorClass );
```
gsmet: Wouldn't it be better to do this in the constructor and have a final variable with it?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ClassBasedValidatorDescriptor.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	@Override
+	public EnumSet<ValidationTarget> getValidationTargets() {
+		SupportedValidationTarget supportedTargetAnnotation = validatorClass.getAnnotation(
```
gsmet: Same here?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorDescriptor.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.EnumSet;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorFactory;
+import javax.validation.constraintvalidation.ValidationTarget;
+
+import org.hibernate.validator.cfg.context.ConstraintDefinitionContext.ValidationCallable;
+
+/**
+ * Represents a specific validator type.
```
gsmet: Not sure the Javadoc is accurate (see my remark about `validatorTypes` earlier).

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorDescriptor.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @author Gunnar Morling
+ */
+public interface ConstraintValidatorDescriptor<A extends Annotation> extends Serializable {
+
+	Class<? extends ConstraintValidator<A, ?>> getValidatorClass();
```
gsmet: Maybe a little more formatting and javadoc love is in order?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
```diff
@@ -115,13 +115,13 @@ public ConstraintValidatorManager(ConstraintValidatorFactory constraintValidator
 			}
 		}
 
-		Class<? extends ConstraintValidator<?, ?>> validatorClass = findMatchingValidatorClass(
+		ConstraintValidatorDescriptor<A> validatorTypeDescriptor = findMatchingValidatorClass(
```
gsmet: `validatorDescriptor` would be nice to be consistent with the naming a few lines below.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
```diff
@@ -238,19 +241,28 @@ public ConstraintHelper() {
 		putConstraint( tmpConstraints, ScriptAssert.class, ScriptAssertValidator.class );
 		putConstraint( tmpConstraints, URL.class, URLValidator.class );
 
+
 		this.builtinConstraints = Collections.unmodifiableMap( tmpConstraints );
 	}
 
-	private static <A extends Annotation> void putConstraint(Map<Class<? extends Annotation>, List<? extends Class<?>>> validators, Class<A> constraintType, Class<? extends ConstraintValidator<A, ?>> validatorType) {
-		validators.put( constraintType, Collections.singletonList( validatorType ) );
+	private static <A extends Annotation> void putConstraint(Map<Class<? extends Annotation>, List<ConstraintValidatorDescriptor<?>>> validators, Class<A> constraintType, Class<? extends ConstraintValidator<A, ?>> validatorType) {
```
gsmet: Good to have something more type safe now.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
```diff
@@ -320,39 +332,35 @@ private boolean isBuiltinConstraint(Class<? extends Annotation> annotationType)
...
+		return validatorClass.getValidationTargets().contains( target );
 	}
 
 	/**
 	 * Registers the given validator classes with the given constraint
 	 * annotation type.
 	 *
 	 * @param annotationType The constraint annotation type
-	 * @param definitionClasses The validators to register
+	 * @param validatorTypes The validator types to register
```
gsmet: Same remark about validator types.

engine/src/main/java/org/hibernate/validator/internal/metadata/descriptor/ConstraintDescriptorImpl.java
```diff
@@ -256,23 +260,14 @@ public ConstraintTarget getValidationAppliesTo() {
 
 	/**
 	 * Return all constraint validators classes (either generic or cross-parameter) which are registered for the
-	 * constraint of this despriptor.
+	 * constraint of this descriptor.
 	 *
 	 * @return The validators applying to type of this constraint.
 	 */
-	public List<Class<? extends ConstraintValidator<T, ?>>> getMatchingConstraintValidatorClasses() {
+	public List<ConstraintValidatorDescriptor<T>> getMatchingConstraintValidatorClasses() {
```
gsmet: Shouldn't we change the name to be clear it return descriptors?

engine/src/main/java/org/hibernate/validator/internal/util/TypeHelper.java
```diff
@@ -295,25 +297,25 @@ private static Type getResolvedSuperclass(Type type) {
 	 * @return Return a Map&lt;Class, Class&lt;? extends ConstraintValidator&gt;&gt; where the map
 	 *         key is the type the validator accepts and value the validator class itself.
 	 */
-	public static <A extends Annotation> Map<Type, Class<? extends ConstraintValidator<A, ?>>> getValidatorsTypes(
+	public static <A extends Annotation> Map<Type, ConstraintValidatorDescriptor<A>> getValidatorTypes(
```
gsmet: Same remark about validator types not being accurate.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ClassBasedValidatorDescriptor.java
gsmet: Btw, just thinking out loud, but any chance we could use ClassMate instead of exposing a new `TypeHelper` method?

gunnarmorling:
@gsmet Thanks for reviewing. I've pushed some more commits for

* Addressing your remarks about the naming (in a way "type" could be applied to the descriptors, too, but I agree that "class" was too far off. Should be fixed everywhere now
* Adding a new annotation `@Incubating` (see [HV-1141](https://hibernate.atlassian.net/projects/HV/issues/HV-1141)) and putting it to the new API members
* Adding docs

gunnarmorling:
Taking a step back, there are two loose ends remaining:

* currently `validateType` only takes a `Class` object, but I think we should have one for `Type`, too. Because a user may wish to define an expression for `List<String>` instead of raw `List` (e.g. an equivalent to `SizeValidatorForCollection`)
* currently there is no way to declare a cross-parameter constraint using a Lambda expression (as the supported validation target(s) is controlled via an annotation on `ConstraintValidator` implementations)

I think we can do these two separately as there always is the current way of doing things.

documentation/src/main/asciidoc/ch11.asciidoc
```diff
@@ -6,6 +6,13 @@ in addition to the functionality defined by the Bean Validation specification. T
 fail fast mode, the API for programmatic constraint configuration and the boolean composition of
 constraints.
 
+New APIs or SPIs are tagged with the `org.hibernate.validator.Incubating` annotation as long as they
+are under development. This means that such elements (e.g. packages, types, methods, constants etc.)
+may be incompatible altered - or removed - in subsequent releases. Usage of incubating API/SPI members
```
gsmet: incompatibly?

engine/src/main/java/org/hibernate/validator/Incubating.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * Marks the annotated element as incubating. The contract of incubating elements (e.g. packages, types, methods,
+ * constants etc.) is under active development and may be incompatible altered - or removed - in subsequent releases.
```
gsmet: incompatibly?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ClassBasedValidatorDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+
+		//by default constraints target the annotated element
+		if ( supportedTargetAnnotation == null ) {
+			return EnumSet.of( ValidationTarget.ANNOTATED_ELEMENT );
+		}
+		else {
+			return EnumSet.copyOf( Arrays.asList( supportedTargetAnnotation.value() ) );
+		}
+	}
+	@Override
```
gsmet: Missing new line just above.

engine/src/main/java/org/hibernate/validator/internal/util/TypeHelper.java
gsmet: Looks like you missed this one.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ClassBasedValidatorDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+	private static final long serialVersionUID = -8207687559460098548L;
+	private static final Log LOG = LoggerFactory.make();
+
+	private final Class<? extends ConstraintValidator<A, ?>> validatorClass;
+	private final Type validatedType;
+	private final EnumSet<ValidationTarget> validationTargets;
+
+	public ClassBasedValidatorDescriptor(Class<? extends ConstraintValidator<A, ?>> validatorClass) {
+		this.validatorClass = validatorClass;
+		this.validatedType = TypeHelper.extractType( validatorClass );
```
gsmet: Did you see my comment about avoiding exposing a new TypeHelper method and use ClassMate instead? Maybe it does not make sense here and has a risk of not being consistent with the rest but I thought I might as well ask.

gsmet:
> I think we can do these two separately as there always is the current way of doing things.

Could you create JIRA issues for them?

gsmet:
@gunnarmorling I added a few more comments on your changes. I can fix them on merging if you prefer!

gunnarmorling:
Sweet, thanks! I'll rework my last commits to address your final remarks and then do the push.

gunnarmorling:
> Could you create JIRA issues for them?

Done: [HV-1142](https://hibernate.atlassian.net/browse/HV-1142).

gunnarmorling:
Merged it, yippie :)

PR #543, opened 2016-10-07
HV-1121 Fix property path generation for type argument constraints

gsmet:
* https://hibernate.atlassian.net/browse/HV-1121

We will have to revisit it later once BVAL-508 is included but I think it's a good first step.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -744,26 +744,11 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 					valueContext.getCurrentGroup(),
 					valueContext.getPropertyPath()
 			) ) {
-				validateTypeArgumentConstraints( context, valueContext, value, typeArgumentsConstraint );
+				// Type arguments
+				validateTypeArgumentConstraints( context, buildNewLocalExecutionContext( valueContext, value ), value, typeArgumentsConstraint );
 
 				// Cascade validation
-				ValueContext<?, Object> newValueContext;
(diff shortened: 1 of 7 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValueContext.java
```diff
@@ -211,6 +216,11 @@ public String toString() {
 		return sb.toString();
 	}
 
+	@Override
+	protected Object clone() {
```
gunnarmorling: When do we clone it?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -772,11 +757,31 @@ else if ( isIndexable ) {
 		}
 	}
 
+	private ValueContext<?, Object> buildNewLocalExecutionContext(ValueContext<?, Object> valueContext, Object value) {
```
gunnarmorling: No actual change here besides the method extraction, right?

engine/src/main/java/org/hibernate/validator/internal/engine/ValueContext.java
gsmet: Oops, we don't. It's a leftover of some investigation.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gsmet: Yes, no change but I needed to use the exact same code a few lines above so I extracted it.

gsmet:
Merged and backported to 5.3.

PR #544, opened 2016-10-08
HV-983 Support @Pattern for numerical types

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-983

added support of Pattern constraint on numbers

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/pattern/BasePatternValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/pattern/BasePatternValidator.java
@@ -4,9 +4,8 @@
  * License: Apache License, Version 2.0
  * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
  */
-package org.hibernate.validator.internal.constraintvalidators.bv;
+package org.hibernate.validator.internal.constraintvalidators.bv.pattern;
 
-import java.util.regex.Matcher;
 import java.util.regex.PatternSyntaxException;
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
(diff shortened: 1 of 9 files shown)


engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/pattern/BasePatternValidator.java
```diff
@@ -40,12 +43,4 @@ public void initialize(Pattern parameters) {
 		}
 	}
 
-	@Override
-	public boolean isValid(CharSequence value, ConstraintValidatorContext constraintValidatorContext) {
-		if ( value == null ) {
-			return true;
-		}
-		Matcher m = pattern.matcher( value );
-		return m.matches();
```
gsmet: Maybe you could even keep this method here (well a generified version of it) and have a `CharSequence getCharSequenceRepresentation(T value)` method to override?

marko-bekhta: yes, I'll do that.

marko-bekhta:
> But all in all, I think it's the responsibility of the user to test his regexp and make sure it's working with the given type

Yes, that was my thought as well. Another thought was to make a different annotation for numbers something like `@NumberPattern` and add a parameter to that annotation which will provide a number format, which will be used to convert the annotated number to string and then tested by regexp.

gsmet:
@marko-bekhta could you experiment with your numberFormat idea? I think it might be a better fit, especially for BigDecimal. The number format could be optional and we would use toString() (and maybe toPlainString() for BigDecimal?) if not provided.

I think the best is to keep this PR as is and open another PR with the result of your experimentation if it ends well.

marko-bekhta:
@gsmet I've updated a JavaDoc of `BasePatternValidator` a bit to reflect the move of isValid method. And yes, I'll try to implement this numberFormat approach and create another PR for it.

gunnarmorling:
As discussed on the JIRA issue, I'm going to close it as we don't see the use case really. Thanks for your efforts either way, @marko-bekhta!

PR #545, opened 2016-10-10
[5.3] HV-1123 Setting up japicmp instead of clirr for API/SPI change reporting

gunnarmorling:
Report can be created by running `mvn japicmp:cmp`. It's only enabled for _engine_ and _cdi_.

This is for 5.3, should be ported to master, too.


gsmet:
I added a commit to fix the indentation and add a comment. I merged it as is in 6.0, will update the base version once 5.3.0.Final is released.

PR #547, opened 2016-10-10
HV-983 Support @Pattern for numerical types

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-983

Here's another approach to support regex patterns for numerical types based on discussion in https://github.com/hibernate/hibernate-validator/pull/544

I've done it on top of implementation from that previous PR. 

@gsmet  what do you think ? is this what you had in mind ?


gsmet:
OK, so, after giving it some thoughts, I think I prefer the NumberPattern approach. I wouldn't make both available and would limit the patch to adding NumberPattern.

As adding a new constraint is something big, I would like to have @gunnarmorling 's feedback on this approach before going further.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/NumberPatternValidator.java
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Checks that given {@link Number} matches provided regexp.
+ *
+ * @author Marko Bekhta
+ */
+public class NumberPatternValidator implements ConstraintValidator<NumberPattern, Number> {
+
+	private static final Log log = LoggerFactory.make();
+
+	private java.util.regex.Pattern pattern;
```
gsmet: Once you have removed the dependency to Pattern.Flag, you'll be able to remove the fqcn.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/NumberPatternValidator.java
```diff
@@ -0,0 +1,90 @@
...
+
+		try {
+			this.pattern = java.util.regex.Pattern.compile( parameters.regexp(), intFlag );
+		}
+		catch (PatternSyntaxException e) {
+			throw log.getInvalidRegularExpressionException( e );
+		}
+		if ( !parameters.numberFormat().trim().isEmpty() ) {
+			try {
+				this.numberFormat = new DecimalFormat( parameters.numberFormat() );
```
gsmet: You can't do that as `DecimalFormat` is not thread safe and `ConstraintValidator`s are shared.

What you should do is check if the format is valid and initialize a string if it is and use this string to instantiate a new `DecimalFormat` on validation (I don't think using a `ThreadLocal` would be a good idea).

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/NumberPatternValidator.java
```diff
@@ -0,0 +1,90 @@
...
+			intFlag = intFlag | flag.getValue();
+		}
+
+		try {
+			this.pattern = java.util.regex.Pattern.compile( parameters.regexp(), intFlag );
+		}
+		catch (PatternSyntaxException e) {
+			throw log.getInvalidRegularExpressionException( e );
+		}
+		if ( !parameters.numberFormat().trim().isEmpty() ) {
```
gsmet: See `StringHelper#isNullOrEmptyString`

engine/src/main/java/org/hibernate/validator/internal/xml/MetaConstraintBuilder.java
```diff
@@ -140,7 +140,7 @@ private Object getElementValue(ElementType elementType, Class<?> returnType, Str
 					return "";
 				}
 				else {
-					throw log.getEmptyElementOnlySupportedWhenCharSequenceIsExpectedExpection();
+					throw log.getEmptyElementOnlySupportedWhenCharSequenceIsExpectedException();
```
gsmet: Mmmmh, I can't see the difference here, what did you change?

marko-bekhta: there was a typo : `ExpeCtion` -> `ExCeption`

marko-bekhta:
@gsmet I've cleaned everything so only NumberPattern related code was left and force pushed a clean commit with all of your comments applied.

engine/src/main/java/org/hibernate/validator/internal/xml/MetaConstraintBuilder.java
gsmet: Raaah, I spent 2 minutes trying to find it :).

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
```diff
@@ -28,7 +27,8 @@ org.hibernate.validator.constraints.Range.message                   = must be be
 org.hibernate.validator.constraints.SafeHtml.message                = may have unsafe html content
 org.hibernate.validator.constraints.ScriptAssert.message            = script expression "{script}" didn't evaluate to true
 org.hibernate.validator.constraints.URL.message                     = must be a valid URL
+org.hibernate.validator.constraints.NumberPattern.message           = must match "{regexp}"
```
gsmet: Wondering if we should find a way to include the format here. WDYT?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/NumberPatternValidator.java
```diff
@@ -0,0 +1,93 @@
...
+		}
+		Matcher m = pattern.matcher( getCharSequenceRepresentation( value ) );
+		return m.matches();
+	}
+
+	/**
+	 * Converts given {@link Number} value to string representation to which regexp will be applied.
+	 * Uses number format if one was provided.
+	 *
+	 * @param value which is validated.
```
marko-bekhta: so it'll be like :
`@param value value which is validated.`
maybe it'll be better to change the parameter `value` -> `number` and then it'll be 
`@param number value which is validated.`
?

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
marko-bekhta: it'll be a longer message than, as that format parameter is optional and we would need to somehow mention that right ? So for example it might be something like:
`String representation of a number, formatted with {numberFormat} (if one was provided)  must match "{regexp}"`

But I don't know ...

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/NumberPatternValidator.java
gsmet: Yes, you can do it, it will be clearer. That being said, keep in mind it's formatted "value - value which is validated" in the resulting javadoc.

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
gsmet: See javax.validation.constraints.DecimalMax.message for a conditional usage. I'm not sure either about this. Let it alone for now, we can improve that later if users complain.

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
```diff
@@ -28,7 +27,8 @@ org.hibernate.validator.constraints.Range.message                   = must be be
...
 
 org.hibernate.validator.constraints.br.CNPJ.message                 = invalid Brazilian corporate taxpayer registry number (CNPJ)
 org.hibernate.validator.constraints.br.CPF.message                  = invalid Brazilian individual taxpayer registry number (CPF)
-org.hibernate.validator.constraints.br.TituloEleitoral.message      = invalid Brazilian Voter ID card number
+org.hibernate.validator.constraints.br.TituloEleitoral.message      = invalid Brazilian Voter ID card number
```
gsmet: We usually keep the empty line at the end of the file.

gunnarmorling:
As discussed on the JIRA issue, I'm going to close it as we don't see the use case really. Thanks for your efforts either way, @marko-bekhta!

PR #550, opened 2016-10-12
HV-1002 Fix collectMetaConstraintsForPath logic in the iterable case

gsmet:
* https://hibernate.atlassian.net/browse/HV-1002

The logic was wrong as it supposed that getIndex() returned the index on
the iterable node while getIndex() returns the index for the child of
the iterable node.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -1494,37 +1494,49 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		BeanMetaData<?> metaData = beanMetaDataManager.getBeanMetaData( clazz );
 		PropertyMetaData property = metaData.getMetaDataFor( elem.getName() );
 
-		// use precomputed method list as ReflectionHelper#containsMember is slow
 		if ( property == null ) {
 			throw log.getInvalidPropertyPathException( elem.getName(), metaData.getBeanClass() );
 		}
-		else if ( !propertyIter.hasNext() ) {
+
+		// if we are in the case of an iterable and we want to validate an element of this iterable, we have to get the
+		// element value
+		if ( property.isCascading() && elem.isIterable() && propertyIter.hasNext() ) {
+			elem = (NodeImpl) propertyIter.next();
+			newValue = newValue == null ? null : getValue( newValue, validationContext, property );
+
+			if ( newValue != null && elem.getIndex() != null ) {
+				newValue = ReflectionHelper.getIndexedValue( newValue, elem.getIndex() );
+			}
+			else if ( newValue != null && elem.getKey() != null ) {
+				newValue = ReflectionHelper.getMappedValue( newValue, elem.getKey() );
+			}
+			else if ( newValue != null ) {
+				throw log.getPropertyPathMustProvideIndexOrMapKeyException();
+			}
+			clazz = (Class<?>) ReflectionHelper.getIndexedType( property.getType() );
+
+			metaData = beanMetaDataManager.getBeanMetaData( clazz );
+			property = metaData.getMetaDataFor( elem.getName() );
+
+			if ( property == null ) {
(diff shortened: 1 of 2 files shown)


engine/src/test/java/org/hibernate/validator/test/internal/engine/ValidatorTest.java
```diff
@@ -325,6 +344,20 @@ public String getFoo() {
 		}
 	}
 
+	class I {
+		@Valid
+		Map<String, J> map = new HashMap<String, J>();
```
marko-bekhta: HI, it's more of a question :) - I've noticed that there's this `CollectionHelper` util class and sometimes in the code it is used to create different collections with for example `CollectionHelper#newHashMap` and some times jus simple instantiation is used like here. Is there any preferences/recommendations when to use what, or is it more of style thing ?

gsmet: CollectionHelper was created at the time where the diamond didn't exist so it was a convenient method to initialize a generic collection. It also has helpers to create sized collection. It has a little less sense now that we have the diamond but we still use it as the primary way to initialize collection.

In this case, I must admit that I like to keep it simple for this sort of test but it's a matter of preference. And I didn't use the diamond operator as I plan to backport this on 5.3/5.4.

marko-bekhta: Thanks for the explanation ! 

Everything else looks good to me, but I'm not as familiar with this part of code as any of you are :) so I'm probably not the one who should merge it in :)

gunnarmorling:
I'm starting to become skeptical about the entire thing. The spec clearly says "@Valid is not honored by this method" for `validateValue()` and `validateProperty()`. It seems we violating this paradigm by supporting nested paths? In hindsight I think it should not have been done, or rather via some explicit option to opt-in.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -1494,37 +1494,49 @@ private int validateConstraintsForGroup(ValidationContext<?> validationContext,
...
+			if ( newValue != null && elem.getIndex() != null ) {
+				newValue = ReflectionHelper.getIndexedValue( newValue, elem.getIndex() );
+			}
+			else if ( newValue != null && elem.getKey() != null ) {
+				newValue = ReflectionHelper.getMappedValue( newValue, elem.getKey() );
+			}
+			else if ( newValue != null ) {
+				throw log.getPropertyPathMustProvideIndexOrMapKeyException();
+			}
+			clazz = (Class<?>) ReflectionHelper.getIndexedType( property.getType() );
```
gunnarmorling: Instead of `property.getType()` the actual value type should be used (in the case of `validateProperty()`, as BV dicates to use runtime type rather than static type for constraint retrieval.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -1494,37 +1494,49 @@ private int validateConstraintsForGroup(ValidationContext<?> validationContext,
...
+		}
+
+		if ( !propertyIter.hasNext() ) {
 			metaConstraintsList.addAll( property.getConstraints() );
 			typeArgumentConstraints.addAll( property.getTypeArgumentsConstraints() );
 		}
 		else {
 			if ( property.isCascading() ) {
-				Type type = property.getType();
 				newValue = newValue == null ? null : getValue( newValue, validationContext, property );
```
gunnarmorling: `getValue()` is called once before already (a few lines up). It may do the right thing, but I'm struggling to grasp what this method is doing. There are so many conditionals, partly checking repeatedly on the same conditions. I feel some more thorough clean-up is due should we decide to keep this functionality. Also it's not so nice that it's implemented recursively instead of purely iterative.

gsmet:
Closing as I will push another one.

PR #551, opened 2016-10-12
Enabling integration tests on WF@JDK9; WildFly Arquillian update

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1128
https://hibernate.atlassian.net/browse/HV-1129

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,7 +129,7 @@
         <wildfly.version>10.1.0.Final</wildfly.version>
         <cdi-api.version>1.2</cdi-api.version>
         <weld.version>2.3.5.Final</weld.version>
-        <wildfly-arquillian.version>1.0.2.Final</wildfly-arquillian.version>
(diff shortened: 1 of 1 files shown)


gunnarmorling:
@Sanne As you asked me: you may try the same in HSEARCH. Just make sure to use a recent JDK 9 (ideally, the latest, b139). Earlier ones expose a bug in conjunction with Weld.

Sanne:
@gunnarmorling ah, nice! I was aware of the Weld issue, didn't know it was fixed in the JDK. Thanks!

pom.xml
```diff
@@ -896,6 +896,9 @@
             <properties>
                 <surefire.argLine.extension>--add-modules java.xml.bind</surefire.argLine.extension>
             </properties>
+            <modules>
+                <module>integration</module>
+            </modules>
```
gsmet: Wouldn't it make more sense to add back the module to the global list?

gsmet:
Merged! Thanks!

PR #552, opened 2016-10-14
HV-991 Add custom weigths parameter to Mod11Check (and Mod10Check)

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-991

added optional weights parameter to Mod11Check annotation 
added custom annotation for 3 Polish identification numbers - NIP, PESEL, REGON
added tests for new annotations
added checks to annotation processor


engine/src/main/java/org/hibernate/validator/constraints/pl/REGON.java
```diff
@@ -0,0 +1,56 @@
...
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import javax.validation.Constraint;
+import javax.validation.Payload;
+import javax.validation.ReportAsSingleViolation;
+
+
+/**
+ * Validates a REGON number - (9/14 digit number from Rejestr Gospodarki Narodowej ).
```
gunnarmorling: Could you add an (English) description of "Rejestr Gospodarki Narodowej"?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/pl/NIPValidator.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.constraintvalidators.hv.pl;
+
+import java.util.List;
+
+import org.hibernate.validator.constraints.pl.NIP;
+
+/**
+ * A class for NIP validator.
```
gunnarmorling: "Validator for @NIP", it's apparent that it is a class :)

engine/src/main/java/org/hibernate/validator/internal/util/ModUtil.java
```diff
@@ -101,4 +101,33 @@ public static int calculateMod11Check(final List<Integer> digits, final int thre
...
+
+	/**
+	 * Calculate Modulo {@code moduloParam} checksum with given weights. If no weights are provided then weights similar to Modulo 11 checksum will be used.
+	 * In case when there will be not enough weights provided the ones provided will be used in a looped manner.
+	 *
+	 * @param digits the digits for which to calculate the checksum
+	 * @param moduloParam modulo parameter to be used
+	 * @param weights weights for the sum.
+	 *
+	 * @return
```
gunnarmorling: Better to omit it.

engine/src/main/java/org/hibernate/validator/constraints/Mod11Check.java
```diff
@@ -122,6 +122,15 @@
 	ProcessingDirection processingDirection() default ProcessingDirection.RIGHT_TO_LEFT;
 
 	/**
+	 * @return an array of custom weights to be used in sum calculation of Mod11 algorithm.
+	 * If the size of provided weights array is less than needed ({@code array.length < endIndex - startIndex})
+	 * the weights will be used repeatedly. For example if constant weight is needed the weights array
+	 * can be initialized as {@code weights = { someConstant }}.
+	 * If weights array is left blank (which is a default value) a classical Mod11 weights are used.
```
gunnarmorling: How about that: "By default a weight of 1 will be used for each digit".

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
```diff
@@ -32,3 +31,8 @@ org.hibernate.validator.constraints.URL.message                     = must be a
 org.hibernate.validator.constraints.br.CNPJ.message                 = invalid Brazilian corporate taxpayer registry number (CNPJ)
 org.hibernate.validator.constraints.br.CPF.message                  = invalid Brazilian individual taxpayer registry number (CPF)
 org.hibernate.validator.constraints.br.TituloEleitoral.message      = invalid Brazilian Voter ID card number
+
+org.hibernate.validator.constraints.pl.REGON.message = Invalid Polish Taxpayer Identification Number (REGON)
```
gunnarmorling: Seems we don't have a Polish translations file yet. Maybe - as a separate issue -you could create one (assuming you speak the language based on the fact that you created these constraints :) )?

engine/src/main/java/org/hibernate/validator/constraints/Mod11Check.java
gunnarmorling: Thinking more about it, is this actually used for the new constraints? If not, I don't think we should add it.

gunnarmorling: If these new constraints cannot be written based on top of the existing ones, I wouldn't bother to add these weights then (and let's change the title of the issue to reflect that it is about adding these three Polish constraints).

marko-bekhta:
I've removed that custom weights parameter and cleaned up JavaDocs. Also I've updated the JIRA ticket. As for the translations - I'll need to create separate JIRA tickets for those, right ?

gunnarmorling:
Yes, separate tickets for those changes would be great.

gunnarmorling:
Rebased and applied. Thanks, @marko-bekhta! Could you send in another PR for adding the new constraints to the docs as well (see [here](https://github.com/hibernate/hibernate-validator/blob/master/documentation/src/main/asciidoc/ch02.asciidoc#additional-constraints))? Thanks again!

marko-bekhta:
@gunnarmorling https://github.com/hibernate/hibernate-validator/pull/567 here's a PR for doc. Is that what you had in mind ? Or should there be more information ? :)

PR #554, opened 2016-10-14
HV-455 The annotation processor should verify the groups parameter of a constraint declaration

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-455

added check for groups parameter to verify that groups are represented by interfaces.

One more for annotation processor :)

### DIFF
--- a/annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationParametersGroupsCheck.java
+++ b/annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationParametersGroupsCheck.java
@@ -12,6 +12,8 @@ import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
 import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
 
 import org.hibernate.validator.ap.checks.ConstraintCheckError;
 import org.hibernate.validator.ap.util.AnnotationApiHelper;
@@ -24,11 +26,9 @@ import org.hibernate.validator.ap.util.CollectionHelper;
  */
 public class AnnotationParametersGroupsCheck extends AnnotationParametersAbstractCheck {
 
-
 	public AnnotationParametersGroupsCheck(AnnotationApiHelper annotationApiHelper) {
 		super( annotationApiHelper );
 	}
-
 
 	@Override
 	protected boolean canCheckThisAnnotation(AnnotationMirror annotation) {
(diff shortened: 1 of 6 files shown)


annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationParametersGroupsCheck.java
```diff
@@ -0,0 +1,53 @@
...
+		return true;
+	}
+
+	@Override
+	protected Set<ConstraintCheckError> doCheck(Element element, AnnotationMirror annotation) {
+		List<? extends AnnotationValue> annotationValue = annotationApiHelper.getAnnotationArrayValue( annotation, "groups" );
+		Set<ConstraintCheckError> issues = CollectionHelper.newHashSet();
+
+		for ( AnnotationValue value : annotationValue ) {
+			if ( value.getValue() instanceof DeclaredType && !( (DeclaredType) value.getValue() ).asElement().getKind().isInterface() ) {
```
gsmet: Wouldn't `!( value.getValue() instanceof DeclaredType ) || ...` be a better test?

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/ValidGroupsParameters.java
```diff
@@ -0,0 +1,66 @@
...
+package org.hibernate.validator.ap.testmodel.annotationparameters;
+
+import javax.validation.constraints.Size;
+
+/**
+ * @author Marko Bekhta
+ */
+public class ValidGroupsParameters {
+
+
```
gsmet: Unnecessary whitespace.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationParametersGroupsCheck.java
marko-bekhta: hmm... this `instanceof` is more for safety of next conversion. I've added it more for the cases when the annotation is of some different kind and groups parameter is present but contains something else ... But if it's only for constraint annotations it can be probably be removed completely. WDYT ?

gunnarmorling: The current check implementation would miss if e.g. `int.class` was given as a group. I think the following would be right:

* Safely assume a `TypeMirror` is given as a value (`groups` only expects class objects, so the user couldn't provide a String or long or so)
* If `getKind()` is not DECLARED, flag it as error (preventing something like `int.class` to be given)
* If kind is DECLARED but the type is not an interface, flag it as error (as you already do)

gunnarmorling:
Rebased and applied. Many thanks, @marko-bekhta!

PR #556, opened 2016-10-17
HV-1131 Deferring retrieval of default MI

gunnarmorling:
@gsmet I've found an alternative approach for the EL issue. The idea is to defer the retrieval of the default message interpolator until it's actually needed (either by calling `getDefaultMessageInterpolator()` or when bootstrapping the VF and no other MI has been given). That way the instantiation of the expression factory can simply be done in the 'ctor of `ResourceBundleMessageInterpolator`, not requiring any synchronization.

I've also rewritten the test so it doesn't need reflection in the actual test routine. I've reworked your commit so it adds the new test and your fixes around the test class loader, but ommitted your original approach for fixing the issue for the sake of a clean history.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
@@ -72,7 +72,13 @@ public class ConfigurationImpl implements HibernateValidatorConfiguration, Confi
 	private static final Log log = LoggerFactory.make();
 
 	private final ResourceBundleLocator defaultResourceBundleLocator;
-	private final MessageInterpolator defaultMessageInterpolator;
+
+	/**
+	 * Built lazily so RBMI and its dependency on EL is only initialized if actually needed
+	 */
+	private MessageInterpolator defaultMessageInterpolator;
+	private MessageInterpolator messageInterpolator;
+
 	private final TraversableResolver defaultTraversableResolver;
 	private final ConstraintValidatorFactory defaultConstraintValidatorFactory;
 	private final ParameterNameProvider defaultParameterNameProvider;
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -400,7 +404,7 @@ public ClassLoader getExternalClassLoader() {
 
 	@Override
 	public final MessageInterpolator getDefaultMessageInterpolator() {
-		return defaultMessageInterpolator;
+		return new ResourceBundleMessageInterpolator( defaultResourceBundleLocator );
```
gsmet: I thought about this approach but I must admit I really disliked the fact that we instantiate a new `ResourceBundleMessageInterpolator` each time we call this method (and `getMessageInterpolator`). I would expect us to return the same object for each call.

Thinking about it and as I noticed `Configuration` is not supposed to be thread safe, couldn't we store the interpolators the first time the methods are called? That would solve this issue.

gsmet:
> I've also rewritten the test so it doesn't need reflection in the actual test routine.

Nice!

gunnarmorling:
@gsmet Pushed an update.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
gsmet: Just to be clear on my worries: it's named like a getter and it used to work like a getter so I didn't think it was such a trivial change to make it return a different instance for each call. I don't think it's an inappropriate expectation to have it behave like a getter.

gunnarmorling:
Ok, I see. FWIW, I don't think that the JavaBeans spec mandates that a
getter always returns the same value. But it's changed as per your request
now.

gsmet:
Merged, thanks!

PR #559, opened 2016-10-18
HV-1134 Add the sourcedir attributes to the output-docbook execution

gsmet:
https://hibernate.atlassian.net/browse/HV-1134

~~No need to backport, it's new in 6.~~ in fact, it was pre existing to 6 so we should probably backport it too (see HipChat for more discussion about this) -> **I will backport it as part of the backport of the source includes patch.**

The real commit is this one https://github.com/hibernate/hibernate-validator/commit/abeb19295ee26ae5096c87fbb1f06555527b1d85 - the other one only fixes the indentation.


gunnarmorling:

PR #560, opened 2016-10-18
HV-1133 The order is not predictible so we'd better use assertj

gsmet:
@gunnarmorling this one should be merged quickly as it will cause CI failures.


gunnarmorling:
Merged; thanks!

PR #561, opened 2016-10-18
HV-1002 Fix collectMetaConstraintsForPath logic in the iterable case

gsmet:
Here is the second attempt at fixing HV-1002. I separated the `collectMetaConstraintsForPath` in 2 as it was becoming unreadable to manage both cases in the very same method. They are not recursive anymore.

It should also return better errors.

I added a second commit to rename the `getValue` methods.

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/internal/engine/ValidatorTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/internal/engine/ValidatorTest.java
@@ -6,10 +6,21 @@
  */
 package org.hibernate.validator.test.internal.engine;
 
+import static org.hibernate.validator.testutil.ConstraintViolationAssert.assertCorrectConstraintTypes;
+import static org.hibernate.validator.testutil.ConstraintViolationAssert.assertCorrectPropertyPaths;
+import static org.hibernate.validator.testutil.ConstraintViolationAssert.assertNumberOfViolations;
+import static org.hibernate.validator.testutils.ValidatorUtil.getValidator;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertSame;
+import static org.testng.Assert.assertTrue;
+
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+
 import javax.validation.ConstraintViolation;
 import javax.validation.GroupSequence;
 import javax.validation.Valid;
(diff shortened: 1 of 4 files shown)


engine/src/test/java/org/hibernate/validator/test/internal/engine/ValidatorTest.java
```diff
@@ -211,6 +214,57 @@ public void testValidateValueWithNestedPath() {
...
+		assertCorrectPropertyPaths( constraintViolationsK, "foo.genericProperty" );
+	}
+
+	@Test(expectedExceptions = { ValidationException.class }, expectedExceptionsMessageRegExp = "^HV000195:.*")
+	@TestForIssue(jiraKey = "HV-1002")
+	public void testValidatePropertyWithNestedPathAndNullPropertyInTheWay() {
+		Validator validator = getValidator();
+		X someX = new X();
+		Set<ConstraintViolation<X>> constraintViolationsX = validator.validateProperty( someX, "list[0].foo" );
+		assertNumberOfViolations( constraintViolationsX, 1 );
```
gunnarmorling: Seems these assertions are never reached due to the (expected) exception before?

engine/src/test/java/org/hibernate/validator/test/internal/engine/ValidatorTest.java
```diff
@@ -211,6 +214,57 @@ public void testValidateValueWithNestedPath() {
...
+		assertCorrectPropertyPaths( constraintViolationsX, "list[0].foo" );
+	}
+
+	@Test(expectedExceptions = { IllegalArgumentException.class }, expectedExceptionsMessageRegExp = "^HV000039:.*")
+	@TestForIssue(jiraKey = "HV-1002")
+	public void testValidatePropertyWithNestedPathAndMissingValid() {
+		Validator validator = getValidator();
+		M someM = new M();
+		Set<ConstraintViolation<M>> constraintViolationsM = validator.validateProperty( someM, "foo.baz" );
+		assertNumberOfViolations( constraintViolationsM, 1 );
```
gunnarmorling: Same here.

gsmet:
> I'm a bit concerned about ValidatorImpl growing more and more, it's becoming very tough to grok.

Agreed. It's becoming unmanageable.

gsmet:
I fixed your comments, rebased and force pushed.

gunnarmorling:
Rebased and applied to master. I tried to cherry-pick to 5.3 but it doesn't cleanly apply. Could you do it as you've done the original change?

gsmet:
@gunnarmorling of course, will do.

PR #562, opened 2016-10-19
HV-1032 Avoid infinite loop in TypeHelper.isAssignable

gsmet:
* https://hibernate.atlassian.net/browse/HV-1032

Should be backported to 5.3.


gunnarmorling:
Rebased and applied to master + 5.3. Thanks, @gsmet!

PR #564, opened 2016-10-20
HV-1137 Simplifying implementation of parameter and return value vali…

gunnarmorling:
…dation a bit

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -1147,9 +1147,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		}
 	}
 
-	private <T> int validateParametersForGroup(ValidationContext<T> validationContext, Object[] parameterValues, Group group) {
-		int numberOfViolationsBefore = validationContext.getFailingConstraints().size();
-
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -1173,32 +1171,29 @@ private int validatePropertyForNonDefaultGroup(ValueContext<?, Object> valueCont
 			while ( defaultGroupSequence.hasNext() ) {
 				Sequence sequence = defaultGroupSequence.next();
 				for ( GroupWithInheritance expandedGroup : sequence ) {
-					int numberOfViolationsOfCurrentGroup = 0;
+					int numberOfViolations = validationContext.getFailingConstraints().size();
```
gsmet: Couldn't this statement be outside of the loop?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -1387,33 +1378,29 @@ private int validatePropertyForNonDefaultGroup(ValueContext<?, Object> valueCont
...
```

gsmet:
See my comments inline.

I have mixed feeling about this. It does not really simplify the logic and we lose an information (which is admittedly useless at the time). No strong opinion though so if you prefer it like that, let's go for it.

gunnarmorling:
Reworked the commit to address your remark and merged. Thanks for reviewing, @gsmet!

PR #566, opened 2016-10-21
fix on StringHelper#decapitalize method by providing English locale

mulderbaba:


gunnarmorling:
ok to test.

engine/src/main/java/org/hibernate/validator/internal/util/StringHelper.java
```diff
@@ -81,7 +82,7 @@ public static String decapitalize(String string) {
 			return string;
 		}
 		else {
-			return string.substring( 0, 1 ).toLowerCase() + string.substring( 1 );
+			return string.substring( 0, 1 ).toLowerCase(Locale.ENGLISH) + string.substring( 1 );
```
gsmet: Good catch!

engine/src/test/java/org/hibernate/validator/test/internal/util/StringHelperTest.java
```diff
@@ -87,6 +88,12 @@ public void decapitalizeShouldReturnDecapizalizedWord() {
 		assertEquals( StringHelper.decapitalize( "Giraffe" ), "giraffe" );
 	}
 
+    @Test
+    public void decapitalizeShouldReturnDecapizalizedWordOnTurkishLocale() {
+        Locale.setDefault(new Locale("tr", "TR"));
+        assertEquals( StringHelper.decapitalize( "IsIsolationLevelGuaranteed" ), "isIsolationLevelGuaranteed" );
```
gsmet: Could you get the default locale before changing it and restore it after the test? It would be cleaner to let the state consistent after the test.

gsmet:
@mulderbaba Good catch! It's weird nobody got the problem before.

Apart from my comments inline:
- could you fix the formatting issues? Our code style requires space padding for method arguments (run mvn checkstyle:check to verify everything is OK - we have IDE code styles here https://github.com/hibernate/hibernate-ide-codestyles/ if you're interested in contributing on a regular basis)
- could you prefix your commit message with the JIRA number?
- could you add the `@TestForIssue` annotation with the JIRA number to your test?
- and last but not least, could you sign the CLA for the Hibernate Validator project here: https://cla.jboss.org/ and get back to me with the user name you used? It's a very simple form to fill once you have a JBoss account.

Thanks for your contribution!

mulderbaba:
@gsmet all set I guess
username for CLA is: mulderbaba
Thanks.

engine/src/main/java/org/hibernate/validator/internal/util/StringHelper.java
```diff
@@ -81,7 +82,7 @@ public static String decapitalize(String string) {
 			return string;
 		}
 		else {
-			return string.substring( 0, 1 ).toLowerCase() + string.substring( 1 );
+			return string.substring( 0, 1 ).toLowerCase( Locale.ENGLISH ) + string.substring( 1 );
```
gsmet: After giving it some thoughts, could you change it to use Locale.ROOT? I think it emphasizes better what we want to do.

After that, we're good to go!

Thanks!

mulderbaba:
@gsmet good call, done.

gsmet:
@mulderbaba OK, perfect. I'm waiting for the CLA approval and I'll merge it. It will be included in 5.3.1 we should release very soon.

mulderbaba:
great @gsmet. do you think it'd be possible to cherry-pick this into 5.2 branch also? I need this to be used by hk2 project in order get Payara working properly. looks like hk2 guys tend to stay on 5.2.4 atm.

gunnarmorling:
Hi @mulderbaba, thanks for your fix! We don't have plans for another 5.2 release at this point, but we can push it to 5.3. We've planned the release of 5.3.1 early next week, and there shouldn't be any issue with upgrading to that from 5.2.4.

gsmet:
Merged in master and backported to the 5.3 branch. Thanks!

PR #568, opened 2016-10-25
HV-1143 Removing ExecutableElement

gunnarmorling:
@gsmet A fall-out of my efforts for simplifying `ValidatorImpl`. No big thing, but it has a nice ratio of added vs. removed lines, so I thought I'd go for it.


engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ExecutableMetaData.java
```diff
@@ -122,15 +124,6 @@ private ExecutableMetaData(
...
 	public ParameterMetaData getParameterMetaData(int parameterIndex) {
-		if ( parameterIndex < 0 || parameterIndex > parameterMetaDataList.size() - 1 ) {
-			throw log.getInvalidExecutableParameterIndexException(
-					ExecutableElement.getExecutableAsString(
-							getType().toString() + "#" + getName(),
-							parameterTypes
-					), parameterTypes.length
-			);
-		}
-
```
gsmet: Any reason why you remove this test?

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedExecutable.java
```diff
@@ -211,10 +211,10 @@ public boolean hasParameterConstraints() {
 	 *         otherwise.
 	 */
 	public boolean isGetterMethod() {
-		return executable.isGetterMethod();
+		return ReflectionHelper.isGetterMethod( executable );
```
gsmet: Maybe add a final attribute for this?

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
```diff
@@ -128,6 +119,40 @@ public static String getSignature(String name, Class<?>[] parameterTypes) {
...
+
+		for ( Class<?> parameterType : parameterTypes ) {
+			if ( !isFirst ) {
+				sb.append( ", " );
+			}
+			else {
+				isFirst = false;
+			}
+
+			sb.append( parameterType.getSimpleName() );
```
gsmet: Surprised you're not using some Java 8 magic?

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableParameterNameProvider.java
```diff
@@ -0,0 +1,38 @@
...
+public class ExecutableParameterNameProvider {
+
+	private final ParameterNameProvider delegate;
+
+	public ExecutableParameterNameProvider(ParameterNameProvider delegate) {
+		this.delegate = delegate;
+	}
+
+	public List<String> getParameterNames(Executable executable) {
+		if ( executable instanceof Method ) {
```
gsmet: It would have been nice to change the `ParameterNameProvider` contract but I suppose we cannot do it taking into account the discussion we had yesterday.

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
```diff
@@ -128,6 +119,40 @@ public static String getSignature(String name, Class<?>[] parameterTypes) {
...
+	 * Returns a string representation of an executable with the given name and parameter types in the form
+	 * {@code <name>(<parameterType 0> ...  <parameterType n>)}, e.g. for logging purposes.
+	 *
+	 * @param name the name of the executable
+	 * @param parameterTypes the types of the executable's parameters
+	 *
+	 * @return A string representation of the given executable.
+	 */
+	public static String getExecutableAsString(String name, Class<?>... parameterTypes) {
+		StringBuilder sb = new StringBuilder( name );
```
gsmet: I know it will probably be harder to test but I would have expected a `getExecutableAsString(Executable executable)` method.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ExecutableMetaData.java
gunnarmorling: Yes, it wasn't needed really. The only invocation path not originating from tests already makes sure that the given parameter values match the length of the validated executable (see `ValidatorImpl#validateParametersForGroup()`). Hence this check never triggered anyways.

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedExecutable.java
gunnarmorling: +1, will do.

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
gunnarmorling: It is just copied/moved code, but I'll look into using the new joiner facility.

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
gunnarmorling: There are cases were we don't have an actual `Executable`, e.g. when configuring a method using the fluent API. Then this one is used to report a non-existing method. But see `ExecutableFormatter` which is doing what you say for cases where we have an executable (and makes invocations of `Log` a bit simpler).

gsmet:
Looks good apart from my minor comments. Not a big fan of wrapping the ParameterNameProvider though...

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ExecutableMetaData.java
gsmet: OK.

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableParameterNameProvider.java
gsmet: Well, we are writing the new BV 2 specification :).

gunnarmorling: Yes, but do you have a good idea how it could look like without breaking existing implementations? We cannot remove the existing methods from PNP (as that'd break existing implementations) so we only could add `getParameterNames(Executable e)`, but I think that'd only obfuscate the contract more than it helps.

gunnarmorling:
> Not a big fan of wrapping the ParameterNameProvider though...

I'm open for all suggestions :) It helps to avoid "if instanceof then" switches at several places.

gunnarmorling:
@gsmet Pushed two more commits for addressing your remarks and for benefiting a bit more from Java8 goodness.

gsmet:
Looks good. Waiting for CI to validate it and I will merge it.

gsmet:
Rebased and merged! Thanks!

gunnarmorling:
I've pushed one more commit related to the removal of the previous grouping functionality right to master: https://github.com/hibernate/hibernate-validator/commit/576ceb8891f6bb4121741bd7011e90e5f264454e. After being to the fresh air I've realized that the previous approach was les than ideal ;)

PR #569, opened 2016-10-26
HV-451 The annotation processor should check @GroupSequence

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-451

Added a check to annotation processor for @GroupSequence values:
 -the class list contains only interface
 -the defined group sequence is expandable (no cyclic definition)
 -the class list contains the hosting bean class (for default group sequence re-definition)

I'm not sure about the 3 option though, if it is implemented as expected or not.


annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationParametersGroupSequenceCheck.java
```diff
@@ -0,0 +1,124 @@
...
+ * Checks that the GroupSequence definition is valid.
+ * <ul>
+ * <li>the class list contains only interface</li>
+ * <li>the defined group sequence is expandable (no cyclic definition)</li>
+ * <li>the class list contains the hosting bean class (for default group sequence re-definition)</li>
+ * </ul>
+ *
+ * @author Marko Bekhta
+ */
+public class AnnotationParametersGroupSequenceCheck extends AnnotationParametersAbstractCheck {
```
gsmet: Let's call it `GroupSequenceCheck`.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationParametersGroupSequenceCheck.java
```diff
@@ -0,0 +1,124 @@
...
+			TypeMirror typeMirror = (TypeMirror) value.getValue();
+			boolean isClassAndSameAsAnnotatedElement = isClassForRedefiningGroupSequence( annotatedElement, typeMirror );
+
+			if ( !annotationApiHelper.isInterface( typeMirror ) && !isClassAndSameAsAnnotatedElement ) {
+				return CollectionHelper.asSet( ConstraintCheckIssue.error(
+						element, annotation, "INVALID_GROUP_SEQUENCE_VALUE_NOT_INTERFACES_ANNOTATION_PARAMETERS"
+				) );
+			}
+			// 2. the defined group sequence is expandable (no cyclic definition)
+			if ( isFromHierarchy( annotatedElement, typeMirror ) && !isClassAndSameAsAnnotatedElement ) {
```
gsmet: Do we agree that you don't detect cyclic definitions defined with `@GroupSequence`s on interfaces defining a group sequence? AFAICS, you only check the hierarchy, not the `@GroupSequence`s.

I'm thinking about something like:
```java
@GroupSequence(Group2.class)
public interface Group1 {
}

@GroupSequence(Group1.class)
public interface Group2 {
}
```

Not sure we can do anything about it but this is probably the most common case of cyclic definition (it can of course be far more complicated to detect than this very simple example).

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/ConstraintCheckFactory.java
```diff
@@ -239,7 +239,10 @@ public ConstraintCheckFactory(Types typeUtils, Elements elementUtils, Constraint
 						new AnnotationParametersGroupsCheck( annotationApiHelper )
 				)
 		);
-		nonAnnotationTypeChecks.put( AnnotationType.NO_CONSTRAINT_ANNOTATION, NULL_CHECKS );
+		nonAnnotationTypeChecks.put(
+				AnnotationType.NO_CONSTRAINT_ANNOTATION,
```
marko-bekhta: There's no such value present so I'll add it in the next commit.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationParametersGroupSequenceCheck.java
marko-bekhta: Correct ! :) I'm not very familiar with the group sequences, and apparently I didn't understood everything about them after reading the doc :) 

I've provided a solution for such problem in the next commit

annotation-processor/src/main/resources/org/hibernate/validator/ap/ValidationProcessorMessages.properties
```diff
@@ -40,3 +40,6 @@ INVALID_DIGITS_ANNOTATION_PARAMETERS=Invalid annotation parameters. Fraction and
 INVALID_DECIMAL_MIN_MAX_ANNOTATION_PARAMETERS=Invalid annotation parameters. Value should be a valid BigDecimal string representation.
 INVALID_GROUPS_VALUE_ANNOTATION_PARAMETERS=Invalid annotation parameters. Groups should be represented by interfaces.
 INVALID_MESSAGE_VALUE_ANNOTATION_PARAMETERS=Message seems to be malformed. Did you miss a '{}' in the message key?
+INVALID_GROUP_SEQUENCE_VALUE_NOT_INTERFACES_ANNOTATION_PARAMETERS=Invalid annotation parameters. Value should be represented by interfaces.
+INVALID_GROUP_SEQUENCE_VALUE_CYCLIC_DEFINITION_ANNOTATION_PARAMETERS=Invalid annotation parameters. The defined group sequence should be expandable (no cyclic definition).
+INVALID_GROUP_SEQUENCE_VALUE_HOSTING_BEAN_ANNOTATION_PARAMETERS=Invalid annotation parameters. The class list should contain the hosting bean class (for default group sequence re-definition).
```
gunnarmorling: The part "(for default group sequence re-definition)" could be omitted, it should be clear from the context of the error message.

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/InvalidGroupSequenceParameters.java
```diff
@@ -0,0 +1,176 @@
...
+	 * Case 1: class present in a list of groups
+	 */
+	public static class Case1 {
+		public interface Group1 {
+		}
+
+		public interface Group2 {
+		}
+
+		@GroupSequence(value = Group2.class)
```
gunnarmorling: Are `Group2`, `Group3` and `YetOtherBean` actually needed for this case? All tests pass if I comment them out, so they could go away it seems?

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/InvalidGroupSequenceParameters.java
```diff
@@ -0,0 +1,176 @@
...
+		public interface Group3 {
+		}
+
+		public class InvalidGroup1 {
+		}
+
+		public class InvalidGroup2 {
+		}
+
+		@GroupSequence(value = { Group1.class, InvalidGroup1.class })
```
gunnarmorling: What's the difference between this and `SomeOtherBean`? Seems one should be enough for the test?

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/InvalidGroupSequenceParameters.java
```diff
@@ -0,0 +1,176 @@
...
+
+		@GroupSequence(value = { Group1.class, int.class })
+		public class SomeBean {
+		}
+
+		@GroupSequence(value = boolean.class)
+		public class SomeOtherBean {
+		}
+
+		@GroupSequence(value = double.class)
```
gunnarmorling: I'd say one test for primitive is enough?

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/InvalidGroupSequenceParameters.java
```diff
@@ -0,0 +1,176 @@
...
+		public class YetOtherBean {
+
+		}
+	}
+
+	/**
+	 * Case 3: Cyclic groups
+	 */
+	public static class Case3 {
```
gunnarmorling: I'm a bit confused by this test :) Couldn't a simple cycle test look like this:

    @GroupSequence(value = {Group1.class, Group2.class, Group1.class})
    public interface MySequence {}

I wouldn't consider the case where a group sequence inherits from another interface. It's discouraged by the spec, so if anything, we could raise a warning when detecting it.

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/InvalidGroupSequenceParameters.java
```diff
@@ -0,0 +1,176 @@
...
+		public interface Group6 extends Group5 {
+		}
+
+	}
+
+
+	/**
+	 * Case 4: Example of redefining a group sequence without using a class in a list
+	 */
+	public static class Case4 {
```
gunnarmorling: Maybe this one could be slimmed down a bit?

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/GroupSequenceCheck.java
```diff
@@ -0,0 +1,150 @@
...
+			return CollectionHelper.asSet( ConstraintCheckIssue.error(
+					element, annotation, "INVALID_GROUP_SEQUENCE_VALUE_HOSTING_BEAN_ANNOTATION_PARAMETERS"
+			) );
+		}
+
+		return Collections.emptySet();
+	}
+
+	/**
+	 * Checks if a given type element ({@link TypeElement}) is not a part of definition of a group sequence annotation declared on
```
gunnarmorling: Remove the "not"?

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/GroupSequenceCheck.java
```diff
@@ -0,0 +1,150 @@
...
+	 * a given group element {@link TypeMirror}
+	 *
+	 * @param typeElement an element to check if it is present or not in definition of a group sequence
+	 * @param groupElement an type mirror at which to look for group sequence annotation
+	 *
+	 * @return {@code true} if given element is part of group sequence definition, and {@code false} otherwise
+	 */
+	private boolean isPartOfGroupSequence(TypeElement typeElement, TypeMirror groupElement) {
+		List<? extends AnnotationValue> annotationValue = null;
+		for ( AnnotationMirror annotationMirror : typeUtils.asElement( groupElement ).getAnnotationMirrors() ) {
```
gunnarmorling: This could be structured a bit nicer with a separate method:

    AnnotationMirror groupSequence = getGroupSequence( groupElement );
    annotationValue = annotationApiHelper.getAnnotationArrayValue( annotationMirror, "value" );

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/GroupSequenceCheck.java
```diff
@@ -0,0 +1,150 @@
...
+			if ( JAVAX_VALIDATION_GROUP_SEQUENCE.equals( ( (TypeElement) annotationMirror.getAnnotationType()
+					.asElement() ).getQualifiedName().toString() ) ) {
+				annotationValue = annotationApiHelper.getAnnotationArrayValue( annotationMirror, "value" );
+				break;
+			}
+		}
+		if ( annotationValue != null ) {
+			for ( AnnotationValue value : annotationValue ) {
```
gunnarmorling: Wouldn't you also have to check whether a given element has been specified in the sequence already?

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/annotationparameters/InvalidGroupSequenceParameters.java
marko-bekhta: Same as before :) you are right it's just more to cover different values. Maybe the implementation is just looking for `boolean` :) But again based on the fact that we know that it's determining if the provided class is a primitive or not we can leave just one :)

marko-bekhta:
Thanks for the review ! I'll apply the changes and when you respond to the comment on cycles I'll apply the changes based on it and push it all.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/GroupSequenceCheck.java
marko-bekhta: This method is receiving an annotated element (`typeElement`) with one of its values from `GroupSequence` annotation, so `groupElement` is already a part of one GroupSequence and we are checking if there's a GroupSequence annotation on it and if so then there shouldn't be an annotated element (`typeElement`) present within its values ... 

Or did you mean the case like this:
```
		@GroupSequence({Group2.class, Group2.class, Group3.class,Group2.class})
		public interface Group1 {
		}
```
where a group (Group2) is present multiple times ?

marko-bekhta:
Hi @gunnarmorling I've read all the comments once more with a fresh head :) and found all the answers :)
I've added a warning for a case when interfaces used in group sequences are extending other interfaces. Added a check if the interface was not already used in a given group sequence already. And also done some clean up. It'll be great if you have a look when you have time. Thanks !

Also I was thinking more about these cycles and here's a case when I think we should produce an error, please let me know if I'm right :)

```
public static class Case7 {
		public interface Group1 {
		}

		public interface Group2 {
		}

		@GroupSequence(value = { Group1.class, Group4.class })
		public interface Group3 {
		}

		@GroupSequence(value = Group5.class)
		public interface Group4 {
		}

		@GroupSequence(value = { Group2.class, Group3.class })
		public interface Group5 {
		}
	}
```

In this case it'll be Group3 -> Group4 -> Group5 -> Group3. Is this a valid case ?

If so I'll need to update the code to go through such chains and find all interfaces. Sorry for all these questions but as I said I'm not very familiar with `GroupSequence`s :)

gunnarmorling:
> In this case it'll be Group3 -> Group4 -> Group5 -> Group3. Is this a valid case ?

it seems not valid indeed. You could try to valid this (using the actual HV engine), it should raise an error.

marko-bekhta:
@gunnarmorling great, thanks for the answer! I'll do what you suggested with engine and make the changes to `GroupSequenceCheck` to handle such cases as well

marko-bekhta:
@gunnarmorling I checked that such case is failing when using an engine, and I've modified code to look for such cases as well. Let me know if anything else should be done for this task.

gsmet:
@marko-bekhta so I squashed and merged it with some additional cleanup (see https://github.com/hibernate/hibernate-validator/commit/a4d77518222c6822c2ba27b969e990d8dbaac5ed)

The main issue was that the following is perfectly valid:
```java
	/**
	 * Case 3: Hierarchy of groups
	 */
	public static class Case3 {
		public interface Group1 {
		}

		public interface Group2 extends Group1 {
		}

		public interface Group3 {
		}

		@GroupSequence(value = { Group2.class, Group3.class })
		public interface GroupSequence1 {
		}
	}
```

You can define hierarchy of groups. It's even one of the first examples in the specification: http://beanvalidation.org/latest-draft/spec/#constraintdeclarationvalidationprocess-groupsequence-groupinheritance .

What is not recommended is having a group sequence inheriting another interface so I changed it.

This misunderstanding had another consequence: you do not check that group inheritance creates cyclic definitions. I think you had something along those lines but checking group **sequences** inheritance once in your code and it was removed due to one of Gunnar's comment.

I did not have time to fix it so I would appreciate if you could do it.

I added a test at the bottom of InvalidGroupSequenceParameters which should throw an error (see the XXX).

We will release 5.4.0.Beta1 without this. It would be nice if you could fix it for CR1!

I created a specific issue to track it: https://hibernate.atlassian.net/browse/HV-1180 .

Thanks again!

marko-bekhta:
@gsmet Great ! I'll look into HV-1180.

PR #571, opened 2016-10-28
Adding BV 2.0 to module ZIP

gunnarmorling:
Hey @gsmet, this may be a possible answer to the module conundrum. I couldn't make it work to add BV 2.0 / HV 6.0 _next to_ the existing modules. While I could override the sub system injecting BV into the deployment, there are just too many fixed dependencies on _javax.validation.api:main_. This causes all sorts of ClassCastExceptions when the injected BV 2 `ValidatorFactory` instance is assigned into a variable of the BV 1 `ValidatorFactory` type.

But I've found a way to add BV 2.0 without overwriting the provided version, and with an easy way to go back to the latter if needed. The idea is to provide a separate "layered product". This is another module hierarchy in _modules/system/layers/bean-validation-2.0_ which is enabled via _modules/layers.conf_. Its modules take precedence over the ones from the _base_ module. To switch back to the provided ones, just comment the line in _layers.conf_. The nice thing is that this doesn't actually touch the container-provided modules.

In the long term we may need to have a way to have both versions of BV available at the same time, but I think the suggestion is a good way forward for us as well as folks interested in testing BV 2.0 without messing with existing modules.

@Sanne This may be interesting for your issue in HSEARCH, too. We may use that mechanism to safely override the JPA subsystem with a version that doesn't do that dreaded injection of Javassist. The same restriction as above applies - you couldn't have ORM 5.2 _and_ the default one - but you could add 5.2 into the mix with a safe way to go back. This could be done in a quick 5.2.5 release, which of course would only make sense if there wouldn't be a WF release soon containing the actual fix we've discussed.


gsmet:
Jenkins, retest this please

gsmet:
+1 for this approach. It is really nice to be able to test HV 6 and BV 2 throughout the whole WildFly stack.

Do you plan to fix the TODO before merging? Or should I merge it?

integration/pom.xml
```diff
@@ -152,7 +152,7 @@
                                     <classifier>wildfly-10-dist</classifier>
                                     <type>zip</type>
                                     <overWrite>false</overWrite>
-                                    <outputDirectory>${project.build.directory}/wildfly-${wildfly.version}/modules/system/layers/base/</outputDirectory>
```
gunnarmorling: This was a bad practice anyways. Custom modules should go directly under _modules_.

gsmet:
I changed the module name as discussed and merged. Thanks!

PR #572, opened 2016-10-31
HV-1147 Providing patch file with HV module update

gunnarmorling:
Hi @gsmet, that's an alternative approach for the module business. I think it's not mergeable as is, but eventually it may be the preferred route. Instead of a "dumb" ZIP this is now creating a patch file as consumeable by the WildFly patching infrastructure. It goes like this:

* In _modules_ the patch file is created; this takes a fresh WF 10.1 and another WF 10.1 with the updated modules in place; the patch gen tool then will create a patch file from the diff of the two dirs and an input descriptor
* In _integration_, _jboss-cli.sh patch apply path/to/patch/_ is used to apply the patch to the WF used for testing

Patching allows the user to roll back patches, see a patch history etc. So it's the recommended way to roll out such updates AFAIU, e.g. Weld is providing a patch file.

Unfornately, creating patches is more troublesome than it should be at this point: the _patch-gen_ tool doesn't exist in any repo I'm aware of and, worse, there is no Maven plug-in for it; so I did a lot of hackery to make it work in this PoC. Hence I'd not go down this route at least for the time being. But I think the approach is worth being pursued and we may get a discussion started with the folks maintaining the patch-gen tool.

WDYT?


modules/pom.xml
```diff
@@ -22,7 +22,9 @@
         <hibernate.validator.module.xml>${module.xml.basedir}/org/hibernate/validator/module.xml</hibernate.validator.module.xml>
         <hibernate.validator.cdi.module.xml>${module.xml.basedir}/org/hibernate/validator/cdi/module.xml</hibernate.validator.cdi.module.xml>
         <validation.api.module.xml>${module.xml.basedir}/javax/validation/api/module.xml</validation.api.module.xml>
-        <layer.name>hibernate-validator-6.0</layer.name>
+        <layer.name>base</layer.name>
```
gsmet: Looks like we can remove `layer.name`. I don't think it's used anymore.

gsmet:
Interesting (and impressive use of Maven hackery). I'm curious though to know how widespread the usage of this patch feature is.

If it is widespread, I agree with you it would be the way to go. And apart from the fact we need to download the tool from GitHub, it's not that bad.

gsmet:
Jenkins, retest this please

modules/pom.xml
gsmet: @gunnarmorling this (very minor) comment still stands. Other than that, it looks good!

gunnarmorling:
@gsmet Removed the layer property and a few others.

With that, I consider the actual work done. The only thing missing is a doc update which I'll send tomorrow.

gunnarmorling:
It may be we need to change the jobs on Jenkins to include settings-example.xml, I'm not 100% sure whether the plug-in is propagated to Maven Central, too.

gunnarmorling:
@gsmet pushed a commit for the docs.

gsmet:
Merged, thanks!

PR #573, opened 2016-11-01
HV-1148 Preparations for further code unification in ValidatorImpl

gunnarmorling:
This essentially pushes some logic into `Cascadable` and `ConstraintLocation` implementations (for appending path nodes), which hopefully will eventually allow us to have a unified validation routine in `ValidatorImpl`.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -33,7 +33,6 @@ import javax.validation.ConstraintValidatorFactory;
 import javax.validation.ConstraintViolation;
 import javax.validation.ElementKind;
 import javax.validation.MessageInterpolator;
-import javax.validation.ParameterNameProvider;
 import javax.validation.Path;
 import javax.validation.TraversableResolver;
 import javax.validation.Validator;
@@ -158,7 +157,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 			MessageInterpolator messageInterpolator,
 			TraversableResolver traversableResolver,
 			BeanMetaDataManager beanMetaDataManager,
-			ParameterNameProvider parameterNameProvider,
+			ExecutableParameterNameProvider parameterNameProvider,
 			TimeProvider timeProvider,
 			TypeResolutionHelper typeResolutionHelper,
 			List<ValidatedValueUnwrapper<?>> validatedValueHandlers,
@@ -168,7 +167,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		this.messageInterpolator = messageInterpolator;
 		this.traversableResolver = traversableResolver;
 		this.beanMetaDataManager = beanMetaDataManager;
-		this.parameterNameProvider = new ExecutableParameterNameProvider( parameterNameProvider );
+		this.parameterNameProvider = parameterNameProvider;
 		this.timeProvider = timeProvider;
 		this.typeResolutionHelper = typeResolutionHelper;
 		this.validatedValueHandlers = validatedValueHandlers;
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -65,6 +65,7 @@ import org.hibernate.validator.internal.util.ConcurrentReferenceHashMap;
 import org.hibernate.validator.internal.util.ConcurrentReferenceHashMap.ReferenceType;
 import org.hibernate.validator.internal.util.Contracts;
 import org.hibernate.validator.internal.util.ExecutableHelper;
+import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
 import org.hibernate.validator.internal.util.ReflectionHelper;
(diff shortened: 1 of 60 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -167,7 +168,7 @@ public ValidatorImpl(ConstraintValidatorFactory constraintValidatorFactory,
 		this.messageInterpolator = messageInterpolator;
 		this.traversableResolver = traversableResolver;
 		this.beanMetaDataManager = beanMetaDataManager;
-		this.parameterNameProvider = parameterNameProvider;
+		this.parameterNameProvider = new ExecutableParameterNameProvider( parameterNameProvider );
```
gsmet: Shouldn't we wrap it once and for all in `ValidatorFactoryImpl` instead of doing it this way? I'm especially worried about the fact that we might inadvertently wrap it several times.

engine/src/main/java/org/hibernate/validator/internal/engine/ValueContext.java
```diff
@@ -126,17 +124,9 @@ public final void setPropertyPath(PathImpl propertyPath) {
...
-		}
-		else if ( node.getKind() == ElementKind.PARAMETER ) {
-			propertyPath.addParameterNode( node.getName(), ( (ParameterMetaData) node ).getIndex() );
-		}
-		else if ( node.getKind() == ElementKind.RETURN_VALUE ) {
-			propertyPath.addReturnValueNode();
-		}
+		PathImpl newPath = PathImpl.createCopy( propertyPath );
+		node.appendTo( newPath );
+		propertyPath = newPath;
```
gsmet: Any reason why you create a temporary variable?

engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		if ( ReflectionHelper.isIterable( type ) || ReflectionHelper.isMap( type ) ) {
+			this.propertyName = null;
+		}
+		else {
+			this.propertyName = ReflectionHelper.getPropertyName( member );
+		}
+
+		if ( typeOfAnnotatedElement instanceof Class && ( (Class<?>) typeOfAnnotatedElement ).isPrimitive() ) {
```
gsmet: Didn't you introduce a new helper method just for that?

gsmet:
OK, looks good overall apart from my minor comments/questions.

I saw that you fixed the behavior of the type arguments constraints regarding the collection element node? Maybe an additional test would be in order.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Type argument constraint location.
+ *
+ * @author Hardy Ferentschik
+ * @author Gunnar Morling
+ */
+// TODO the behavior should not be based on the property name; e.g. there will not be a property name for a type
+// constraint on a method parameter
+class TypeArgumentConstraintLocation implements ConstraintLocation {
```
gunnarmorling: It's part of my upcoming work.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
gunnarmorling: Good catch, that's a left-over.

gunnarmorling:
> I saw that you fixed the behavior of the type arguments constraints regarding the collection element node? 

There should be no change in behavior. Can you elaborate a bit on where you think it has changed?

gsmet:
Maybe it's not triggered but I'm thinking about:
https://github.com/hibernate/hibernate-validator/pull/573/files#diff-af7dc46566c31e14b067152de18040d7R785
and
https://github.com/hibernate/hibernate-validator/pull/573/files#diff-e5cd06a63133ca330a97be728cde7d93R74

Before that, we added the collection element node unconditionally. But maybe you can't have a propertyName set if you're dealing with a type argument constraint?

gunnarmorling:
> But maybe you can't have a propertyName set if you're dealing with a type argument constraint?

You can have a property name with a type argument constraint (e.g. `private Optional<@Min(1) Integer> myPositiveInteger`). But the code for method validation does its own path appending atm., it doesn't talk to `ConstraintLocation#appendNode()` (as it eventually should). Hence I think everything fits together as of this PR, but for sure it's in an intermediate state I plan to fix in follow-up PRs.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gsmet: So, we just have this last question pending.

gunnarmorling:
@gsmet Pushed one more commit to address your last concern about the proliferation of EPNP instances. It happens only in `ValidatorFactoryImpl` and `ValidatorContextImpl` now (the latter in case a specific PNP is given for a single `Validator`).

gsmet:
Cool, merged!

PR #575, opened 2016-11-07
HV-1147 Some adjustments to enable usage of the patching mechanism on…

gunnarmorling:
… JDK 9

### DIFF
--- a/integration/pom.xml
+++ b/integration/pom.xml
@@ -214,6 +214,18 @@
                         <artifactId>wildfly-patching</artifactId>
                         <version>1.0.1.Final</version>
                     </dependency>
+                    <!-- The exclusion is needed to have the build work with JDK 9 -->
+                    <dependency>
+                        <groupId>org.wildfly.core</groupId>
+                        <artifactId>wildfly-cli</artifactId>
+                        <version>1.0.1.Final</version>
+                        <exclusions>
+                            <exclusion>
+                                <groupId>sun.jdk</groupId>
+                                <artifactId>jconsole</artifactId>
+                            </exclusion>
+                        </exclusions>
+                    </dependency>
                 </dependencies>
             </plugin>
         </plugins>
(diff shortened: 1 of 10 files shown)


integration/pom.xml
```diff
@@ -214,6 +214,18 @@
                         <artifactId>wildfly-patching</artifactId>
                         <version>1.0.1.Final</version>
                     </dependency>
+                    <!-- The exclusion is needed to have the build work with JDK 9 -->
+                    <dependency>
+                        <groupId>org.wildfly.core</groupId>
+                        <artifactId>wildfly-cli</artifactId>
+                        <version>1.0.1.Final</version>
+                        <exclusions>
+                            <exclusion>
```
gsmet: Couldn't you put the exclusion directly on the wildfly-patching artifact?

gunnarmorling:
Pushed another commit which makes the _documentation_ module pass, too. Only remaining issue is in _distribution_ related to JavaDocs.

gsmet:
> * amending Surefire arg line with required module exports

I don't see this part in the commit? Did I miss something?

And did you test locally (on a multi-core machine) that removing the 4 threads for the engine tests doesn't slow down the build?

gunnarmorling:
> I don't see this part in the commit? Did I miss something?

I happens by virtue of the `surefire.argLine` and `surefire.argLine.extension` properties in the parent POM. There is a "jdk-9" profile which does it once for all sub-modules.

> And did you test locally (on a multi-core machine) that removing the 4 threads for the engine tests doesn't slow down the build?

Ah, seems I went a bit too far by trying to centralize the Surefire config. I've added this back.

I've also pushed another commit which makes the distribution build work. JavaDoc can be generated now on 9 without a build failure, but it looks broken. I reckon the style sheet was changed in Java 9 again. I'll leave that for another time (going to open an issue).

gsmet:
Merged, thanks!

PR #576, opened 2016-11-09
HV-1135 Improved java.time support

gsmet:
See beanvalidation/beanvalidation-api#70 for a general presentation.

See https://github.com/beanvalidation/beanvalidation-tck/pull/77 for the TCK changes.


engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -66,14 +65,6 @@
 	 */
 	String CONSTRAINT_MAPPING_CONTRIBUTORS = "hibernate.validator.constraint_mapping_contributors";
 
-	/**
-	 * Property corresponding to the {@link #timeProvider(TimeProvider)} method. Accepts a String with the
-	 * fully-qualified class name of a {@link TimeProvider} implementation.
-	 *
-	 * @since 5.2
-	 */
-	String TIME_PROVIDER = "hibernate.validator.time_provider";
```
gunnarmorling: I think we should leave it in for now (circling back to our discussion about compatability). But deprecate it and remove the documentation.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -169,20 +160,6 @@
...
-	 * time and the current default time zone will be used when validating these constraints.
-	 *
-	 * @param timeProvider the time provider to register. Must not be {@code null}
-	 *
-	 * @return {@code this} following the chaining method pattern
-	 *
-	 * @hv.experimental This API is considered experimental and may change in future revisions
-	 * @since 5.2
-	 */
-	HibernateValidatorConfiguration timeProvider(TimeProvider timeProvider);
```
gunnarmorling: Same here.

engine/src/main/java/org/hibernate/validator/HibernateValidatorContext.java
```diff
@@ -68,20 +67,6 @@
...
 	/**
-	 * Registers the given time provider with the bootstrapped validator. This provider will be used to obtain the
-	 * current time when validating {@code @Future} and {@code @Past} constraints. If not set or if {@code null} is
-	 * passed as a parameter, the time provider of the {@link javax.validation.ValidatorFactory} is used.
-	 *
-	 * @param timeProvider the time provider to register.
-	 *
-	 * @return {@code this} following the chaining method pattern
-	 *
-	 * @hv.experimental This API is considered experimental and may change in future revisions
```
gunnarmorling: Ah, I see we marked it as experimental. Then it should be alright to remove.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/future/FutureValidatorForReadableInstant.java
```diff
@@ -1,41 +0,0 @@
...
-import javax.validation.ConstraintValidator;
-import javax.validation.ConstraintValidatorContext;
-import javax.validation.constraints.Future;
-
-import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorContext;
-import org.hibernate.validator.spi.time.TimeProvider;
-import org.joda.time.ReadableInstant;
-
-/**
- * Check if Joda Time type who implements {@code import org.joda.time.ReadableInstant} is in the future.
```
gunnarmorling: Are we removing the Joda support?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractEpochBasedTimeValidator.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Base class for all time validators that use an epoch to be compared to the time reference.
+ *
+ * @author Alaa Nassef
+ * @author Guillaume Smet
+ */
+public abstract class AbstractEpochBasedTimeValidator<C extends Annotation, T> implements ConstraintValidator<C, T> {
+
+	private static final Log log = LoggerFactory.make();
```
gunnarmorling: I know it's done inconsistently, but I'd prefer to have this named `LOG` (that's how I do it in new code) as it's a constant.

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter11/context/MyFutureValidator.java
```diff
@@ -12,14 +10,14 @@
 import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorContext;
 
 //tag::include[]
-public class MyFutureValidator implements ConstraintValidator<Future, Date> {
+public class MyFutureValidator implements ConstraintValidator<Future, Instant> {
```
gsmet: I thought it was a good idea for our examples to use the new java.time API. Nothing more.

engine/src/main/java/org/hibernate/validator/HibernateValidatorContext.java
gsmet: Yes, we discussed about it a few weeks ago on HipChat and we decided to remove it as:
- it was experimental
- having both ClockProvider and TimeProvider would be confusing

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/future/FutureValidatorForReadableInstant.java
gsmet: No, we are still supporting it. I moved the class and changed the implementation so git wasn't able to detect it's a move.

gsmet:
Pushed a rebase.

@gunnarmorling Interesting issues to discuss when you'll get there:
 * https://hibernate.atlassian.net/browse/HV-898
 * https://hibernate.atlassian.net/browse/HV-820

cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidatorFactoryBean.java
```diff
@@ -51,7 +53,7 @@
 
 	public ValidatorFactoryBean(BeanManager beanManager, ValidationProviderHelper validationProviderHelper) {
 		this.beanManager = beanManager;
-		this.destructibleResources = newHashSet( 4 );
+		this.destructibleResources = newHashSet( 5 );
```
gunnarmorling: I kinda regret we started doing this for sets and maps.

Unfortunately, initializing them with the number of intended elements will cause a resize. The reason being we don't take the load factor into account which is 0.75 by default. So a resize will happen after inserting the 3rd or 4th element. So in order to let it hold 5 elements, we'd have to initialize it with 8 (8 * 0.75 > 5). But it's all such micro optimization, that I think we could go for the default constructor equally well.

cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidatorFactoryBean.java
```diff
@@ -182,6 +185,25 @@ private ParameterNameProvider createParameterNameProvider(Configuration<?> confi
...
+	private ClockProvider createClockProvider(Configuration<?> config) {
+		BootstrapConfiguration bootstrapConfiguration = config.getBootstrapConfiguration();
+		String clockProviderFqcn = bootstrapConfiguration.getClockProviderClassName();
+
+		if ( clockProviderFqcn == null ) {
+			return config.getDefaultClockProvider();
+		}
+
+		@SuppressWarnings("unchecked")
+		Class<ClockProvider> clockProviderClass = (Class<ClockProvider>) run(
```
gunnarmorling: Technically, wouldn't it be `Class<? extends ClockProvider>`?

documentation/src/main/asciidoc/ch08.asciidoc
```diff
@@ -257,6 +257,45 @@ for obtaining parameter names at runtime. Refer to <<section-paranamer-parameter
 to learn more about this specific implementation.
 ====
 
+[[section-clock-provider]]
+==== `ClockProvider`
+
+For time related validation (`@Past` and `@Future` constraints for instance), it might be useful to define what is
+considered `now`.
+
+This is especially important when you want to test your constraints in a stable environment.
```
gunnarmorling: stable environment=reliable manner

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter08/BootstrappingTest.java
```diff
@@ -65,6 +67,16 @@ public void traversableResolver() {
 		//end::traversableResolver[]
 	}
 
+	public void clockProvider() {
+		//tag::clockProvider[]
+		ValidatorFactory validatorFactory = Validation.byDefaultProvider()
+				.configure()
+				.clockProvider( new FixedClockProvider( ZonedDateTime.now() ) )
```
gunnarmorling: Might make more sense to actually show usage of a fixed date: `ZonedDateTime.of("2016-...")`.

engine/src/main/java/org/hibernate/validator/HibernateValidatorContext.java
gunnarmorling: +1

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractJavaTimeValidator.java
```diff
@@ -0,0 +1,68 @@
...
+
+		if ( isPast ) {
+			return isPresentValid ? ( result <= 0 ) : ( result < 0 );
+		}
+		else {
+			return isPresentValid ? ( result >= 0 ) : ( result > 0 );
+		}
+	}
+
+	protected abstract T getReferenceValue(Clock reference);
```
gunnarmorling: A small comment would be nice.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractJavaTimeValidator.java
```diff
@@ -0,0 +1,68 @@
...
+			return true;
+		}
+
+		Clock reference;
+
+		try {
+			ClockProvider clockProvider = context.getClockProvider();
+			reference = clockProvider.getClock();
+		}
+		catch (Exception e) {
```
gunnarmorling: Is there any specific reason for wrapping this exception here?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractJavaTimeValidator.java
```diff
@@ -0,0 +1,68 @@
...
+			throw LOG.getUnableToGetCurrentTimeFromClockProvider( e );
+		}
+
+		int result = value.compareTo( getReferenceValue( reference ) );
+
+		if ( isPast ) {
```
gunnarmorling: Re-use is desirable, but I'd prefer specific implementations. One if pushed from the execution path to solely validator resolution time.

engine/src/main/java/org/hibernate/validator/internal/engine/DefaultClockProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author Guillaume Smet
+ */
+public class DefaultClockProvider implements ClockProvider {
+
+	private static final DefaultClockProvider INSTANCE = new DefaultClockProvider();
+
+	private DefaultClockProvider() {
+	}
+
+	public static final DefaultClockProvider getInstance() {
```
gunnarmorling: Usually we just make the `INSTANCE` member public. No strong preference, just noticed the slight derivation from the usual pattern.

gunnarmorling: Ah, I guess that stems from the original class.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -681,4 +682,11 @@ ConstraintDeclarationException getInconsistentValueUnwrappingConfigurationBetwee
 
 	@Message(id = 199, value = "Cannot extract value for type %1$s with more than one type parameters.")
 	ValidationException getUnableToExtractValueForTypeWithMultipleTypeParametersException(Type type);
+
+	@LogMessage(level = INFO)
```
gunnarmorling: Should it be just DEBUG?

engine/src/main/java/org/hibernate/validator/internal/xml/BootstrapConfigurationImpl.java
```diff
@@ -164,6 +173,7 @@ public String toString() {
 		sb.append( ", messageInterpolatorClassName='" ).append( messageInterpolatorClassName ).append( '\'' );
```
gunnarmorling: The new toShortName() facility could be used here.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
```diff
@@ -204,6 +215,22 @@ private void setParameterNameProvider(String parameterNameProviderFqcn, ClassLoa
...
+			try {
+				@SuppressWarnings("unchecked")
+				Class<? extends ClockProvider> clazz = (Class<? extends ClockProvider>) run(
+						LoadClass.action( clockProviderFqcn, externalClassLoader )
+				);
+				clockProvider = run( NewInstance.action( clazz, "clock provider class" ) );
+				log.usingClockProvider( clazz );
+			}
+			catch (ValidationException e) {
+				throw log.getUnableToInstantiateParameterNameProviderClassException( clockProviderFqcn, e );
```
gunnarmorling: ParameterNameProvider=ClockProvider

Or, better yet, let's just move to `getUnableToInstantiateException()`, I don't think there's much value in having the dedicated versions really.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/bv/time/ClockProviderFutureTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+	@BeforeMethod
+	public void setupValidator() {
+		validator = validatorFactory.getValidator();
+	}
+
+	@Test
+	public void clockProviderIsUsedForFutureOnReadableInstant() {
+		Order order = new Order();
+		order.shipmentDateAsReadableInstant = new DateTime( 2099, 2, 15, 4, 0, 0 );
```
gunnarmorling: Puh, I was wondering whether the test may begin to fail in 2100, but it looks safe :)

cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidatorFactoryBean.java
gsmet: You lost your right to make this sort of comment when you voted against my PR to adapt CollectionHelper to avoid the resize :) - see https://github.com/hibernate/hibernate-ogm/pull/683.

I really think we should do it once and for all and be done with it.

gunnarmorling:
Ah, didn't remember that we had discussed it before.

Personally I'd just use the default constructor in the very most cases or
pass a size big enough for the few cases where it seems useful.

cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidatorFactoryBean.java
gunnarmorling: Ah, didn't remember that we had discussed it before. Personally I'd just use the default constructor in the very most cases or pass a size big enough for the few cases where it seems useful.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gsmet: It's in line with the others. See https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java#L80 for instance.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
gsmet: The usage of getUnableToInstantiateException and the other similar one is in fact very specific. I changed it to getUnableToInstantiateClockProviderClassException.

cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidatorFactoryBean.java
gsmet: Yeah, the issue is that we often do CollectionHelper.newCollection( otherCollection.size() ). And I won't expect people to do the math by themselves.

I'll cherry-pick the OGM commit to fix this once and for all.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractJavaTimeValidator.java
gsmet: Yes, as I don't know what the implementor will do, I thought it was better to catch and wrap the exception.

cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidatorFactoryBean.java
gsmet: Fixed.

engine/src/main/java/org/hibernate/validator/internal/engine/DefaultClockProvider.java
gsmet: Yup, coming from `TimeProvider`. Fixed anyway.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractJavaTimeValidator.java
gsmet: Used an abstract method instead.

documentation/src/main/asciidoc/ch08.asciidoc
gsmet: Fixed.

cdi/src/main/java/org/hibernate/validator/internal/cdi/ValidatorFactoryBean.java
gsmet: Fixed.

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter08/BootstrappingTest.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractJavaTimeValidator.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/xml/BootstrapConfigurationImpl.java
gsmet: Not really. We have a String not a Type or a Class. We could implement one to work on String but I don't think it's really worth it in this case.

gsmet:
I think I addressed all the comments I found valid. Feel free to insist on the ones I discarded if you feel strongly about it.

gunnarmorling:
Merged; thanks, @gsmet!

PR #577, opened 2016-11-09
HV-1148 further unification of previously redundant code

gunnarmorling:
Hey @gsmet, here's the next chunk coming up :)

Let me know what you think, I feel the effort is beginning to pay off as the places in `ValidatorImpl` where we'd do more or less the same things are getting less.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
@@ -147,6 +147,11 @@ public class ValidationContext<T> {
 
 	private final TimeProvider timeProvider;
 
+	/**
+	 * The name of the validated (leaf) property in case of a validateProperty()/validateValue() call.
+	 */
+	private String validatedProperty;
+
 	private ValidationContext(ConstraintValidatorManager constraintValidatorManager,
 			MessageInterpolator messageInterpolator,
 			ConstraintValidatorFactory constraintValidatorFactory,
@@ -394,6 +399,14 @@ public class ValidationContext<T> {
 		}
 
 		return null;
+	}
+
+	public String getValidatedProperty() {
+		return validatedProperty;
+	}
+
+	public void setValidatedProperty(String validatedProperty) {
+		this.validatedProperty = validatedProperty;
 	}
 
 	@Override
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -11,15 +11,10 @@ import static org.hibernate.validator.internal.util.CollectionHelper.newHashMap;
 import static org.hibernate.validator.internal.util.logging.Messages.MESSAGES;
 
 import java.lang.annotation.ElementType;
-import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Executable;
-import java.lang.reflect.Field;
-import java.lang.reflect.Member;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
-import java.security.AccessController;
-import java.security.PrivilegedAction;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -27,7 +22,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.ConcurrentMap;
 
 import javax.validation.ConstraintValidatorFactory;
 import javax.validation.ConstraintViolation;
@@ -60,8 +54,6 @@ import org.hibernate.validator.internal.metadata.aggregated.ReturnValueMetaData;
 import org.hibernate.validator.internal.metadata.core.MetaConstraint;
 import org.hibernate.validator.internal.metadata.facets.Cascadable;
 import org.hibernate.validator.internal.metadata.facets.Validatable;
-import org.hibernate.validator.internal.util.ConcurrentReferenceHashMap;
-import org.hibernate.validator.internal.util.ConcurrentReferenceHashMap.ReferenceType;
 import org.hibernate.validator.internal.util.Contracts;
 import org.hibernate.validator.internal.util.ExecutableHelper;
 import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
@@ -70,9 +62,6 @@ import org.hibernate.validator.internal.util.TypeHelper;
 import org.hibernate.validator.internal.util.TypeResolutionHelper;
 import org.hibernate.validator.internal.util.logging.Log;
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
-import org.hibernate.validator.internal.util.privilegedactions.GetDeclaredField;
-import org.hibernate.validator.internal.util.privilegedactions.GetDeclaredMethod;
-import org.hibernate.validator.internal.util.privilegedactions.SetAccessibility;
 import org.hibernate.validator.spi.time.TimeProvider;
 import org.hibernate.validator.spi.valuehandling.ValidatedValueUnwrapper;
 
@@ -148,11 +137,6 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 	 */
 	private final List<ValidatedValueUnwrapper<?>> validatedValueHandlers;
 
-	/**
-	 * Keeps an accessible version for each non-accessible member whose value needs to be accessed during validation.
-	 */
-	private final ConcurrentMap<Member, Member> accessibleMembers;
-
 	public ValidatorImpl(ConstraintValidatorFactory constraintValidatorFactory,
 			MessageInterpolator messageInterpolator,
 			TraversableResolver traversableResolver,
@@ -175,12 +159,6 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		this.failFast = failFast;
 
 		validationOrderGenerator = new ValidationOrderGenerator();
-
-		this.accessibleMembers = new ConcurrentReferenceHashMap<>(
-				100,
-				ReferenceType.SOFT,
-				ReferenceType.SOFT
-		);
 	}
 
 	@Override
@@ -572,7 +550,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 			}
 		}
 
-		validationSuccessful = validateMetaConstraint( validationContext, valueContext, metaConstraint );
+		validationSuccessful = validateMetaConstraint( validationContext, valueContext, valueContext.getCurrentBean(), metaConstraint );
 
 		// reset the unwrapping mode
 		valueContext.setUnwrapMode( UnwrapMode.AUTOMATIC );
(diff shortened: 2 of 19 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -853,19 +851,20 @@ private void validateTypeArgumentConstraints(ValidationContext<?> context,
 	 *
 	 * @return The number of constraint violations raised when validating the {@code ValueContext} current group.
 	 */
-	private int validatePropertyForCurrentGroup(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, List<MetaConstraint<?>> metaConstraints, List<MetaConstraint<?>> typeUseConstraints) {
```
gsmet: We should update the javadoc here too.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ValidatableParametersMetaData.java
```diff
@@ -18,14 +25,41 @@
...
+	private final List<ParameterMetaData> parameterMetaData;
 	private final Iterable<Cascadable> cascadables;
 
-	public ValidatableParametersMetaData(Iterable<? extends Cascadable> cascadables) {
-		this.cascadables = CollectionHelper.<Cascadable>newHashSet( cascadables );
+	public ValidatableParametersMetaData(List<ParameterMetaData> parameterMetaData) {
+		this.parameterMetaData = Collections.unmodifiableList( parameterMetaData );
+		this.cascadables = Collections.unmodifiableList( parameterMetaData.stream()
+			.filter( p -> p.isCascading() )
+			.collect( Collectors.toList() ) );
```
gsmet: You changed the impl from a `Set` to a `List`. This is what you intended to do?

engine/src/main/java/org/hibernate/validator/internal/metadata/location/PropertyConstraintLocation.java
```diff
@@ -68,6 +81,20 @@ public void appendTo(ExecutableParameterNameProvider parameterNameProvider, Path
 	}
 
 	@Override
+	// TODO Probably this should be done with a security check to prevent direct usage by external clients
+	public Object getValue(Object parent) {
```
gsmet: Do you have an idea on how to secure this?

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
```diff
@@ -175,7 +176,15 @@ public BeanMetaDataImpl(Class<T> beanClass,
 				cascadedProperties.add( propertyMetaData );
 			}
 			else {
-				allMetaConstraints.addAll( propertyMetaData.getTypeArgumentsConstraints() );
+				Set<MetaConstraint<?>> typeConstraints = propertyMetaData.getTypeArgumentsConstraints().stream()
```
gsmet: Shouldn't we do this sooner? I'm a bit worried we transform them along the way.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -1461,6 +1460,7 @@ else if ( propertyPathNode.getKey() != null ) {
 
 		metaConstraints.addAll( propertyMetaData.getConstraints() );
 		typeArgumentConstraints.addAll( propertyMetaData.getTypeArgumentsConstraints() );
+		propertyPath.removeLeafNode();
```
gsmet: Not sure I understand the rationale of this commit. Maybe a more explicit commit message and a comment around the two `propertyPath.removeLeafNode()` would help understand what is going on?

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
```diff
@@ -323,8 +325,10 @@ public boolean defaultGroupSequenceIsRedefined() {
 
 	@Override
 	public UnwrapMode getUnwrapMode(ConstraintLocation location) {
-		if ( location.getPropertyName() != null ) {
-			return propertyMetaDataMap.get( location.getPropertyName() ).unwrapMode();
+		PropertyConstraintLocation propertyLocation = getPropertyConstraintLocation( location );
```
gsmet: Nice, I like this change.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -147,6 +147,11 @@
 
 	private final TimeProvider timeProvider;
 
+	/**
```
gsmet: Typo in the commit message of this particular commit f3a0154406b015961d50b0e6a6ae03a9857b7784: s/implmentation/implementation/

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -1335,15 +1125,11 @@ private void validatePropertyForNonDefaultGroup(ValidationContext<?> validationC
 	 *
 	 * @param validationContext The validation context.
 	 * @param propertyPath The property path for which constraints have to be collected.
-	 * @param metaConstraints A list where {@code MetaConstraint}s which match the given path are saved.
-	 * @param typeArgumentConstraints A list where type arguments {@code MetaConstraint}s which match the given path are saved.
 	 *
 	 * @return Returns an instance of {@code ValueContext} which describes the local validation context associated to the given property path.
 	 */
 	private <V> ValueContext<?, V> collectMetaConstraintsForPathWithValue(ValidationContext<?> validationContext,
```
gsmet: We should change the name of the method as it's not collecting constraints anymore.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -1415,15 +1200,11 @@ else if ( propertyPathNode.getKey() != null ) {
 	 *
 	 * @param validationContext The validation context.
 	 * @param propertyPath The property path for which constraints have to be collected.
-	 * @param metaConstraints A list where {@code MetaConstraint}s which match the given path are saved.
-	 * @param typeArgumentConstraints A list where type arguments {@code MetaConstraint}s which match the given path are saved.
 	 *
 	 * @return Returns an instance of {@code ValueContext} which describes the local validation context associated to the given property path.
 	 */
 	private <V> ValueContext<?, V> collectMetaConstraintsForPathWithoutValue(ValidationContext<?> validationContext,
```

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -1601,4 +1388,18 @@ private Object getBeanPropertyValue(ValidationContext<?> validationContext, Obje
 
 		return value;
 	}
+
+	private String getPropertyName(ConstraintLocation location) {
```
gsmet: I like https://github.com/hibernate/hibernate-validator/pull/577/commits/4fe2bfb21d7c40054371c99446cf096bd7aedfe2 but it's a bit unfortunate we have to duplicate this logic several times.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gsmet: Withdrawn, you fixed it in a follow-up commit.

gsmet:
I added a few comments here and there, mostly minor things.

I must admit that I would like an explanation for this commit https://github.com/hibernate/hibernate-validator/pull/577/commits/4f98f83d06a3f0f4100951be61bbdf16c88e8794 which I did not understand fully.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ValidatableParametersMetaData.java
gunnarmorling: I don't think it makes any effective difference at this point. I kinda liked that it keeps the order of the parameters when iterating through the cascaded ones.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gunnarmorling: Previously, the path would already have contained the node for the property of interest. Now we'll add that node when validating the constraint (via that new single `validateMetaConstraint()` method). Hence I'm removing it from the path so we don't end up with having it twice.

One could argue that the new approach is a bit less efficient (as we don't pre-select the relevant constraints), but I preferred that for the sake of less redundant code. We can look into optimizing things later on if needed.

Eventually we also should visit the constraints in order of their location. I.e. first all constraints on Foo#bar, then all on Foo#baz etc. Same for method parameters. This will allow us to do less appendNode()/removeNode() calls, i.e. not do it per constraint but per location. It also should help with that BVAL issue about ordered validation of constraints. I think we are getting to a better base for that.

gunnarmorling:
> I must admit that I would like an explanation for this commit 4f98f83 which I did not understand fully.

Previously, the path would already have contained the node for the property of interest. Now we'll add that node when validating the constraint (via that new single validateMetaConstraint() method). Hence I'm removing it from the path so we don't end up with having it twice.

One could argue that the new approach is a bit less efficient (as we don't pre-select the relevant constraints), but I preferred that for the sake of less redundant code. We can look into optimizing things later on if needed.

Eventually we also should visit the constraints in order of their location. I.e. first all constraints on Foo#bar, then all on Foo#baz etc. Same for method parameters. This will allow us to do less appendNode()/removeNode() calls, i.e. not do it per constraint but per location. It also should help with that BVAL issue about ordered validation of constraints. I think we are getting to a better base for that.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ValidatableParametersMetaData.java
gsmet: Yes, I think using List would be beneficial in a lot of places. When I looked at the constraints validation ordering, having Sets everywhere annoyed me. I think it would be nice to have the result of a validation as deterministic as possible.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
gsmet: My question was more: is this special casing specific to this place or should we move it a bit sooner to get PropertyMetaData return these transformed type parameter constraints?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gsmet: OK. Could you add a comment at the end of the `propertyPath.removeLeafNode();` lines so that we remember why we remove this last node?

It wasn't obvious when I read the code so I'm pretty sure it won't be in 2 years when I'll read that code again!

engine/src/main/java/org/hibernate/validator/internal/metadata/location/PropertyConstraintLocation.java
gunnarmorling: I've filed [HV-1157](https://hibernate.atlassian.net/browse/HV-1157).

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
gunnarmorling: Ideally, we'd get rid of the special handling of type parameter constraints altogether. But my preference is to leave this as is for now and look into it again when implementing the container stuff proper from BV 2.0. The delegate business in `TypeParameterLocation` is kind of a hack, but I think it's good enough for now and allows us to move on.

gunnarmorling:
@gsmet, rebased, amended a few commit messages and force pushed.

gsmet:
Rebased and merged! Nice progress on the clean-up!

PR #578, opened 2016-11-10
HV-1153 Missing javax.el impl should throw a ValidationException

gsmet:
* https://hibernate.atlassian.net/browse/HV-1153

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
@@ -640,8 +640,8 @@ public interface Log extends BasicLogger {
 	ValidationException getNoUnwrapperFoundForTypeException(Type type);
 
 	@Message(id = 183,
-			value = "Unable to load 'javax.el.ExpressionFactory'. Check that you have the EL dependencies on the classpath, or use ParameterMessageInterpolator instead")
-	ValidationException getMissingELDependenciesException();
+			value = "Unable to initialize 'javax.el.ExpressionFactory'. Check that you have the EL dependencies on the classpath, or use ParameterMessageInterpolator instead")
+	ValidationException getUnableToInitializeELExpressionFactoryException(@Cause Throwable e);
 
 	@LogMessage(level = WARN)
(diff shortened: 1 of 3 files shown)


skovtunenko:
@gsmet , but this dependency on javax.el implementation:
`
<dependency>
    <groupId>org.glassfish.web</groupId>
    <artifactId>javax.el</artifactId>
    <version>2.2.4</version>
</dependency>
`
licensed under 2 OSI approved licenses :
> COMMON DEVELOPMENT AND DISTRIBUTION LICENSE (CDDL - Version 1.1)
> GNU General Public License (GPL - Version 2, June 1991) with the Classpath Exception

That means that users of **Hibernate-validator** library should distribute their sourcode. How we can use new **Hibernate-validator** versions with our closed-source application?

gsmet:
Hi @skovtunenko ,

First we didn't really change anything except we are now failing early instead of failing at runtime.

Second, the CDDL allows you to distribute the javax.el dependency in a proprietary software.

Finally, if you really don't need it, you can use another MessageInterpolator:
```
Validator validator = Validation.byDefaultProvider()
				.configure()
				.messageInterpolator( new ParameterMessageInterpolator() )
				.buildValidatorFactory()
				.getValidator();
```

All in all, I don't think it's an issue.

gunnarmorling:
> That means that users of Hibernate-validator library should distribute their sourcode.

No, that's not true. That's exactly what GPL *with the Classpath Exception* prevents. Check out the very last box on https://uel.java.net/license.html.

skovtunenko:
Hi @gsmet , @gunnarmorling ! 
Thanks for your quick response, will read carefully.
My initial assumptions based on this CDDL statement:
> 3.1. Availability of Source Code.
> Any Covered Software that You distribute or otherwise make available in Executable form must also 
> be made available in Source Code form and that Source Code form must be distributed only under 
> the terms of this License.

gunnarmorling:
AFAIU, this clause only affects the files under CDDL themselves, but not any other "work" using or integrating the CDDL-licensed files.

But either way, UEL is dual licensed, CDDL *and* GPL with the Classpath Exception which allows you to create an executable under your own terms.

Note that there also is [JUEL](http://juel.sourceforge.net/) which is licensed under the Apache License 2.0. You have plenty of options here :)

PR #580, opened 2016-11-20
HV-1155 Classloading issues in modularized environments

gsmet:

Note: this PR is against the 5.3 branch as the primary goal was to fix an issue for WF.next.

It shouldn't be applied as is as we should discuss the addition of `com.sun.el` in the Import-Package declaration of hibernate-validator.


gsmet:
Superseded by https://github.com/hibernate/hibernate-validator/pull/581

PR #581, opened 2016-11-21
HV-1155 Class loading issues in modular environments

gsmet:

Long story short...

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/SimpleELContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/SimpleELContext.java
@@ -7,11 +7,13 @@
 package org.hibernate.validator.internal.engine.messageinterpolation.el;
 
 import java.lang.reflect.Method;
+
 import javax.el.ArrayELResolver;
 import javax.el.BeanELResolver;
 import javax.el.CompositeELResolver;
 import javax.el.ELContext;
 import javax.el.ELResolver;
+import javax.el.ExpressionFactory;
 import javax.el.ListELResolver;
 import javax.el.MapELResolver;
 import javax.el.ResourceBundleELResolver;
@@ -37,7 +39,14 @@ public class SimpleELContext extends ELContext {
 	private final VariableMapper variableMapper;
 	private final ELResolver resolver;
 
-	public SimpleELContext() {
+	public SimpleELContext(ExpressionFactory expressionFactory) {
+		// In javax.el.ELContext, the ExpressionFactory is extracted from the context map. If it is not found, it
+		// defaults to ELUtil.getExpressionFactory() which, if we provided the ExpressionFactory to the
+		// ResourceBundleMessageInterpolator, might not be the same, thus potentially instantiating another
+		// ExpressionFactory outside of the boundaries of the class loader safe loading mechanism of
+		// ResourceBundleMessageInterpolator.
+		putContext( ExpressionFactory.class, expressionFactory );
+
 		functions = new MapBasedFunctionMapper();
 		variableMapper = new MapBasedVariableMapper();
 		resolver = DEFAULT_RESOLVER;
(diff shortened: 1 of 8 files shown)


osgi/integrationtest/src/test/java/org/hibernate/validator/osgi/integrationtest/OsgiIntegrationTest.java
```diff
@@ -212,7 +215,7 @@ private ExpressionFactory buildExpressionFactory() {
 		Thread.currentThread().setContextClassLoader( getClass().getClassLoader() );
 
 		try {
-			return ExpressionFactory.newInstance();
+			return ELManager.getExpressionFactory();
```
gunnarmorling: Couldn't we even simplify that entire test by simply relying on the default interpolator?

gsmet: I supposed the purpose of this test was to check that passing an external `ExpressionFactory` would work.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/SetClassLoader.java
```diff
@@ -0,0 +1,35 @@
...
+import java.security.PrivilegedAction;
+
+import org.hibernate.validator.internal.util.Contracts;
+
+/**
+ * Privileged action used to set the Thread context class loader.
+ *
+ * @author Guillaume Smet
+ */
+public final class SetClassLoader implements PrivilegedAction<Void> {
```
gunnarmorling: I'd name it `SetContextClassLoader`.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/SetClassLoader.java
```diff
@@ -0,0 +1,35 @@
...
+	private final ClassLoader classLoader;
+
+	public static SetClassLoader ofContext(ClassLoader classLoader) {
```
gunnarmorling: We named these methods `action` in the case where there are not several ones with different semantics.

osgi/integrationtest/src/test/java/org/hibernate/validator/osgi/integrationtest/OsgiIntegrationTest.java
```diff
@@ -116,7 +119,7 @@ public void canObtainValidatorFactoryAndPerformValidation() {
 				.validate( new Customer() );
 
 		assertEquals( 1, constraintViolations.size() );
-		assertEquals( "must be greater than or equal to 1", constraintViolations.iterator().next().getMessage() );
+		assertEquals( "must be greater than or equal to 2", constraintViolations.iterator().next().getMessage() );
```
gunnarmorling: Where is that coming from?

gunnarmorling:
@gsmet looks good, added some minor final comments.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/SetClassLoader.java
gsmet: Yup, I just made it a symmetry of GetClassLoader. I'll change it.

gunnarmorling:
I think that it more was a requirement to use EL at all, it wouldn't have
worked with the default interpolator. I remember that we added that test
now :) (see HV-1039).

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/SimpleELContext.java
```diff
@@ -37,7 +39,9 @@
 	private final VariableMapper variableMapper;
 	private final ELResolver resolver;
 
-	public SimpleELContext() {
+	public SimpleELContext(ExpressionFactory expressionFactory) {
+		putContext( ExpressionFactory.class, expressionFactory );
```
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/SetClassLoader.java
gsmet: Done.

gsmet:
I'll work on trying to put together a test for that.

gsmet:
@gunnarmorling I tried to write a test using our `IgnoringPackageClassLoader` but I didn't succeed. I think the only way to get it working would be to isolate the test in its own JVM.

So for now, I think I would let it like that. It exercises the fact that we can create a specific ExpressionFactory but we can't really test that it is used.

gunnarmorling:
@gsmet can you update it with my change from #583? I think then we are good to go. Thanks a ton for fighting through this craziness!

PR #582, opened 2016-11-21
HV-1155 Backport for 5.3

gsmet:

Note: this is not perfect as there is still an `ExpressionFactory` initialized outside of the class loader safe portion in the `BeanELResolver` class of the 2.4 version of javax.el-api...

WildFly 10.1+ (and maybe before) is using the 3.0.1-b8 version which is not affected by this issue but it's there.

We should definitely move to 3.0.1-b8 for HV 5.4.


gunnarmorling:
Superseded by #583. Thanks, @gsmet, we are almost there :)

PR #583, opened 2016-11-22
[5.3] HV-1155 - some refinements

gunnarmorling:
Hi @gsmet, it's largely based on your work, only changed some small bits. Most notably the it avoids the TCCL change for the bundle locators, in order to avoid any behavioral change here.


gsmet:
Merged! Thanks!

PR #584, opened 2016-11-22
HV-840 [follow-up] Annotation Processor doesn't detect errors in parameter constraints in inheritance hierarchies

gsmet:
* https://hibernate.atlassian.net/browse/HV-840
 * Follow-up of: https://github.com/hibernate/hibernate-validator/pull/524

Hi @marko-bekhta ,

So I did a couple of fixes and thought we might as well start with a clean review slate instead of having dozens of comments piling up.

This is all based on what you did. As you suspected it, the inheritance tree was what I had in mind. I fixed a couple of issues though, the most important one being that you ended the exploration when a superclass did not have the method, while you should also have explored the superclasses of this superclass.

I think we now have something working pretty well.

Sorry again about the delay, we had a few bad surprises with our upgrade in WildFly our it drained a lot of time and efforts.

Comments welcome!

-- 
Guillaume


annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/AbstractMethodOverrideCheck.java
```diff
@@ -88,105 +79,88 @@ public AbstractMethodOverrideCheck(Elements elementUtils, Types typeUtils, Const
...
 			return null;
 		}
-
-		for ( Element element : elementUtils.getAllMembers( otherTypeElement ) ) {
+		for ( Element element : elementUtils.getAllMembers( typeElement ) ) {
 			if ( !element.getKind().equals( ElementKind.METHOD ) ) {
 				continue;
 			}
-			if ( elementUtils.overrides( currentMethod, (ExecutableElement) element, currentTypeElement ) ) {
+			if ( elementUtils.overrides( currentMethod, (ExecutableElement) element, getEnclosingTypeElement( currentMethod ) ) ) {
```
marko-bekhta: maybe get enclosing element before the loop ? or this operation is not very expensive ? Something like:
```
TypeElement methodEnclosingType = getEnclosingTypeElement( currentMethod );
for (...){
....
if ( elementUtils.overrides( currentMethod, (ExecutableElement) element, methodEnclosingType ) ) {
```

marko-bekhta:
hi @gsmet !

All looks great ! You definitely are much better at naming methods and variables :) and a great improvement to MethodInheritanceTree as well.

Also I can see that it seems like I still need to do some changes to my IDE formatting as there still are some differences :)

As for the delays - it's all fine and I completely understand it. I myself got stuck with my work and wasn't able to spent much time on HV related items :)

Have a nice day,
Marko

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/AbstractClassCheck.java
```diff
@@ -20,15 +22,15 @@
...
+ * All check methods not overridden will return an empty set.
  * </p>
  *
  * @author Marko Bekhta
  */
 public abstract class AbstractClassCheck implements ClassCheck {
 
 	@Override
-	public Collection<ConstraintCheckIssue> checkMethod(ExecutableElement element) {
+	public Set<ConstraintCheckIssue> checkMethod(ExecutableElement element) {
```
gsmet: Yes, it's just for the sake of consistency. I must admit I wouldn't have changed it if I didn't have other changes to do.

To be honest, if I had to do it from the ground up, I think I would have used ArrayList everywhere to have the result as predictable as possible. I don't think Sets are of much value here.

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/AbstractMethodOverrideCheck.java
```diff
@@ -88,105 +79,88 @@ public AbstractMethodOverrideCheck(Elements elementUtils, Types typeUtils, Const
...
+			MethodInheritanceTree.Builder methodInheritanceTreeBuilder) {
+		for ( TypeMirror implementedInterface : currentTypeElement.getInterfaces() ) {
+			TypeElement interfaceTypeElement = (TypeElement) typeUtils.asElement( implementedInterface );
+			ExecutableElement overriddenMethod = getOverriddenMethod( overridingMethod, interfaceTypeElement );
+			if ( overriddenMethod != null ) {
+				methodInheritanceTreeBuilder.addOverriddenMethod( overridingMethod, overriddenMethod );
+				collectOverriddenMethodsInInterfaces( overriddenMethod, interfaceTypeElement, methodInheritanceTreeBuilder );
+			}
+			else {
+				collectOverriddenMethodsInInterfaces( overridingMethod, interfaceTypeElement, methodInheritanceTreeBuilder );
```
gsmet: There's a slight difference in the calls: one is using `overriddenMethod` and the other `overridingMethod`. I thought it was clearer this way than to declare another variable. Your question makes me think I might be wrong. WDYT?

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/AbstractMethodOverrideCheck.java
gsmet: Right! I'll fix it, thanks for noticing.

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/AbstractMethodOverrideCheck.java
gsmet: OK, I fixed it, it seems to be clearer now.

gsmet:
@marko-bekhta thanks for the review. I pushed a commit fixing your remarks. Is it better now? (I'll probably squash it before merging)

annotation-processor/src/main/java/org/hibernate/validator/ap/classchecks/AbstractMethodOverrideCheck.java
marko-bekhta: ahhaa ! when I was looking at it these two lines looked the same to me :)

marko-bekhta:
@gsmet yes ! looks great !

annotation-processor/src/test/java/org/hibernate/validator/ap/ConstraintValidationProcessorTest.java
```diff
@@ -96,10 +100,48 @@ public void fieldLevelValidationUsingBuiltInConstraints() {
...
+				new DiagnosticExpectation( Kind.ERROR, 191 ),
+				new DiagnosticExpectation( Kind.ERROR, 219 ),
+				new DiagnosticExpectation( Kind.ERROR, 373 ),
+				new DiagnosticExpectation( Kind.ERROR, 387 ),
+				new DiagnosticExpectation( Kind.ERROR, 409 ),
+				new DiagnosticExpectation( Kind.ERROR, 434 )
+		);
+
+		assertEquals( diagnostics.getDiagnostics().get( 0 ).getMessage( Locale.getDefault() ),
+				"Parameters of method \"doSomething((@javax.validation.constraints.NotNull :: java.lang.String))\" do not respect the inheritance rules. " +
```
gunnarmorling: The error will be shown in the context of the method. Do we need to repeat the entire signature? Shouldn't the compiler itself point it out?

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/overriding/MethodOverridingTests.java
```diff
@@ -0,0 +1,479 @@
...
+	public static class MethodOverridingTestCase2 {
+		public void doSomething(@NotNull @NotBlank @Size(max = 10) String param) {
+
+		}
+	}
+
+	public static class MethodOverridingTestCase2Sub extends MethodOverridingTestCase2 {
+
+		@Override
+		public void doSomething(@NotNull @NotBlank @Size(max = 10) String param) {
```
gunnarmorling: Is there a strong need to check for this? Technically, it doesn't represent any constraint strengthening.

annotation-processor/src/main/java/org/hibernate/validator/ap/AbstractElementVisitor.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Set;
+import javax.lang.model.util.ElementKindVisitor6;
+
+import org.hibernate.validator.ap.checks.ConstraintCheckIssue;
+import org.hibernate.validator.ap.util.CollectionHelper;
+import org.hibernate.validator.ap.util.Configuration;
+import org.hibernate.validator.ap.util.MessagerAdapter;
+
+/**
+ * An abstract {@link javax.lang.model.element.ElementVisitor} that should be used for implementation
```
gunnarmorling: No biggie, but we can use simple names in JavaDocs (and import them), too. Reads a tad nicer.

annotation-processor/src/test/java/org/hibernate/validator/ap/testmodel/overriding/MethodOverridingTests.java
gsmet: Well, it does not respect the inheritance rules we defined. The rule is that you cannot define constraints in the method hierarchy if there are already one method defining constraints in the method hierarchy. That's what we implemented here.

And it's definitely easier to throw an error for this than trying to determine what is or what is not a constraint strengthening.

gunnarmorling: At runtime, we'd allow that one. See `OverridingMethodMustNotAlterParameterConstraints` and how it's calling `!method.isEquallyParameterConstrained()`. That's primarily due to the behavior of Weld proxy classes though which would copy the annotations from proxied classes verbatim to the method definition on the proxy.

I'd prefer behavior of the AP to closely mimic the runtime engine, but if it's too cumbersome to do, we can leave as you've proposed it. It's in the sense of the spec for sure.

gsmet:
@marko-bekhta ... and merged! Thanks for your tenacity on this subject, we finally got this in!

marko-bekhta:
@gsmet Great ! :)

PR #585, opened 2016-11-25
HV-644 Use Jandex for annotation discovery and repository

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-644

This is far-far-far away from ready :) . I've just started to look into this task and wanted to show what I have and to ask if I'm going in the right direction.

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -94,6 +94,11 @@
             <artifactId>money-api</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.jboss</groupId>
+            <artifactId>jandex</artifactId>
+            <optional>true</optional>
+        </dependency>
 
         <!--
         Test dependencies
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/jandex/AbstractConstrainedElementJandexBuilder.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/jandex/AbstractConstrainedElementJandexBuilder.java
@@ -8,6 +8,8 @@ package org.hibernate.validator.internal.metadata.jandex;
 
 import java.lang.annotation.Annotation;
 import java.lang.annotation.ElementType;
+import java.lang.reflect.Executable;
+import java.lang.reflect.Field;
 import java.lang.reflect.Member;
 import java.util.Arrays;
 import java.util.Collection;
@@ -16,25 +18,21 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Stream;
-
 import javax.validation.Valid;
 import javax.validation.groups.ConvertGroup;
 
 import org.hibernate.validator.internal.engine.valuehandling.UnwrapMode;
 import org.hibernate.validator.internal.metadata.core.AnnotationProcessingOptions;
-import org.hibernate.validator.internal.metadata.core.AnnotationProcessingOptionsImpl;
 import org.hibernate.validator.internal.metadata.core.ConstraintHelper;
 import org.hibernate.validator.internal.metadata.core.MetaConstraint;
 import org.hibernate.validator.internal.metadata.descriptor.ConstraintDescriptorImpl;
 import org.hibernate.validator.internal.metadata.jandex.util.JandexHelper;
 import org.hibernate.validator.internal.metadata.location.ConstraintLocation;
 import org.hibernate.validator.internal.util.CollectionHelper;
-import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
 import org.hibernate.validator.internal.util.annotationfactory.AnnotationDescriptor;
 import org.hibernate.validator.internal.util.annotationfactory.AnnotationFactory;
 import org.hibernate.validator.internal.util.logging.Log;
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
-import org.hibernate.validator.parameternameprovider.ParanamerParameterNameProvider;
 import org.hibernate.validator.valuehandling.UnwrapValidatedValue;
 
 import org.jboss.jandex.AnnotationInstance;
@@ -54,16 +52,13 @@ public abstract class AbstractConstrainedElementJandexBuilder {
 
 	protected final JandexHelper jandexHelper;
 
-	protected final ExecutableParameterNameProvider parameterNameProvider;
-
 	protected final AnnotationProcessingOptions annotationProcessingOptions;
 
-	protected AbstractConstrainedElementJandexBuilder(ConstraintHelper constraintHelper, JandexHelper jandexHelper) {
+	protected AbstractConstrainedElementJandexBuilder(ConstraintHelper constraintHelper, JandexHelper jandexHelper,
+			AnnotationProcessingOptions annotationProcessingOptions) {
 		this.jandexHelper = jandexHelper;
 		this.constraintHelper = constraintHelper;
-		//TODO: init correctly this is just for testing
-		this.parameterNameProvider = new ExecutableParameterNameProvider( new ParanamerParameterNameProvider() );
-		this.annotationProcessingOptions = new AnnotationProcessingOptionsImpl();
+		this.annotationProcessingOptions = annotationProcessingOptions;
 	}
 
 	/**
@@ -77,7 +72,7 @@ public abstract class AbstractConstrainedElementJandexBuilder {
 	 */
 	protected UnwrapMode determineUnwrapMode(Type type, Collection<AnnotationInstance> annotationInstances, boolean typeArgumentAnnotated) {
 		boolean indexable = jandexHelper.isIndexable( type );
-		Optional<AnnotationInstance> unwrapValidatedValue = findAnnotation( annotationInstances, UnwrapValidatedValue.class );
+		Optional<AnnotationInstance> unwrapValidatedValue = jandexHelper.findAnnotation( annotationInstances, UnwrapValidatedValue.class );
 
 		if ( !unwrapValidatedValue.isPresent() && typeArgumentAnnotated && !indexable ) {
 			return UnwrapMode.UNWRAP;
@@ -131,17 +126,16 @@ public abstract class AbstractConstrainedElementJandexBuilder {
 	}
 
 	/**
-	 * Finds and converts constraint annotations to {@link MetaConstraint}
+	 * Finds and converts constraint annotations to {@link ConstraintDescriptorImpl}
 	 *
 	 * @param annotationInstances a collection of annotation instances
 	 * @param member a {@link Member} under investigation
 	 *
 	 * @return a stream of {@link MetaConstraint}s based on input parameters
 	 */
-	protected Stream<MetaConstraint<?>> findConstraints(Collection<AnnotationInstance> annotationInstances, Member member) {
+	protected Stream<ConstraintDescriptorImpl<?>> findConstraints(Collection<AnnotationInstance> annotationInstances, Member member) {
 		return findConstrainAnnotations( annotationInstances )
-				.flatMap( annotationInstance -> findConstraintAnnotations( member, annotationInstance ) )
-				.map( descriptor -> createMetaConstraint( member, descriptor ) );
+				.flatMap( annotationInstance -> findConstraintAnnotations( member, annotationInstance ) );
 	}
 
 	/**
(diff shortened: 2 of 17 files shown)


gunnarmorling:
Very nice to see this, @marko-bekhta! Could you maybe describe a bit the current status? I.e. what's working, what is not, any larger road blocks you see?

marko-bekhta:
Hi @gunnarmorling ! Well so here's how I see this:
1. create a `JandexMetaDataProvider` (some code is already present) which will receive jandex index and go through classes/interfaces and look for constraints.
2. to create `BeanConfiguration`s there will be a set of builders (similar to what is present for `XmlMetaDataProvider`). Currently I've started to work on `ConstrainedFieldJandexBuilder`. It seems to me that these builders will have a few methods in common so most likely there will be one abstract and a few implementations for Type, Field, Method, Constructor
3. with having a way to create `BeanConfiguration` from step 2, create a map with all configurations. 

So right now I'm looking into how to create `ConstraintElement`s from Jandex index. There are a few things left that I'm thinking about at this point. They are marked as TODOs in the code. Here they are:

- is there a way to find out if an annotation is a constraint or not based on `AnnotationInstance` (Jandex class)
- how to convert `AnnotationInstance` to `Annotation` which is used to init `ConstraintDescriptorImpl` (for this one I think I'll try to do something similar to what is present in `MetaConstraintBuilder#buildMetaConstraint` with `AnnotationDescriptor` and `AnnotationFactory.create`

Also I've created a very simple `ConstrainedFieldJandexBuilderTest` that creates a jandex index form a bean to help write/debug all this jandex related code. 

So basically there's not much done for now :( . And I just wanted to have a place to discuss these kind of things. So I that I don't write a lot of code that will be wrong :)  Also one more question - is it ok to use Java 8 here or should we stay with 7 and ignore Streams and lambdas ?

gsmet:
@marko-bekhta just a quick answer on the Java 8 side: we won't backport this so feel free to use streams and lambdas and whatever fancy Java 8 features you want!

gunnarmorling:
> is there a way to find out if an annotation is a constraint or not based on AnnotationInstance (Jandex class)

You'd have to mimic what's done by `ConstraintHelper#isConstraintAnnotation()`. In the first iteration it should be enough to check for presence of the `Constraint` (meta) annotation. I've [just asked](http://lists.jboss.org/pipermail/wildfly-dev/2016-December/005561.html) on wildfly-dev how to obtain meta-annotations, as it isn't clear from the API to me.

> how to convert AnnotationInstance to Annotation

We'd have to create an annotation proxy ourselve as we do it for XML-based configuration, too. See `org.hibernate.validator.internal.xml.MetaConstraintBuilder.buildMetaConstraint()`. It's not clear to me yet how to get the actual annotation type, but the answer may be provided with my question above on wildfly-dev.

gunnarmorling:
@marko-bekhta, I think you are on the right track overall. Let's see what comes out of my question on wildfly-dev.

marko-bekhta:
@gunnarmorling  thanks for the info! I've just pushed a new commit and there's Annotation creation in `ConstrainedFieldJandexBuilder#instanceToAnnotation` method.

marko-bekhta:
hi @gunnarmorling  @gsmet,

I'm looking at the next line:
```
if ( annotationProcessingOptions.areParameterConstraintsIgnoredFor( executable, i ) ) {
   metadata.add(....);
}
```

and I'm wondering if it should be with `!` or not ? it's in `AnnotationMetaDataProvider:455`

marko-bekhta:
Hi @gunnarmorling and @gsmet 
here's an update on this task - I've finished with the builders for constraints and with the Jandex provider (in general). There are some things left to be done like get the index from somewhere and a few things from TODOs. I'll list them below:

- how the index should be received by provider ?
- change class search (as @gunnarmorling  mentioned `JandexHelper.getClassForName`). I've made a `JandexHelper` and it's methods nonstatic so the inside logic of `getClassForName`can be changed without changes to the method itself. I plan to look into this change next. 
- there seems to be a problem with Jandex index and parameter annotations. Please see test `ConstrainedFieldJandexBuilderTest.validAnnotationIsMissing`. It looks that Jandex is ignoring the `@Valid` annotation on a parameter. As a result - right now jandex metadata provider is throwing an  for cases when there's a type argument constraint for method parameter.
- there are a couple of methods like `isIndexable`, `isMap`, `isIterable` which are reusing `ReflectionHelper` methods, and I cannot see a way to mimic the logic using Jandex classes
- I haven't found any possibility to determine if annotation is a constraint or not based on Jandex classes, and I can see that for now there's no answer on wildfly-dev about it.
- in `GroupSequenceJandexHelper` there are two methods copied from `AnnotationMetaDataProvider` to obtain GroupSequenceProvider. Based on what is done in them - not sure if it'll be possible to do the same with jandex classes.
- don't see a way to filter out synthetic methods. 
- methods and constructors are returned in one collection and they don't have any differences (at least I didn't find any obvious one :) ). Constructor names are in such form `<init>` so I've added a logic to check if method name match this form then it's a constructor and we should look for a constructor rather that for a method in such case, when looking for a class member (reflection class)
- maybe `parameterNameProvider` can be improved to consume not only Method/Constructor ? But as` Executable executable` is already available at the point when `parameterNameProvider` is used - not sure if such change would be of any use.

That seems to summarize it all :)

gunnarmorling:
Hey @marko-bekhta, we still need to follow up in more depth, but one quick answer to the last one:

> maybe parameterNameProvider can be improved to consume not only Method/Constructor ? 

We cannot really change that one as it's part of the BV API, but check out `ExecutableParameterNameProvider`. It does what you want.

marko-bekhta:
HI @gunnarmorling ! Thanks for the answer. I've used the `ExecutableParameterNameProvider` in the code already - for example you can see at this [line](https://github.com/hibernate/hibernate-validator/pull/585/files#diff-023c77d1e50bc32478c8febc35517603R175). It was just a thought to maybe save some execution time, but as at that point `Executable` is already available it probably will not do any change....

gsmet:
@marko-bekhta FYI, I'm taking a look at it. I'll probably push some fixes.

marko-bekhta:
@gsmet Great! thanks for the update.

gsmet:
@marko-bekhta 

> there seems to be a problem with Jandex index and parameter annotations. Please see test ConstrainedFieldJandexBuilderTest.validAnnotationIsMissing. It looks that Jandex is ignoring the @Valid annotation on a parameter. As a result - right now jandex metadata provider is throwing an for cases when there's a type argument constraint for method parameter.

AFAICS, the test is working. Did I miss something?

marko-bekhta:
@gsmet 
for me it gives next:
```
java.lang.AssertionError: 
Expected size:<3> but was:<2> in:
<["javax.validation.constraints.NotNull", "javax.validation.constraints.Size"]>

at org.hibernate.validator.test.internal.metadata.jandex.ConstrainedFieldJandexBuilderTest.validAnnotationIsMissing(ConstrainedFieldJandexBuilderTest.java:94)
```
So it seems that `@Valid` is ignored ...

gsmet:
@marko-bekhta it's weird. I don't have this error at all. It's working correctly here.

gsmet:
> how the index should be received by provider?

This is something that is still undecided AFAIK. We need to see how we could get the index from WildFly.

I'm also wondering how we could test this Jandex stuff with all our tests (simply replacing the `AnnotationMetaDataProvider` to see how it behaves). It's really not easy considering you have to add the indexed classes one at a time.

If we can find a way to access the Jandex index of WildFly, we could run the TCK in container mode (-Dincontainer) to exercise the Jandex implementation.

> change class search (as @gunnarmorling mentioned JandexHelper.getClassForName). I've made a JandexHelper and it's methods nonstatic so the inside logic of getClassForNamecan be changed without changes to the method itself. I plan to look into this change next.

I haven't taken a look at this. I think it can be left alone for now.

> there seems to be a problem with Jandex index and parameter annotations. Please see test ConstrainedFieldJandexBuilderTest.validAnnotationIsMissing. It looks that Jandex is ignoring the @Valid annotation on a parameter. As a result - right now jandex metadata provider is throwing an for cases when there's a type argument constraint for method parameter.

As mentioned, I don't have this issue at all. The test is passing here (I'm running the test in Eclipse if it matters).

> there are a couple of methods like isIndexable, isMap, isIterable which are reusing ReflectionHelper methods, and I cannot see a way to mimic the logic using Jandex classes

Yeah, they are less used now due to changes Gunnar made recently but we still need them.

I think I'll rebase your patch on top of our recent work and see how it goes from there.

> I haven't found any possibility to determine if annotation is a constraint or not based on Jandex classes, and I can see that for now there's no answer on wildfly-dev about it.

I think I did make it work.

> in GroupSequenceJandexHelper there are two methods copied from AnnotationMetaDataProvider to obtain GroupSequenceProvider. Based on what is done in them - not sure if it'll be possible to do the same with jandex classes.

Let's keep them for now, we'll see if we need a parent class or something once we have a working experiment.

> don't see a way to filter out synthetic methods.

I did it the hard way. We'll need to open a Jandex issue so that they expose a method for it.

> methods and constructors are returned in one collection and they don't have any differences (at least I didn't find any obvious one :) ). Constructor names are in such form <init> so I've added a logic to check if method name match this form then it's a constructor and we should look for a constructor rather that for a method in such case, when looking for a class member (reflection class)

Yeah, I think it's as good as we can do for now. I moved it to its own method to be more clear. We will need to open a Jandex issue.

> maybe parameterNameProvider can be improved to consume not only Method/Constructor ? But asExecutable executable is already available at the point when parameterNameProvider is used - not sure if such change would be of any use.

Haven't taken a look to this yet.

Next comment will be about the changes I made.

marko-bekhta:
> it's weird. I don't have this error at all. It's working correctly here.

@gsmet It definitely is... I've just tried to run the test with maven and it fails:

```
Results :

Failed tests: 
  ConstrainedFieldJandexBuilderTest.testGetConstrainedFields:68 » Validation HV0...
  ConstrainedFieldJandexBuilderTest.validAnnotationIsMissing:94 
Expected size:<3> but was:<2> in:
<["javax.validation.constraints.NotNull", "javax.validation.constraints.Size"]>

Tests run: 2, Failures: 2, Errors: 0, Skipped: 0

```

as well as if I run the test through the IDEA ... and it's the behavior that I have on two different computers ...

gsmet:
My current tree is here: https://github.com/gsmet/hibernate-validator/tree/pr/585 .

I fixed a couple of bugs, the provider is now working (I haven't inspected the result yet, just that there's some BeanConfiguration returned).

Note that one important change is that it does not scan all the classes of the index but just the constrained beans. I'm still unsure my implementation is sufficient, especially regarding class hierarchy.

Apart from that, I mainly cleaned up/reorganized the code. Here are a few general comments that might help you:
* always make your methods/fields private by default and make them protected only if needed
* put the methods only required by a subclass in the subclass itself, not in the parent
* no need for commenting everything: we document the API and what might be unclear but, for a private method, if you understand what it does just by reading its name and parameters, there's no need for javadoc

We are on Holiday shutdown starting this evening but I think I'll try to revisit a few things next week. I'd like to rebase this on the current master as I'm not sure it's going to be an easy task. And I still need to understand a few things.

Finally I would really like to find a way to build a usable index.

marko-bekhta:
Hi @gsmet, great improvements to the implementation! I'm not sure either about filtering constrained beans, I suppose - we'll be more sure when we have a real index at our hands, right ? And good catch with those stream concatenations :). 

The only thing that I'd probably change is a return type of `JandexMetaDataProvider#findConstrainedBeans` from `Set` to `Stream`. But it's just my personal preference - I try to return `Streams` instead of collections wherever possible to omit returning modifiable collections (to prevent change of state of some bean enclosing that collection) and to prevent calls to `stream()` right after receiving a collection. I don't know if that would be more efficient than what we have right now but I'd probably would done something like this:

```
private static Stream<ClassInfo> findConstrainedBeans(IndexView indexView, List<DotName> constraintAnnotations) {
		// TODO HV-644: not sure this is totally accurate. I'm especially wondering how it will behave with subclasses and such.
		return constraintAnnotations.stream()
				.flatMap( constraintAnnotation -> indexView.getAnnotations( constraintAnnotation ).stream() )
				.map( ai -> ai.target() )
				.map( JandexMetaDataProvider::annotationInstanceToClassInfo )
				.distinct();
	}
```

Also thanks for those points on visibility modifiers and documentation! 

I've tried to run the tests on one more machine - and it's still the same for me - the tests are failing with the same error. (I tried to run the code from your branch as well). 

Thanks again and Happy New Year to you! :)

gsmet:
Closing this one. I'll create another one superseding this one to continue the discussion.

gsmet:
@marko-bekhta personally I'm all for using streams when you expect to work on the result. But if not, I must admit my preference is to use good old collection types.

It's only an implementation detail anyway. If we get to just having that to discuss, I think we would have come a great way ;).

PR #586, opened 2016-11-25
Initialization of the 5.4 branch

gsmet:
This is the proposal for the initialization of the 5.4 branch.


gsmet:
@gunnarmorling force pushed an update fixing the javax.el commit. It's the only one that has this issue.

gunnarmorling:
Pushed all remaining commits. Thanks a lot, @gsmet!

PR #588, opened 2016-11-25
HV-1161 Use the correct checkstyle conf for OSGi integration tests

gsmet:
https://hibernate.atlassian.net/browse/HV-1161

Should be backported to 5.4.


gunnarmorling:

PR #589, opened 2016-11-25
HV-1163 Update performance tests to 5.3.3.Final

gsmet:
* https://hibernate.atlassian.net/browse/HV-1163

Should be backported to 5.4.


gunnarmorling:
Merged and backported.

PR #590, opened 2016-11-25
HV-1168 Parameterize patch.xml file - HV-1167 Use the patch file in tck-runner

gsmet:
* https://hibernate.atlassian.net/browse/HV-1168
 * https://hibernate.atlassian.net/browse/HV-1167

The goal of this patch is to parameterize the patch.xml file to avoid us the maintenance task of updating the HV and WF versions manually.

It also uses the patch in the tck-runner module.


gsmet:
Pushed a new commit to remove the module files as they are not useful anymore.

modules/src/main/modules/patch.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <patch-config xmlns="urn:jboss:patch-config:1.0">
-    <name>wildfly-10-hibernate-validator-6.0</name>
-    <description>This patch upgrades Hibernate Validator 6.0 within WildFly 10 installation</description>
+    <name>wildfly-10-hibernate-validator-${project.version}</name>
```
gunnarmorling: Cool.

Also make the WF version more flexible?

gsmet:
Well, they are not useful anymore because we use the ones from the WildFly distribution and we update them.

So we don't have to maintain our own.

gunnarmorling:
Where is this update happening?

It does for the TCK runner, but eventually we should be using the patch
file there, too.

gunnarmorling:
I think just the major should be fine (10, 11), and that one would
implicitly target the latest minor of that major.

gsmet:
Are you sure it does?

AFAICS, the `setupModules.groovy` script from the TCK runner also updates the files in place.

gunnarmorling:
Just had another look :) You are right that there is a script now that updates the existing descriptors in place (in _modules/target/wildfly-patched_) when creating the patch file. So the copies are not needed any more.

For the TCK runner we have a script (essentially the same one as above) which updates the descriptors of the WF unzipped for testing. Instead, it should use the patch created as described above and apply it in the same way the integration tests are doing it. I should have done it right when moving to the patching mechanism, but then I had forgotten about it.

gsmet:
Created https://hibernate.atlassian.net/browse/HV-1167 to track the use of the module file.

gsmet:
@gunnarmorling pushed an update addressing both HV-1167 and HV-1168

gunnarmorling:
Merged; thanks, @gsmet. Added one more commit for removing the now obsolete Groovy patching script from the TCK runner module.

PR #592, opened 2016-11-29
HV-1166 (PoC for cascading SPI)

gunnarmorling:
Hi @emmanuelbernard, that's a PoC for the value extractor SPI. You may find it helpful to play with to get a better feeling for the issue. It allows to plug in different `ValueExtractor` s which then are used to obtain the elements of the cascaded value.

What it does:

* Supports custom extractors, with examples for map key extraction and Guava multimap extraction
* Extractors are applied for cascaded validation only; no type-level constraints atm.

What needs to be done:

* Refine extractor SPI (currently it's quite static, more or less as in our call; it'd need a way to describe the `Node` for the value or something like that
* Refine type binding mechanism; it's very rough atm. and works for the happy path, but there are many loose ends
* The extractor needs to be determined per _value type_, i.e. some caching is in order as it's otherwise a big perf penalty

Maybe you can take a look at the tests, add some test scenarios? Or use it otherwise to tinker around. Anyways, feedback welcome. @gsmet, from you, too, of course :)


engine/src/main/java/org/hibernate/validator/internal/engine/cascading/LegacyMapValueExtractor.java
```diff
@@ -0,0 +1,28 @@
...
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.engine.cascading;
+
+import java.util.Map;
+
+import org.hibernate.validator.spi.cascading.ValueExtractor;
+
+class LegacyMapValueExtractor implements ValueExtractor<Map<?, ?>> {
```
gunnarmorling: Currently it's not needed, but I'm wondering whether we should require `@ExtractedValue` on the actual type itself:

    class LegacyMapValueExtractor implements ValueExtractor<@ExtractedValue Map<?, ?>> { ... }

It seems nice for consistency purposes and makes this extractor's purpose more apparent.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractors.java
```diff
@@ -0,0 +1,109 @@
...
+
+		tmpValueExtractors.add( new ValueExtractorDescriptor( LegacyIterableValueExtractor.INSTANCE ) );
+		tmpValueExtractors.add( new ValueExtractorDescriptor( IterableValueExtractor.INSTANCE ) );
+
+		tmpValueExtractors.add( new ValueExtractorDescriptor( ObjectValueExtractor.INSTANCE ) );
+
+		valueExtractors = Collections.unmodifiableList( tmpValueExtractors );
+	}
+
+	public List<ValueExtractor<?>> getCascadedValueExtractors(Class<?> valueType, Class<?> declarationType, List<TypeVariable<?>> cascadingTypeParameters) {
```
gunnarmorling: That thing is nightmare-ish. It works for the general case now, but eventually, the logic would have to be refined. Specifically, it works due to the order of registration of built-in extractors (more specific ones before more general ones). It didn't seem needed to implement it 100% for now.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/CustomValueExtractorTest.java
```diff
@@ -0,0 +1,128 @@
...
+import org.hibernate.validator.HibernateValidator;
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ListMultimap;
+
+/**
+ * @author Gunnar Morling
+ */
+public class CustomValueExtractorTest {
```
gunnarmorling: That test is interesting as basis for tinkering around with it.

gunnarmorling:
Note that this requires https://github.com/beanvalidation/beanvalidation-api/pull/72 (allowing `@Valid` to be used on type parameters).

engine/src/main/java/org/hibernate/validator/spi/cascading/ValueExtractor.java
```diff
@@ -0,0 +1,23 @@
...
+
+public interface ValueExtractor<T> {
+
+	void extractValues(T originalValue, ValueReceiver receiver);
+
+	interface ValueReceiver {
+
+		void objectValue(Object object);
+
+		void listValue(int i, Object object);
```
gsmet: Wondering if `listItem` and `iterableItem` (or `listElement` and `iterableElement`) would be more descriptive? It's just a detail but I was surprised to see an index for `listValue` and then I understood it was only an item, not the list.

engine/src/test/java/org/hibernate/validator/test/internal/util/ExecutableFormatterTest.java
```diff
@@ -22,15 +22,15 @@
 
 	@Test
 	public void executableAsStringShouldReturnMethodNameWithBracesForParameterlessMethod() throws Exception {
-		assertEquals( new ExecutableFormatter( Foo.class.getMethod( "zap" ) ).toString(), "zap()" );
+		assertEquals( new ExecutableFormatter( Foo.class.getMethod( "zap" ) ).toString(), "Foo#zap()" );
 		assertEquals( new ExecutableFormatter( Bar.class.getConstructor() ).toString(), "Bar()" );
 	}
 
 	@Test
 	public void executableAsStringShouldReturnMethodNameWithSimpleParamerTypeNames() throws Exception {
```
gsmet: Nothing related to your code but we have a typo here: s/Paramer/Parameter/

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -283,6 +286,15 @@ public HibernateValidatorConfiguration addValidatedValueHandler(ValidatedValueUn
...
 	}
 
+	@Override
+	public HibernateValidatorConfiguration addCascadedValueExtractor(ValueExtractor<?> extractor) {
+		Contracts.assertNotNull( extractor, MESSAGES.parameterMustNotBeNull( "extractor" ) );
+		cascadedValueExtractors.add( extractor );
+
+		return this;
+	}
+
```
gsmet: Superfluous new line.

engine/src/main/java/org/hibernate/validator/internal/util/TypeVariableBindings.java
```diff
@@ -0,0 +1,87 @@
...
+		for ( TypeVariable<?> typeVariable : subTypeParameters ) {
+			currentBindings.put( typeVariable, typeVariable );
+		}
+		allBindings.put( type, currentBindings );
+
+		collectTypeBindings( type, allBindings, currentBindings );
+
+		allBindings.put( Object.class, Collections.emptyMap() );
+
+		return allBindings;
```
gsmet: Maybe make it unmodifiable?

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractors.java
```diff
@@ -64,13 +65,15 @@ public ValueExtractors(Iterable<ValueExtractor<?>> externalExtractors) {
...
+					if ( !cascadedValueExtractors.containsKey( cascadingValueTypeParameter ) ) {
+						cascadedValueExtractors.put( cascadingValueTypeParameter, extractorDescriptor.getValueExtractor() );
+					}
 				}
 			}
 		}
 
-		// TODO implement selection of most specific extractor per requested type parameter
-		return cascadedValueExtractors.size() > 1 ? Collections.singletonList( cascadedValueExtractors.iterator().next() ) : cascadedValueExtractors;
+		return new ArrayList<>( cascadedValueExtractors.values() );
```
gsmet: Just to be sure, the order of the extractors is not important?

It should be made unmodifiable.

gunnarmorling:
> I'm wondering if I should extract the unrelated fixes into proper JIRAs and get them committed? It would keep the patch smaller. WDYT?

Yes, that would be nice. The first few commits (in actual order within my branch, the shown order in the PR differs a bit) are un-related.

engine/src/main/java/org/hibernate/validator/spi/cascading/ValueExtractor.java
gunnarmorling: Yes, "listItem" would be better.

That said, the entire contract needs to change anyways, it needs to be more generic. That's one of the next things to look into. But essentially we don't know the "nature" of the association navigated into, so we cannot work with a fixed set of pre-defined methods. One idea is to the creation of `Node` instances to that contract, but we'll have to see.

engine/src/main/java/org/hibernate/validator/internal/util/TypeVariableBindings.java
gunnarmorling: +1

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractors.java
gunnarmorling: It currently is important, but eventually it shouldn't be. In the end I think we need to keep the most specific extractor per type parameter, e.g. if we have an extractor for `Iterable<@ExtractedValue E` and one for `List<@ExtractedValue E` and the validated value is an `ArrayList`, only keep the latter. But I didn't get there yet. Currently it will simply keep the first extractor found per type variable, hence the order of insertion in the constructor is relevant. It's a hack to make things work a bit.

engine/src/main/java/org/hibernate/validator/spi/cascading/ValueExtractor.java
gsmet: OK, additional comment (but probably useless if it goes away): as you inversed the order of key and value in `mapValue`, I think you should do the same for `listValue` and have the index as second parameter. It would be more consistent.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractors.java
gsmet: That's what I thought and with this new implementation with an `HashMap`, you lose the order.

gsmet:
@gunnarmorling it should be rebased now that I extracted the unrelated commits. I also fixed the typos in the test method names and it might have created a conflict here too.

engine/src/main/java/org/hibernate/validator/spi/cascading/ValueExtractor.java
```diff
@@ -0,0 +1,23 @@
...
+
+	interface ValueReceiver {
+
+		void value(Object object, String nodeName);
+
+		void iterableValue(Object object, String nodeName);
+
+		void indexedValue(Object object, String nodeName, int i);
+
+		void keyedValue(Object object, String nodeName, Object key);
```
emmanuelbernard: So this model does not support the IntegerAndList kind of model, correct? Or is `value(...)` the one we would use?

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/CustomValueExtractorTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+		Validator validator = Validation.byProvider( HibernateValidator.class )
+				.configure()
+				.addCascadedValueExtractor( new MapKeyExtractor() )
+				.buildValidatorFactory()
+				.getValidator();
+
+		Set<ConstraintViolation<CustomerWithCascadingKeys>> violations = validator.validate( bob );
+
+		assertCorrectPropertyPaths( violations, "addressByType[key(too short)].value", "addressByType[key(too small)].value" );
```
emmanuelbernard: I think I would have preferred something like

```
"addressByType[key('too short')].value"
```

At least for strings and any type that we don't know and will be `toString`ed.
We make exceptions for numbers, booleans, dates

engine/src/test/java/org/hibernate/validator/test/internal/util/TypeVariableBindingsTest.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+public class TypeVariableBindingsTest {
+
+	@Test
+	public void canObtainBindingForList() {
+		Map<Class<?>, Map<TypeVariable<?>, TypeVariable<?>>> bindings = TypeVariableBindings.getTypeVariableBindings( List.class );
+
+		Map<TypeVariable<?>, TypeVariable<?>> listBindings = bindings.get( List.class );
+		assertThat( listBindings ).containsOnly(
+			entry( List.class.getTypeParameters()[0], List.class.getTypeParameters()[0] )
```
emmanuelbernard: Twice the same stuff?

emmanuelbernard: hum I think I got it.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/AnnotatedObject.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.hibernate.validator.internal.engine.cascading;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedType;
+import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
+
+/**
+ * A pseudo type variable used for to express that the annotated element itself (e.g. a list) has been marked for
```
gsmet: s/used for to/used to/

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+	private static Type getExtractedType(Class<?> extractorImplementationType) {
+		// TODO deal with indirect implementations (MyExtractor -> BaseExtractor -> ValueExtractor)
+		AnnotatedType genericInterface = extractorImplementationType.getAnnotatedInterfaces()[0];
+		AnnotatedType extractedType = ( (AnnotatedParameterizedType) genericInterface ).getAnnotatedActualTypeArguments()[0];
+		return extractedType.getType();
+	}
+
+	Type getExtractedType() {
+		return extractedType;
+	}
```
gsmet: Missing new line.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -591,19 +593,19 @@ private void validateCascadedConstraints(ValidationContext<?> validationContext,
 					elementType
 			) ) {
 
-				Object value = getCascadableValue( validationContext, valueContext.getCurrentBean(), cascadable );
+					Object value = getCascadableValue( validationContext, valueContext.getCurrentBean(), cascadable );
```
gsmet: Any reason why you indented with 2 tabs?

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractors.java
```diff
@@ -50,23 +50,33 @@ public ValueExtractors(Iterable<ValueExtractor<?>> externalExtractors) {
...
 		valueExtractors = Collections.unmodifiableList( tmpValueExtractors );
 	}
 
 	public ValueExtractor<?> getCascadedValueExtractor(Class<?> valueType, TypeVariable<?> cascadingTypeParameter) {
-		Map<Class<?>, Map<TypeVariable<?>, TypeVariable<?>>> allBindings = TypeVariableBindings.getTypeVariableBindings( valueType );
+		Map<Class<?>, Map<TypeVariable<?>, TypeVariable<?>>> allBindings = null;
+
+		if ( cascadingTypeParameter != AnnotatedObject.INSTANCE && cascadingTypeParameter != ArrayElement.INSTANCE ) {
+			allBindings = TypeVariableBindings.getTypeVariableBindings( (Class<?>) cascadingTypeParameter.getGenericDeclaration() );
+		}
```
gsmet: AFAICS it's not a case that can happen but I think I would prefer it to be initialized to an empty map in the other case.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -683,4 +683,7 @@ ConstraintDeclarationException getInconsistentValueUnwrappingConfigurationBetwee
 
 	@Message(id = 196, value = "Unable to convert the Type %s to a Class.")
 	ValidationException getUnableToConvertTypeToClassException(Type type);
+
+	@Message(id = 197, value = "No value extractor found type parameter '%2$s' of type %1$s.")
```
gsmet: s/found/found for/ ?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -616,14 +692,11 @@ private void validateCascadedConstraints(ValidationContext<?> validationContext,
 
 	@SuppressWarnings({ "rawtypes", "unchecked" })
 	private void validateCascadedValues(Object value, ValidationContext<?> context, ValueContext<?, ?> valueContext, Cascadable cascadable, ValidationOrder validationOrder) {
-		Set<TypeVariable<?>> cascadingOrConstrainedTypeParameters = new HashSet<>( cascadable.getCascadingTypeParameters() );
-		cascadingOrConstrainedTypeParameters.addAll( cascadable.getTypeArgumentsConstraints().keySet() );
-
-		for ( TypeVariable<?> cascadingOrConstrainedTypeParameter : cascadingOrConstrainedTypeParameters ) {
+		for ( TypeVariable<?> cascadingTypeParameter : cascadable.getCascadingTypeParameters() ) {
 			List<TypeVariable<?>> cascadingOrConstrainedTypeParametersOfValueType = getCorrespondingTypeParametersInSubType(
```
gsmet: Maybe remove the `OrConstrained` here too?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -616,14 +692,11 @@ private void validateCascadedConstraints(ValidationContext<?> validationContext,
...
 					value.getClass(),
 					TypeHelper.getErasedReferenceType( cascadable.getCascadableType() ),
-					cascadingOrConstrainedTypeParameter
+					cascadingTypeParameter
 			);
 
 			for ( TypeVariable<?> cascadingOrConstrainedTypeParameterOfValueType : cascadingOrConstrainedTypeParametersOfValueType ) {
```

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gunnarmorling: No, just an oversight.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -443,15 +447,18 @@ private String interpolate(String messageTemplate,
...
 			return false;
 		}
 
-		for ( PathImpl p : pathSet ) {
-			if ( path.isRootPath() || p.isRootPath() || isSubPathOf( path, p ) || isSubPathOf( p, path ) ) {
-				return true;
+		for ( PathAndTypeParameter p : pathSet ) {
+			if ( path.isRootPath() || p.path.isRootPath() || isSubPathOf( path, p.path ) || isSubPathOf( p.path, path ) ) {
+				if ( typeParameter == p.typeParameter ||
+						( p.typeParameter == null && ( typeParameter == AnnotatedObject.INSTANCE || typeParameter == ArrayElement.INSTANCE ) ) ) {
```
gsmet: Care to explain the second part of the condition? Thanks.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gunnarmorling: Ok.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
gunnarmorling: It has become obsolete with my latest commit (going to push it asap). It was needed to prevent repeated validation of an object which was part of the graph (e.g. the validated bean) _and_ contained within an array visited during cascaded validation.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractors.java
gunnarmorling: It's only needed within this if block?

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorDescriptor.java
gunnarmorling: Ok.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -177,7 +182,7 @@ public ValidatorFactoryImpl(ConfigurationState configurationState) {
 		boolean tmpAllowParallelMethodsDefineParameterConstraints = false;
 
 		List<ValidatedValueUnwrapper<?>> tmpValidatedValueHandlers = newArrayList( 5 );
-
+		List<ValueExtractor<?>> cascadedValueExtractors = null;
```
gunnarmorling: Ok.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -141,17 +134,6 @@ private ConfigurationImpl() {
 		);
 	}
 
-	private ValidatedValueUnwrapper<?> createJavaFXUnwrapperClass(TypeResolutionHelper typeResolutionHelper) {
-		try {
-			Class<?> jfxUnwrapperClass = run( LoadClass.action( JFX_UNWRAPPER_CLASS, getClass().getClassLoader() ) );
```
gsmet: You can remove the JFX_UNWRAPPER_CLASS constant at the top of the class.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -592,11 +589,76 @@ private boolean validateMetaConstraint(ValidationContext<?> validationContext, V
 		return success;
 	}
 
+	// TODO the value extractor should be retrieved at meta-data instantation and be cached (unlike cascades, constraint
+	// validator resolution is based on the static type of an element
```
gsmet: You know it's dangerous to let a parenthesis open!

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/LegacyOptionalValueExtractor.java
```diff
@@ -0,0 +1,30 @@
...
+package org.hibernate.validator.internal.engine.cascading;
+
+import java.util.Optional;
+
+import org.hibernate.validator.spi.cascading.ValueExtractor;
+
+/**
+ * Extractor for supporting the legacy construct {@code @Valid Optional<Foo>}.
+ * @author Gunnar Morling
+ *
```
gsmet: Sorry for the nitpicking but it's the only thing you let me: looks like you got the new line wrong in the javadoc comment.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/LegacyOptionalValueExtractor.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+// TODO should we keep that, or only support {@code Optional<@Valid Foo>}.
```
gsmet: If we plan to remove it someday, I think the best is to remove it now.

But if it does not cost us, I think we should probably keep it (or throw an error if we consider it's not a valid pattern anymore)

engine/src/main/java/org/hibernate/validator/spi/cascading/ValueExtractor.java
gunnarmorling: Yes, in that case you'd invoke `indexedValue()` for each list element and `value()` for the single int property.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/LegacyIterableValueExtractor.java
```diff
@@ -0,0 +1,26 @@
...
+class LegacyIterableValueExtractor implements ValueExtractor<Iterable<?>> {
+
+	static final LegacyIterableValueExtractor INSTANCE = new LegacyIterableValueExtractor();
+
+	private LegacyIterableValueExtractor() {
+	}
+
+	@Override
+	public void extractValues(Iterable<?> originalValue, ValueReceiver receiver) {
+		receiver.value( originalValue, null );
```
gunnarmorling: The legacy one will be used for `@Valid Set<Order> orders`, and the new one will be used for `Set<@Valid Order> orders`. The difference being that the former would validate any property constraints on a custom `Set` implementation, whereas the latter would only validate the actual iterable elements.

Selection is happening via the type parameter to which extractors are bound: the non-legacy one is bound to `<T>` of Iterable, the legacy one is bound to an internal pseudo type parameter `ANNOTATED_ELEMENT`. Depending on where `@Valid` is spotted, one or the other will be used.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractors.java
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * @author Gunnar Morling
+ */
+public class ValueExtractors {
+
+	private final List<ValueExtractorDescriptor> valueExtractors;
+
+	public ValueExtractors(Iterable<ValueExtractor<?>> externalExtractors) {
+		List<ValueExtractorDescriptor> tmpValueExtractors = new ArrayList<>();
```
gunnarmorling: :)

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/CustomValueExtractorTest.java
gunnarmorling: Ok, I've added this to the follow-up issue HV-1190.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
gunnarmorling: Thanks, done.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gunnarmorling: Ha, but you didn't notice the typeo in "instantation" :)

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractors.java
gsmet: I was talking about the `allBindings` map. I don't feel comfortable having it set to null in some cases while we are using it below without a null check. For the time being, I don't think we can be in the case where it is null and we need it in the for loop but I'd better be safe and not have this pattern.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/MetaDataBuilder.java
```diff
@@ -73,7 +75,11 @@ public void add(ConstrainedElement constrainedElement) {
 		constraints.addAll( adaptConstraints( constrainedElement.getKind(), constrainedElement.getConstraints() ) );
 		constraints.addAll( adaptConstraints( constrainedElement.getKind(), new HashSet<>( constrainedElement.getTypeArgumentConstraints().values() ) ) );
 		isCascading = isCascading || constrainedElement.isCascading();
-		unwrapMode = constrainedElement.unwrapMode();
+
+		if ( unwrapMode == null || constrainedElement.getSource().getPriority() > unwrapModeSource.getPriority() ) {
+			unwrapMode = constrainedElement.unwrapMode();
+			unwrapModeSource = constrainedElement.getSource();
```
gunnarmorling: Yes, it's a bug fix; it was uncovered in the course of implementing the new mechanism. I wouldn't bother backporting unless a user reports it.

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter11/valuehandling/PropertyValueExtractor.java
```diff
@@ -7,28 +7,18 @@
 //tag::include[]
 package org.hibernate.validator.referenceguide.chapter11.valuehandling;
 
-//end::include[]
```
gsmet: you're missing an //end::include[]

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/LegacyOptionalValueExtractor.java
gunnarmorling: Ok.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -686,4 +683,10 @@ ConstraintDeclarationException getInconsistentValueUnwrappingConfigurationBetwee
 
 	@Message(id = 197, value = "No value extractor found type parameter '%2$s' of type %1$s.")
 	ValidationException getNoValueExtractorFoundForTypeException(@FormatWith(ClassObjectFormatter.class) Class<?> type, TypeVariable<?> typeParameter);
+
+	@Message(id = 198, value = "No suitable value extractor found for type %1$s.")
+	ValidationException getNoValueExtractorFoundForUnwrapException(Type type);
+
+	@Message(id = 199, value = "Cannot extract value for type %1$s with more than one type parameters.")
+	ValidationException getUnableToExtractValueForTypeWithMultipleTypeParameterException(Type type);
```
gunnarmorling: Ok.

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter11/valuehandling/PropertyValueExtractor.java
gunnarmorling: Added it back.

gsmet:
Jenkins, retest this please

gsmet:
Merged, thanks!

PR #595, opened 2016-12-05
Various fixes for type arguments support

gsmet:
* https://hibernate.atlassian.net/browse/HV-1165
 * https://hibernate.atlassian.net/browse/HV-1175
 * https://hibernate.atlassian.net/browse/HV-1176

Better to validate commit by commit as they are quite unrelated.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
@@ -18,7 +18,6 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.lang.reflect.WildcardType;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
@@ -398,8 +397,7 @@ public final class ReflectionHelper {
 			iter = ( (Iterable<?>) value ).iterator();
 		}
 		else if ( TypeHelper.isArray( type ) ) {
-			List<?> arrayList = Arrays.asList( value );
-			iter = arrayList.iterator();
+			iter = CollectionHelper.iteratorFromArray( value );
 		}
 		else {
 			return null;
@@ -491,4 +489,5 @@ public final class ReflectionHelper {
 
 		return wrapperType;
 	}
+
 }
--- a/engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
@@ -34,6 +34,7 @@ import org.hibernate.validator.internal.util.logging.LoggerFactory;
  * @author Hardy Ferentschik
  * @author Gunnar Morling
  * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
+ * @author Guillaume Smet
  */
 public final class ReflectionHelper {
 
@@ -259,36 +260,19 @@ public final class ReflectionHelper {
 	}
 
 	/**
-	 * Indicates if the type is considered indexable (ie is an {@code Iterable}, an array or a {@code Map}).
-	 *
-	 * @param type the type to inspect.
-	 *
-	 * @return Returns true if the type is indexable.
+	 * Indicates whether the given type represents a collection of elements or not (i.e. whether it is an
+	 * {@code Iterable}, {@code Map} or array type).
 	 */
-	public static boolean isIndexable(Type type) {
-		boolean isIndexable = false;
-		if ( ReflectionHelper.isList( type ) ) {
-			isIndexable = true;
-		}
-		else if ( ReflectionHelper.isMap( type ) ) {
-			isIndexable = true;
-		}
-		else if ( TypeHelper.isArray( type ) ) {
-			isIndexable = true;
-		}
-		return isIndexable;
+	public static boolean isCollection(Type type) {
+		return isIterable( type ) ||
+				isMap( type ) ||
+				TypeHelper.isArray( type );
 	}
 
-
 	/**
-	 * Determines the type of elements of an <code>Iterable</code>, array or the value of a <code>Map</code>.
-	 *
-	 * @param type the type to inspect
-	 *
-	 * @return Returns the type of elements of an <code>Iterable</code>, array or the value of a <code>Map</code>. <code>
-	 *         null</code> is returned in case the type is not indexable (in the context of JSR 380).
+	 * Determines the type of the elements of an {@code Iterable}, array or the value of a {@code Map}.
 	 */
-	public static Type getIndexedType(Type type) {
+	public static Type getCollectionElementType(Type type) {
 		Type indexedType = null;
 		if ( isIterable( type ) && type instanceof ParameterizedType ) {
 			ParameterizedType paramType = (ParameterizedType) type;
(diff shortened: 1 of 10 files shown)


engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
```diff
@@ -259,36 +260,27 @@ public static Object getValue(Method method, Object object) {
...
-			isIndexable = true;
-		}
-		else if ( ReflectionHelper.isMap( type ) ) {
-			isIndexable = true;
-		}
-		else if ( TypeHelper.isArray( type ) ) {
-			isIndexable = true;
-		}
-		return isIndexable;
+	public static boolean isElementCollection(Type type) {
```
gunnarmorling: I'd prefer `isCollection()` or `isCollectionType()`.

"element collection" reminds too much of the construct with the same name in JPA (`@ElementCollection`). Here I'd refer to any collection-kind type by "collection" including maps, i.e. not `java.util.Collection`.

engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
```diff
@@ -259,36 +260,27 @@ public static Object getValue(Method method, Object object) {
...
 	 * Determines the type of elements of an <code>Iterable</code>, array or the value of a <code>Map</code>.
 	 *
 	 * @param type the type to inspect
 	 *
 	 * @return Returns the type of elements of an <code>Iterable</code>, array or the value of a <code>Map</code>. <code>
-	 *         null</code> is returned in case the type is not indexable (in the context of JSR 380).
+	 *         null</code> is returned in case the type is not a collection of elements.
 	 */
-	public static Type getIndexedType(Type type) {
+	public static Type getElementCollectionElementType(Type type) {
```
gunnarmorling: getCollectionElementType().

engine/src/test/java/org/hibernate/validator/test/internal/engine/typeannotationconstraint/TypeAnnotationConstraintTest.java
```diff
@@ -250,48 +646,38 @@ public void unsupported_use_of_type_constraints_logs_warning() {
...
-	static class A1 {
+	// List
+
+	static class List1 {
 		@Valid
 		List<@NotNull @NotBlank String> names;
 	}
 
-	static class A2 {
+	static class List2 {
```
gunnarmorling: Maybe `TypeWithList2`? When looking at the usage above, I first thought the type itself would be a list type.

engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
```diff
@@ -499,4 +498,85 @@ public static Type boxedType(Type type) {
...
+	 *
+	 * @param object a given array
+	 * @return an {@code Iterator} iterating over the array
+	 */
+	private static Iterator<?> buildIteratorFromArray(Object object) {
+		Iterator<?> iterator;
+		if ( Object.class.isAssignableFrom( object.getClass().getComponentType() ) ) {
+			iterator = Arrays.asList( (Object[]) object ).iterator();
+		}
+		else if ( object.getClass() == boolean[].class ) {
```
gunnarmorling: Ok, if we need to do all that, how about avoiding the conversion to list then alltogether and have explicite iterator types:

    private static class BooleanArrayIterator implements Iterator<Boolean> {

        private boolean[] values;
        private int current = 0;

        public BooleanArrayIterator(boolean[] values) {
            this.values = values;
        }

        @Override
        public boolean hasNext() {
            return current < values.length;
        }

        @Override
        public Boolean next() {
            boolean result = values[current];
            current++;
            return result;
        }
    }

gsmet:
Force-pushed an update addressing your comments.

engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
```diff
@@ -259,36 +260,27 @@ public static Object getValue(Method method, Object object) {
 	}
 
 	/**
-	 * Indicates if the type is considered indexable (ie is an {@code Iterable}, an array or a {@code Map}).
+	 * Indicates if the type is considered a collection of elements (ie is an <code>Iterable</code>, an array or
```
gunnarmorling: We prefer to use `{@code ...}` over `<code>...</code>`. The reasoning is layed out in the contributor guide.

engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
```diff
@@ -259,36 +260,27 @@ public static Object getValue(Method method, Object object) {
...
+	 * a <code>Map</code>).
 	 *
-	 * @param type the type to inspect.
-	 *
-	 * @return Returns true if the type is indexable.
+	 * @param type the type to check.
+	 * @return returns <code>true</code> if <code>type</code> can be considered a collection of elements
```
gunnarmorling: Do you think `@param` and `@return` add much value here?

I'm all for great documentation, but I think not stating the obvious also is a virtue as it reduces the time to read. Hence I often only put a method comment and omit the others. I'd only add them if there is something specific to say about them. That's what I'd put here:

    /**
     * Whether the given type represents a collection of elements or not (i.e. whether it's an
     * {@code Iterable}, {@code Map} or array type).
     */

I know it's nit-picking, but I thought I'd bring it up so we may converge (I've also changed my style over time, so you may see examples I wrote using the style I'm critizing here ;)).

engine/src/test/java/org/hibernate/validator/test/internal/engine/typeannotationconstraint/TypeAnnotationConstraintTest.java
```diff
@@ -95,82 +129,349 @@ public void getter_constraint_provided_on_type_parameter_of_a_list_gets_validate
...
+		);
+		assertCorrectConstraintTypes(
+				constraintViolations,
+				NotBlank.class,
+				NotBlank.class,
+				NotNull.class
+		);
+	}
+
+	@Test
```
gunnarmorling: Wow, very thorough testing overall :) 👍

engine/src/test/java/org/hibernate/validator/test/internal/engine/typeannotationconstraint/TypeAnnotationConstraintTest.java
```diff
@@ -250,48 +646,38 @@ public void unsupported_use_of_type_constraints_logs_warning() {
...
-		assertThat( firstNode.getIndex() ).isNull();
-		assertThat( firstNode.getName() ).isEqualTo( "names" );
-		assertThat( firstNode.getKind() ).isEqualTo( ElementKind.PROPERTY );
-
-		Path.Node secondNode = propertyPathIterator.next();
-		assertThat( secondNode.getIndex() ).isEqualTo( 0 );
-		assertThat( secondNode.getName() ).isEqualTo( "<collection element>" );
-		assertThat( secondNode.getKind() ).isEqualTo( ElementKind.PROPERTY );
+	@SafeVarargs
+	private static <T> Set<T> newSet(T... elements) {
```
gunnarmorling: You could use `CollectionHelper#asSet()` instead.

gunnarmorling:
Regarding backporting, I think HV-1165 and HV-1176 would be worth a 5.3.4?

engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
gsmet: Weird. That's what I usually do. I think I copy pasted it from somewhere else. Will fix.

engine/src/main/java/org/hibernate/validator/internal/util/ReflectionHelper.java
gunnarmorling: I'd prefer more useful docs over making a tool happy. I don't know whether it's possible, but maybe one could have different rules/policies for API/SPI (where I can get behind a 0 warnings policy) and internal parts (where that seems over the top).

PR #599, opened 2016-12-07
HV-1181 Adding reference docs to distribution; excluding some unwante…

gunnarmorling:
…d files

Needs backporting to 5.4, too.


gunnarmorling:
Applied to master and 5.4.

PR #600, opened 2016-12-07
HV-1182 Update WildFly patch name in the documentation

gsmet:
* https://hibernate.atlassian.net/browse/HV-1182

To be backported to 5.4 before release.


gunnarmorling:
Applied to master and 5.4.

PR #602, opened 2016-12-08
HV-1180 Detect cyclic definition of group sequence due to group inheritance in the annotation processor

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1180

added check of interface inheritance in group sequences, and also added one more case that uses interface inheritance and "GroupSequence inheritance"


gsmet:
@marko-bekhta added one more commit and merged. I also backported it to 5.4.

marko-bekhta:
@gsmet Great! :) 
Also please look at Jandex PR when you have time for it :) And in a meantime if there's something else for me to look into - just let me know, thanks!

PR #605, opened 2016-12-14
Type annotations access for AP

marko-bekhta:
It's a simple example on how type annotations can be processed in AP, to check for things like: does next variable  `List<@NotNull String> someVar` has a `@Valid` annotation on it or not. 

I've seen in https://hibernate.atlassian.net/browse/HV-879 and also (I think) on one of the JDK dev-mail-lists  a discussion that there are a few bugs in implementation for getting such annotations. I've briefly looked inside the Checker framework and also done some "debugging" and it seems to me that in Checker they are casting Elements to `com.sun.tools.javac.code.Symbol` classes, like for example here - https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/util/element/MethodTypeParamApplier.java#L60

So this PR is an attempt to get those annotations somehow. What do you think, would something like this be a suitable approach ? And will it be of any value to spent more time on this ? 

I've put everything inside one test for now.


gunnarmorling:
Hi @marko-bekhta, the entire area of type-level constraints is in flux for BV 2.0 and semantics may change. So I'd suggest to not put effort into this right now. E.g. the `@Valid` annotation will not be needed for `List<@Email String>` in BV 2.0 anymore.

What concerns casting to internal types, we should avoid it as it will render the AP un-usable in Eclipse if not done carefully. We'd have to check the environment and do the right "hack" for each. We do it in MapStruct (a side of project of mine), e.g. [here](https://github.com/mapstruct/mapstruct/blob/master/processor/src/main/java/org/mapstruct/ap/internal/util/workarounds/EclipseAsMemberOfWorkaround.java) but also there it's a measure of last resort only.

gunnarmorling:
Hey @marko-bekhta, I think we should close this one for now as type constraints are just too much in flux right now and it'd be sort of a moving target to keep track of it with the AP. Let's revisit once the dust has settled in that area. Thanks for your efforts nevertheless!

PR #606, opened 2016-12-15
HV-1187 Support JDK9 ea+148

gsmet:
* https://hibernate.atlassian.net/browse/HV-1187

Waiting for https://github.com/jbossas/patch-gen/pull/17 to be integrated and a new alpha released.

Uses Groovy 2.4.8-SNAPSHOT as it has important fixes regarding JDK9 ea+148 support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ There are more build options available as well. For more information refer to [C
 
 To build Hibernate Validator with JDK 9, export the following environment variable:
 
-    export MAVEN_OPTS="--add-modules java.annotations.common,java.xml.bind"
+    export MAVEN_OPTS="--add-modules java.annotations.common,java.xml.bind --add-opens=java.base/java.security=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED"
 
 Then the build can be started like this:
 
--- a/README.md
+++ b/README.md
@@ -79,15 +79,18 @@ There are more build options available as well. For more information refer to [C
 
 To build Hibernate Validator with JDK 9, export the following environment variable:
 
-    export MAVEN_OPTS="--add-modules java.annotations.common,java.xml.bind --add-opens=java.base/java.security=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED"
+    export MAVEN_OPTS="--add-modules java.annotations.common,java.xml.bind --add-opens=java.base/java.security=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED"
 
 Then the build can be started like this:
 
-    mvn -s settings-example.xml clean install -DdisableDocumentationBuild=true -DdisableDistributionBuild
+    mvn -s settings-example.xml clean install
 
-The documentation and distribution modules are known to not work on Java 9 for the time being, hence they need to be excluded.
-Also the integration tests on WildFly will fail on Java 9 currently, hence this "integration" module is excluded automatically when building on JDK 9.
+Also the OSGi integration tests will fail on Java 9 currently, hence the "osgi" module is excluded automatically when building on JDK 9. We are waiting for the release of Karaf 4.1.0.
 
+Here are the reasons why we added the various --add-opens options:
+
+ * java.security: required by wildfly-maven-plugin:execute-commands (for the WildFly integration tests and the TCK runner running in container mode)
+ * java.lang: required by JRuby for Asciidoc processing
 
 ## Continuous Integration
 
--- a/cdi/pom.xml
+++ b/cdi/pom.xml
@@ -191,4 +191,16 @@
             </plugin>
         </plugins>
     </build>
+    <profiles>
+        <profile>
+            <id>jdk9</id>
+            <activation>
+                <jdk>9</jdk>
+            </activation>
+            <properties>
+                <!-- required by Weld -->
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -79,7 +79,7 @@ There are more build options available as well. For more information refer to [C
 
 To build Hibernate Validator with JDK 9, export the following environment variable:
 
-    export MAVEN_OPTS="--add-modules java.annotations.common,java.xml.bind"
+    export MAVEN_OPTS="--add-modules java.annotations.common,java.xml.bind --add-opens=java.base/java.security=ALL-UNNAMED --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED"
```
gunnarmorling: Could you add a comment saying which is needed for what? So we can track it more easily and check whether we can remove any after a component or plug-in update.

cdi/pom.xml
```diff
@@ -191,4 +191,15 @@
...
         </plugins>
     </build>
+    <profiles>
+        <profile>
+            <id>jdk9</id>
+            <activation>
+                <jdk>9</jdk>
+            </activation>
+            <properties>
+                <maven-surefire-plugin.argLine.add-opens>--add-opens java.base/java.lang=ALL-UNNAMED</maven-surefire-plugin.argLine.add-opens>
```
gunnarmorling: Same here.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
```diff
@@ -178,7 +178,7 @@ public BeanMetaDataImpl(Class<T> beanClass,
 			if ( propertyMetaData.isCascading() ) {
 				cascadedProperties.add( propertyMetaData );
 			}
-			else {
+			else if ( !propertyMetaData.getTypeArgumentsConstraints().isEmpty() ) {
```
gunnarmorling: Is it an optimization to avoid creating a stream instance?

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -634,31 +649,8 @@ else if ( constraintHelper.isMultiValueConstraint( annotationType ) ) {
...
-
-	private <A extends Annotation> MetaConstraint<A> createReturnValueMetaConstraint(Executable member,
-			ConstraintDescriptorImpl<A> descriptor) {
-		return new MetaConstraint<>( descriptor, ConstraintLocation.forReturnValue( member ) );
-	}
-
-	private <A extends Annotation> MetaConstraint<A> createCrossParameterMetaConstraint(Executable member,
-			ConstraintDescriptorImpl<A> descriptor) {
-		return new MetaConstraint<>( descriptor, ConstraintLocation.forCrossParameter( member ) );
+	private <A extends Annotation> MetaConstraint<?> createMetaConstraint(ConstraintDescriptorImpl<A> descriptor, ConstraintLocation location) {
```
gunnarmorling: Do we even need that any longer?

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -847,4 +838,57 @@ else if ( TypeHelper.isAssignable( AnnotatedArrayType.class, annotatedType.getCl
...
+	private static class TypeArgumentPropertyLocation implements TypeArgumentLocation {
+		private final Member member;
+
+		private TypeArgumentPropertyLocation(Member member) {
+			this.member = member;
+		}
+
+		@Override
+		public ConstraintLocation toConstraintLocation() {
+			return ConstraintLocation.forProperty( member );
```
gunnarmorling: Shouldn't it be a `TypeArgumentConstraintLocation`?

gsmet: No, it's the one which will be wrapped later. It's strictly equivalent to the existing code.

gsmet:
Pushed a new version with fixes and more comments.

gsmet:
> On the sharing of locations, it seems not to happen across metadata providers, right? Not a big problem, I'm still wondering though whether we only should switch to accessible members when creating the aggregated model. Then we could do it exactly once. Not sure though whether it's worth the effort. Maybe something for later (also to reduce conflicts with my concurrent work).

Yes, that might be something we want to do in the future but it's orthogonal to this patch.

pom.xml
```diff
@@ -951,8 +971,41 @@
                 <jdk>9</jdk>
             </activation>
             <properties>
-                <surefire.argLine.extension>--add-modules ${jigsaw.modules}</surefire.argLine.extension>
+                <maven-surefire-plugin.argLine.add-modules>--add-modules=${maven-surefire-plugin.jigsaw.modules}</maven-surefire-plugin.argLine.add-modules>
+                <arquillian.javaVmArguments.add-opens>
```
gunnarmorling: Wow, that's a whole lot. They are all required by WildFly?

gsmet: I took them from the WF patch referenced by Tomaz. I think it's better to be in line with their expectations. I didn't play hide and seek for this one.

gunnarmorling:
+1. Maybe only add a comment on where it's from.

gsmet:
@gunnarmorling pushed a new version:
- only use Groovy 2.4.8-SNAPSHOT for JDK 9
- upgrade patch-gen-maven-plugin to 2.0.1.Alpha3 that Alexey released

I think it's ready to go now. I'll work on the backports once this is in.

gunnarmorling:
Looks reat now! Merged, thanks, @gsmet!

PR #608, opened 2016-12-16
HV-1188 Switch to LinkedHashSet to preserve order of constraints and messages

mkurz:
- https://hibernate.atlassian.net/browse/HV-1188

The only place where `new HashSet` or `import java.util.HashSet` is left is in the tests and in the `hibernate-validator-performance` and `hibernate-validator-test-utils` projects.
I think about to also replace this occurrences with `LinkedHashSet` - so it's obvious for everybody that no `HashSet` are used anymore...

PS: I signed the CLA already.

### DIFF
--- a/annotation-processor/src/main/java/org/hibernate/validator/ap/AbstractElementVisitor.java
+++ b/annotation-processor/src/main/java/org/hibernate/validator/ap/AbstractElementVisitor.java
@@ -46,8 +46,8 @@ public class AbstractElementVisitor<T, V> extends ElementKindVisitor6<T, V> {
 	 * @param foundIssues a collection of issues to be reported
 	 */
 	protected void reportIssues(Collection<ConstraintCheckIssue> foundIssues) {
-		Set<ConstraintCheckIssue> warnings = CollectionHelper.newHashSet();
-		Set<ConstraintCheckIssue> errors = CollectionHelper.newHashSet();
+		Set<ConstraintCheckIssue> warnings = CollectionHelper.newLinkedHashSet();
+		Set<ConstraintCheckIssue> errors = CollectionHelper.newLinkedHashSet();
 
 		for ( ConstraintCheckIssue issue : foundIssues ) {
 			if ( issue.isError() ) {
--- a/annotation-processor/src/main/java/org/hibernate/validator/ap/ClassVisitor.java
+++ b/annotation-processor/src/main/java/org/hibernate/validator/ap/ClassVisitor.java
@@ -108,7 +108,7 @@ public class ClassVisitor extends AbstractElementVisitor<Void, Void> {
 
 	private void processClassChecks(Element element) {
 		try {
-			Set<ConstraintCheckIssue> allIssues = CollectionHelper.newHashSet();
+			Set<ConstraintCheckIssue> allIssues = CollectionHelper.newLinkedHashSet();
 			Collection<ClassCheck> classChecks = factory.getClassChecks( element );
 			for ( ClassCheck classCheck : classChecks ) {
 				allIssues.addAll( classCheck.execute( element ) );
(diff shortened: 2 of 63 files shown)


gunnarmorling:
Ok to Test

mkurz:
I also have pull request for the `5.4` branch eagerly waiting in the line :smile:

gsmet:
Hi @mkurz,

Quick question: did you analyze the change you made or did you just change `newHashSet` to `newLinkedHashSet` everywhere? Same for `asSet`.

`LinkedHashSet` consumes slightly more memory so we should evaluate this change for each use case e.g. do it only when we want to maintain the order.

That being said, we are heading towards adding more order to constraints evaluation (see http://beanvalidation.org/proposals/BVAL-248/) so it would make sense to keep the order of `ConstraintViolation`s - even if the annotation order was not guaranteed.

Thanks for your feedback!

mkurz:
Hi @gsmet,

thanks a lot for your feedback. At first I tried to just change `newHashSet`s which do effect the ordering - but I wasn't sure if I wasn't missing some (I am not 100% familiar with the source). I did some testing and found out I did miss some so I just went ahead and changed everything. I thought the impact is so minimal it maybe doesn't matter and maybe we can just completly switch to `LinkedHashSet`. Like make it a policy "Use `LinkedHashSet` everywere, never use `HashSet`" so we are sure order is kept everywhere. As in my initial comment I think/thougt about to also change the remaining `HashSet` in the tests to `LinkedHashSet`s as well to make this a consistent policy.

If you don't agree just switching to `LinkedHashSet`s maybe you want to take half an hour and go through my changes and comment which lines you think is ok to stay `HashSet`s. I could have a look at this after the weekend as well.

Also do you have any memory benchmarks on switching to `LinkedHashSet`s? What's the real life impact? Is it that bad?

mkurz:
I asked around a bit:
http://stackoverflow.com/q/41202793/810109

It's **8 bytes** more per entry. We can do the math now. That's 7,8 kB difference per 1000 constraint entries.

gunnarmorling:
Hi @mkurz, thanks a lot for your efforts! Keeping an order of the constraints is one of the issues being discussed for BV 2.0 (see the [BVAL-248 proposal](http://beanvalidation.org/proposals/BVAL-248/) for the discussions that happened so far). I think we should first converge on this at the spec level before putting an implementation in place. My concern is that people would rely on the behavior of such ad-hoc implementation, resulting in portability issues, not the least when moving to the BV 2.0 solution proper.

That being said, I'm not fully sure whether the conclusions in the SO answer you linked are correct. There's a lot of implicitness in there :) It'd be great to have confirmation from the JDK team that there indeed is a reliable order (apart from the one for repeatable annotations, which most of the SO discussion relates to) and that they are committing to retain that.

gunnarmorling:
Hi @mkurz, after some more consideration I'm going to not merge this change for now.

I think we should first look into a solution on the spec level. Your contribution to this would be very welcome, e.g. join us on the [mailing list](http://lists.jboss.org/pipermail/beanvalidation-dev/) and get a discussion started. Getting an authoritative answer from the JDK team whether there is a guaranteed order of retrieved annotations would be a great thing.

Thanks for your effort around this topic; also if we are not going for the suggested change right now, I think we'll come to a good solution in the BV 2.0 time frame.

mkurz:
I will try to get in touch with the JDK team after the holidays. Is there an ETA for BV 2.0? Will it be finished in 2017?
Thank you.

gunnarmorling:
> I will try to get in touch with the JDK team after the holidays.

Cool, I'm quite curious about the outcome.

> Is there an ETA for BV 2.0? Will it be finished in 2017?

Yes, we are aiming for a release in 2017. The first preview will be out
very early in the year, and BV 2.0 Final is also planned to be part of Java
EE 8 (scheduled for 2017, too).

PR #610, opened 2016-12-19
HV-1158 Use standard EL classes added in EL 3 rather than our own

gsmet:
* https://hibernate.atlassian.net/browse/HV-1158


engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/SimpleELContext.java
```diff
@@ -6,74 +6,52 @@
...
-import javax.el.VariableMapper;
+import javax.el.StandardELContext;
+import javax.el.StaticFieldELResolver;
 
 /**
  * @author Hardy Ferentschik
+ * @author Guillaume Smet
  */
-public class SimpleELContext extends ELContext {
+public class SimpleELContext extends StandardELContext {
```
gunnarmorling: Couldn't we even get rid of this altogether and use the new `ELManager`?

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/RootResolver.java
```diff
@@ -30,11 +28,9 @@
 	public static final String FORMATTER = "formatter";
 	private static final String FORMAT = "format";
 
-	private final Map<String, Object> map = Collections.synchronizedMap( new HashMap<String, Object>() );
-
 	@Override
 	public Class<?> getCommonPropertyType(ELContext context, Object base) {
-		return base == null ? String.class : null;
+		return null;
```
gunnarmorling: Is it right to return null here?

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/RootResolver.java
```diff
@@ -44,47 +40,33 @@
...
-		return false;
+		return true;
 	}
 
 	@Override
 	public void setValue(ELContext context, Object base, Object property, Object value) {
-		if ( resolve( context, base, property ) ) {
-			setProperty( (String) property, value );
-		}
+		throw new PropertyNotWritableException();
```
gunnarmorling: Probably we should wrap it into a `ValidationException` (not sure if it ever can occur).

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/SimpleELContext.java
gunnarmorling: Or, if we want to avoid `ELManager` due to the classloading issues we discussed a while ago, why is it that we need a custom `ELContext`?

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/RootResolver.java
gsmet: AFAICS it's not something that should occur. And all the other resolvers throw a `PropertyNotWritableException` so we'd better be consistent.

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/RootResolver.java
gsmet: Yes. In fact, it was already returning null before simplication as base was null when it was called.

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/SimpleELContext.java
gsmet: There are 2 reasons why I kept our specific context:
- first, we provide a way to inject a specific `ExpressionFactory` and believe it or not, the one we passed to `StandardELContext` is not used, except to initialize the `StreamELResolver` (see the constructor of `StandardELContext`). See ELContext#convertToType to see how it defaults to the static one if it is not found in the context. So if we want to be sure it's used, we need to put it in the context.
- the second reason is that, with a `StandardELContext`, we instantiate new resolvers for each interpolation, while they are stateless and can be shared.

Note that to keep the list of resolvers static, I didn't add the `StreamELResolver` which didn't look as much value to me (and wasn't present before). However, I added the `StaticFieldELResolver`.

gunnarmorling: > So if we want to be sure it's used, we need to put it in the context.

Yes, but couldn't it be done like this:

    ELContext ctx  = new StandardELContext( factory );
    ctx.putContext( ExpressionFactory.clss, factory );

> we instantiate new resolvers for each interpolation

Mh, that sounds bad. Where is it happening?

gsmet: As discussed on HipChat, we instantiate a new context for each message interpolation (which is the way it should be done). In our implementation, we use a static `CompositeELResolver` which is shared while `StandardELContext` instantiate new resolvers.

gunnarmorling:

PR #614, opened 2016-12-23
HV-1193 Avoid systematic resizing of collections when we provide a size

gsmet:
* https://hibernate.atlassian.net/browse/HV-1193

Even if it's just to stop having this conversation over and over again, I think it's worth it. I expect the calculation to be inlined so it shouldn't cost a lot.

Having a systematic resizing when we provide a size is definitely something we should avoid.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/CollectionHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/CollectionHelper.java
@@ -34,7 +34,7 @@ public final class CollectionHelper {
 	}
 
 	public static <K, V> HashMap<K, V> newHashMap(int size) {
-		return new HashMap<K, V>( size );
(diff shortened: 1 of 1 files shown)


gsmet:
Note that the build is going to be broken until we merge the java.time PR.

engine/src/main/java/org/hibernate/validator/internal/util/CollectionHelper.java
```diff
@@ -76,7 +70,7 @@ private CollectionHelper() {
 	}
 
 	public static <T> ArrayList<T> newArrayList(int size) {
-		return new ArrayList<T>( size );
+		return new ArrayList<T>( getInitialCapacityFromExpectedSize( size ) );
```
gunnarmorling: Why are you doing it for a list type?

PR #616, opened 2016-12-28
Corrected documentation URL for hibernate-validator-annotation-processor

kdombeck:
Current URL goes to a not found page.


gsmet:
Thanks for taking the time to report this and propose a fix. I proposed a slightly different update to point to the html_single documentation which is now our canonical reference (trying to improve SEO here).

Closing this one.

PR #622, opened 2017-01-07
HV-1185 Expand tests for URL validator

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1185

I've added those urls from that [link](https://mathiasbynens.be/demo/url-regex) mentioned in the issue description. 

Not all of the cases from that page pass though... I've commented out ones that do not pass. Should we update the regexp in RegexpURLValidator (update the one that is already in there or maybe replace with the one by `@diegoperini`) so all the cases will pass. WDYT?

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
@@ -72,7 +72,7 @@ public class EmailValidator implements ConstraintValidator<Email, CharSequence> 
 			return false;
 		}
 
-		return DomainNameUtil.isValidDomainAddress( domainPart );
+		return DomainNameUtil.isValidEmailDomainAddress( domainPart );
 	}
 
 	/**
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
@@ -6,6 +6,10 @@
  */
 package org.hibernate.validator.internal.constraintvalidators.hv;
 
+import static java.util.regex.Pattern.CASE_INSENSITIVE;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
 
@@ -28,6 +32,18 @@ import org.hibernate.validator.internal.util.DomainNameUtil;
  * @author Guillaume Smet
  */
 public class EmailValidator implements ConstraintValidator<Email, CharSequence> {
+
+	private static final int MAX_LOCAL_PART_LENGTH = 64;
+
+	private static final String LOCAL_PART_ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~\u0080-\uFFFF-]";
+	private static final String LOCAL_PART_INSIDE_QUOTES_ATOM = "([a-z0-9!#$%&'*.(),<>\\[\\]:;  @+/=?^_`{|}~\u0080-\uFFFF-]|\\\\\\\\|\\\\\\\")";
+	/**
+	 * Regular expression for the local part of an email address (everything before '@')
+	 */
+	private static final Pattern LOCAL_PART_PATTERN = Pattern.compile(
+			"(" + LOCAL_PART_ATOM + "+|\"" + LOCAL_PART_INSIDE_QUOTES_ATOM + "+\")" +
+					"(\\." + "(" + LOCAL_PART_ATOM + "+|\"" + LOCAL_PART_INSIDE_QUOTES_ATOM + "+\")" + ")*", CASE_INSENSITIVE
+	);
 
 	@Override
 	public void initialize(Email annotation) {
@@ -52,12 +68,26 @@ public class EmailValidator implements ConstraintValidator<Email, CharSequence> 
 		String localPart = stringValue.substring( 0, splitPosition );
 		String domainPart = stringValue.substring( splitPosition + 1 );
 
-		if ( !DomainNameUtil.isValidEmailLocalPart( localPart ) ) {
+		if ( !isValidEmailLocalPart( localPart ) ) {
 			return false;
 		}
 
 		return DomainNameUtil.isValidDomainAddress( domainPart );
 	}
 
+	/**
+	 * Checks validity of emails local part.
+	 *
+	 * @param localPart an email local part to check
+	 *
+	 * @return {@code true} if a given emails local part is valid, {@code false} otherwise
+	 */
+	private boolean isValidEmailLocalPart(String localPart) {
+		if ( localPart.length() > MAX_LOCAL_PART_LENGTH ) {
+			return false;
+		}
+		Matcher matcher = LOCAL_PART_PATTERN.matcher( localPart );
+		return matcher.matches();
+	}
 
 }
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
@@ -6,16 +6,11 @@
  */
 package org.hibernate.validator.internal.constraintvalidators.hv;
 
-import static java.util.regex.Pattern.CASE_INSENSITIVE;
-
-import java.net.IDN;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
 
 import org.hibernate.validator.constraints.Email;
+import org.hibernate.validator.internal.util.DomainNameUtil;
 
 /**
  * Checks that a given character sequence (e.g. string) is a well-formed email address.
@@ -33,34 +28,6 @@ import org.hibernate.validator.constraints.Email;
  * @author Guillaume Smet
  */
 public class EmailValidator implements ConstraintValidator<Email, CharSequence> {
-	private static final String LOCAL_PART_ATOM = "[a-z0-9!#$%&'*+/=?^_`{|}~\u0080-\uFFFF-]";
-	private static final String LOCAL_PART_INSIDE_QUOTES_ATOM = "([a-z0-9!#$%&'*.(),<>\\[\\]:;  @+/=?^_`{|}~\u0080-\uFFFF-]|\\\\\\\\|\\\\\\\")";
-	private static final String DOMAIN_LABEL = "[a-z0-9!#$%&'*+/=?^_`{|}~-]";
-	private static final String DOMAIN = DOMAIN_LABEL + "+(\\." + DOMAIN_LABEL + "+)*";
-	private static final String IP_DOMAIN = "\\[[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\]";
-	//IP v6 regex taken from http://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses
-	private static final String IP_V6_DOMAIN = "(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))";
-	private static final int MAX_LOCAL_PART_LENGTH = 64;
-	/**
-	 * This is the maximum length of a domain name. But be aware that each label (parts separated by a dot) of the
-	 * domain name must be at most 63 characters long. This is verified by {@link IDN#toASCII(String)}.
-	 */
-	private static final int MAX_DOMAIN_PART_LENGTH = 255;
-
-	/**
-	 * Regular expression for the local part of an email address (everything before '@')
-	 */
-	private static final Pattern LOCAL_PART_PATTERN = Pattern.compile(
-			"(" + LOCAL_PART_ATOM + "+|\"" + LOCAL_PART_INSIDE_QUOTES_ATOM + "+\")" +
-					"(\\." + "(" + LOCAL_PART_ATOM + "+|\"" + LOCAL_PART_INSIDE_QUOTES_ATOM + "+\")" + ")*", CASE_INSENSITIVE
-	);
-
-	/**
-	 * Regular expression for the domain part of an email address (everything after '@')
-	 */
-	private static final Pattern DOMAIN_PATTERN = Pattern.compile(
-			DOMAIN + "|" + IP_DOMAIN + "|" + "\\[IPv6:" + IP_V6_DOMAIN + "\\]", CASE_INSENSITIVE
-	);
 
 	@Override
 	public void initialize(Email annotation) {
@@ -75,7 +42,7 @@ public class EmailValidator implements ConstraintValidator<Email, CharSequence> 
 		// cannot split email string at @ as it can be a part of quoted local part of email.
 		// so we need to split at a position of last @ present in the string:
 		String stringValue = value.toString();
-		int splitPosition = stringValue.lastIndexOf( "@" );
+		int splitPosition = stringValue.lastIndexOf( '@' );
 
 		// need to check if
 		if ( splitPosition < 0 ) {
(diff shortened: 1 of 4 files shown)


engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
```diff
@@ -290,10 +290,91 @@ private void runUrlContainerValidation(Validator validator, URLContainer contain
...
-		assertTrue( validator.isValid( "ftp://abc.de", null ) );
+		assertFalse( validator.isValid( "//", null ) );
+		assertFalse( validator.isValid( "//a", null ) );
+		assertFalse( validator.isValid( "///", null ) );
+		assertFalse( validator.isValid( "///a", null ) );
+		assertFalse( validator.isValid( "foo.com", null ) );
+		assertFalse( validator.isValid( ":// should fail", null ) );
+//		assertFalse( validator.isValid( "rdar://1234", null ) );
+//		assertFalse( validator.isValid( "ftps://foo.bar/", null ) );
+//		assertFalse( validator.isValid( "h://test", null ) );
```
gsmet: I don't think these ones are an issue: we should support exotic protocols.

Note that the purpose of the regexps presented in https://mathiasbynens.be/demo/url-regex is to create a URL shortener, not to follow the standard, so all the examples presented shouldn't be considered valid.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
```diff
@@ -290,10 +290,91 @@ private void runUrlContainerValidation(Validator validator, URLContainer contain
...
+//			assertFalse( validator.isValid( "http://", null ) );
+//			assertFalse( validator.isValid( "http://.", null ) );
+//			assertFalse( validator.isValid( "http://..", null ) );
+//			assertFalse( validator.isValid( "http://../", null ) );
+//			assertFalse( validator.isValid( "http://?", null ) );
+//			assertFalse( validator.isValid( "http://??", null ) );
+//			assertFalse( validator.isValid( "http://??/", null ) );
+//			assertFalse( validator.isValid( "http://#", null ) );
+//			assertFalse( validator.isValid( "http://##", null ) );
+//			assertFalse( validator.isValid( "http://##/", null ) );
```
gsmet: All the ones above are indeed an issue.

If we decide to fix this, we might do it by using the domain validation we apply in EmailValidator to the domain part. I think it should fix most of the issues.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
```diff
@@ -290,10 +290,91 @@ private void runUrlContainerValidation(Validator validator, URLContainer contain
...
+			assertFalse( validator.isValid( "http://foo.bar?q=Spaces should be encoded", null ) );
+//			assertFalse( validator.isValid( "http:///a", null ) );
+			assertFalse( validator.isValid( "http:// shouldfail.com", null ) );
+			assertFalse( validator.isValid( "http://foo.bar/foo(bar)baz quux", null ) );
+//			assertFalse( validator.isValid( "http://-error-.invalid/", null ) );
+//			assertFalse( validator.isValid( "http://a.b--c.de/", null ) );
+//			assertFalse( validator.isValid( "http://-a.b.co", null ) );
+//			assertFalse( validator.isValid( "http://a.b-.co", null ) );
```
gsmet: The four above with dashes might be considered an issue too.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
```diff
@@ -290,10 +290,91 @@ private void runUrlContainerValidation(Validator validator, URLContainer contain
...
+//			assertFalse( validator.isValid( "http://0.0.0.0", null ) );
+//			assertFalse( validator.isValid( "http://10.1.1.0", null ) );
+//			assertFalse( validator.isValid( "http://10.1.1.255", null ) );
+//			assertFalse( validator.isValid( "http://224.1.1.1", null ) );
+//			assertFalse( validator.isValid( "http://1.1.1.1.1", null ) );
+//			assertFalse( validator.isValid( "http://123.123.123", null ) );
+//			assertFalse( validator.isValid( "http://3628126748", null ) );
+//			assertFalse( validator.isValid( "http://.www.foo.bar/", null ) );
```
gsmet: The ones with malformed IP addresses might also be considered an issue, even if I don't think it's that important to be thorough on IP addresses based URLs.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
```diff
@@ -290,10 +290,91 @@ private void runUrlContainerValidation(Validator validator, URLContainer contain
...
+		assertTrue( validator.isValid( "http://foo.com/blah_(wikipedia)#cite-1", null ) );
+		assertTrue( validator.isValid( "http://foo.com/blah_(wikipedia)_blah#cite-1", null ) );
+		assertTrue( validator.isValid( "http://foo.com/unicode_(✪)_in_parens", null ) );
+		assertTrue( validator.isValid( "http://foo.com/(something)?after=parens", null ) );
+		assertTrue( validator.isValid( "http://☺.damowmow.com/", null ) );
+		assertTrue( validator.isValid( "http://code.google.com/events/#&product=browser", null ) );
+		assertTrue( validator.isValid( "http://j.mp", null ) );
+		assertTrue( validator.isValid( "ftp://foo.bar/baz", null ) );
+		assertTrue( validator.isValid( "http://foo.bar/?q=Test%20URL-encoded%20stuff", null ) );
+		assertTrue( validator.isValid( "http://مثال.إختبار", null ) );
```
gsmet: I'm surprised the IDN URLs pass with java.net.URI. I used to think that they were not supported (I had to use galimatias to parse URIs with IDN). Did it change recently?

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
```diff
@@ -290,10 +290,91 @@ private void runUrlContainerValidation(Validator validator, URLContainer contain
...
+			assertFalse( validator.isValid( "http:// shouldfail.com", null ) );
+			assertFalse( validator.isValid( "http://foo.bar/foo(bar)baz quux", null ) );
+//			assertFalse( validator.isValid( "http://-error-.invalid/", null ) );
+//			assertFalse( validator.isValid( "http://a.b--c.de/", null ) );
+//			assertFalse( validator.isValid( "http://-a.b.co", null ) );
+//			assertFalse( validator.isValid( "http://a.b-.co", null ) );
+//			assertFalse( validator.isValid( "http://0.0.0.0", null ) );
+//			assertFalse( validator.isValid( "http://10.1.1.0", null ) );
+//			assertFalse( validator.isValid( "http://10.1.1.255", null ) );
+//			assertFalse( validator.isValid( "http://224.1.1.1", null ) );
```
gsmet: The ones with valid IP addresses should be considered valid.

gunnarmorling: Why is it the expectation that these are invalid? Is it some illegal IP range or something?

I'd be fine if we just rejected those with a wrong number of octets but don't apply any further validation on the actual ranges (if that's the issue).

gsmet: @gunnarmorling the link from which the URLs were extracted was used to have an URL validator for link shortener. Thus they have very special requirements. That's why we should be careful with the examples given.

marko-bekhta: here's some explanation from the author of that regexp:

> I have added simple network ranges validation, the rules I used are:
> - valid range 1.0.0.0 - 223.255.255.255, network adresses above and including 224.0.0.0 are reserved addresses
> - first and last IP address of each class is excluded since they are used as network broadcast addresses
> since I don't think this is worth implementing completely in a regular expression, a following pass should exclude the Intranet address space:
> 10.0.0.0 - 10.255.255.255
> 172.16.0.0 - 172.31.255.255
> 192.168.0.0 - 192.168.255.255
> the loopback and the automatic configuration address space:
> 127.0.0.0 - 127.255.255.255
> 169.254.0.0 - 169.254.255.255
> while the local, multicast and and the reserved address spaces:
> 0.0.0.0 - 0.255.255.255 (SPECIAL-IPV4-LOCAL-ID-IANA-RESERVED)
> 224.0.0.0 - 239.255.255 (MCAST-NET)
> 240.0.0.0 - 255.255.255.255 (SPECIAL-IPV4-FUTURE-USE-IANA-RESERVED)
> should already be excluded by the above regular expression.

I've extracted the validation used in EmailValidator as @gsmet suggested in a few comments above, and left commented out those cases that still do not pass.

engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	/**
+	 * Checks validity of emails local part.
+	 *
+	 * @param localPart an email local part to check
+	 *
+	 * @return {@code true} if a given emails local part is valid, {@code false} otherwise
+	 */
+	public static boolean isValidEmailLocalPart(String localPart) {
```
gsmet: I would keep this in the EmailValidator as it's in no way related to the domain validation.

marko-bekhta: Right, I thought that regexps for local part and domain had same subparts so I moved both out - but after closer look - they don't. I'll move the local part back. Also I found out that for the email domain ip addresses are in brackets for example - `jsmith@[192.168.2.1] or jsmith@[IPv6:2001:db8::1]` so the same regexp wouldn't work for domain and email. I'll see what I can do about it and let you know when I have something.

marko-bekhta:
So I've finally made a few more changes :). Now it should catch more invalid cases. I can't see other way than having two methods (one for email another for URLs) to validate domain part, as for the email IPv4 should be in `[]` and for URL it shouldn't (also for IPv6 for email it should be like `[IPv6:SOME_IP_GOES_HERE]`, and for URL - IPv6 is just in `[]`). 

Another question is about dotless domains. Even though they are prohibited (see [here](https://www.icann.org/news/announcement-2013-08-30-en)) for example emails with such domains are considered "valid" (see [here](https://en.wikipedia.org/wiki/Email_address)). If we disallow such domains in that case URL validation for files fails. From one of the cases present in tests - for `file://Users/foobar/tmp` the "host" part is `Users` and it is dotless. Should we leave it as is? Or should we do something about it ? WDYT?

I also added a simple valid/invalid IPv6 case, as these were missing for the URLs.

engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
```diff
@@ -23,36 +23,67 @@
...
 
 	/**
-	 * Regular expression for the domain part of an email address (everything after '@')
+	 * Regular expression for the domain part of an URL
 	 */
+	// A host string must be a domain string, an IPv4 address string, or "[", followed by an IPv6 address string, followed by "]".
 	private static final Pattern DOMAIN_PATTERN = Pattern.compile(
-			DOMAIN + "|" + IP_DOMAIN + "|" + "\\[IPv6:" + IP_V6_DOMAIN + "\\]", CASE_INSENSITIVE
+			DOMAIN + "|\\[" + IP_V6_DOMAIN + "\\]", CASE_INSENSITIVE
```
gsmet: So, the IP domain is now covered by `DOMAIN`?

engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
```diff
@@ -23,36 +23,67 @@
...
 		// the regular expression match would take care of this, but IDN.toASCII drops the trailing '.'
 		if ( domain.endsWith( "." ) ) {
 			return false;
 		}
 
+		Matcher matcher = pattern.matcher( domain );
+		if ( !matcher.matches() ) {
+			return false;
+		}
+
```
gsmet: I'm a bit concerned about this change. To validate the domain name with the regexp, I think we should convert to punycode first. At least, that's what I see everywhere.

Or you did change the regexp to also accept the range of characters from IDN?

All in all, I think the commit would merit a longer commit message so that we can understand what you're doing.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
```diff
@@ -363,22 +364,16 @@ private void assertValidUrls(ConstraintValidator<URL, CharSequence> validator) {
...
-//			assertFalse( validator.isValid( "http://a.b-.co", null ) );
-//			assertFalse( validator.isValid( "http://0.0.0.0", null ) );
-//			assertFalse( validator.isValid( "http://10.1.1.0", null ) );
-//			assertFalse( validator.isValid( "http://10.1.1.255", null ) );
-//			assertFalse( validator.isValid( "http://224.1.1.1", null ) );
-//			assertFalse( validator.isValid( "http://1.1.1.1.1", null ) );
+			assertFalse( validator.isValid( "http://-error-.invalid/", null ) );
+			assertFalse( validator.isValid( "http://a.b--c.de/", null ) );
+			assertFalse( validator.isValid( "http://-a.b.co", null ) );
+			assertFalse( validator.isValid( "http://a.b-.co", null ) );
```
gsmet: AFAICS, you changed the regexp to make these cases invalid?

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
```diff
@@ -363,22 +364,16 @@ private void assertValidUrls(ConstraintValidator<URL, CharSequence> validator) {
...
 //			assertFalse( validator.isValid( "http://123.123.123", null ) );
 //			assertFalse( validator.isValid( "http://3628126748", null ) );
 			assertFalse( validator.isValid( "http://.www.foo.bar/", null ) );
 			assertFalse( validator.isValid( "http://www.foo.bar./", null ) );
 			assertFalse( validator.isValid( "http://.www.foo.bar./", null ) );
-//			assertFalse( validator.isValid( "http://10.1.1.1", null ) );
-//			assertFalse( validator.isValid( "http://10.1.1.254", null ) );
+			assertFalse( validator.isValid( "http://2001:0db8:0a0b:12f0:0000:0000:0000:0001", null ) );
```
gsmet: In the regexp, it looks like you put `[]` around the IPv6 IP address but here you just put the address without any wrapper? Did I miss something?

engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
marko-bekhta: well as the regexp for the domain is built from chars and digits IPv4 is picked up with it as well, so I thought that as additional `[]` are not needed for IPv4 then we can skip the part` | IP_DOMAIN`

engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
marko-bekhta: Yes, this part might be a bit confusing ... As you've guessed I've changed the regexp to allow other chars as well (not only ASCII) and moved the regexp check before `IDN.toASCII`.
The only reason for this change was to be able to catch invalid cases that you commented on next:
```
http://a.b--c.de
```
so as I understand these two dashes `--` makes this url invalid. But if we do `IDN.toASCII` first and there were non ASCII symbols it might generate for example `xn--` which would be caught with the regexp as invalid (in case we update regexp to catch `--`), even though it's not.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
marko-bekhta: yes, it's this change:
`DOMAIN_LABEL = "(" + DOMAIN_CHARS_WITHOUT_DASH + "-?)*" + DOMAIN_CHARS_WITHOUT_DASH + "+"`

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/URLValidatorTest.java
marko-bekhta: right it's an invalid case - so there's `assertFalse`  on this line. there's a valid case at line 332:
`assertTrue( validator.isValid( "http://[2001:0db8:0a0b:12f0:0000:0000:0000:0001]", null ) );`

gsmet: Ah, right, sorry!

gunnarmorling:
@gsmet, @marko-bekhta, where are we with this one? Does it require any further changes?

marko-bekhta:
Hi @gunnarmorling , I've replied to @gsmet comments and as I understand he'll make his final review. I don't have any more changes to add from my side.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/EmailValidator.java
```diff
@@ -85,42 +68,26 @@ public boolean isValid(CharSequence value, ConstraintValidatorContext context) {
...
 
-	private boolean matchLocalPart(String localPart) {
+	/**
+	 * Checks validity of emails local part.
+	 *
+	 * @param localPart an email local part to check
+	 *
+	 * @return {@code true} if a given emails local part is valid, {@code false} otherwise
+	 */
+	private boolean isValidEmailLocalPart(String localPart) {
```
gsmet: I'll remove this comment in the final version committed as it's useless (private method and method name and parameter very explicit).

engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
```diff
@@ -0,0 +1,102 @@
...
+	private static final String IP_DOMAIN = "[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}";
+	//IP v6 regex taken from http://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses
+	private static final String IP_V6_DOMAIN = "(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))";
+
+	/**
+	 * Regular expression for the domain part of an URL
+	 */
+	// A host string must be a domain string, an IPv4 address string, or "[", followed by an IPv6 address string, followed by "]".
```
gsmet: I will include that in the previous comment.

engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
```diff
@@ -0,0 +1,102 @@
...
+
+	private DomainNameUtil() {
+	}
+
+	/**
+	 * Checks validity of domain name used in email. To be valid it should be either valid host name, or IP address in []
+	 *
+	 * @param domain domain to check for validity
+	 *
+	 * @return {@code true} if provided string is a valid domain, {@code false} otherwise
```
gsmet: Note that when we reformat, `@param` and `@return` are not separated by a new line (at least in Eclipse). I'll fix that.

gsmet:
@marko-bekhta merged, thanks!

engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
marko-bekhta: In such moments I sometimes think - maybe I should use Eclipse (at least for HV), to not have these formatting differences... But then I understand that I'm to used to IntelliJ :)
need to find and update this to match :)

PR #623, opened 2017-01-09
HV-1124 Add integration test for javax.money validation on WildFly

gsmet:
* https://hibernate.atlassian.net/browse/HV-1124

### DIFF
--- a/modules/src/script/setupModules.groovy
+++ b/modules/src/script/setupModules.groovy
@@ -2,9 +2,13 @@ println "[INFO] ----------------------------------------------------------------
 println "[INFO] UPDATING BEAN VALIDATION RELATED WILDFLY MODULES                        ";
 println "[INFO] ------------------------------------------------------------------------";
 
-def processFileInplace(file, Closure processText) {
+def processFileInplace(File file, Closure processText) {
     def text = file.text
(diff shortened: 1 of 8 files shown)


modules/src/script/setupModules.groovy
```diff
@@ -24,6 +28,7 @@ println "[INFO] Using HV version " + hvArtifactName;
 processFileInplace( hvModuleXml ) { text ->
     text.replaceAll( /hibernate-validator.*jar/, hvArtifactName )
 }
+appendDependency( hvModuleXml, "javax.money.api", true )
```
gunnarmorling: I may misremember, but didn't we want to add this optional dependency to the descriptor to begin with? Then a user would only have to drop in the modules (and add the API module in jboss-deployment-structure.xml), but not fiddle with any existing module.xml.

gsmet: Isn't it what it does? This optional dependency is included in our patch file.

gunnarmorling: Ah, of course. Not sure why, but somehow I thought that'd be the setup code within the integration test module.

gsmet: :)

PR #624, opened 2017-01-11
HV-1202 Use Instant for Date and Calendar validation

gsmet:
* https://hibernate.atlassian.net/browse/HV-1202


gunnarmorling:

PR #626, opened 2017-01-13
HV-1207 - HV-1208 Cache the ValueExtractor determination in the metadata, per constraint unwrapping and implicit unwrapping

gsmet:
* https://hibernate.atlassian.net/browse/HV-1207
 * https://hibernate.atlassian.net/browse/HV-1208

Note: it doesn't compile on CI as it needs the related BV changes: https://github.com/beanvalidation/beanvalidation-api/pull/77 .


engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
```diff
@@ -41,13 +45,28 @@
 	private final ConstraintLocation location;
 
 	/**
+	 * The {@link ValueExtractor} used to extract the value for validation.
+	 */
+	private final ValueExtractorDescriptor valueExtractorDescriptor;
```
gunnarmorling: I'm wondering whether it shouldn't be on `ConstraintLocation` instead.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
```diff
@@ -41,13 +45,28 @@
...
+
+	/**
+	 * The type of the validated element.
+	 */
+	private final Type typeOfValidatedElement;
```
gunnarmorling: How is it different from `ConstraintLocation#getTypeForValidatorResolution()`? Having all these types with slightly different semantics makes me a bit concerned.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
gsmet: Nope. `ConstraintLocation` is the object defining a location, it can be shared by several different constraints, some of them requiring unwrapping, some of them not.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
gsmet: Yeah... The fact is that it's currently very hard to understand what exactly the various types are for. I kept the behavior as it was before. It may be suboptimal but I supposed the existing code was here for a reason. If you want to disambiguate them, I would appreciate some explanation/comments because I was not able to do it myself.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
gsmet: Note that in the MetaConstraint, the extractor is only set if there is some extraction going on.

I think it will be more clear with the next commit coming.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
gsmet: The `typeOfValidatedElement` is the type of the validation object. It might not be the type used for validator resolution as it might be the wrapped type.

Note that it's exactly the same behavior as before. The code has been moved but is exactly identical to the existing logic.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
gunnarmorling: Ah, ok, so that's already planning for allowing different validation-applies-to settings for the constraint of one element. Makes sense.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/ConstraintMappingContextImplBase.java
```diff
@@ -54,25 +57,28 @@ protected void addConstraint(ConfiguredConstraint<?> constraint) {
 	/**
 	 * Converts all constraints managed by this creational context into {@link MetaConstraint}s.
 	 *
-	 * @param constraintHelper constraint helper required for creation of meta constraints.
-	 *
-	 * @return All constraints of this context as meta constraints.
+	 * @param constraintHelper constraint helper required for creation of meta constraints
```
gunnarmorling: No need for these @params if you ask me.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -314,7 +314,7 @@ public boolean isFailFast() {
 		return failFast;
 	}
 
-	ValueExtractors getValueExtractors() {
+	ValueExtractorManager getValueExtractors() {
```
gunnarmorling: `getValueExtractorManager()`

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -349,15 +349,17 @@ Validator createValidator(ConstraintValidatorFactory constraintValidatorFactory,
...
 			MethodValidationConfiguration methodValidationConfiguration) {
 
 		BeanMetaDataManager beanMetaDataManager;
 		if ( !beanMetaDataManagerMap.containsKey( parameterNameProvider ) ) {
 			beanMetaDataManager = new BeanMetaDataManager(
 					constraintHelper,
 					executableHelper,
+					typeResolutionHelper,
 					parameterNameProvider,
+					valueExtractors,
```
gunnarmorling: The var name should be adjusted, too.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ObservableValueExtractor.java
```diff
@@ -19,6 +20,7 @@
  * @author Gunnar Morling
  */
 @IgnoreJava8Requirement
```
gunnarmorling: Just seeing it, `@IgnoreJava8Requirement` can go away.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorDescriptor.java
```diff
@@ -85,21 +89,29 @@ private static Type getExtractedType(Class<?> extractorImplementationType) {
...
-	public TypeVariable<?> extractedTypeParameter() {
+	public TypeVariable<?> getExtractedTypeParameter() {
 		return extractedTypeParameter;
 	}
 
 	public ValueExtractor<?> getValueExtractor() {
 		return valueExtractor;
 	}
 
+	public boolean unwrapByDefault() {
```
gunnarmorling: `isUnwrapByDefault()`, sticking to your change above?

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyMetaData.java
```diff
@@ -185,26 +183,6 @@ public boolean accepts(ConstrainedElement constrainedElement) {
 		public final void add(ConstrainedElement constrainedElement) {
 			super.add( constrainedElement );
 
-			// HV-925
-			// Trying to detect inconsistent value unwrapping configuration between a property field and its getter.
-			// If a field or getter explicitly uses @UnwrapValidatedValue, the corresponding getter / field needs to either
-			// not use @UnwrapValidatedValue or use the same value for the annotation.
-			UnwrapMode newUnwrapMode = constrainedElement.unwrapMode();
```
gunnarmorling: Nice!

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
gunnarmorling: Where has it been before?

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraints.java
```diff
@@ -0,0 +1,108 @@
...
+		ValueExtractorDescriptor valueExtractorDescriptor;
+		Type typeOfValidatedElement;
+
+		if ( location instanceof TypeArgumentConstraintLocation ) {
+			TypeArgumentConstraintLocation typeArgumentConstraintLocation = (TypeArgumentConstraintLocation) location;
+
+			Class<?> declaredType = TypeHelper.getErasedReferenceType( typeArgumentConstraintLocation.getDelegate().getTypeForValidatorResolution() );
+			TypeVariable<?> typeParameter = typeArgumentConstraintLocation.getTypeParameter();
+
+			valueExtractorDescriptor = valueExtractors.getCascadedValueExtractor( declaredType, typeParameter );
```
gunnarmorling: We should rename `getCascadedValueExtractor` into `getValueExtractor` as it's not used for cascading only anymore.

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedExecutable.java
```diff
@@ -295,15 +286,6 @@ public ConstrainedExecutable merge(ConstrainedExecutable other) {
 		List<TypeVariable<?>> mergedCascadingTypeParameters = new ArrayList<>( cascadingTypeParameters );
 		mergedCascadingTypeParameters.addAll( other.cascadingTypeParameters );
 
-		// TODO - Is this the right way of handling the merge of unwrapMode? (HF)
```
gunnarmorling: I like that we get rid of things like this :)

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/JavaFXPropertyValueUnwrapperTest.java
```diff
@@ -40,10 +38,11 @@
...
 
 /**
- * Tests for {@link org.hibernate.validator.internal.engine.valuehandling.JavaFXPropertyValueUnwrapper}.
+ * Tests for {@link ObservableValueExtractor}.
  *
  * @author Khalid Alqinyah
  * @author Hardy Ferentschik
+ * @author Guillaume Smet
  */
 public class JavaFXPropertyValueUnwrapperTest {
```
gunnarmorling: Should be renamed then, too.

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/UnwrapValidatedValueTest.java
```diff
@@ -118,87 +113,6 @@ public void shouldRaiseExceptionIfNoMatchingUnwrapperIsFound() {
 		validator.validate( new Order() );
 	}
 
-	@Test
```
gunnarmorling: Why are these gone?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -147,7 +143,7 @@
 	 */
 	private final TypeResolutionHelper typeResolutionHelper;
```
gunnarmorling: This seems unused now.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ObservableValueExtractor.java
gsmet: No, it can't. I remember trying to remove it but animal-sniffer complains as JavaFX is not part of the Java 8 signatures.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
gsmet: Done

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorDescriptor.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gsmet: Good catch. Removed.

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/JavaFXPropertyValueUnwrapperTest.java
gsmet: Done.

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/UnwrapValidatedValueTest.java
gsmet: `.unwrapValidatedValue()` is gone and everything is managed in the constraint definitions.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/ConstraintMappingContextImplBase.java
gsmet: Removed.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraints.java
gsmet: Done.

gsmet:
@gunnarmorling I just pushed a version addressing all your comments.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -747,7 +671,7 @@ private void validateCascadedValues(Object value, ValidationContext<?> context,
 			);
 
 			for ( TypeVariable<?> cascadingTypeParameterOfValueType : cascadingTypeParametersOfValueType ) {
-				ValueExtractorDescriptor extractor = valueExtractors.getCascadedValueExtractor(
+				ValueExtractorDescriptor extractor = valueExtractorManager.getTypeParameterValueExtractor(
```
gunnarmorling: Why not just `getValueExtractor()`? It's used for type parameter constraints and cascaded validation.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
```diff
@@ -172,7 +175,6 @@ private void throwExceptionForNullValidator(Type validatedValueType, String path
 
 	private <T> ConstraintValidator<A, ?> getInitializedConstraintValidator(ValidationContext<T> validationContext,
 			ValueContext<?, ?> valueContext) {
-		Type validatedValueType = valueContext.getDeclaredTypeOfValidatedElement();
```
gunnarmorling: Nice you could do it :)

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/LegacyOptionalValueExtractor.java
```diff
@@ -17,6 +18,7 @@
  * @author Gunnar Morling
  */
 // TODO should we keep that, or only support {@code Optional<@Valid Foo>}.
+@UnwrapByDefault
```
gunnarmorling: Why is that actually? As `Optional` has a type parameter, constraints applying to the wrapped value should be given on the type argument, hence no need for default-unwrapping for element-level constraints. Same for the non-legacy extractor.

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/UnwrapValidatedValueTest.java
gunnarmorling: Ok, but shouldn't we have a test for specifying the option (via the API) on the constraint level? Or are there other tests for that?

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/LegacyOptionalValueExtractor.java
gsmet: Removed.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gsmet: Renamed.

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/UnwrapValidatedValueTest.java
gsmet: I added some new tests to test the programmatic configuration with payload.

gsmet:
@gunnarmorling updated to use payload and fixed all your comments.

PR #627, opened 2017-01-13
HV-1201 Allow to specify a field for reporting constraint violations from @ScriptAssert

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1201

added reportOn annotation parameter to ScriptAssert. Also updated annotation processor to check for this new parameter.

### DIFF
--- a/documentation/src/main/asciidoc/ch02.asciidoc
+++ b/documentation/src/main/asciidoc/ch02.asciidoc
@@ -750,16 +750,20 @@ optionally allowed attributes using the annotation `@SafeHtml.Tag`.
 
 |`@ScriptAssert(lang=, +
               script=, +
-              alias=)`
+              alias=, +
+              reportOn=)`
 |Any type
 |Checks whether the given script can successfully be
 evaluated against the annotated element. In order to use this
 constraint, an implementation of the Java Scripting API as
 defined by JSR 223 ("Scripting for the
-Java^TM^ Platform") must part of the
+Java^TM^ Platform") must be a part of the
 class path. The expressions to be evaluated can be written in
 any scripting or expression language, for which a JSR 223
-compatible engine can be found in the class path.
+compatible engine can be found in the class path. Even though
+this is a type level constraint, one can use `reportOn` attribute
+to be able to report a constraint violation on a specified property
+rather than the whole object.
 |None
 
 |`@URL(protocol=, +
--- a/engine/src/main/java/org/hibernate/validator/cfg/defs/ScriptAssertDef.java
+++ b/engine/src/main/java/org/hibernate/validator/cfg/defs/ScriptAssertDef.java
@@ -32,4 +32,9 @@ public class ScriptAssertDef extends ConstraintDef<ScriptAssertDef, ScriptAssert
 		addParameter( "alias", alias );
 		return this;
 	}
+
+	public ScriptAssertDef reportOn(String reportOn) {
+		addParameter( "reportOn", reportOn );
+		return this;
+	}
 }
--- a/engine/src/main/java/org/hibernate/validator/constraints/ScriptAssert.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/ScriptAssert.java
@@ -111,6 +111,15 @@ public @interface ScriptAssert {
 	String alias() default "_this";
 
 	/**
+	 * @return The name of the property for which you would like to report a validation error.
+	 * If given, the resulting constraint violation will be reported on the specified property.
+	 * If not given, the constraint violation will be reported on the annotated bean.
+	 *
+	 * @since 5.4
+	 */
+	String reportOn() default "";
+
+	/**
 	 * Defines several {@code @ScriptAssert} annotations on the same element.
 	 */
 	@Target({ TYPE })
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ScriptAssertValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ScriptAssertValidator.java
@@ -6,13 +6,13 @@
  */
 package org.hibernate.validator.internal.constraintvalidators.hv;
 
+import static org.hibernate.validator.internal.util.logging.Messages.MESSAGES;
+
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
 
 import org.hibernate.validator.constraints.ScriptAssert;
 import org.hibernate.validator.internal.util.Contracts;
-
-import static org.hibernate.validator.internal.util.logging.Messages.MESSAGES;
 
 /**
  * Validator for the {@link ScriptAssert} constraint annotation.
@@ -20,10 +20,13 @@ import static org.hibernate.validator.internal.util.logging.Messages.MESSAGES;
  * @author Gunnar Morling
  * @author Hardy Ferentschik
  * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
+ * @author Marko Bekhta
  */
 public class ScriptAssertValidator implements ConstraintValidator<ScriptAssert, Object> {
 
 	private String alias;
+	private String reportOn;
+	private String message;
 	private ScriptAssertContext scriptAssertContext;
 
 	@Override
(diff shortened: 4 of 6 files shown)


annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationParametersScriptAssertCheck.java
```diff
@@ -23,25 +24,49 @@
...
+			for ( Element member : elementUtils.getAllMembers( (TypeElement) element ) ) {
+				if ( !member.getKind().isField() ) {
+					continue;
+				}
+				if ( member.getSimpleName().toString().equals( reportOn ) ) {
+					hasReportOnField = true;
+				}
+			}
+			if ( !hasReportOnField ) {
+				issues.add(
```
gunnarmorling: It's a good consideration to check presence of the field, but actually it's not needed. `getPropertyNode()` in the node builder API doesn't mandate that the given property actually exists, it's intended to pass "fake fields", e.g. for map keys. So we shouldn't enforce this here.

engine/src/main/java/org/hibernate/validator/constraints/ScriptAssert.java
```diff
@@ -111,6 +111,11 @@
 	String alias() default "_this";
 
 	/**
+	 * @return The name of the field on which you would like to report a validation error for.
```
gunnarmorling: "field" = "property". In general, BV abstracts from fields/getters and exposes them as properties in a unified manner. Also the default behavior should be mentioned. Something along these lines:

> If given, the resulting constraint violation will be reported on the specified property. If not given, the constraint violation will be reported on the annotated bean.

Ah, and could you add "@since 5.4"?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ScriptAssertValidator.java
```diff
@@ -6,37 +6,61 @@
...
  * Validator for the {@link ScriptAssert} constraint annotation.
  *
  * @author Gunnar Morling
  * @author Hardy Ferentschik
  * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
+ * @author Marko Bekhta
  */
 public class ScriptAssertValidator implements ConstraintValidator<ScriptAssert, Object> {
 
+	private static final Log log = LoggerFactory.make();
```
gunnarmorling: log=LOG (I know we have it lowercase in many places, but in new code we tend to use the proper casing for constants).

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ScriptAssertValidator.java
```diff
@@ -6,37 +6,61 @@
...
+		this.message = constraintAnnotation.message();
 		this.scriptAssertContext = new ScriptAssertContext( constraintAnnotation.lang(), constraintAnnotation.script() );
 	}
 
 	@Override
 	public boolean isValid(Object value, ConstraintValidatorContext constraintValidatorContext) {
-		return scriptAssertContext.evaluateScriptAssertExpression( value, alias );
+		boolean validationResult = scriptAssertContext.evaluateScriptAssertExpression( value, alias );
+
+		if ( !reportOn.isEmpty() ) {
```
gunnarmorling: See my comment above, we shouldn't check for the presence of the field or property getter.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/ScriptAssertValidatorTest.java
```diff
@@ -135,9 +175,21 @@ public void scriptExpressionReturningNoBooleanRaisesException() throws Exception
...
 	 */
 	private ConstraintValidator<ScriptAssert, Object> getInitializedValidator(String lang, String script) {
+		return getInitializedValidator( lang, script, null );
+	}
+
+	/**
+	 * @param lang the script type
+	 * @param script the actual script
+	 * @param alias the alias name of the this object
+	 * @param reportOn a filed name on which to report an error
```
gunnarmorling: filed=property. Though you could omit the param tags altogether, they don't provide any further insight.

marko-bekhta:
@gunnarmorling thanks for the review! I'll make corresponding changes tomorrow.

annotation-processor/src/main/java/org/hibernate/validator/ap/checks/annotationparameters/AnnotationParametersScriptAssertCheck.java
marko-bekhta: So maybe make it as a warning to let users know that they might mistyped a property name? or just remove it ?

gunnarmorling:
Let's remove it. We cannot really know whether it's a typo or a "virtual"
property on purpose.

engine/src/main/java/org/hibernate/validator/constraints/ScriptAssert.java
```diff
@@ -111,6 +111,14 @@
 	String alias() default "_this";
 
 	/**
+	 * @since 5.4
+	 * @return The name of the property on which you would like to report a validation error for.
```
gsmet: - s/on which/for which/
- remove the for at the end of the sentence
- the `@since 5.4` should be after the `@return`

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/ScriptAssertValidatorTest.java
```diff
@@ -113,6 +118,35 @@ public void scriptExpressionReturningNoBooleanRaisesException() throws Exception
 		validator.isValid( new Object(), null );
 	}
 
+	@Test
+	public void reportOnField() {
```
gsmet: Please add the `@TestForIssue` annotation. It's less important than for bugs but it's a good habit.

documentation/src/main/asciidoc/ch02.asciidoc
```diff
@@ -750,7 +750,8 @@ optionally allowed attributes using the annotation `@SafeHtml.Tag`.
 
 |`@ScriptAssert(lang=, +
               script=, +
-              alias=)`
+              alias=, +
+              reportOn=)`
```
gsmet: I think it would be nice to add a sentence about thre `reportOn` option in the description below.

engine/src/main/java/org/hibernate/validator/cfg/defs/ScriptAssertDef.java
```diff
@@ -32,4 +32,9 @@ public ScriptAssertDef alias(String alias) {
 		addParameter( "alias", alias );
 		return this;
 	}
+
+	public ScriptAssertDef reportOn(String reportOn) {
+		addParameter( "reportOn", reportOn );
+		return this;
+	}
```
gsmet: Is the `ScriptAssertDef` class tested somewhere? If so, it would be nice to also test the `reportOn` option.

marko-bekhta:
@gsmet thanks for review! I'll make corresponding changes.

marko-bekhta:
yes @gunnarmorling , I've made all the changes, and added a test for the ScriptAssertDef as I couldn't find one. So I'd say it's good to go :)

gunnarmorling:
Ok, cool. I'll take another look and merge.

documentation/src/main/asciidoc/ch02.asciidoc
```diff
@@ -750,16 +750,20 @@ optionally allowed attributes using the annotation `@SafeHtml.Tag`.
...
 |Any type
 |Checks whether the given script can successfully be
 evaluated against the annotated element. In order to use this
 constraint, an implementation of the Java Scripting API as
 defined by JSR 223 ("Scripting for the
-Java^TM^ Platform") must part of the
+Java^TM^ Platform") must be a part of the
```
gsmet: You see, you also start to see typos everywhere ;)

gunnarmorling:
Rebased and applied. Added one more commit for some clean-up, most notably using the common assertion methods on `ConstraintViolationAssert`. Thanks, @marko-bekhta!

gunnarmorling:
@marko-bekhta I've applied it to 5.4, too. It's a bit against the spririt of CRs, as no features should be added after a CR has been put out. But this feature is very insulated, so I don't think there's much risk to it and it's nice to have it out.

PR #628, opened 2017-01-13
HV-1204 Supporting to put @Currency constraint via API

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1204

I noticed btw. we are missing some more constraint def types (e.g. for all the country specific constraints). Some review is due. It's no blocker though as there is the generic constraint def.


gsmet:
Fixed the JIRA issue number in the commit message and merged, thanks.

PR #629, opened 2017-01-15
HV-1218 added missing programmatic constraint definitions

marko-bekhta:
This PR is based on https://github.com/hibernate/hibernate-validator/pull/628
I've added all missing Defs if that's needed :)

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/cfg/defs/LuhnCheckDef.java
+++ b/engine/src/main/java/org/hibernate/validator/cfg/defs/LuhnCheckDef.java
@@ -1,0 +1,41 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.cfg.defs;
+
+import org.hibernate.validator.cfg.ConstraintDef;
+import org.hibernate.validator.constraints.LuhnCheck;
+
+/**
+ * @author Marko Bekta
+ */
+public class LuhnCheckDef extends ConstraintDef<LuhnCheckDef, LuhnCheck> {
+
+	public LuhnCheckDef() {
+		super( LuhnCheck.class );
+	}
+
+	public LuhnCheckDef startIndex(int index) {
+		addParameter( "startIndex", index );
+		return this;
+	}
+
+	public LuhnCheckDef endIndex(int index) {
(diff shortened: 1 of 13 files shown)


gsmet:
@marko-bekhta thanks! Could you create a global JIRA for that with a fixVersion to 5.4-next?

It would be nice if you could add some tests similar to what @gunnarmorling did in #628 to exercise the parameter handling.

marko-bekhta:
@gsmet yes, I'll do that.

gunnarmorling:
Nice. Thinking of it, would have been a nice example for a small annotation processor to generate the def types based on the actual constraints :)

gsmet:
@marko-bekhta while we're at it, could you add a package.html in the pl specific constraints package?

marko-bekhta:
@gunnarmorling, @gsmet  I think this one is not very urgent right ? if that's so I'd like to try the processor approach to cover generation of all these Defs. Otherwise I can finish this one - add a few tests for each Def and add a package file that @gsmet mentioned. And after that look into annotation processor Def generation. WDYT ?

gunnarmorling:
Hey, it's not very urgent, but I'm not sure whether an AP really is worth it. It's not that we add new constraints every day. But then, such processor might quickly be implemented. One problem is that we need the defs for the BV types which are not subject to compilation within HV.

I'll leave it to your judgement which route to pursue, but probably the AP will never pay off if it takes more than a few hours to write it.

gsmet:
Although it's not critical, I would like to have it for our next 5.4 so I think for now the manual approach is good enough (and you already did most of the work). We could pursue the annotation processor approach as a parallel pet project.

marko-bekhta:
I've added a few tests and missing `package.html`  and squashed everything together. Later I'll look into annotation processor option as well, as for me it'll be an interesting thing to implement  :)

gsmet:
@marko-bekhta so I merged it on master but, while backporting it on 5.4, I found a bug that one of your tests reveals.

I'll fix the bug then merge on 5.4.

marko-bekhta:
@gsmet ok, great! And what was the bug ?

gsmet:
@marko-bekhta this one: https://hibernate.atlassian.net/browse/HV-1220

gsmet:
@marko-bekhta and here is the corresponding PR: https://github.com/hibernate/hibernate-validator/pull/640/files

It's normal you didn't see it as it was coincidentally fixed in master. Nice you triggered it by accident, very few chances we would have found it otherwise.

marko-bekhta:
@gsmet Thanks! I was just looking at that PR :)

PR #630, opened 2017-01-16
HV-1205 ("Don't use @Repeatable on some constraints")

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1205

The second commit should be merged to master, too.


gsmet:
@gunnarmorling surprising the build does not fail.

gunnarmorling:
We have AnimalSniffer set up on 5.4, but for a reason that I don't remember it's not run as part of the default build. But it seems it wouldn't even have caught that one, I just tried it out. Maybe it's not working with annotations or something?

PR #631, opened 2017-01-16
HV-1206 Enforcing that @ExtractedValue is given exactly once per extr…

gunnarmorling:
…actor type

https://hibernate.atlassian.net/browse/HV-1206


gsmet:
Merged, thanks!

PR #632, opened 2017-01-17
Switch to Asciidoctor for PDF and epub generation

gsmet:
If someone wants to play with it:
```
cd documentation
mvn clean install
```

I also tuned the HTML output to be more practical (especially enabled section anchor and the side toc).

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -93,12 +93,64 @@
                             </attributes>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>output-pdf</id>
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>process-asciidoc</goal>
+                        </goals>
+                        <configuration>
+                            <backend>pdf</backend>
+                            <outputDirectory>${basedir}/target/asciidoc-pdf</outputDirectory>
+                            <sourceHighlighter>coderay</sourceHighlighter>
+                            <attributes>
+                                <sourcedir>${basedir}/src/test/java</sourcedir>
+                                <engine-sourcedir>${basedir}/../engine/src/main/java</engine-sourcedir>
+                                <icons>font</icons>
+                                <pagenums/>
+                                <toc/>
+                                <idprefix/>
+                                <idseparator>-</idseparator>
+                            </attributes>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>output-epub</id>
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>process-asciidoc</goal>
+                        </goals>
+                        <configuration>
+                            <backend>epub3</backend>
+                            <outputDirectory>${basedir}/target/asciidoc-epub</outputDirectory>
+                            <sourceHighlighter>coderay</sourceHighlighter>
+                            <attributes>
+                                <sourcedir>${basedir}/src/test/java</sourcedir>
+                                <engine-sourcedir>${basedir}/../engine/src/main/java</engine-sourcedir>
+                                <icons>font</icons>
+                                <pagenums/>
+                                <toc/>
+                                <idprefix/>
+                                <idseparator>-</idseparator>
+                            </attributes>
+                        </configuration>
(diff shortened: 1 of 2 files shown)


Sanne:
looks great! and epub too :+1:

documentation/pom.xml
```diff
@@ -125,108 +177,4 @@
...
-                            </formats>
-                            <options>
-                                <xincludeSupported>true</xincludeSupported>
-                                <xmlTransformerType>saxon</xmlTransformerType>
-                                <!-- needed for uri-resolvers; can be ommitted if using 'current' uri scheme -->
-                                <!-- could also locate the docbook dependency and inspect its version... -->
-                                <!-- docbookVersion>1.72.0</docbookVersion -->
-                                <localeSeparator>-</localeSeparator>
-                                <injectionDateFormat>yyyy-MM-dd</injectionDateFormat>
-                                <transformerParameters>
```
gunnarmorling: Is the analytics snippet still contained in the rendered output?

gunnarmorling: Emmanuel told me that @ge0ffrey knows how to do it in case we need a pointer.

ge0ffrey: It's easy in asciidoc:
1. Add a docinfo footer file:
  https://github.com/droolsjbpm/optaplanner/blob/master/optaplanner-docs/src/main/asciidoc/index-docinfo-footer.html
2. Set these in the attributes element
  https://github.com/droolsjbpm/optaplanner/blob/master/optaplanner-docs/pom.xml#L50

yrodiere:
Gah. I replied to your email before I saw there was a PR. Copying my comments here:

Overall it looks great, very clean. There are a few really strange issues, though. If the previous PDF wasn't perfect either, I agree we can ship the PDF as is, but otherwise we may want to look at it a bit...

The main issues:
 * Borders for code examples seem to generate strange artifacts when there is a page break just before. See for instance the bottom of pages 55, 62, 97 and 112.
 * Is there any way to tune page breaks? Or to tune the font size for code examples so that they fit better? Sometimes its a bit weird:
   * Pages 18-19, I would have put the page break just before the "2.1.5" title
   * Page 97, I think scaling down the font size just a bit would make the code example fit inside the page.

The nitpicking:

 * I find the margins aggressively small, but maybe it's just me
 * Some images seem way too big to fit one page: I noticed page 96, but maybe there are others.

ge0ffrey:
"Some images seem way too big to fit one page". There's a setting to tweak this, see:
  https://github.com/asciidoctor/asciidoctor-pdf#image-scaling
but the default setting shouldn't do this. The weird thing is that in optaplanner-docs, I have plenty of big images that get scaled down to fit nicely, but in another project (same asciidoc versions, some config etc), I couldn't get the big images to scale down to fit, even with playing with that param. This problem entirely eluded me. If you find out how to fix it, let me know pls.

yrodiere:
@gsmet Actually, I just had the look at the [Docbook output](https://docs.jboss.org/hibernate/validator/5.4/reference/en-US/pdf/hibernate_validator_reference.pdf), and there are really strange things too, such as page breaks right in the middle of code examples (p. 76) or images overflowing out of the page (p. 85).

So yes, you're right, it seems better with asciidoc anyway :) We can always merge as is, and tune it later.

documentation/pom.xml
gunnarmorling: Cool; thanks, @ge0ffrey!

gsmet:
@gunnarmorling yes, it was my plan to move the pdf + epub generation to a profile as for the docbook output.

gunnarmorling:
Ok, cool. You also could add something like <skip>${skip.pdf}</skip> to the
execution configuration block.

gsmet:
Looks like we all agree it's the way to go and the output is good enough. I'll pursue this as a pet project.

gunnarmorling:
Nice. I'll close it for now so to keep the list of in-flight PRs lower. Looking forward to the final version of it. Would great to have it in Alpha1.

PR #633, opened 2017-01-18
HV-1210 Upgrade Groovy to 2.4.8

gsmet:
* https://hibernate.atlassian.net/browse/HV-1210

It should be backported to 5.4.


gunnarmorling:
Merged and backported to 5.4 as well as 5.3. Thanks, @gsmet!

PR #634, opened 2017-01-18
HV-1211 Remove empty ConstraintValidator#initialize implementations

gsmet:
It's not mandatory but let's show the good example in our code.

 * https://hibernate.atlassian.net/browse/HV-1211


gunnarmorling:

PR #636, opened 2017-01-18
HV-1214 Switch to Asciidoctor for the documentation output

gsmet:
* https://hibernate.atlassian.net/browse/HV-1214

I spend some time on it this evening. I think it's ready but it would be nice if someone could take a look at it.

To generate all the outputs:
```
cd documentation
mvn -Pdocumentation-pdf clean install
```

We might want to tune the style of the PDF a little in the future but I think it's already a good basis.

And, to be honest, I refrained from tuning the output too much to limit the maintenance burden in the future.

I created https://hibernate.atlassian.net/browse/HV-1215 as I'll have to adapt a little the release scripts to take this into account (mostly preventing the release scripts to add meta content to the HTML doc as it's now already included).


documentation/pom.xml
```diff
@@ -22,7 +22,15 @@
     <description>The Hibernate Validator reference manual</description>
 
     <properties>
-        <jdocbook.ignoreTranslations>false</jdocbook.ignoreTranslations>
+        <asciidoctor.base-outputdir>${basedir}/target/asciidoctor</asciidoctor.base-outputdir>
+        <!--
+        With the asciidoctor Maven plugin 1.6, we are going to put back the images into the src/main/asciidoc
```
gunnarmorling: OOI, what's the issue here that's fixed in that new version? A link to the issue in their tracker would be nice.

documentation/src/main/resources/docinfo/index-docinfo-footer.html
```diff
@@ -0,0 +1,16 @@
...
+	(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+		new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+		j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='//www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+	})(window,document,'script','dataLayer','{html.google-tag-manager-id}');
+</script>
+<noscript><iframe src="//www.googletagmanager.com/ns.html?id={html.google-tag-manager-id}" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+<!-- /Google Analytics -->
+
+<!-- HibernateDoc.OutdatedContent -->
+<script src="//code.jquery.com/jquery-3.1.0.min.js" integrity="sha256-cCueBR6CsyA4/9szpPfrX3s49M9vUU5BgtiJj06wt/s=" crossorigin="anonymous"></script>
```
gunnarmorling: That URL looks weird. Should be https://code.jquery.com/jquery-3.1.0.min.js I reckon?

gunnarmorling: (though I personally dislike not getting it from our own server to begin with, but arguable there are pros/cons for both approaches)

gsmet: It's a protocol relative URL. Meaning the protocol adapts depending on the current protocol of the page.

gunnarmorling:
Hum, ok. What's the benefit of that, why not always go for HTTPS?

documentation/pom.xml
gsmet: See the commented part with resources. Asciidoctor currently copies all the resources from src/main/asciidoc so you end up with all the images/css/docinfo files, even in the PDF or epub case. 1.6 provides a resources option allowing to filter the elements copied.

I'll make that more clear.

gsmet:
> The PDF still has some oddities, (e.g. page 15, 98, or on page 45 the first one or two listings may appear before the page break), but I think it's definitely good enough.

Unfortunately, apart from opening an issue, I don't think there's much I can do about fixing it. That being said, as mentioned earlier by Yoann, our current PDF output is far from being perfect either.

> Maybe we can find a "less styled" font for listings in the PDF, too? Something more neutral would be nice, but for sure no blocker either.

OK, so you have the same problem I have. I was wondering if maybe I was missing a common font. I agree it's something we should fix.

> The epub version also has some broken image references (e.g. next to the author name at the beginning of each chapter, maybe we can just get rid of this; but also the XML schema). I wouldn't spend much time on tuning epub output. It's nice to have, but not really needed, so I'd rather omit it if it takes to much effort to fix.

I'll check that. Maybe it's too alpha for us.

gsmet:
Jenkins, retest this please

gsmet:
@gunnarmorling I modified the fonts. We now have a specific theme which is mostly the default with a few variables changed. I also put the default in the theme directory so that we can easily compare with the reference if we need to update it one day.

I removed the epub output. It's definitely not in a good shape. I didn't have an epub viewer to test it but now that I did, it's far from being ready for prime time.

I also changed the comments.

documentation/src/main/resources/docinfo/index-docinfo-footer.html
gsmet: > Hum, ok. What's the benefit of that, why not always go for HTTPS?

To be honest, I also use that to be sure the JS is not loaded when the user is browsing it locally. If he has downloaded the doc of a specific version, I'm pretty sure he doesn't want to be notified that it's an old one.

gsmet:
@gunnarmorling the release scripts are ready and so is the outdated content script on docs.jboss.org.

The only thing left will be to remove the link for the multi page html output on the website.

documentation/src/main/resources/theme/default-theme.yml
```diff
@@ -0,0 +1,271 @@
+font:
```
gunnarmorling: Can you add a reference to the original source of this (GitHub link). And instead having the two files side by side, how about adding the original one verbatim in one commit and then adapt it as per our needs in a separate commit. That's how we handled other cases where started from a copied resource.

pom.xml
```diff
@@ -153,6 +153,12 @@
 
         <groovy.version>2.4.8</groovy.version>
 
+        <asciidoctor-maven-plugin.version>1.5.3</asciidoctor-maven-plugin.version>
+        <jruby.version>1.7.26</jruby.version>
+        <asciidoctorj.version>1.5.4.1</asciidoctorj.version>
+        <asciidoctorj-pdf.version>1.5.0-alpha.11</asciidoctorj-pdf.version>
+        <asciidoctorj-epub3.version>1.5.0-alpha.6</asciidoctorj-epub3.version>
```
gunnarmorling: Seems not needed anymore?

documentation/src/main/resources/theme/default-theme.yml
gsmet: Because, if I want to update it and base it on a new default-theme.yml file, it's going to be messy. This way, it's really easy to keep track of the changes.

I did that a lot in ${previous-job} and it's really more practical like that.

I'll add the reference to the GitHub file at the top of our theme but I prefer to keep the original file too to be on the safe side.

pom.xml
gsmet: Yeah, it does not do any harm so I let it there as this way, we can enable it quickly to do more tests if we want to.

documentation/src/main/resources/theme/default-theme.yml
gunnarmorling: Ok, works for me, too.

gsmet: Done.

gunnarmorling:
Rebased and applied. Thanks, @gsmet! Can you merge it to 5.4, too?

gsmet:
Merged to 5.4.

PR #638, opened 2017-01-19
HV-1212 Removing AnimalSniffer set-up;

gunnarmorling:
It's not needed anymore as we are requiring Java 8 at runtime, too.
In the future, where we may require Java 9 for building but still want to
keep runtime compatability with Java 8, the signatures built into the JDK
itself (added in Java 9) can be used.

https://hibernate.atlassian.net/browse/HV-1212


gsmet:
Merged, thanks!

PR #640, opened 2017-01-20
HV-1220 Make programmatically defined cross parameter constraints on the parameters of a method returning void work as expected

gsmet:
* https://hibernate.atlassian.net/browse/HV-1220

I'll check if a backport to 5.3 is necessary as soon as this one is in. And I'll forward port the test to master.


gsmet:
Pushed an update to rename `forExecutable` to `forReturnValue`.

gunnarmorling:
Rebased and applied. Thanks for spotting and fixing, @marko-bekhta / @gsmet!

On the backporting, let's apply it to 5.3. No need for an eager release, though; it's a very special case and not user-reported.

PR #642, opened 2017-01-24
HV-1223 Setting up jqAssistant

gunnarmorling:
@gsmet That sets up the tooling we discussed.

Run it like this:

     mvn jqassistant:server -pl engine -Pjqassistant

For running the server, allowing to explore the project structure interactively (go to http://localhost:7474/browser/)

Run this to run the rule check:

    mvn clean install -pl engine -Pjqassistant -DskipTests=true

E.g. create a public constant exposing an internal type to see it in action.

I've put it into a separate profile as it takes some build time, but we can have job on CI executing it.

### DIFF
--- a/jqassistant/rules.xml
+++ b/jqassistant/rules.xml
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jqa:jqassistant-rules xmlns:jqa="http://www.buschmais.com/jqassistant/core/analysis/rules/schema/v1.0">
+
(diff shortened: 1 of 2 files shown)


gsmet:
@gunnarmorling looks good and a nice thing to have.

But, if we set up a CI job, shouldn't we set failOnViolations to true?

gunnarmorling:
> But, if we set up a CI job, shouldn't we set failOnViolations to true?

Yes, that's true. I've pushed one more commit for this.

gsmet:
Rebased, merged and Jenkins job created. Thanks!

PR #643, opened 2017-01-24
added check for Currency units in annotation processor

marko-bekhta:
Not sure if this one will be of any value. But I put it together while thinking about those domain validation changes (for Url validation) so I thought I'll share this with you, and you'll decide if it's needed or not :)

### DIFF
--- a/annotation-processor/pom.xml
+++ b/annotation-processor/pom.xml
@@ -40,6 +40,17 @@
             <artifactId>joda-time</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.javamoney</groupId>
+            <artifactId>moneta</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>javax.money</groupId>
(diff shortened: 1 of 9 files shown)


gsmet:
Hi Marko,

Looks nice but I'm not very fond of us maintaining a list of currencies in the long time. I agree it does not change a lot but it does change. i'm also wondering if it may be possible to register other currencies in Moneta, which would make null and void the whole approach.

So, I'm -1 on this.

@gunnarmorling ?

gunnarmorling:
Yes, hard-coding this list of currencies seems not desirable. In fact, javax.money defines an SPI for providing currencies: `CurrencyProviderSpi`. I.e. the set of available currencies really depends on the context of the specific application and registered currency providers.

One thing we could consider is to obtain the list of currencies exposed by the JDK itself (check out `java.util.Currency.getAvailableCurrencies()`, but it still wouldn't accomodate for the case of a custom currency provider implementation.

All in all I think that's not something we should validate.

gsmet:
Agreed. Let's close it. Thanks for your efforts, Marko!

PR #647, opened 2017-01-30
HV-1216 Avoid usage of tables in reference docs where feasible

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1216

migrated tables to labeled lists in documentation. Resulting pdf file is attached to the JIRA.


gunnarmorling:
Thanks, @marko-bekhta. I've added another commit, mainly for getting rid of the "uses" sub-definition and pulling the actual element description right at the beginning. I'll merge that one as soon as the build has finished. Just let me know in case you see any further issues. Thanks again, it's much better for the PDF now.

marko-bekhta:
great! Happy to help :)

PR #648, opened 2017-01-31
HV-1224 Update the groupId in the Karaf features file

gsmet:
* https://hibernate.atlassian.net/browse/HV-1224 (follow-up)
 * https://hibernate.atlassian.net/browse/HV-1229

The important part is the first commit fixing the groupId in the Karaf features.xml file.

I upgraded the dependencies while trying to diagnose the issue so I thought I might as well push it.


gunnarmorling:
Looks good. Thanks, @gsmet!

PR #652, opened 2017-02-01
HV-966 ParameterNameProvider is called on methods with no parameter as well as Object methods

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-966

So obviously that approach that I've mentioned in JIRA didn't work out :). The reason for that is that if we exit from validation that fast some of the checks will never be performed and no exceptions will be thrown. For example:
```
Log#getMethodOrConstructorNotDefinedByValidatedTypeException
Log#getInvalidParameterCountForExecutableException // for the case when no parameters are passed
```

So the place where I found that method call could be skipped is in `ExecutableParameterNameProvider`. 
I also made some other changes related to this - I'll comment inline to explain why the change was made. 

Another thing that came to my mind is that maybe we should throw an exception (`IllegalArgumentException`) if a parameterless method is passed for parameter method validation, WDYT?

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/cfg/context/ExecutableConstraintMappingContextImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/cfg/context/ExecutableConstraintMappingContextImpl.java
@@ -125,6 +125,7 @@ abstract class ExecutableConstraintMappingContextImpl {
 	private List<ConstrainedParameter> getParameters(ConstraintHelper constraintHelper, TypeResolutionHelper typeResolutionHelper,
 			ExecutableParameterNameProvider parameterNameProvider, ValueExtractorManager valueExtractorManager) {
 		List<ConstrainedParameter> constrainedParameters = newArrayList();
+		List<String> parameterNames = parameterNameProvider.getParameterNames( executable );
 
 		for ( int i = 0; i < parameterContexts.length; i++ ) {
 			ParameterConstraintMappingContextImpl parameter = parameterContexts[i];
@@ -138,7 +139,7 @@ abstract class ExecutableConstraintMappingContextImpl {
 								executable,
 								ReflectionHelper.typeOf( executable, i ),
 								i,
-								parameterNameProvider.getParameterNames( executable ).get( i )
+								parameterNames.get( i )
 						)
 				);
 			}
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/hibernate/validator/internal/cfg/context/ExecutableConstraintMappingContextImpl.java
```diff
@@ -125,6 +125,7 @@ public ConstrainedElement build(ConstraintHelper constraintHelper, TypeResolutio
 	private List<ConstrainedParameter> getParameters(ConstraintHelper constraintHelper, TypeResolutionHelper typeResolutionHelper,
 			ExecutableParameterNameProvider parameterNameProvider, ValueExtractorManager valueExtractorManager) {
 		List<ConstrainedParameter> constrainedParameters = newArrayList();
+		List<String> parameterNames = parameterNameProvider.getParameterNames( executable );
```
marko-bekhta: `parameterNameProvider.getParameterNames( executable )` was called inside of the loop so I thought that extracting it out of it would be a bit faster.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -259,11 +259,6 @@ public ValidatorImpl(ConstraintValidatorFactory constraintValidatorFactory,
 	}
 
 	private <T> Set<ConstraintViolation<T>> validateParameters(T object, Executable executable, Object[] parameterValues, Class<?>... groups) {
-		//this might be the case for parameterless methods
-		if ( parameterValues == null ) {
-			return Collections.emptySet();
-		}
-
```
marko-bekhta: This removed if statement seems to be redundant - this private method is called from two places. And in both cases there's 
```
Contracts.assertNotNull( parameterValues, MESSAGES.validatedParameterArrayMustNotBeNull() );
```
So `null` cannot be passed to this method.

gunnarmorling:
Looks good, thanks, @marko-bekhta!

> Another thing that came to my mind is that maybe we should throw an exception (IllegalArgumentException) if a parameterless method is passed for parameter method validation, WDYT?

I don't think we can do this, as the BV spec doesn't forbid passing such method. So we cannot apply stricter rules in the RI.

gunnarmorling:
Rebased, applied and added one more commit for some clean-up/simplification. Thanks, Marko!

marko-bekhta:
Great! Happy to help :)

PR #653, opened 2017-02-02
HV-1233 Convert tabs to space in rendered doc

gsmet:
It allows to have a consistent behavior across all browsers.

 * https://hibernate.atlassian.net/browse/HV-1233

Let's backport it to 5.4 in case we do a bugfix release.


gunnarmorling:
Thanks, that was quick :) Merged to master and 5.4.

PR #655, opened 2017-02-03
Upgrading to BV API 2.0.0.Alpha1

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1236


gsmet:
@gunnarmorling  Hmmm, what's the point of this? We would have to move back to the snapshot as soon as we make a new change to the API.

gunnarmorling:
Yes, it'll be changed again soon. But HV 6.0.0.Alpha1 should be built
against it.

gsmet:
Added the JIRA issue number to the commit message and merged. Thanks!

PR #656, opened 2017-02-06
HV-1238

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1238

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
@@ -201,7 +201,7 @@ public class ConstraintHelper {
 
 		putConstraints( tmpConstraints, Digits.class, DigitsValidatorForCharSequence.class, DigitsValidatorForNumber.class );
 
-		List<Class<? extends ConstraintValidator<Future, ?>>> futureValidators = newArrayList( 11 );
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
```diff
@@ -201,7 +200,7 @@ public ConstraintHelper() {
 
 		putConstraints( tmpConstraints, Digits.class, DigitsValidatorForCharSequence.class, DigitsValidatorForNumber.class );
 
-		List<Class<? extends ConstraintValidator<Future, ?>>> futureValidators = newArrayList( 11 );
+		List<Class<? extends ConstraintValidator<Future, ?>>> futureValidators = new ArrayList<>( 18 );
```
gsmet: With this, you don't avoid the resizing. Please keep the `newArrayList` call and change only the parameter value, it will take care of it.

Same for the other changes below.

gunnarmorling: That's an array list, not a set with a load factor (assuming that's what you have in mind).

I'd prefer to phase out `newArrayList() ` as it provides no advantage any longer as we can rely on diamond. It's different for `newHashSet() `.

gsmet:
Rebased and merged.

PR #658, opened 2017-02-06
HV-1088 Fixing ConstraintValidatorManager for concurrent access

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1088

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
@@ -80,7 +80,7 @@ public class ConstraintValidatorManager {
 
 	/**
 	 * @param validatedValueType the type of the value to be validated. Cannot be {@code null}.
-	 * @param descriptor the constraint descriptor for which to get an initalized constraint validator. Cannot be {@code null}
+	 * @param descriptor the constraint descriptor for which to get an initialized constraint validator. Cannot be {@code null}
 	 * @param constraintFactory constraint factory used to instantiate the constraint validator. Cannot be {@code null}.
 	 * @param <A> the annotation type
 	 *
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Should be backported to 5.4.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
```diff
@@ -94,87 +94,52 @@ public ConstraintValidatorManager(ConstraintValidatorFactory constraintValidator
...
-			ConstraintValidatorFactory constraintFactory,
-			ConstraintValidatorDescriptor<A> validatorDescriptor,
-			ConstraintDescriptor<A> descriptor) {
+	@SuppressWarnings("unchecked")
+	private <A extends Annotation> ConstraintValidator<A, ?> createAndInitializeValidator(Type validatedValueType,
+			ConstraintDescriptorImpl<A> descriptor,
+			ConstraintValidatorFactory constraintFactory) {
+
+		ConstraintValidatorDescriptor<A> validatorDescriptor = findMatchingValidatorDescriptor( descriptor, validatedValueType );
+		ConstraintValidator<A, ?> constraintValidator = null;
```
gsmet: We shouldn't initialize `constraintValidator` here, it's useless. I'll change that and merge.

gsmet:
Rebased and merged. Nice work, thanks!

exbe:
@gunnarmorling  looks like it was never back ported to 5.4.

Is there any plans/effort to make it?

gunnarmorling:
Hey @exbe, we don't have any plans on doing any further 5.x releases at this point in HV, the community project. Should you have a support contract with Red Hat, it's different; in that case please reach out through the usual channels. Thanks a lot for your understanding!

gsmet:
Please let's continue this discussion in the open PR instead of having it buried there: https://github.com/hibernate/hibernate-validator/pull/1014 .

gunnarmorling:
It's our general strategy to support only one release line in community at a given time. Instead of maintaining older versions, we try to focus on adding new features in the current stable line and fixing bugs in there. 5.4 hasn't seen a release in more than a year, and that it's still mentioned on the downloads page is mainly for the reason that it is the last Bean Validation 1.1 compatible implementation (we probably should remove it there).

That said, I'd *highly* recommend the upgrade to HV 6, which does not only come with all the BV 2.0 features, but also many bug fixes applied since 5.4.x and a *much* better performance: http://in.relation.to/2018/03/20/bean-validation-benchmark-rerevisited/. The upgrade will be smooth (no-op really if you solely use the BV spec'ed APIs). Only if you use some of the experimental/incubating features in HV 5.x, you'd have to do some (not too large) adjustments in your code base.

I'll leave it to @gsmet (who took over as the HV project lead) to decide on your PR. Thanks for your efforts in any case.

PR #659, opened 2017-02-06
HV-1237 Nested type parameters and nested cascading

gsmet:
* https://hibernate.atlassian.net/browse/HV-1237


engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
```diff
@@ -0,0 +1,150 @@
...
+			emailAddressMap.map.put( "valid", validEmailAddresses );
+			emailAddressMap.map.put( "invalid", emailAddressesContainingInvalidEmails );
+
+			return emailAddressMap;
+		}
+	}
+
+	private static class CinemaEmailAddresses {
+
+		private final Map<@NotNull @Valid Optional<@NotNull @Valid Cinema>, @Valid List<@Valid EmailAddress>> map = new HashMap<>();
```
gunnarmorling: `Optional<@NotNull ...>` seems not sensible. Also there are some `@Valid` too much I think. This would be more realistic:

    Map<@NotNull Optional<@Valid Cinema>, List<@Valid EmailAddress>> map;

I prefer us to stick to realistic examples so we are sure we are on the right track to solve actual issues (I don't mean to sound harsh I case I do ;).

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
```diff
@@ -0,0 +1,150 @@
...
+		assertCorrectPropertyPaths(
+				constraintViolations,
+				"map[Optional[Cinema[cinema3]]].visitor.name",
+				"map[Optional.empty].<map key>"
+		);
+	}
+
+	private static class EmailAddressMap {
+
+		private final Map<String, @Valid List<@Valid EmailAddress>> map = new HashMap<>();
```
gunnarmorling: Why does it have two `@Valid`?

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
```diff
@@ -0,0 +1,150 @@
...
+		Set<ConstraintViolation<CinemaEmailAddresses>> constraintViolations = validator.validate( CinemaEmailAddresses.validCinemaEmailAddresses() );
+
+		assertNumberOfViolations( constraintViolations, 0 );
+
+		constraintViolations = validator.validate( CinemaEmailAddresses.invalidCinemaEmailAddresses() );
+
+		assertNumberOfViolations( constraintViolations, 2 );
+		assertCorrectPropertyPaths(
+				constraintViolations,
+				"map[Optional[Cinema[cinema2]]].[1].email",
```
gunnarmorling: Ah, the usage of square brackets in the toString() methods was confusing me on first look (thought some sort of node nesting was going on). Maybe worth to have some assertions on the actual path nodes (also their types etc.)

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
```diff
@@ -0,0 +1,150 @@
...
+	public void testNestedValid() {
+		Validator validator = getValidator();
+
+		Set<ConstraintViolation<EmailAddressMap>> constraintViolations = validator.validate( EmailAddressMap.validEmailAddressMap() );
+
+		assertNumberOfViolations( constraintViolations, 0 );
+
+		constraintViolations = validator.validate( EmailAddressMap.invalidEmailAddressMap() );
+
+		assertNumberOfViolations( constraintViolations, 2 );
```
gunnarmorling: I found that the assertion on the number often is not helpful. Let's say it's 1 actually, you don't know which one is missing. Whereas if you just omit this assertion, you'll see the actual paths in the next statement.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
```diff
@@ -0,0 +1,150 @@
...
+		assertCorrectPropertyPaths(
+				constraintViolations,
+				"map[invalid].[1].email",
```
gunnarmorling: There is an empty-named node in here (after [1], before email). Is that intentional? Let's assert on the actual paths to have more explicitness of what we expect.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
```diff
@@ -26,11 +26,20 @@
...
+	private final Type containerType;
+	private final ConstraintLocation outterDelegate;
 
 	TypeArgumentConstraintLocation(ConstraintLocation delegate, TypeVariable<?> typeParameter, Type typeOfAnnotatedElement) {
 		this.delegate = delegate;
 		this.typeParameter = typeParameter;
 		this.typeForValidatorResolution = ReflectionHelper.boxedType( typeOfAnnotatedElement );
+		this.containerType = delegate.getTypeForValidatorResolution();
+
+		ConstraintLocation outterDelegate = delegate;
```
gunnarmorling: outter=outer

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
```diff
@@ -46,22 +47,22 @@
 	private final ConstraintLocation location;
 
 	/**
-	 * The {@link ValueExtractor} used to extract the value for validation.
+	 * The {@link ValueExtractor}s used to extract the value for validation.
 	 */
-	private final ValueExtractorDescriptor valueExtractorDescriptor;
+	private final List<ValueExtractorDescriptor> valueExtractorDescriptors;
```
gunnarmorling: Can you add an explantion why this is a list now? I reckon it's to navigate from the outermost container to the innermost?

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gsmet: Yeah, I had the same thought lately. I added them mostly to be consistent. I'd also prefer to get rid of them.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
gsmet: Exactly. Seems clear enough then :). I'll add a comment.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gunnarmorling: Also I'm wondering about the types of nodes. Should the intermediary ones be of kind `TYPE_ARGUMENT` instead of `PROPERTY` as it's the case now.

gunnarmorling:
Ok, cool :)

I'm a bit concerned about adding a list also for those many cases where
it's not needed. We may consider to encapsulate it behind dedicated
`ValueExtractorDescriptor` implementations, e.g. there would be a
NestedValueExtractorDescriptor which deals with the nested case, allowing
to use the plain and simple one for the normal case.

But we can leave that as optimization for following.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gsmet: Well, in the current implementation, we need a full chain of `@Valid`. IIRC, we discussed it the other day and you agreed it was acceptable as a first step.

To not require this `@Valid` chain, we would need to continue the `@Valid` discovery on the whole parameterized type and either add a `isCascading` to `CascadingTypeParameter` or rebuild the necessary tree once we made the discovery.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gsmet: To be honest, I don't want to spend too much time in it before we agree on the structure of the path. Let's discuss that.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gunnarmorling: @gsmet if you check out http://staging.beanvalidation.org/latest-draft/spec/#validationapi-constraintviolation, note that key/index/inIterabler are only set on the following node. This may be the cause of confusion.

So I think I'd expect these three nodes here (not totally sure about the kind and name of the 2nd node, but that seems sensible):

    node( kind = PROPERTY, name = "map", key = null, index = null, inIterable = false )
    node( kind = TYPE_ARGUMENT, name = "<map value>", key = "invalid", index = null, inIterable = true )
    node( kind = PROPERTY, name = "email", key = null, index = 1, inIterable = true )

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gsmet: Removed.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gsmet: This is now what we have. Except the second one is a PROPERTY as we don't have the TYPE_ARGUMENT node type (but it's already done as a type argument).

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gsmet: Changed the toString and added assertions on the structure of the path.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gsmet: Done.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
gsmet: Fixed.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
gsmet: I explained a bit more the purpose of this list.

gsmet:
@gunnarmorling so I think we are in pretty good shape now. I'll work tomorrow morning on the array support.

gsmet:
@gunnarmorling it's ready for review, I think I fixed all the pending issues.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
```diff
@@ -0,0 +1,234 @@
...
+			);
+
+			return cinemaEmailAddresses;
+		}
+	}
+
+	@SuppressWarnings({ "unused", "unchecked" })
+	private static class CinemaArray {
+
+		private Optional<@NotNull @Valid Cinema>[] array;
```
gunnarmorling: Using `@NotNull` within `Optional` is weird. Also it's for the sake of the example only, I'd prefer if we found a more realistic usage.

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -696,27 +698,66 @@ else if ( constraintHelper.isMultiValueConstraint( annotationType ) ) {
...
+					: CascadingTypeParameter.annotatedObject( ReflectionHelper.typeOf( executable ) ) );
 		}
 
-		return cascadingTypeParameters;
+		return cascadingTypeParameters.isEmpty() ? Collections.emptyList() : cascadingTypeParameters;
 	}
 
-	private List<TypeVariable<?>> getCascadingTypeParameters(TypeVariable<?>[] typeParameters, AnnotatedType annotatedType) {
-		List<TypeVariable<?>> cascadingTypeParameters = new ArrayList<>();
+	private List<CascadingTypeParameter> getCascadingTypeParameters(TypeVariable<?>[] typeParameters, AnnotatedType annotatedType) {
```
gunnarmorling: That method became quite long. The two cases it handles are good candidates for extracting two methods.

engine/src/main/java/org/hibernate/validator/internal/metadata/cascading/package.html
```diff
@@ -0,0 +1,14 @@
+<!--
```
gsmet: Done.

test-utils/src/main/java/org/hibernate/validator/testutil/ConstraintViolationAssert.java
```diff
@@ -462,6 +462,11 @@ public PathExpectation returnValue() {
 			return this;
 		}
 
+		public PathExpectation typeArgument(String name, boolean inIterable, Object key, Integer index) {
+			nodes.add( new NodeExpectation( name, ElementKind.PROPERTY, inIterable, key, index, null ) );
```
gsmet: Created a JIRA issue and added a TODO in the code.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ValidatableParametersMetaData.java
```diff
@@ -46,7 +46,7 @@ private ParameterConstraintLocation getParameterConstraintLocation(ConstraintLoc
 			return (ParameterConstraintLocation) location;
 		}
 		else if ( location instanceof TypeArgumentConstraintLocation ) {
-			return (ParameterConstraintLocation) ( (TypeArgumentConstraintLocation) location ).getDelegate();
```
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
gsmet: Done.

engine/src/test/java/org/hibernate/validator/test/internal/engine/cascaded/NestedCascadedConstraintsTest.java
gsmet: OK, I went the List route, even if it's not that good either.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
```diff
@@ -46,22 +47,23 @@
 	private final ConstraintLocation location;
 
 	/**
-	 * The {@link ValueExtractor} used to extract the value for validation.
+	 * The sequence of {@link ValueExtractor}s used to navigate from the outermost container to the innermost container
+	 * and extract the value for validation.
 	 */
-	private final ValueExtractorDescriptor valueExtractorDescriptor;
+	private final List<ValueExtractorDescriptor> valueExtractorDescriptors;
```
gsmet: Yeah, well, I tried to do that and in the end I thought the best solution was to use `Collections.singletonList`. I introduced a new helper called `CollectionHelper.toImmutableList` (same for `Set`) that we should use to make a list immutable and that make some naive optimization. See this commit https://github.com/hibernate/hibernate-validator/pull/659/commits/662003de38b0428a7a13be68b0d30bf50ea0ec21 .

I think it should be generalized to all the immutable list/set in the metadata. Tell me what you think about it.

gsmet:
@gunnarmorling pushed 2 additional commits to address your comments.

Waiting for your feedback on the `List` of `ValueExtractorDescriptor` optimization.

gunnarmorling:
> Waiting for your feedback on the List of ValueExtractorDescriptor optimization.

I think it's good, merged it (and added the `@Immutable` stereotype).

My thinking was not only about memory, though, but also about complexity of usage. Now `ValidatorImpl` needs to deal with a list of extractors, whereas I hoped this complexity could be encapsulated within `MetaConstraint`.

PR #660, opened 2017-02-07
Fixing mentioning wrong annotations in JavaDoc for annotation Lists (EAN, Currency)

marko-bekhta:
As in the title. The others seems to be OK. 

I was just making the same 'typo' while copying an annotation for DurationMin :)


gunnarmorling:
Merged; thanks, @marko-bekhta!

PR #661, opened 2017-02-08
HV-1240 Don't call ValueExtractors on null values anymore

gsmet:
As a side effect, we only report 1 violation on Optionals when the
Optional is null and the wrapper and the wrapped value have a @NotNull
constraint.


gunnarmorling:

Can you add a short note to the [migration guide](https://developer.jboss.org/wiki/HibernateValidatorMigrationGuide) about the null Optional case? Thanks!

gsmet:
@gunnarmorling done.

PR #663, opened 2017-02-08
HV-1242 Hibernate Validator 5.4 implements JSR 349

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1242


gsmet:
Nice catch! Merged!

PR #664, opened 2017-02-08
HV-1232 Implement constraints for Period and Duration

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1232

Hi,

So this is the initial things that I've got for these new constraints. For `Duration` I've created new annotations as well as added validators for existing Min/Max (assuming that the values are in nanos). For the new `Duration` annotations user can choose time units that are more suitable.
As for the `Period` ... It is more trickier as there's no comparator for these classes and we need to make some sort of assumption on how many days are in the month to be able to normalize these objects and compare them. Initially I've just put a constant for 30 days in month in a validator but I think that providing user with a possibility to decide how many days are in the month is a better option (29 ? 30 ? 31)

This is not yet finished, things like tests (for validators and for AP), documentation, JavaDoc explaining those "30 days in month" are missing for now. I thought that first we should agree on the annotations themselves and then I'll finish it all. WDYT ?


engine/src/main/java/org/hibernate/validator/constraints/time/PeriodMax.java
```diff
@@ -0,0 +1,73 @@
...
+
+	Class<? extends Payload>[] payload() default { };
+
+	int years() default 0;
+
+	int months() default 0;
+
+	int days() default 0;
+
+	/**
```
gunnarmorling: Do we really need this? Couldn't we build a `Period` from the given years/months/days, subtract the annotated period via `minus()` and test `isNegative()` on the result?

Also an `inclusive()` flag seems useful (similar to `@DecimalMax`).

engine/src/main/java/org/hibernate/validator/constraints/time/PeriodMax.java
```diff
@@ -0,0 +1,73 @@
...
+
+import org.hibernate.validator.constraints.time.PeriodMax.List;
+
+/**
+ * Annotated {@link Period} element's length must be less than or equal to
+ * the one constructed from {@link PeriodMax#years()}, {@link PeriodMax#months()},
+ * {@link PeriodMax#days()} values.
+ *
+ * @author Marko Bekhta
+ * @hv.experimental
```
gunnarmorling: Can you use the `@Incubating` annotation instead?

engine/src/main/java/org/hibernate/validator/constraints/time/PeriodRange.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+@Documented
+@Constraint(validatedBy = { })
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
+@Retention(RUNTIME)
+@Repeatable(List.class)
+@PeriodMin
+@PeriodMax
+@ReportAsSingleViolation
+public @interface PeriodRange {
```
gunnarmorling: I'd just skip that in the favor of two separate `@PeriodMax`/`@PeriodMin` annotations.

engine/src/main/java/org/hibernate/validator/constraints/time/DurationMin.java
```diff
@@ -0,0 +1,64 @@
...
+
+	String message() default "{org.hibernate.validator.constraints.time.DurationMin.message}";
+
+	Class<?>[] groups() default { };
+
+	Class<? extends Payload>[] payload() default { };
+
+	long value() default 0;
+
+	ChronoUnit units() default ChronoUnit.NANOS;
```
gunnarmorling: `SECOND` may be a more commonly used value and thus a better default?

engine/src/main/java/org/hibernate/validator/constraints/time/PeriodRange.java
gunnarmorling: Some for the `DurationRange`.

engine/src/main/java/org/hibernate/validator/constraints/time/PeriodMax.java
gsmet: Yeah, I thought it might work yesterday evening and then I read the Javadoc and the code for `minus`:
```java
    public Period minus(TemporalAmount amountToSubtract) {
        Period isoAmount = Period.from(amountToSubtract);
        return create(
                Math.subtractExact(years, isoAmount.years),
                Math.subtractExact(months, isoAmount.months),
                Math.subtractExact(days, isoAmount.days));
    }
```

It simply substract each part of the `Period`, so you could end up with `1 year -2 months 4 days` and this would indeed be considered negative.

I'm wondering if it really makes sense to consider `Period` baring these limitations.

engine/src/main/java/org/hibernate/validator/constraints/time/PeriodRange.java
gsmet: +1 ti skip the `*Range` annotations.

engine/src/main/java/org/hibernate/validator/constraints/time/DurationMin.java
gsmet: To be honest, I think I would rather have `days`, `hours`, `minutes` and `seconds` - I don't think nanos would be useful but we can add it for consistency. I think limiting to one single unit is not that good.

@gunnarmorling WDYT about that proposal?

annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
```diff
@@ -247,8 +249,8 @@ public ConstraintHelper(Types typeUtils, AnnotationApiHelper annotationApiHelper
 				BeanValidationTypes.FUTURE,
 				JAVA_TIME_TYPES_SUPPORTED_BY_FUTURE_AND_PAST_ANNOTATIONS
 		);
-		registerAllowedTypesForBuiltInConstraint( BeanValidationTypes.MAX, Number.class, String.class );
-		registerAllowedTypesForBuiltInConstraint( BeanValidationTypes.MIN, Number.class, String.class );
+		registerAllowedTypesForBuiltInConstraint( BeanValidationTypes.MAX, Number.class, String.class, Duration.class );
+		registerAllowedTypesForBuiltInConstraint( BeanValidationTypes.MIN, Number.class, String.class, Duration.class );
```
gsmet: I don't think I would add them here. We have a nice dedicated annotation, let's use that.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/MaxValidatorForDuration.java
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	@Override
+	public boolean isValid(Duration value, ConstraintValidatorContext constraintValidatorContext) {
+		// null values are valid
+		if ( value == null ) {
+			return true;
+		}
+
+		return duration.compareTo( value ) <= 0;
```
gsmet: `>=`?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/MaxValidatorForDuration.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Check that the validated {@link Duration} is less than or equal to the minimum
+ * value specified.
+ *
+ * @author Marko Bekhta
+ */
+public class MaxValidatorForDuration implements ConstraintValidator<Max, Duration> {
+
+	private Duration duration;
```
gsmet: Maybe call them `maxDuration` (and `minDuration` for the `MinValidatorForDuration`). It would be easier to apprehend the comparison below.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/time/BasePeriodMinMaxValidator.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.hibernate.validator.internal.util.logging.Log;
+import org.hibernate.validator.internal.util.logging.LoggerFactory;
+
+/**
+ * Base class for PeriodMinMax validators containing comparing method for {@link Period}.
+ *
+ * @author Marko Bekhta
+ */
+public abstract class BasePeriodMinMaxValidator<T extends Annotation> implements ConstraintValidator<T, Period> {
```
gsmet: Note: the comments in this class are to be taken into account only if we keep the `Period` validators.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/time/BasePeriodMinMaxValidator.java
```diff
@@ -0,0 +1,48 @@
...
+
+	private static final Log log = LoggerFactory.make();
+
+	private static int MONTHS_IN_YEAR = 12;
```
gsmet: You don't need that. See `Period.toTotalMonths`.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/time/BasePeriodMinMaxValidator.java
```diff
@@ -0,0 +1,48 @@
...
+	private int remainingDays;
+	private int daysInMonth;
+
+	protected final void initialize(int years, int months, int days, int daysInMonth) {
+		if ( daysInMonth < 1 ) {
+			log.getDaysInMonthCannotBeNegativeException( daysInMonth );
+		}
+		this.daysInMonth = daysInMonth;
+		this.remainingDays = days % daysInMonth;
+		this.totalMonths = years * MONTHS_IN_YEAR + months + days / daysInMonth;
```
gsmet: I think you'd better transform everything to days. It will be easier.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/time/DurationMaxValidator.java
```diff
@@ -0,0 +1,39 @@
...
+		this.duration = Duration.of( constraintAnnotation.value(), constraintAnnotation.units() );
+	}
+
+	@Override
+	public boolean isValid(Duration value, ConstraintValidatorContext context) {
+		// null values are valid
+		if ( value == null ) {
+			return true;
+		}
+		return duration.compareTo( value ) > -1;
```
gsmet: Let's be consistent and use 0 as a reference.

(But as mentioned earlier, I vote for removing these ones)

engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
```diff
@@ -297,6 +308,11 @@ public ConstraintHelper() {
 		putConstraint( tmpConstraints, ScriptAssert.class, ScriptAssertValidator.class );
 		putConstraint( tmpConstraints, URL.class, URLValidator.class );
 
+		putConstraint( tmpConstraints, DurationMin.class, DurationMinValidator.class );
+		putConstraint( tmpConstraints, DurationMax.class, DurationMaxValidator.class );
+		putConstraint( tmpConstraints, PeriodMin.class, PeriodMinValidator.class );
+		putConstraint( tmpConstraints, PeriodMax.class, PeriodMaxValidator.class );
```
gsmet: They usually are added by alphabetical order.

engine/src/main/java/org/hibernate/validator/constraints/time/DurationMin.java
gunnarmorling: I thought about it, too.

But then `Duration` itself also only allows creation from the single elements, i.e. there is no method `of(days, hours, minutes, ...)`. So I guess I'd stay consistent with that, but I don't have a strong preference really.

engine/src/main/java/org/hibernate/validator/constraints/time/PeriodMax.java
gunnarmorling: > I'm wondering if it really makes sense to consider Period baring these limitations.

Yes, I'm getting doubts as well.

I first thought one could construct something around `normalize()`,  but after all you cannot really tell whether "2 months" is longer or shorter than "60 days", as you don't know the number of days in a months. So it seems for a good reason that the java.time API doesn't expose a way for comparing periods. We then shouldn't try to invent it.

+1 for supporting `Duration` only.

engine/src/main/java/org/hibernate/validator/constraints/time/DurationMin.java
gsmet: Yeah, looks like you only have the `parse` method for that. Or you have to use the `plus*` methods. That being said, we would do the operation only once at startup so I don't think it's that bad.

@marko-bekhta WDYT?

As for me, I find it more readable to say 4 hours 30 minutes rather than converting everything to minutes but maybe it's just me.

gunnarmorling:
Yeah, that sounds reasonable indeed.

marko-bekhta:
I've read through the discussion - 
I wasn't able to think of anything better for Period than what you've seen as you were saying that there's no way to say what is a larger period 2 month or 60 days (so I can only see that option of allowing users to determine how long is the month). But in general - I would say that if an object itself cannot be compared to another in one way we should not be adding a constraint for it. So +1 for having Duration only.

As for the duration - as you were saying it doesn't seem to be very readable to have just one unit and would be better to be able to put something like `@MinDuration(minutes=5 , seconds=3)`. But in that case we would need to put a parameter for all `ChronoUnit`s (or at least for all smaller ones). I think it'll be not good if for example we have only `hours` `minutes` and `seconds` but no `milliseconds` or `nanoseconds`.
Because what if you have a duration of 1:11:12.123456 ? you wouldn't be able to specify it right ?

So on the one hand - having only one `ChronoUnit` is not very readable, on the other hand we would need to add a lot of annotation parameters. I really don't have a strong opinion on which option is better so I'll just trust your judgment on this one :)

gsmet:
Yes, you would have to add all of them (even if I don't expect the smallest ones to be used). I won't expect someone to require a duration of 4 nanoseconds. As for the names, keep the names used in `Duration` so `nanos`, `millis` to be consistent.

It's a bit more verbose but it's always a pain to specify duration in only one units. I did that a lot in property files and it's really a pain to convert the values in your head.

marko-bekhta:
So to summarize:
- support of Duration only
- add all parameters till `years` (don't think that `ERAS` or `MILLENNIA` are that used :) )
- drop the range annotation
- remove support of Duration for `@Min` `@Max`

did I miss anything ?

gunnarmorling:
@marko-bekhta sounds great!

gsmet:
> add all parameters till years (don't think that ERAS or MILLENNIA are that used :) )

I think for a first version I would stick to what is available in `Duration` (the `plus*` methods basically) and nothing more. You will have the same issue we discussed earlier with months and years.

> did I miss anything ?

 * The bug I reported on the sign of one of the comparisons but you will find it anyway when writing tests.
 * The inclusive option similar to what is done in `DecimalMin`/`DecimalMax`.

gunnarmorling:
Looks great. I'll just apply some typo-fixes and merge it.

marko-bekhta:
great, thanks!

gunnarmorling:
Rebased and applied. Thanks, @marko-bekhta!

PR #665, opened 2017-02-10
HV-1105 Avoiding non-final field and synchronized map in ValidatorFac…

gunnarmorling:
…toryImpl;

* XmlMetaDataProvider doesn't need to be nulled out; refactoring its structure to make more apparent what state it holds
* replaced synchronized map by CHM; making sure only one BeanMetaDataManager is built per parameter name provider

https://hibernate.atlassian.net/browse/HV-1105


gsmet:
Merged, thanks!

PR #666, opened 2017-02-10
HV-1243 Throw an exception if no value extractor is found during cascaded validation

gsmet:
* https://hibernate.atlassian.net/browse/HV-1243


gunnarmorling:
Hoping that PR isn't jinxed ;)

PR #672, opened 2017-02-14
HV-1241 Use a proper LRU cache in ConstraintValidatorManager

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1241

Created LRU cache based on LinkedHashMap and replaced cache in ConstraintValidatorManager with it.

Will such `LinkedHashMap` based cache be a proper one ? and what should be a max capacity for this cache ?


gunnarmorling:
@marko-bekhta I've been thinking about it and as you say it's just not possible to find a reasonable size for the cache. And exposing it as option wouldn't be great either; it'd be one of those knobs people don't know what to set to. So I'm inclined currently to leave things as is.

gunnarmorling:
Oki, closing it. Thanks nevertheless for investigating it, @marko-bekhta!

marko-bekhta:
Happy to help!

PR #674, opened 2017-02-15
HV-1249 Allocating smaller sets/maps in ConstraintTree code path; mor…

gunnarmorling:
…e than 5 composing constraints should be a very rare exception

https://hibernate.atlassian.net/browse/HV-1249


gsmet:
Merged, thanks!

PR #675, opened 2017-02-16
HV-1251 Check that we don't use snapshot dependencies when we release

gsmet:
* https://hibernate.atlassian.net/browse/HV-1251


gunnarmorling:
Nice!

PR #680, opened 2017-02-21
HV-1258 Scoping BeanMetaDataManager per parameter name provider *and* method validation configuration

gunnarmorling:
Before we'd have ignored a new method validation configuration if the parameter name provider was the same.

This also unties metadata providers from the parameter name provider. There should never have been a link between the two in the first place.

It's a fallout of the work on HV-1256/HV-1257; sending it separately as its good to go and probably also worth backporting.

https://hibernate.atlassian.net/browse/HV-1258

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
@@ -93,7 +93,7 @@ public class ConfigurationImpl implements HibernateValidatorConfiguration, Confi
 	private boolean failFast;
 	private final List<ValueExtractor<?>> cascadedValueExtractors = new ArrayList<>();
 	private ClassLoader externalClassLoader;
-	private final MethodValidationConfiguration methodValidationConfiguration = new MethodValidationConfiguration();
+	private final MethodValidationConfiguration.Builder methodValidationConfigurationBuilder = new MethodValidationConfiguration.Builder();
 
 	public ConfigurationImpl(BootstrapState state) {
 		this();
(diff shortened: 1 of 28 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
```diff
@@ -232,13 +241,7 @@ public BeanMetaDataImpl(Class<T> beanClass,
...
 	@Override
 	public boolean hasConstraints() {
-		if ( beanDescriptor.isBeanConstrained()
-				|| !beanDescriptor.getConstrainedConstructors().isEmpty()
-				|| !beanDescriptor.getConstrainedMethods( MethodType.NON_GETTER, MethodType.GETTER ).isEmpty() ) {
-			return true;
-		}
-
-		return false;
+		return hasConstraints;
```
gsmet: This one would have been better in a separate commit and JIRA I think. We can do with it though, I think.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -213,36 +213,36 @@ public final HibernateValidatorConfiguration failFast(boolean failFast) {
...
 	@Override
 	public HibernateValidatorConfiguration allowOverridingMethodAlterParameterConstraint(boolean allow) {
-		this.methodValidationConfiguration.allowOverridingMethodAlterParameterConstraint( allow );
+		this.methodValidationConfigurationBuilder.allowOverridingMethodAlterParameterConstraint( allow );
 		return this;
 	}
 
 	public boolean isAllowOverridingMethodAlterParameterConstraint() {
-		return this.methodValidationConfiguration.isAllowOverridingMethodAlterParameterConstraint();
+		return this.methodValidationConfigurationBuilder.isAllowOverridingMethodAlterParameterConstraint();
```
gsmet: Not sure I'm really convinced by how we deal with the builder state in the getter here. The pattern is kinda weird. Not that I can think of something better though.

gunnarmorling:
> I'm a bit annoyed by the fact that you changed the way hasConstraints is determined while it's really totally unrelated. I think it should have its own JIRA and a proper commit.

Ok, I'll create a JIRA and extract it into a separate commit. I had a larger change originally (not creating the descriptor model eagerly any more), and this was part of it. I reverted that largely, but found that specific change worth keeping, being a bit lazy only :)

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
gunnarmorling: Who says a builder is write-only ;)

gsmet:
Yeah, I'm a bit cautious about this one because we had a bug in this part not so long ago with `hasConstraints()` returning false instead of true in certain cases. So I prefer to be on the safe side if we have to dig changes in this area again :).

gunnarmorling:
Extracted https://hibernate.atlassian.net/browse/HV-1259 for hasConstraints and reworked the PR.

gsmet:
Cool, waiting for CI to go green and I'll merge it.

gunnarmorling:
Now that's a long CI run ;)

gsmet:
Ah, yeah, sorry, I merged it and kept it open for the 5.4 backport which didn't happen!

Closing!

PR #681, opened 2017-02-23
HV-1257 et al.: Allowing to pass in value extractors during bootstrap

gunnarmorling:
* https://hibernate.atlassian.net/browse/HV-1257 (`Configuration#addValueExtractor()`, `ValidatorContext#addValueExtractor()`)
* https://hibernate.atlassian.net/browse/HV-1260 (_validation.xml_)
* https://hibernate.atlassian.net/browse/HV-1261 (service loader)

Requires https://github.com/beanvalidation/beanvalidation-api/pull/89

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
@@ -14,7 +14,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -91,7 +90,6 @@ public class ConfigurationImpl implements HibernateValidatorConfiguration, Confi
 	// HV-specific options
 	private final Set<DefaultConstraintMapping> programmaticMappings = newHashSet();
 	private boolean failFast;
-	private final List<ValueExtractor<?>> cascadedValueExtractors = new ArrayList<>();
 	private ClassLoader externalClassLoader;
 	private final MethodValidationConfiguration.Builder methodValidationConfigurationBuilder = new MethodValidationConfiguration.Builder();
 
@@ -198,6 +196,19 @@ public class ConfigurationImpl implements HibernateValidatorConfiguration, Confi
 	}
 
 	@Override
+	public HibernateValidatorConfiguration addValueExtractor(ValueExtractor<?> extractor) {
+		Contracts.assertNotNull( extractor, MESSAGES.parameterMustNotBeNull( "extractor" ) );
+
+		if ( log.isDebugEnabled() ) {
+			log.debug( "Adding value extractor " + extractor );
+		}
+
+		validationBootstrapParameters.addValueExtractor( extractor );
+
+		return this;
+	}
+
+	@Override
 	public final HibernateValidatorConfiguration addMapping(InputStream stream) {
 		Contracts.assertNotNull( stream, MESSAGES.inputStreamCannotBeNull() );
 
@@ -263,14 +274,6 @@ public class ConfigurationImpl implements HibernateValidatorConfiguration, Confi
 		if ( value != null ) {
 			validationBootstrapParameters.addConfigProperty( name, value );
 		}
-		return this;
-	}
-
-	@Override
-	public HibernateValidatorConfiguration addCascadedValueExtractor(ValueExtractor<?> extractor) {
-		Contracts.assertNotNull( extractor, MESSAGES.parameterMustNotBeNull( "extractor" ) );
-		cascadedValueExtractors.add( extractor );
-
 		return this;
 	}
 
@@ -387,8 +390,9 @@ public class ConfigurationImpl implements HibernateValidatorConfiguration, Confi
 		return validationBootstrapParameters.getClockProvider();
 	}
 
-	public List<ValueExtractor<?>> getCascadedValueExtractors() {
-		return cascadedValueExtractors;
+	@Override
+	public Set<ValueExtractor<?>> getValueExtractors() {
+		return validationBootstrapParameters.getValueExtractors();
 	}
 
 	@Override
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
@@ -14,9 +14,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.ServiceLoader;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.validation.BootstrapConfiguration;
 import javax.validation.ClockProvider;
@@ -34,6 +38,7 @@ import javax.validation.valueextraction.ValueExtractor;
 import org.hibernate.validator.HibernateValidatorConfiguration;
 import org.hibernate.validator.cfg.ConstraintMapping;
 import org.hibernate.validator.internal.cfg.context.DefaultConstraintMapping;
+import org.hibernate.validator.internal.engine.cascading.ValueExtractorDescriptor;
 import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorFactoryImpl;
 import org.hibernate.validator.internal.engine.resolver.DefaultTraversableResolver;
 import org.hibernate.validator.internal.util.Contracts;
@@ -92,6 +97,7 @@ public class ConfigurationImpl implements HibernateValidatorConfiguration, Confi
 	private boolean failFast;
 	private ClassLoader externalClassLoader;
 	private final MethodValidationConfiguration.Builder methodValidationConfigurationBuilder = new MethodValidationConfiguration.Builder();
+	private final Map<ValueExtractorDescriptor.Key, ValueExtractorDescriptor> valueExtractors = new HashMap<>();
 
 	public ConfigurationImpl(BootstrapState state) {
 		this();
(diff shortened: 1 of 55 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorManager.java
```diff
@@ -31,6 +32,34 @@
...
+
+			LegacyMapValueExtractor.DESCRIPTOR,
+			MapValueExtractor.DESCRIPTOR,
+			MapKeyExtractor.DESCRIPTOR,
+
+			LegacyIterableValueExtractor.DESCRIPTOR,
+			IterableValueExtractor.DESCRIPTOR,
+
+			OptionalValueExtractor.DESCRIPTOR,
+			ObjectValueExtractor.DESCRIPTOR
```
gsmet: So you don't include `ObservableValueValueExtractor` here? I can understand that it's a tough choice :).

gsmet: Considering how useful `getDefaultValueExtractors` is, I would have made it a Set of VED instead of VE and reuse it in the constructor. That could avoid forgetting to add one at one place or the other.

And we could stream the result directly in `getDefaultValueExtractors`.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
```diff
@@ -136,6 +142,15 @@ public void setClockProvider(ClockProvider clockProvider) {
 		this.clockProvider = clockProvider;
 	}
 
+	public Map<Key, ValueExtractorDescriptor> getValueExtractors() {
+		return valueExtractors;
+	}
+
+	public void addValueExtractor(ValueExtractor<?> valueExtractor) {
```
gsmet: I would pass the descriptor here. You're forced to unwrap it then rewrap it in 2 places.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextImpl.java
```diff
@@ -107,6 +111,12 @@ public HibernateValidatorContext clockProvider(ClockProvider clockProvider) {
 	}
 
 	@Override
+	public HibernateValidatorContext addValueExtractor(ValueExtractor<?> extractor) {
+		valueExtractors.add( extractor );
```
gsmet: Maybe we could change them to VED here and throw the error early if it's not a valid VE?

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorManager.java
```diff
@@ -9,70 +9,119 @@
...
+	@Immutable
+	private final Map<ValueExtractorDescriptor.Key, ValueExtractorDescriptor> valueExtractors;
+
+	public ValueExtractorManager(Set<ValueExtractor<?>> externalExtractors) {
+		LinkedHashMap<ValueExtractorDescriptor.Key, ValueExtractorDescriptor> tmpValueExtractors = new LinkedHashMap<>();
 
 		for ( ValueExtractor<?> valueExtractor : externalExtractors ) {
-			tmpValueExtractors.add( new ValueExtractorDescriptor( valueExtractor ) );
+			ValueExtractorDescriptor descriptor = new ValueExtractorDescriptor( valueExtractor );
+			tmpValueExtractors.put( descriptor.getKey(), descriptor );
```
gsmet: I wonder if it wouldn't be better to add the user defined VE after the default ones, thus providing a way to override the default ones.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
```diff
@@ -136,6 +142,15 @@ public void setClockProvider(ClockProvider clockProvider) {
 		this.clockProvider = clockProvider;
 	}
 
+	public Map<Key, ValueExtractorDescriptor> getValueExtractors() {
```
gsmet: Wondering if it would be better to announce clearly it returns VED, not VE.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -198,6 +203,22 @@ public HibernateValidatorConfiguration clockProvider(ClockProvider clockProvider
 	}
 
 	@Override
+	public HibernateValidatorConfiguration addValueExtractor(ValueExtractor<?> extractor) {
+		Contracts.assertNotNull( extractor, MESSAGES.parameterMustNotBeNull( "extractor" ) );
+
+		// TODO prevent duplicates
```
gsmet: Let's decide what to do about this before merging and implement it. We have added a lot of TODOs lately.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextImpl.java
gunnarmorling: I'll take a look. It may not work for `Configuration` though as the public method mandates `VE` there.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -513,6 +514,7 @@ public int hashCode() {
 			int result = 1;
 			result = prime * result + ( ( methodValidationConfiguration == null ) ? 0 : methodValidationConfiguration.hashCode() );
 			result = prime * result + ( ( parameterNameProvider == null ) ? 0 : parameterNameProvider.hashCode() );
+			result = prime * result + ( ( valueExtractorManager == null ) ? 0 : valueExtractorManager.hashCode() );
```
gunnarmorling: Will change it.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorManager.java
gunnarmorling: Hum, no it should be part of it.

gunnarmorling: Good idea.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorManager.java
gunnarmorling: Yes, that's how it should be done eventually. Currently this all relies on the order here.

engine/src/main/java/org/hibernate/validator/internal/xml/BootstrapConfigurationImpl.java
```diff
@@ -54,18 +55,21 @@
...
 
+
 	private BootstrapConfigurationImpl() {
 		this.defaultProviderClassName = null;
 		this.constraintValidatorFactoryClassName = null;
 		this.messageInterpolatorClassName = null;
 		this.traversableResolverClassName = null;
 		this.parameterNameProviderClassName = null;
 		this.clockProviderClassName = null;
+		this.valueExtractorClassNames = new HashSet<>( 0 );
```
gunnarmorling: Agreed.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
```diff
@@ -246,6 +263,24 @@ private void setClockProvider(String clockProviderFqcn, ClassLoader externalClas
...
 
+	private void setValueExtractors(Set<String> valueExtractorFqcns, ClassLoader externalClassLoader) {
+		for ( String valueExtractorFqcn : valueExtractorFqcns ) {
+			try {
+				@SuppressWarnings("unchecked")
+				Class<? extends ValueExtractor<?>> clazz = (Class<? extends ValueExtractor<?>>) run(
+						LoadClass.action( valueExtractorFqcn, externalClassLoader )
+						);
+				ValueExtractor<?> valueExtractor = run( NewInstance.action( clazz, "value extractor class" ) );
+				log.addingValueExtractor( clazz );
```
gunnarmorling: Ok.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextImpl.java
gsmet: Yes, the public methods would add VEs but we would convert them to VED as early as possible to throw an error as early as possible and use as much as possible VED internally.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
gunnarmorling: Yes, of course.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
gunnarmorling: INFO is used in other places in this class (see logging methods 3 - 6).

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextImpl.java
gunnarmorling: Done.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorManager.java
gunnarmorling: Done.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
gunnarmorling: Done.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
gunnarmorling: Done.

engine/src/main/java/org/hibernate/validator/internal/xml/BootstrapConfigurationImpl.java
gunnarmorling: Done.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
gunnarmorling: Added the checks.

gunnarmorling:
@gsmet re-worked/added commits to address your remarks. Hope it's all good now.

gsmet:
Merged, thanks!

PR #682, opened 2017-02-24
HV-1262 Showing concerned validator classes in case there are multiple ones

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1262

@gsmet The actual fix is in the first commit. I'd say let's backport that one to 5.4.


gsmet:
Merged. Backported an amended version of the first commit. I didn't backport the `TypeFormatter` as `Type.getTypeName()` is new in Java 8.

PR #685, opened 2017-02-28
HV-1264 Allow to pass an escaped version of the regexp to the message interpolator

gsmet:
* https://hibernate.atlassian.net/browse/HV-1264

It should be backported to 5.4, probably not exactly in this form so that we don't break the APIs (it's probably better that I work on the backport too).


engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -275,18 +274,17 @@ public void addConstraintFailures(Set<ConstraintViolation<T>> failingConstraintV
...
 				localContext.getCurrentValidatedValue(),
 				descriptor,
+				constraintViolationCreationContext.getMessageParameters(),
 				constraintViolationCreationContext.getExpressionVariables()
 		);
 		// at this point we make a copy of the path to avoid side effects
 		Path path = PathImpl.createCopy( constraintViolationCreationContext.getPath() );
-		//same for expression variables
-		Map<String, Object> expressionVariables =
-				Collections.unmodifiableMap( constraintViolationCreationContext.getExpressionVariables() );
```
gsmet: We don't need a copy here anymore because the maps in constraintViolationCreationContext are now immutable so we don't allow the side effects that might have existed before and have justified this copy.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
```diff
@@ -226,7 +226,7 @@ private void throwExceptionForNullValidator(Type validatedValueType, String path
 			// If not we create a violation
 			// using the error message in the annotation declaration at top level.
 			if ( localViolations.isEmpty() ) {
-				final String message = (String) getDescriptor().getAttributes().get( "message" );
+				final String message = getDescriptor().getMessageTemplate();
```
gsmet: This is from pre 1.1 era. I updated all the occurrences of this pattern.

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/ParameterTermResolver.java
```diff
@@ -38,6 +39,16 @@ public String interpolate(Context context, String expression) {
 		return resolvedExpression;
 	}
 
+	private Object getVariable(Context context, String parameter) {
+		if ( context instanceof HibernateMessageInterpolatorContext ) {
```
gsmet: I wondered if it could be worth it to create the aggregation earlier. I decided against it because `getMessageParameters()` is preexisting and I'm a bit worried about changing the definition of it.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
gunnarmorling: Sounds good.

engine/src/main/java/org/hibernate/validator/messageinterpolation/HibernateMessageInterpolatorContext.java
```diff
@@ -23,4 +27,22 @@
 	 * @return The currently validated root bean type.
 	 */
 	Class<?> getRootBeanType();
+
+	/**
+	 * @return the message parameters added to this context for interpolation
+	 *
+	 * @since 6.0
+	 */
+	default Map<String, Object> getMessageParameters() {
```
gunnarmorling: Why did you decide to make this a default method? The only implementation is our own one.

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/parser/Token.java
```diff
@@ -56,7 +56,7 @@ public void makeELToken() {
 	public void terminate() {
 		value = builder.toString();
 		if ( isEL ) {
-			// HSEARCH-834 curly braces need to be un-escaped prior to be passed to the EL engine
+			// HV-834 curly braces need to be un-escaped prior to be passed to the EL engine
```
gunnarmorling: :)

gunnarmorling:
@gsmet I've pushed one commit with some smaller things I noticed during review. My only question is about the usage of default methods (see inline).

gsmet:
Removed the default methods and pushed to master. Will work on the backport.

PR #686, opened 2017-02-28
HV-1152 Make JavaDoc render correctly with Java 9

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1152

As for the `add-modules` part I was using Java:
```
java version "9-ea"
Java(TM) SE Runtime Environment (build 9-ea+158)
Java HotSpot(TM) 64-Bit Server VM (build 9-ea+158, mixed mode)
```
and it was failing to find `java.annotations.common`. So I've looked and `java.annotation.Generated` was present in `java.xml.ws.annotation` so I've made this change. 

I've also moved that `<finalName>` to a build level from assembly plugin configuration as IDE was complaining about it's location

As for the css part - I didn't want to change current look of JavaDoc, and that's why I ended up adding a new stylesheet. It's mainly because of the borders - if keep them for the tables - some of them start to look vary bad for jdk9 build (if needed can send a screenshot of it :) ). So I've took regular jdk9 JavaDoc and adapted it to Hibernate style.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -177,6 +177,10 @@
 
         <!-- add-opens options required for Arquillian and WildFly -->
         <arquillian.javaVmArguments.add-opens></arquillian.javaVmArguments.add-opens>
+
+        <!-- JavaDoc stylesheet name -->
+        <!-- styles for jdk8 and jdk9 are different-->
+        <javadoc-stylesheet-name>stylesheet.css</javadoc-stylesheet-name>
     </properties>
 
     <prerequisites>
@@ -645,7 +649,7 @@
                     <configuration>
                         <docfilessubdirs>true</docfilessubdirs>
                         <javadocDirectory>${project.basedir}/../src/main/javadoc</javadocDirectory>
-                        <stylesheetfile>stylesheet.css</stylesheetfile>
+                        <stylesheetfile>${javadoc-stylesheet-name}</stylesheetfile>
                         <bottom>
                             <![CDATA[Copyright &copy; ${project.inceptionYear}-{currentYear} <a href="http://redhat.com">Red Hat, Inc.</a> All Rights Reserved]]></bottom>
                     </configuration>
@@ -996,6 +1000,7 @@
                     --add-opens=java.management/javax.management.openmbean=ALL-UNNAMED
                     --add-opens=java.naming/javax.naming=ALL-UNNAMED
                 </arquillian.javaVmArguments.add-opens>
+                <javadoc-stylesheet-name>stylesheet-jdk9.css</javadoc-stylesheet-name>
             </properties>
             <build>
                 <pluginManagement>
--- a/src/main/javadoc/stylesheet-jdk9.css
+++ b/src/main/javadoc/stylesheet-jdk9.css
@@ -1,0 +1,934 @@
+/**
+ * Hibernate O/RM Java 7 compliant JavaDoc stylesheet.
+ */
+body {
+    background: #FFFFFF url(resources/bkg_gradient.gif) repeat-x;
+    margin: 0;
+    font-family: 'Lucida Grande', Geneva, Verdana, Arial, sans-serif;
+    font-size: 12px;
+    padding: 0;
+    color: #333;
+    height: 100%;
+    width: 100%;
+}
+
+iframe {
+    margin: 0;
+    padding: 0;
+    height: 100%;
+    width: 100%;
+    overflow-y: scroll;
+    border: none;
+}
+
+a:link, a:visited {
+    text-decoration: none;
+    color: #455e60;
+}
+
+a:hover, a:focus {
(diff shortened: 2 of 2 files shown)


gunnarmorling:
> So I've looked and java.annotation.Generated was present in java.xml.ws.annotation so I've made this change.

Yes, that was done recently. That area is still in flux, see [our discussion](http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-February/011320.html) on the jigsaw-dev mailing list.

On the style, thanks a ton for coming up with that stylesheet!

The only issue I see is that the dividers between table column/rows have gone (here are the [current docs](http://docs.jboss.org/hibernate/stable/validator/api/) for a comparison). Any chance to add that? At least vertical ones may look a tad better.

And one super-minor thing: when entering something into the search box in the upper right, the results overlay still has the stock-blue header. Maybe this could get a brownish color so to fit better with the rest of theme?

Otherwise, it's awesome! Thanks again!

gsmet:
@marko-bekhta Thanks for taking care of that!

Didn't check the stylesheet but I disagree on moving finalName to the build level. It has nothing to do here. There's something weird with your IDE :).

It would be nice if you could:
- move the add-modules stuff to a different commit as I don't think we will commit that for now
- discard the finalName change
- keep the CSS change in one commit

Thanks!

marko-bekhta:
@gunnarmorling I think the vertical ones can be added, as for the horizontal ones - they were causing some problems, I'll see if anything can be done about it, and I'll look once more through the doc to see if any colors should be updated (like that one for the search that you've mentioned).

@gsmet Ok, I'll do that :) as for the `finalName` it's how IDE is reading the pom file and validation it with the xsd schema. I've looked here - https://maven.apache.org/pom.html and <build> has a finalName in it:
```
 <build>
    <directory>${project.basedir}/target</directory>
    <outputDirectory>${project.build.directory}/classes</outputDirectory>
    <finalName>${project.artifactId}-${project.version}</finalName>
``` 
But I'm fine with where it was previously :)

gsmet:
@marko-bekhta I don't think they have the same meaning. In `<build>`, `finalName` would be the final name of the main artifact whereas in the assembly configuration, it would be the one of the assembly. At least, that's what I think.

marko-bekhta:
I've updated the css a bit more - see the screen below: 
javadoc
There are borders for tables now :) and a search input glow is more Hibernate-looking :)
Also I've split it into two commits first one just with the style related changes and second for the README update.

gunnarmorling:
Very nice, the screenshot looks good. I'll take a closer look tomorrow.
Other Hibernate projects will benefit from this as well, so kudos to you
for taking up this one!

marko-bekhta:
:) I just saw that other PR - https://github.com/hibernate/hibernate-validator/pull/688, so I've removed that last commit for the README. 

Happy to help. Let me know if any changes will be required.

gunnarmorling:
@marko-bekhta Looks good, going to merge it. Out of curiousity, how did you create this, did you take the default JDK 9 stylesheet.css and then modified it? Or did you take our existing one and adapted it?

marko-bekhta:
@gunnarmorling modified the JDK default one. I was using the current one to get colors and some font related values. 
At first I thought to modify the current one but I didn't want to break something for the current doc so (as some modifications were required)...

gunnarmorling:
Ok, thanks for clarifying. I had compared it to the JDK 9 default one and
was overwhelmed by the huge number of differences. But most of them are
actually formatting related (whitespace after ":", an empty line between
declarations etc.). I reckon they originate from your editor?

marko-bekhta:
Yes, that's right .... I reformatted that CSS file, I've probably shouldn't
done that, right?

2017-03-02 10:12 GMT+01:00 Gunnar Morling <notifications@github.com>:

gunnarmorling:
Yes, that would have been a bit better :) As it helps to keep the history
comprehensible. But it's no problem. Sorry for all the questions, just
trying to make sure I understand what's going on.

I've found the following to be the best approach for dealing with this kind
of changes:

* Create a first commit with the original file, unaltered (i.e. commit the
default Java 9 stylesheet.css)
* Apply the actual modifications in a separate commit
* Don't apply any re-formatting (or, if you really, really have to, do it
in a separate commit between the initial one and the one with the actual
changes)

This allows to understand the actual changes later on, e.g. allowing it to
re-apply them another time, if the underlying base has changed.

No worries, though. I'll take care of it.

gunnarmorling:
Applied. As discussed above, I've added the original JDK 9 stylesheet.css in a separate commit before your's and reworked your commit to remove the formatting changes. Also added another commit to change some styling. I'll open up a follow-up issue for ironing out some more glitches.

Thanks a lot @marko-bekhta, that was one of the big missing pieces for building on Java 9!

PR #688, opened 2017-03-01
HV-1266 Updating build instructions to reflect relocation of @Generated

gunnarmorling:
@gsmet Not sure what we have on CI; we can keep it on hold if there's still an older version installed. But I had the change handy so I thought I could well propose it.

https://hibernate.atlassian.net/browse/HV-1266


gsmet:
Rebased and merged, thanks!

PR #689, opened 2017-03-02
HV-1269 Modify JMH performance tests to prevent JVM optimisation skew…

johnaohara:
…ing results

### DIFF
--- a/performance/src/main/java/org/hibernate/validator/performance/cascaded/CascadedValidation.java
+++ b/performance/src/main/java/org/hibernate/validator/performance/cascaded/CascadedValidation.java
@@ -26,6 +26,7 @@ import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.Threads;
 import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -53,7 +54,7 @@ public class CascadedValidation {
 	@Threads(50)
 	@Warmup(iterations = 10)
 	@Measurement(iterations = 50)
-	public void testCascadedValidation(CascadedValidationState state) {
+	public void testCascadedValidation(CascadedValidationState state, Blackhole bh) {
 		// TODO graphs needs to be generated and deeper
 		Person kermit = new Person( "kermit" );
 		Person piggy = new Person( "miss piggy" );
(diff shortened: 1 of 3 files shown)


performance/src/main/java/org/hibernate/validator/performance/cascaded/CascadedValidation.java
```diff
@@ -65,20 +66,24 @@ public void testCascadedValidation(CascadedValidationState state) {
 
 		Set<ConstraintViolation<Person>> violations = state.validator.validate( kermit );
 		assertThat( violations ).hasSize( 0 );
+
+		bh.consume( violations );
 	}
 
+/*
```
gsmet: Hi John,

You commented this, is this what you wanted? BTW, the use of the constant looks weird as it's not made to be used here (it is used in the for loops).

But there's definitely something weird with the constants as we don't use them consistently.

johnaohara: Sorry, my mistake. I intended to remove this test, the for loop is completely optimised away by the JVM, and the test JMH test does not perform any actual work. If a large number of iterations is required as a test, then new Benchmark method with 50 * NUMBER_OF_VALIDATION_ITERATIONS would be appropriate and remove the for loop.  This would allow the JMH framework to manage the number of iterations.

gsmet: OK, will clean it up further and merge.

Thanks!

performance/src/main/java/org/hibernate/validator/performance/simple/SimpleValidation.java
```diff
@@ -67,10 +68,11 @@
 	@Threads(50)
 	@Warmup(iterations = 10)
 	@Measurement(iterations = 50)
-	public void testSimpleBeanValidation(ValidationState state) {
+	public void testSimpleBeanValidation(ValidationState state, Blackhole bh) {
 		DriverSetup driverSetup = new DriverSetup( state );
 		Set<ConstraintViolation<Driver>> violations = state.validator.validate( driverSetup.getDriver() );
 		assertThat( violations ).hasSize( driverSetup.getExpectedViolationCount() );
+		bh.consume( violations );
```
gunnarmorling: Out of curiousity, how comes that the assertion on the size of `violations` on the line before isn't enough to prevent the optimization you mention?

johnaohara: In practice, the effect might be the same. However, the purpose of Blackhole is to prevent Dead-code elimination, False sharing and Write wall, which can skew benchmark results.  I am not sure what effect .hasSize() has on the aforementioned JIT optimisations.

gunnarmorling: Ok, thanks! Better safe than sorry I guess.

gunnarmorling:
@gsmet Are you handling this one?

gsmet:
@gunnarmorling yes I want to make some additional clean up too.

gunnarmorling:
Cool, thanks!

gsmet:
Closing this one, superseded by https://github.com/hibernate/hibernate-validator/pull/694 which integrates your changes.

@johnaohara Thanks

PR #690, opened 2017-03-03
HV-1239 Programmatic API for configuring container element constraints

gunnarmorling:
@gsmet Could you take a first look and let me know what you think? It only supports getters right now, but it puts all the foundations in place, adding the other elements will be very simple.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/cfg/context/ContainerElementConstraintMappingContextImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/cfg/context/ContainerElementConstraintMappingContextImpl.java
@@ -1,0 +1,220 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.cfg.context;
+
+import java.lang.annotation.Annotation;
+import java.lang.annotation.ElementType;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.hibernate.validator.cfg.ConstraintDef;
+import org.hibernate.validator.cfg.context.ConstructorConstraintMappingContext;
+import org.hibernate.validator.cfg.context.ContainerElementConstraintMappingContext;
+import org.hibernate.validator.cfg.context.ContainerElementTarget;
+import org.hibernate.validator.cfg.context.MethodConstraintMappingContext;
+import org.hibernate.validator.cfg.context.PropertyConstraintMappingContext;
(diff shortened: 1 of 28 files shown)


gunnarmorling:
Ok, thanks!

gunnarmorling:
@gsmet, @emmanuelbernard: one thing I've been wondering: should we rename `containerElement()` to `containerElementType()`? It may appear as splitting hair, but the constraints are put to the _type_, not the _element_ itself.

Alternatively, one could split it up into `typeArgument()` (generic containers) and `componentType()` (arrays). The names would be a bit better to comprehend, but both would do the same, so I don't like it too much.

Note my considerations are primarily driven by applying the same concepts/names to the spec XML schema, hence the extensive consideration.

Thoughts?

engine/src/main/java/org/hibernate/validator/internal/cfg/context/ContainerElementConstraintMappingContextImpl.java
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	@Override
+	public ContainerElementConstraintMappingContext containerElement(int index, int... nestedIndexes) {
+		return parent.containerElement( index, nestedIndexes );
+	}
+
+	ContainerElementConstraintMappingContext nestedContainerElement(int[] nestedIndexes) {
+		if ( !( configuredType instanceof ParameterizedType ) && !( TypeHelper.isArray( configuredType ) ) ) {
+			throw LOG.getTypeIsNotAParameterizedNorArrayTypeException(  configuredType );
```
gunnarmorling: Ha, that's why they call you "Eagle Eye Guillaume" :)

gunnarmorling:
@gsmet Ready for review (apart from the renaming discussing above, I'd favor `containerElementType()`. Also there is one glitch with the API, there are some illegal state transitions. E.g. you can do:

    .type( FishTank.class )
        .property( "model", FIELD )
            .containerElement()
                .parameter( 1 )

Going straight to a method parameter doesn't make sense of course. I'm sure it could be resolved (by having multiple `ContainerElementConstraintMappingContext` interfaces, one for property/parameter/return-value.

But I couldn't make it work impl-wise so I gave up on this eventually (a runtime exception is raised when calling the methods in an unsupported location). Should someone else have the energy to give it a try, I'd be happy to merge it as a follow-up, otherwise I think it's good enough.

gunnarmorling:
Btw. that's another project idea I had ;) Define the states of such an internal DSL and the valid transitions (ie. allowed method calls) in some abstract means (external DSL?) and then have a generator for creating the required interfaces for only exposing the allowed ops in each state and class bodies to be implemented. Just dreaming of this, though :)

gunnarmorling:
Rebased. Added one more commit more geared towards HV-1245, too, @gsmet.

gsmet:
> Alternatively, one could split it up into typeArgument() (generic containers) and componentType() (arrays). The names would be a bit better to comprehend, but both would do the same, so I don't like it too much.

So I tend to agree that splitting it would make the names clearer but as it's the exact same thing, it feels weird to have 2 different names. So I concur with you that keeping one name is better.

> Should someone else have the energy to give it a try, I'd be happy to merge it as a follow-up, otherwise I think it's good enough.

Agreed, it's good enough as is.

I added one commit to reindent a test and removed the last commit. Expect a follow-up on this very subject.

PR #691, opened 2017-03-03
HV-1270 Throw an exception if a composed annotation has, for one given

gsmet:
constraint type, a direct annotation and annotations in a List container

 * https://hibernate.atlassian.net/browse/HV-1270

No tests. The test is part of the TCK.


gunnarmorling:
Looks good, rebased and applied. Added one commit for using the class formatter in the log message. Thanks, @gsmet!

PR #693, opened 2017-03-08
HV-1275 Annotation processor should detect if composed annotation has direct and List annotations of the same kind

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1275

Adding check for mix constraint declaration on a composed annotation


gunnarmorling:
Rebased and applied. Pushed one more commit for some clean-up. Thanks, @marko-bekhta!

PR #695, opened 2017-03-09
HV-1277 Improve the performances of HV 6

gsmet:
* https://hibernate.atlassian.net/browse/HV-1277

This is a base of discussion. I already know you won't like some of the changes but it will at least allow to discuss them.

The second commit is what makes a BIG difference.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
@@ -32,6 +32,8 @@ import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintVa
 import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager;
 import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintViolationCreationContext;
 import org.hibernate.validator.internal.engine.path.PathImpl;
+import org.hibernate.validator.internal.metadata.BeanMetaDataManager;
+import org.hibernate.validator.internal.metadata.aggregated.BeanMetaData;
 import org.hibernate.validator.internal.metadata.core.MetaConstraint;
 import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
 import org.hibernate.validator.internal.util.IdentitySet;
@@ -47,6 +49,7 @@ import org.hibernate.validator.internal.util.logging.LoggerFactory;
  * @author Hardy Ferentschik
  * @author Emmanuel Bernard
  * @author Gunnar Morling
+ * @author Guillaume Smet
  */
 public class ValidationContext<T> {
 
@@ -66,6 +69,11 @@ public class ValidationContext<T> {
 	 * The root bean class of the validation.
 	 */
 	private final Class<T> rootBeanClass;
+
+	/**
+	 * The metadata of the root bean.
+	 */
+	private final BeanMetaData<T> rootBeanMetaData;
 
 	/**
 	 * The method of the current validation call in case of executable validation.
@@ -147,6 +155,7 @@ public class ValidationContext<T> {
 			boolean failFast,
 			T rootBean,
 			Class<T> rootBeanClass,
+			BeanMetaData<T> rootBeanMetaData,
 			Executable executable,
 			Object[] executableParameters,
 			Object executableReturnValue) {
@@ -160,6 +169,7 @@ public class ValidationContext<T> {
 
 		this.rootBean = rootBean;
 		this.rootBeanClass = rootBeanClass;
+		this.rootBeanMetaData = rootBeanMetaData;
 		this.executable = executable;
 		this.executableParameters = executableParameters;
 		this.executableReturnValue = executableReturnValue;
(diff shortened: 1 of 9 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -369,7 +372,7 @@ private ValidationOrder determineGroupValidationOrder(Class<?>[] groups) {
 			return Collections.emptySet();
 		}
 
-		BeanMetaData<U> beanMetaData = beanMetaDataManager.getBeanMetaData( valueContext.getCurrentBeanType() );
+		BeanMetaData<U> beanMetaData = valueContext.getCurrentBeanMetaData();
```
gunnarmorling: Nice.

engine/src/main/java/org/hibernate/validator/internal/engine/ValueContext.java
```diff
@@ -93,6 +124,13 @@ public final T getCurrentBean() {
 		return currentBeanType;
 	}
 
+	public final BeanMetaData<T> getCurrentBeanMetaData() {
+		if ( currentBeanType != null && currentBeanMetaData == null ) {
```
gunnarmorling: Why is it done lazily? Seems we'll need it in any case?

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
```diff
@@ -118,18 +121,21 @@ public boolean equals(Object o) {
...
-			return false;
-		}
 
 		return true;
 	}
 
+	private static int buildHashCode(ConstraintDescriptorImpl<?> constraintDescriptor, ConstraintLocation location, List<ValueExtractorDescriptor> valueExtractorDescriptors) {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ( constraintDescriptor != null ? constraintDescriptor.hashCode() : 0 );
```
gunnarmorling: Are the null checks really needed?

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ValueExtractorDescriptor.java
```diff
@@ -98,7 +98,7 @@ public Key getKey() {
 		return key;
 	}
 
-	public Type getExtractedType() {
+	public Class<?> getExtractedType() {
```
gunnarmorling: Seems it was wise to limit it to the wildcard parameterization in the spec :)

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -167,17 +167,19 @@ public ValidatorImpl(ConstraintValidatorFactory constraintValidatorFactory,
 	public final <T> Set<ConstraintViolation<T>> validate(T object, Class<?>... groups) {
 		Contracts.assertNotNull( object, MESSAGES.validatedObjectMustNotBeNull() );
 
-		if ( !beanMetaDataManager.isConstrained( object.getClass() ) ) {
```
gsmet: The issue is not `isConstrained` by itself, that's why I kept it and there's still a usage of it.

The issue was that we used `beanMetaDataManager.isConstrained` which got the bean metadata from the concurrent map and then we got again the bean metadata from the concurrent map a few lines below (either really a few lines below or hidden in a method call a few lines below). We should get the bean metadata from the concurrent map only once.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -68,6 +76,12 @@
 	private final Class<T> rootBeanClass;
 
 	/**
+	 * The metadata of the root bean.
+	 */
+	@Immutable
```
gsmet: Yeah right, I used it as a marker to say that this value was not supposed to change during the object lifecycle and it was supposed to be "final" as the other ones. I'll remove them.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -196,6 +214,13 @@ public T getRootBean() {
 		return rootBeanClass;
 	}
 
+	public BeanMetaData<T> getRootBeanMetaData() {
+		if ( rootBeanClass != null && rootBeanMetaData == null ) {
```
gsmet: So in this case, it's an error. It's just that I started with `ValueContext` and then copied the behavior from `ValueContext` to `ValidationContext`. I'll change it here.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
gsmet: Not really. I'll remove them.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
gunnarmorling: Got it. In OGM we have `@EffectivelyImmutable` for a similar purpose (though it applies to initialization of the field during construction time, e.g. a post-construct callback, not lazy-init as done here). If needed we could add it to HV, too.

We don't need to handle those stereotypes with religious believe, it's just kind of an (open) experiment whether its helpful to us or not.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gsmet: Mmmh, but in fact, we can get rid of these too. I got too conservative and wanted to keep the `ValidationContext` initialization after the `hasConstraints` check but it's a bad idea.

I changed it and got rid of `isConstrained()`.

gsmet:
Force pushed an update that addresses all your concerns.

gunnarmorling:

PR #696, opened 2017-03-10
Some small improvements related to HV-1245

gunnarmorling:


engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraint.java
```diff
@@ -236,22 +235,53 @@ public boolean isSuccess() {
...
+			return new TypeParameterAndExtractor( typeParameter, valueExtractorDescriptor );
+		}
+
+		@Override
+		public String toString() {
+			return "TypeParameterAndExtractor [typeParameter=" + typeParameter + ", valueExtractorDescriptor=" + valueExtractorDescriptor + "]";
+		}
+	}
+
+	private static final class ValueExtractionPathNode {
```
gsmet: As discussed on HipChat, it would be nice to have a specific implementation for the common case where we have only one node.

Other than that, it looks good to me.

gunnarmorling:
@gsmet Reworked that last commit and force pushed.

PR #699, opened 2017-03-15
HV-1278 Add performance test for multi-level container element validation

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1278

There are multiple different tests added. I'd probably leave just `testMultiLevelPreGeneratedValidation` this one with "UnConstrained" version of it, so no additional time is spent on creating maps and lists for each iteration. What would you say ?

Also didn't use `@Email` as it's part of HV and there's a test profile for bVal.

### DIFF
--- a/performance/src/main/java/org/hibernate/validator/performance/BenchmarkRunner.java
+++ b/performance/src/main/java/org/hibernate/validator/performance/BenchmarkRunner.java
@@ -6,10 +6,10 @@
  */
 package org.hibernate.validator.performance;
 
-import java.util.Arrays;
-import java.util.List;
+import java.util.stream.Stream;
 
 import org.hibernate.validator.performance.cascaded.CascadedValidation;
+import org.hibernate.validator.performance.simple.MultiLevelContainerValidation;
 import org.hibernate.validator.performance.simple.SimpleValidation;
 import org.hibernate.validator.performance.statistical.StatisticalValidation;
 import org.openjdk.jmh.results.format.ResultFormatType;
@@ -29,10 +29,11 @@ import org.openjdk.jmh.runner.options.OptionsBuilder;
  */
 public final class BenchmarkRunner {
 
-	private static final List<Class<?>> DEFAULT_TEST_CLASSES = Arrays.asList(
+	private static final Stream<Class<?>> DEFAULT_TEST_CLASSES = Stream.of(
 			SimpleValidation.class,
 			CascadedValidation.class,
-			StatisticalValidation.class
+			StatisticalValidation.class,
+			MultiLevelContainerValidation.class
 	);
 
 	private BenchmarkRunner() {
@@ -49,7 +50,7 @@ public final class BenchmarkRunner {
 			builder.resultFormat( ResultFormatType.JSON );
 		}
 		if ( commandLineOptions.getIncludes().isEmpty() ) {
-			DEFAULT_TEST_CLASSES.stream().forEach( testClass -> builder.include( testClass.getName() ) );
+			DEFAULT_TEST_CLASSES.forEach( testClass -> builder.include( testClass.getName() ) );
 		}
 
 		Options opt = builder.build();
--- a/performance/src/main/java/org/hibernate/validator/performance/BenchmarkRunner.java
+++ b/performance/src/main/java/org/hibernate/validator/performance/BenchmarkRunner.java
@@ -6,12 +6,13 @@
  */
 package org.hibernate.validator.performance;
 
+import java.util.Objects;
 import java.util.stream.Stream;
 
 import org.hibernate.validator.performance.cascaded.CascadedValidation;
-import org.hibernate.validator.performance.simple.MultiLevelContainerValidation;
 import org.hibernate.validator.performance.simple.SimpleValidation;
 import org.hibernate.validator.performance.statistical.StatisticalValidation;
+
 import org.openjdk.jmh.results.format.ResultFormatType;
 import org.openjdk.jmh.runner.Runner;
 import org.openjdk.jmh.runner.RunnerException;
@@ -29,12 +30,19 @@ import org.openjdk.jmh.runner.options.OptionsBuilder;
  */
 public final class BenchmarkRunner {
 
-	private static final Stream<Class<?>> DEFAULT_TEST_CLASSES = Stream.of(
-			SimpleValidation.class,
-			CascadedValidation.class,
-			StatisticalValidation.class,
-			MultiLevelContainerValidation.class
-	);
+	private static final Stream<? extends Class<?>> DEFAULT_TEST_CLASSES = Stream.of(
+			SimpleValidation.class.getName(),
+			CascadedValidation.class.getName(),
+			StatisticalValidation.class.getName(),
+			/* Benchmarks specific to Bean Validation 2.0
+			*
+			* Cannot simply reference BV 2.0 test classes as they may contain incompatible functionality
+			* compared to previous versions of BV. Which would not allow to compile these benchmarks for
+			* those previous versions. So they are located in separate source folder which is only included
+			* during compilation for BV 2.0 implementations.
+			*/
+			"org.hibernate.validator.performance.multilevel.MultiLevelContainerValidation"
+	).map( BenchmarkRunner::classForName ).filter( Objects::nonNull );
 
 	private BenchmarkRunner() {
 	}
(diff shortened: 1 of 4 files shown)


gsmet:
@marko-bekhta I think we should only keep `testMultiLevelPreGeneratedValidation`. I don't think there is a point to measure the case where we don't have any constraint or am I wrong?

I'll review this particular case and will revisit my review if we end up keeping more!

performance/src/main/java/org/hibernate/validator/performance/simple/MultiLevelContainerValidation.java
```diff
@@ -0,0 +1,273 @@
...
+		Set<ConstraintViolation<OtherMapContainer>> violations = state.validator.validate(
+				new OtherMapContainer( prepareOtherTestData( state.random.nextInt( MAX_MAP_ENTRIES ), state.random.nextInt( MAX_LIST_ENTRIES ) ) )
+		);
+		bh.consume( violations );
+	}
+
+	/**
+	 * Contains test classes and test data generator.
+	 */
+	static final class MultiLevelContainerDataGenerator {
```
gsmet: I think you can make this private? Same for the other classes below.

performance/src/main/java/org/hibernate/validator/performance/simple/MultiLevelContainerValidation.java
```diff
@@ -0,0 +1,273 @@
...
+				chars[i] = (char) RANDOM.nextInt();
+			}
+			return String.valueOf( chars );
+		}
+
+		/**
+		 * Model classes.
+		 */
+
+		static class MapContainer {
```
gsmet: I think these classes should be outside the generator class. It's weird to include them in the generator class.

performance/src/main/java/org/hibernate/validator/performance/simple/MultiLevelContainerValidation.java
```diff
@@ -0,0 +1,273 @@
...
+
+			private Map<@NotNull Optional<@Valid Cinema>, List<@NotNull @Valid EmailAddress>> map;
```
gsmet: Let's make it final (just a good habit to take).

performance/src/main/java/org/hibernate/validator/performance/simple/MultiLevelContainerValidation.java
```diff
@@ -0,0 +1,273 @@
...
+			}
+
+			public Map<Cinema, List<EmailAddress>> getMap() {
+				return map;
+			}
+		}
+
+		static class Cinema {
+
+			String name;
```
gsmet: private + final (same below).

performance/src/main/java/org/hibernate/validator/performance/simple/MultiLevelContainerValidation.java
```diff
@@ -0,0 +1,273 @@
...
+
+			Reference<@Valid Visitor> visitor;
+
+			Cinema(String name, Reference<Visitor> visitor) {
+				this.name = name;
+				this.visitor = visitor;
+			}
+
+			static Cinema get() {
```
gsmet: Maybe, call it `generate` instead of `get`? It's more to the point I think. Same below for `Visitor`.

performance/src/main/java/org/hibernate/validator/performance/simple/MultiLevelContainerValidation.java
```diff
@@ -0,0 +1,273 @@
...
+			static Visitor get() {
+				return new Visitor( randomString() );
+			}
+		}
+
+		static class EmailAddress {
+
+			// not using @Email as it will require HV as a dependency and these tests can be run against bVal as well
+			@NotNull
+			@Pattern(regexp = "(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|\"(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21\\x23-\\x5b\\x5d-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])*\")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\\x01-\\x08\\x0b\\x0c\\x0e-\\x1f\\x21-\\x5a\\x53-\\x7f]|\\\\[\\x01-\\x09\\x0b\\x0c\\x0e-\\x7f])+)\\])")
```
gsmet: I would say, just keep the `@NotNull` constraint. The point is to exercise the validator code, not to have an expensive constraint checked. I would have used `@NotBlank` for a change but it's not part of the spec either :).

performance/src/main/java/org/hibernate/validator/performance/simple/MultiLevelContainerValidation.java
```diff
@@ -0,0 +1,273 @@
...
+			private final String email;
+
+			public EmailAddress(String value) {
+				this.email = value;
+			}
+
+			public static EmailAddress get() {
```
gsmet: Same remark about s/get/generate/ for this method and the other one below.

marko-bekhta:
@gsmet Thanks for the review, I'll make the changes later toady.
As for the test without constraints - it was meant to be something to compare the results to. But thinking about it now - it doesn't make much sense :) That's what sometimes happens when writing code late at night :)

gsmet:
@marko-bekhta I pushed an additional commit to your branch, nothing much.

@gunnarmorling we have a major problem with this: it does not compile with BV 1.1 anymore so we can't build the performances tests for the older versions or for BVal anymore.

Maybe we should separate the performance tests in 2 submodules? One for BV 1.x and one for BV 2?

gunnarmorling:
> Maybe we should separate the performance tests in 2 submodules? One for BV 1.x and one for BV 2?

Yes, that or we have a separate source folder for the new ones, e.g. _src/main/java-bv2_. This source folder can be added for the current profile using the build-helper-maven-plugin. The only issue is the reference to the classes from `BenchmarkRunner`, this would have to be changed to deal with names and ignore any benchmarks non present.

Open to both approaches; having everything in one module is nicer from an execution point-of-view, having a single report etc., but the name-based references are a bit more error prone for misses when renaming classes.

marko-bekhta:
Maybe name-based references can be "fixed" with an annotation on the test
classes and some simple annotation processor which will provide a list of
all found annotated tests (as a return value of some generated by AP
method)?

2017-03-20 18:10 GMT+01:00 Gunnar Morling <notifications@github.com>:

gsmet:
I'd go for Gunnar's approach as a good first step. It's not as if we change these names often.

performance/src/main/java/org/hibernate/validator/performance/BenchmarkRunner.java
```diff
@@ -29,12 +26,12 @@
...
 public final class BenchmarkRunner {
 
-	private static final Stream<Class<?>> DEFAULT_TEST_CLASSES = Stream.of(
-			SimpleValidation.class,
-			CascadedValidation.class,
-			StatisticalValidation.class,
-			MultiLevelContainerValidation.class
-	);
+	private static final Stream<? extends Class<?>> DEFAULT_TEST_CLASSES = Stream.of(
+			"org.hibernate.validator.performance.simple.SimpleValidation",
```
gsmet: For all the BV 1 compatible classes, you could use `<Class>.class.getName()` so that we limit the issue with renaming to the BV 2 classes. it's not particularly efficient but it's not that important as it's just a startup cost.

performance/src/main/java/org/hibernate/validator/performance/BenchmarkRunner.java
```diff
@@ -29,11 +26,12 @@
...
-	private static final List<Class<?>> DEFAULT_TEST_CLASSES = Arrays.asList(
-			SimpleValidation.class,
-			CascadedValidation.class,
-			StatisticalValidation.class
-	);
+	private static final Stream<? extends Class<?>> DEFAULT_TEST_CLASSES = Stream.of(
+			"org.hibernate.validator.performance.simple.SimpleValidation",
+			"org.hibernate.validator.performance.cascaded.CascadedValidation",
+			"org.hibernate.validator.performance.statistical.StatisticalValidation",
+			"org.hibernate.validator.performance.multilevel.MultiLevelContainerValidation"
```
gsmet: Maybe add a comment just before starting the BV 2 classes:
`// Benchmarks specific to Bean Validation 2.0`
It would be even more important to explain why they are not added in the same way as the others.

marko-bekhta: @gsmet yes, no problem, just give me a sec :)

gsmet:
Merged, thanks!

PR #700, opened 2017-03-15
HV-1283 (5.4)

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1283

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/resourceloading/PlatformResourceBundleLocator.java
+++ b/engine/src/main/java/org/hibernate/validator/resourceloading/PlatformResourceBundleLocator.java
@@ -6,7 +6,11 @@
  */
 package org.hibernate.validator.resourceloading;
 
+import static org.hibernate.validator.internal.util.CollectionHelper.newHashSet;
+import static org.hibernate.validator.internal.util.logging.Messages.MESSAGES;
+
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/hibernate/validator/resourceloading/PlatformResourceBundleLocator.java
```diff
@@ -164,21 +166,42 @@ private ResourceBundle loadBundle(ClassLoader classLoader, Locale locale, String
...
+			Method getModule = run( GetMethod.action( Class.class, "getModule" ) );
+			// not on Java 9
+			if ( getModule == null ) {
+				return true;
+			}
+
+			// on Java 9, check whether HV is a named module
+			Object module = getModule.invoke( PlatformResourceBundleLocator.class );
+			Method isNamedMethod = run( GetMethod.action( module.getClass(), "isNamed" ) );
+			boolean isNamed = (Boolean) isNamedMethod.invoke( module );
```
gunnarmorling: I tested this manually, which I think - while not perfect - is good enough. And even if things went wrong, the impact is low, so I think that's ok.

gsmet:
Pushed to 5.4 and master. Thanks.

PR #701, opened 2017-03-15
Backports for 5.2.5

gsmet:
@gunnarmorling so here is the backport proposal for 5.2

I took the liberty to include HV-1140 as it really is a bugfix and I don't see how it could lead to a regression. If you feel strongly against it, I'll remove it.

I squashed HV-1280 to have only one commit for product review (I couldn't do it in 5.3 or 5.4 as I had already pushed the first commit).

I'll do the related work on JIRA once we agree on this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Typically *NIX text editors, by default, append '~' to files on saving to make backups
 *~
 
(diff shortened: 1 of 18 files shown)


gunnarmorling:
Merged; thanks.

PR #702, opened 2017-03-16
HV-1285 Use the new toImmutable helpers where it makes sense

gsmet:
Also add the @Immutable marker where it makes sense.

 * https://hibernate.atlassian.net/browse/HV-1285


gunnarmorling:
Rebased and applied. Nice improvement, @gsmet!

PR #703, opened 2017-03-16
HV-1286 Some improvements for the performance tests

gsmet:
* https://hibernate.atlassian.net/browse/HV-1286


marko-bekhta:
Limit to throughput is a nice change :)
as it was taking ages to run this in All mode.

PR #704, opened 2017-03-17
HV-1287 Upgrading to CheckStyle 7.6

gunnarmorling:
The wrong equals() methods are a good spot indeed.

https://hibernate.atlassian.net/browse/HV-1287


gsmet:
Merged, thanks!

PR #706, opened 2017-03-22
HV-1288 Don't use org.testng.AssertJUnit

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1288

Also while looked at this - changed a few static fail imports to `org.testng.Assert.fail`

Wasn't able to configure `ImportControl` for `AssertJUnit` as in case only a class is disallowed checkstyle is not picking up static imports from that class. And there seems to be no option to configure these static imports. So I've just added a Regexp check to look for static imports of `org.testng.AssertJUnit` methods.


gsmet:
Hmmmm. So it hurts to say that **again** but Gunnar was probably right from the beginning :).

If we end up having to use a regexp, we probably use the wrong tool (or in this case, I think the tool is not very well designed for its purpose).

Could you take a look at https://github.com/policeman-tools/forbidden-apis ? Looks like the signature file format is simple: https://github.com/policeman-tools/forbidden-apis/blob/master/src/main/resources/de/thetaphi/forbiddenapis/signatures/jdk-deprecated-1.6.txt . You can disable a method or a class. If it makes a better work, we should use it instead. Note that in passing, we might be able to have less checkstyle configuration files as at least one of them is just to enable JUnit for parts where we rely on JUnit. Note that even if we use JUnit, we should disable the `AssertJUnit` class so maybe we need one signature file for generally disabled API (might be handy in the future) and one for JUnit that we enable only for projects relying on TestNG.

I was a bit worried about how we would share the signature files but it looks like we can reference a file in a jar even if not documented (so we could put the necessary signature files in the build-config jar). See the end of the comment here: https://github.com/policeman-tools/forbidden-apis/blob/master/src/main/java/de/thetaphi/forbiddenapis/maven/AbstractCheckMojo.java#L69 .

marko-bekhta:
:)
Yes, I'll look into this later today

marko-bekhta:
Sooooooo :) here's what I've done:

- Configured forbidden-api plugin with some bundled signature files (`jdk-unsafe`, `jdk-deprecated`, `jdk-non-portable`)
- Removed checkstyle rule for `System.out` and replaced it with forbidden-api bundled `jdk-system-out`
- Added `SuppressForbidden` annotation, so it can be used to suppress some "illegal" classes and methods where it's really needed. Not sure that I've chosen the best place for it though ...
- Removed all different checkstyle configurations and left only one with all the rules
- Removed license header checks from checkstyle configs (so there can be one checkstyle config for all modules) and replaced those checks with `license-maven-plugin`. Also added headers for files that were missing them
- Added `maven-remote-resources-plugin` so that all config related resources from build-config module can be easily used in configurations of other plugins.
- Also did some "cleaning" of issues reported by forbidden-apis plugin. Most of them are using something without specifying Locale/TimeZone/Charset

build-config/src/main/resources/java-header-style.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+  ~ Hibernate Validator, declare and validate application constraints
+  ~
+  ~ License: Apache License, Version 2.0
+  ~ See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+  -->
+<additionalHeaders>
```
gsmet: Oh nice! I missed these configuration files when I used it on the BV TCK.

distribution/pom.xml
```diff
@@ -164,6 +164,10 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
gsmet: Is there a reason why you enable it per project? There are projects where it doesn't make sense?

pom.xml
```diff
@@ -184,6 +184,8 @@
 
         <!-- Forbidden API related properties -->
         <forbiddenapis-junit.path>forbidden-junit.txt</forbiddenapis-junit.path>
+
+        <hibernate-validator-parent.basedir>.</hibernate-validator-parent.basedir>
```
gsmet: I would have expected it to be redefined in each submodule? At least that's what I needed to do for BV TCK. Do you have a clever trick to avoid that?

annotation-processor/src/main/java/org/hibernate/validator/ap/util/MessagerAdapter.java
```diff
@@ -114,7 +114,7 @@ private void report(ConstraintCheckIssue issue, Kind kind) {
 		String message = errorMessages.getString( issue.getMessageKey() );
 
 		if ( issue.getMessageParameters() != null ) {
-			message = MessageFormat.format( message, issue.getMessageParameters() );
+			message = new MessageFormat( message, Locale.ENGLISH ).format( issue.getMessageParameters() );
```
gsmet: Is there a reason not to use `String.format` here?

pom.xml
marko-bekhta: this is removed in later commits with adding of `maven-remote-resources-plugin`
at first I was thinking about adding it to every pom, but I got lazy :) and done it a bit differently in the end. This `maven-remote-resources-plugin` do something similar to `maven-dependency` unpack, and places those configs in target so each module should have all configs available in target folder ...

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalWithPrivateGetterTest.java
```diff
@@ -41,7 +42,7 @@ public void testConstraintCanBePlacedOnPrivateGetter() {
 	public void testConstraintOnPrivateGetterReturnsConstraintViolation() {
 		Validator validator = getValidator();
 
-		LocalDate[][] testData = { { LocalDate.MAX, LocalDate.MIN }, { LocalDate.now(), null }, { null, null } };
+		LocalDate[][] testData = { { LocalDate.MAX, LocalDate.MIN }, { LocalDate.now( ZoneId.systemDefault() ), null }, { null, null } };
```
gsmet: I'm a bit sceptic about this type of changes. @gunnarmorling what's your take on this? I'm not worried about these occurrences but more about how it will be frustrating when we write new tests.

annotation-processor/src/main/java/org/hibernate/validator/ap/util/MessagerAdapter.java
marko-bekhta: I though about it ... the reason why I left it as is is because the messages here use `{0}` such things for inserting parameters. And it will require to update messages to use something like `%s` or `%1$s`. And I wasn't sure if that was a _safe_ change to make... if it is, I'd say - let's change it to use String.format

pom.xml
```diff
@@ -562,15 +560,41 @@
...
+                            <groupId>org.hibernate.validator</groupId>
+                            <artifactId>hibernate-validator-build-config</artifactId>
+                            <version>${project.version}</version>
+                        </dependency>
+                    </dependencies>
+                    <configuration>
+                        <resourceBundles>
+                            <resourceBundle>org.hibernate.validator:hibernate-validator-build-config:${project.version}</resourceBundle>
+                        </resourceBundles>
+                        <!--<outputDirectory></outputDirectory>-->
```
gsmet: Maybe remove the comment?

gsmet: Mmmh maybe we could use a directory which makes more sense, such as build-config-resources or something similar?

pom.xml
```diff
@@ -562,15 +560,41 @@
                     </executions>
                 </plugin>
                 <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
```
gsmet: You can omit the groupId here. Not sure it's done for the existing ones but we try to do it for the new ones.

pom.xml
```diff
@@ -562,15 +560,41 @@
...
+                    <artifactId>maven-remote-resources-plugin</artifactId>
```
gsmet: Interesting. I didn't know this plugin and it's in fact very useful... Nice trick.

build-config/src/main/resources/forbidden-common.txt
```diff
@@ -0,0 +1,115 @@
...
+# License: Apache License, Version 2.0
+#  See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+
+# This file contains forbidden API signatures common to all modules
+
+@ignoreUnresolvable
+@defaultMessage Shouldn't be used
+
+################################################################################################################
+# Some packages that should be used
```
gsmet: + not :)

build-config/src/main/resources/forbidden-common.txt
```diff
@@ -0,0 +1,115 @@
...
+
+################################################################################################################
+# org.testng.Assert#fail() should be used
+org.assertj.core.api.Fail#fail()
+org.testng.FileAssert#fail()
+org.assertj.core.api.Assertions#fail()
+
+################################################################################################################
+# Nobody should be using StringBuffer anymore
+java.lang.StringBuffer
```
gsmet: You need to define all the methods, even if you have defined the types?

build-config/src/main/resources/forbidden-junit.txt
```diff
@@ -0,0 +1,49 @@
...
+@ignoreUnresolvable
+@defaultMessage JUnit is not allowed, use TestNG instead
+
+################################################################################################################
+#packages
+org.junit.**
+
+################################################################################################################
+#Asserts
+org.testng.AssertJUnit
```
gsmet: Same question here about class vs list of methods.

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/JavaFXObservableValueValueExtractorTest.java
```diff
@@ -111,6 +112,7 @@ public void testJavaFXPropertySkipUnwrapping() {
 	}
 
 	@SuppressWarnings("unused")
+	@SuppressForbidden( reason = "Test class that uses JavaFx types.")
```
gsmet: Isn't it possible to add it globally on the class?

annotation-processor/src/main/java/org/hibernate/validator/ap/util/MessagerAdapter.java
gsmet: Yeah OK, let it alone, it's safer.

gsmet:
@marko-bekhta very interesting work, kudos! I added a couple of comments inline.

About the class vs list of methods question, **if** we have to provide the list of methods, we would need to document how to generate this list.

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalWithPrivateGetterTest.java
marko-bekhta: @gsmet  +1 
I'm not very happy about them as well. That's why I've done them all in a separate commit so it can be removed if we decide so. These things comes from those `jdk-unsafe`, `jdk-deprecated`, `jdk-non-portable`. They seems to be useful to have in general (judging  by the names).

build-config/src/main/resources/forbidden-common.txt
marko-bekhta: Not sure about that ... I'll check if it'll work with the class only, and let you know
There's not much about the format of these files (or in other words - I haven't found anything about the format in the documentation of the plugin :) so I just looked at format of provided files, but I haven't checked for such cases ...)

marko-bekhta: ahh.. just remembered that one of the plugins (either forbidden-apis or check-style illegal imports) wasn't picking up the `java.lang.StringBuffer` as it's from the "automatically imported package" ... But not sure about which one it was of these two  ... 
Again - I'll play with different cases and I'll let you what works and what's not :)

distribution/pom.xml
marko-bekhta: There's no licenses in documentation module. Only pom file has a header and all others - don't.

marko-bekhta:
I've tested the "class vs method list" thing - it's enough to just use the class. So I've dropped all methods from the files. 
Also rebased to master and applied a fix for missing headers in new xmls.

gsmet:
OK, looks good. Waiting for @gunnarmorling 's position on adding the locale and other things everywhere.

engine/src/main/java/org/hibernate/validator/internal/SuppressForbidden.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation to be used for suppression of forbidden apis errors.
+ */
+@Retention(RetentionPolicy.CLASS)
```
gunnarmorling: So it actually has to be CLASS retention, SOURCE isn't enough?

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ObservableValueValueExtractor.java
```diff
@@ -26,6 +28,7 @@ private ObservableValueValueExtractor() {
 	}
 
 	@Override
+	@SuppressForbidden(reason = "ObservableValue should be allowed in this case")
```
gunnarmorling: Why is it forbidden to begin with?

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/NestedTypeArgumentsValueExtractorTest.java
```diff
@@ -287,6 +289,7 @@ private static MapOfListsWithAutomaticUnwrapping invalidListElementBar() {
 	}
 
 	@SuppressWarnings({ "unchecked", "unused" })
+	@SuppressForbidden( reason = "Test class that uses JavaFx types.")
```
gunnarmorling: Why is JavaFX forbidden by default?

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalWithPrivateGetterTest.java
gunnarmorling: I'd personally _not_ forbid something like `now()`. It is legit for our testing here. Generally I'd strongly prefer to not go overboard with this kind of checks; it's not that we had an actual problem here.

engine/src/test/resources/org/hibernate/validator/test/internal/xml/containerelementtype/getter-canDeclareContainerElementTypeConstraintsForListContainingArrayType-mapping.xml
```diff
@@ -1,3 +1,9 @@
+<!--
```
gunnarmorling: On the headers of these files, we had left them out intentionally; but then there is no harm by having this small (constant) header, so why not.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ObservableValueValueExtractor.java
marko-bekhta: I believe it comes from `jdk-non-portable`

engine/src/test/resources/org/hibernate/validator/test/internal/xml/containerelementtype/getter-canDeclareContainerElementTypeConstraintsForListContainingArrayType-mapping.xml
marko-bekhta: I did add those as their absence was causing the build to fail. But we can just add an exclude rule for them, if it's preferred not to have headers here. Just say how it should be :)

engine/src/main/java/org/hibernate/validator/internal/SuppressForbidden.java
marko-bekhta: Not sure why it cannot be SOURCE, but I've checked here:
https://github.com/policeman-tools/forbidden-apis/blob/master/src/main/java/de/thetaphi/forbiddenapis/SuppressForbidden.java
and it has `RetentionPolicy.CLASS` for it...

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalWithPrivateGetterTest.java
marko-bekhta: I do agree with you. I've looked at those bundled files and these Locale/TimeZone/Charset things comes from `jdk-unsafe`. As there seems to be nothing else (except locals and timezones,charsets) in those unsafe files I'd say it can be _safely_ :) removed.
Link to the signature files if needed: 
https://github.com/policeman-tools/forbidden-apis/tree/master/src/main/resources/de/thetaphi/forbiddenapis/signatures

test-utils/src/main/java/org/hibernate/validator/testutil/DescriptorAssert.java
```diff
@@ -54,7 +55,7 @@ public void containsConversion(Class<?> from, Class<?> to) {
 			}
 
 			if ( !foundMatchingConversion ) {
-				fail( String.format( "<%s> does not contain a conversion from <%s> to <%s>.", actual, from, to ) );
+				fail( String.format( Locale.getDefault(), "<%s> does not contain a conversion from <%s> to <%s>.", actual, from, to ) );
```
gunnarmorling: @marko-bekhta Is this still a left-over? I.e. are you planning to revert this change as per our discussion above? Same question for some more places where `Locale.getDefault()` is used now.

build-config/src/main/resources/forbidden-common.txt
gunnarmorling: Is there any further action needed here?

distribution/pom.xml
gunnarmorling: We can still enable it globally if that's simpler.

engine/src/main/java/org/hibernate/validator/internal/SuppressForbidden.java
gunnarmorling: ok.

engine/src/main/java/org/hibernate/validator/internal/engine/cascading/ObservableValueValueExtractor.java
gunnarmorling: Can we disable this check then? It seems to do a bit too much.

engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/OptionalWithPrivateGetterTest.java
gunnarmorling: +1 for cutting down the number of changes here; let's limit to the actual issue at hand.

engine/src/test/resources/org/hibernate/validator/test/internal/xml/containerelementtype/getter-canDeclareContainerElementTypeConstraintsForListContainingArrayType-mapping.xml
gunnarmorling: No leave them in, now that they are there :)

build-config/src/main/resources/forbidden-common.txt
marko-bekhta: I've checked that mentioning just a class is enough and updated the commit with that change. So nothing else for this one

test-utils/src/main/java/org/hibernate/validator/testutil/DescriptorAssert.java
marko-bekhta: yes, I was waiting for your response on this one (I was just 90% sure if to remove them or not :) ) - I'll remove `jdk-unsafe` and `jdk-non-portable` from the config as well as that commit with changes caused by it.

gsmet:
@marko-bekhta I merged this one. Thanks for your hard work.

I reworked the commits so that they are a bit more atomic. I editorialized a bit the comments and also renamed the annotation to `@IgnoreForbiddenApisErrors` so that its purpose is more clear.

One thing I also fixed is that you can't use the maven-remote-resources plugin to get the license resources in the parent pom as build-config might not be deployed yet (or might not contain the resources if it's an old version). I included this in your commit.

That being said, I chose to go the variable way in a follow-up commit. Not really fond of special casing the parent.

PR #707, opened 2017-03-24
HV-1291 Support container element type constraints definition via XML

gsmet:
* https://hibernate.atlassian.net/browse/HV-1291

@gunnarmorling so this is my progress on container element type constraints definition via XML. I updated the validation-mapping-2.0.xsd here. Once we agree on it, I'll push a PR to beanvalidation-api.

I think it's feature and test complete (the tests are the same as for the programmatic API) except for the discussion we had about throwing an exception if XML constraints are defined and we try to define programmatic API or annotation based constraints. I think this can be treated as a follow-up. If you agree, I'll open a specific JIRA.


gunnarmorling:
> If you agree, I'll open a specific JIRA.

+1 Sounds reasonable. Looks good overall, going to merge it. Let's see what others say on the schema structure when proposing it for BV:

PR #708, opened 2017-03-29
HV-1292 Add support for the new constraints added to the spec

gsmet:
* https://hibernate.atlassian.net/browse/HV-1292
 * https://hibernate.atlassian.net/browse/HV-1293

HV-1293 should probably be backported to all supported releases.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/NegativeValidatorForNumber.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/NegativeValidatorForNumber.java
@@ -1,0 +1,89 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.constraintvalidators.bv;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.constraints.Negative;
+
+/**
+ * Check that the number being validated is negative.
+ *
+ * @author Hardy Ferentschik
+ * @author Xavier Sosnovsky
+ * @author Guillaume Smet
+ */
+public class NegativeValidatorForNumber implements ConstraintValidator<Negative, Number> {
+
+	private static final short SHORT_ZERO = (short) 0;
+	private static final byte BYTE_ZERO = (byte) 0;
+
+	private boolean strict;
+
+	@Override
+	public void initialize(Negative negative) {
+		this.strict = negative.strict();
+	}
+
+	@Override
+	public boolean isValid(Number value, ConstraintValidatorContext constraintValidatorContext) {
+		// null values are valid
+		if ( value == null ) {
+			return true;
+		}
+		// handling of NaN, positive infinity and negative infinity
+		else if ( value instanceof Double ) {
+			if ( (Double) value == Double.NEGATIVE_INFINITY ) {
+				return true;
+			}
+			else if ( Double.isNaN( (Double) value ) || (Double) value == Double.POSITIVE_INFINITY ) {
+				return false;
+			}
+		}
+		else if ( value instanceof Float ) {
+			if ( (Float) value == Float.NEGATIVE_INFINITY ) {
+				return true;
+			}
+			else if ( Float.isNaN( (Float) value ) || (Float) value == Float.POSITIVE_INFINITY ) {
+				return false;
+			}
+		}
+
+		int comparisonResult;
+		if ( value instanceof BigDecimal ) {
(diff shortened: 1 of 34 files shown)


gsmet:
Jenkins, retest this please

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/NegativeValidatorForNumber.java
```diff
@@ -0,0 +1,89 @@
...
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.constraints.Negative;
+
+/**
+ * Check that the number being validated is negative.
+ *
+ * @author Hardy Ferentschik
+ * @author Xavier Sosnovsky
```
gunnarmorling: Where is Xavier's name coming from?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/NegativeValidatorForNumber.java
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	@Override
+	public boolean isValid(Number value, ConstraintValidatorContext constraintValidatorContext) {
+		// null values are valid
+		if ( value == null ) {
+			return true;
+		}
+		// handling of NaN, positive infinity and negative infinity
+		else if ( value instanceof Double ) {
```
gunnarmorling: I reckon you won't like it, but what do you think of having dedicated validators for the individual type instaed of the huge instanceof switch here. I'd prefer that, as validator resolution is done only once, so we won't pay the price for checking the type for each validator invocation again and again.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/notempty/NotEmptyValidatorForArray.java
```diff
@@ -0,0 +1,36 @@
...
+	 * @param array the array to validate
+	 * @param constraintValidatorContext context in which the constraint is evaluated
+	 * @return returns {@code true} if the array is not {@code null} and the array is not empty
+	 */
+	@Override
+	public boolean isValid(Object[] array, ConstraintValidatorContext constraintValidatorContext) {
+		if ( array == null ) {
+			return false;
+		}
+		return Array.getLength( array ) > 0;
```
gunnarmorling: Interesting, what's the rationale for not using array.length?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/NegativeValidatorForNumber.java
gsmet: It's based on DecimalMin/Max validators so I let the name of the original authors.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/NegativeValidatorForNumber.java
gsmet: I don't have a strong opinion about it. But I think we should keep consistency. I mimicked DecimalMin/Max so if we decide to change it, we should probably do it globally in a later commit.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/notempty/NotEmptyValidatorForArray.java
gsmet: Consistency, it's based on the Size validators. But I agree we should use array.length instead. Pushed an additional commit to fix it for Size and NotEmpty.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/NegativeValidatorForNumber.java
gunnarmorling: Ah, yes. Makes sense.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/NegativeValidatorForNumber.java
gunnarmorling: +1 for splitting up both. Can be done as a follow-up, though, if you prefer.

gsmet: Yes, I prefer getting the release out of the door ASAP. Will create a new JIRA to track this.

gunnarmorling: +1

gsmet: https://hibernate.atlassian.net/browse/HV-1300

gunnarmorling:

PR #710, opened 2017-03-29
HV-1294 Fix OSGi configuration of the cdi module

gsmet:
* https://hibernate.atlassian.net/browse/HV-1294

Looks like the configuration of the manifest in the jar section was necessary to include the OSGi information in the manifest. Don't exactly know why but, this way, we are consistent with the engine module.

Probably something we should backport at least to 5.4, maybe to the other supported versions.

### DIFF
--- a/cdi/pom.xml
+++ b/cdi/pom.xml
@@ -141,6 +141,15 @@
                         <goals>
                             <goal>jar</goal>
                         </goals>
+                        <configuration>
+                            <archive>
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Rebased and applied. It should be backported to 5.4.

gsmet:
Done.

gunnarmorling:
Thx!

PR #714, opened 2017-04-03
HV-1302 @SafeHtml should support adding additional protocols

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1302

Added `additionalProtocols` to `@SafeHtml`.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
@@ -91,10 +91,29 @@ public @interface SafeHtml {
 		String[] attributes() default { };
 
 		/**
-		 * @return list of valid protocols.
+		 * @return list of tag attributes with corresponding allowed protocols which are whitelisted.
 		 * @since 6.0
 		 */
-		String[] protocols() default { };
+		Attribute[] additionalAttributesWithProtocols() default { };
+
+		/**
+		 * Allows to specify attribute with whitelisted protocols.
+		 * @since 6.0
+		 */
+		@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER })
+		@Retention(RUNTIME)
+		@Documented @interface Attribute {
+			/**
+			 * @return the attribute name to whitelist.
+			 */
+			String name();
+
+			/**
+			 * @return list of attribute protocols which are whitelisted.
+			 */
+			String[] protocols() default { };
+
+		}
 	}
 
 	/**
--- a/engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
@@ -60,7 +60,7 @@ public @interface SafeHtml {
 	String[] additionalTags() default { };
 
 	/**
-	 * @return Allows to specify additional whitelist tags with optional attributes.
+	 * @return Allows to specify additional whitelist tags with optional attributes and protocols.
 	 */
 	Tag[] additionalTagsWithAttributes() default { };
 
@@ -89,6 +89,12 @@ public @interface SafeHtml {
 		 * @return list of tag attributes which are whitelisted.
 		 */
 		String[] attributes() default { };
+
+		/**
+		 * @return list of valid protocols.
+		 * @since 6.0
+		 */
+		String[] protocols() default { };
 	}
 
 	/**
--- a/engine/src/test/java/org/hibernate/validator/test/cfg/ProgrammaticConstraintDefinitionsTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/cfg/ProgrammaticConstraintDefinitionsTest.java
@@ -75,12 +75,16 @@ public class ProgrammaticConstraintDefinitionsTest {
 		doProgrammaticTest( new SafeHtmlDef().whitelistType( SafeHtml.WhiteListType.NONE )
 				.additionalTagsWithAttributes( AnnotationFactory.create( tagDescriptor ) ), "<td class='class' id='tableId'>1234qwer</td>", 0 );
 
-		AnnotationDescriptor<SafeHtml.Tag> protocolDescriptor = new AnnotationDescriptor( SafeHtml.Tag.class );
-		protocolDescriptor.setValue( "name", "img" );
-		protocolDescriptor.setValue( "attributes", new String[]{ "src" } );
-		protocolDescriptor.setValue( "protocols", new String[]{ "data" } );
+		AnnotationDescriptor<SafeHtml.Tag.Attribute> attributeDescriptor = new AnnotationDescriptor( SafeHtml.Tag.Attribute.class );
+		attributeDescriptor.setValue( "name",  "src"  );
+		attributeDescriptor.setValue( "protocols", new String[]{ "data" } );
+
+		tagDescriptor = new AnnotationDescriptor( SafeHtml.Tag.class );
+		tagDescriptor.setValue( "name", "img" );
+		tagDescriptor.setValue( "additionalAttributesWithProtocols", new SafeHtml.Tag.Attribute[]{ AnnotationFactory.create( attributeDescriptor ) } );
+
 		doProgrammaticTest( new SafeHtmlDef().whitelistType( SafeHtml.WhiteListType.NONE )
-				.additionalTagsWithAttributes( AnnotationFactory.create( protocolDescriptor ) ), "<img src='data:image/png;base64,100101' />", 0 );
+				.additionalTagsWithAttributes( AnnotationFactory.create( tagDescriptor ) ), "<img src='data:image/png;base64,100101' />", 0 );
 	}
 
 	@Test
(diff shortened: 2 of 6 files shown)


engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
```diff
@@ -65,6 +65,13 @@
 	Tag[] additionalTagsWithAttributes() default { };
 
 	/**
+	 * @return Additional whitelist protocols which are allowed on top of the protocols specified by the
+	 * {@link #whitelistType()} for a specified tag's attribute. Adding a protocol will also whitelists the
+	 * given tag and it's attribute.
+	 */
+	Protocol[] additionalProtocols() default { };
```
gunnarmorling: We already got `additionalTagsWithAttributes()`; did you consider to enhance `@SafeHtml.Tag` instead of adding a new annotation?

`@Tag` accepts multiple attributes, so we'd had to add all protocols for all of them (if more than one attribute is specified), but that should be alright.

@oyvindhorneland, WDYT?

engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
```diff
@@ -65,6 +65,13 @@
 	Tag[] additionalTagsWithAttributes() default { };
 
 	/**
+	 * @return Additional whitelist protocols which are allowed on top of the protocols specified by the
+	 * {@link #whitelistType()} for a specified tag's attribute. Adding a protocol will also whitelists the
```
gunnarmorling: s/whitelists/whitelist/
s/it's/its/

engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
```diff
@@ -84,6 +91,29 @@
...
+	public @interface Protocol {
+		/**
+		 * @return the tag the URL protocol is for.
+		 */
+		String tag();
+
+		/**
+		 * @return the attribute key.
+		 */
+		String key();
```
gunnarmorling: Should we decide to go this route with a new annotation, it should be named `attribute()` for the sake of consistency. Also some `@since` tags on the new annotation and on `additionalProtocols()` would be good.

engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
oyvindhorneland: @gunnarmorling I agree. The name `additionalTagsWithAttributes` is maybe a bit off if you add `String[] protocols() default { };` to it, but that's probably preferable to adding the `@SafeHtml.Protocol`.

marko-bekhta: I'll remove the `Protocol` annotation then and make the corresponding changes.

gunnarmorling:
Cool, thanks!

oyvindhorneland:
@marko-bekhta Looks good and really great to see a request for enhancement addressed this fast!

engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
```diff
@@ -81,6 +81,12 @@
 		 * @return list of tag attributes which are whitelisted.
 		 */
 		String[] attributes() default { };
+
+		/**
+		 * @return list of valid protocols.
+		 * @since 5.4.2
```
gunnarmorling: Hum, adding this in a micro is a bit of an issue (micros shouldn't alter the API at all). Also we don't have plans for another (non-bugfix) 5.x reease at this time. I'd prefer to have this in 6 only for now. Should we then do another 5.x release, it'd be a candidate for backporting of course.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/SafeHtmlValidatorTest.java
```diff
@@ -170,4 +182,16 @@ public Foo(String source) {
 			this.source = source;
 		}
 	}
+
+	public static class Bar {
+		@SafeHtml(
+				whitelistType = WhiteListType.BASIC,
+				additionalTagsWithAttributes = @SafeHtml.Tag(name = "img", attributes = "src", protocols = { "data" })
```
gunnarmorling: Would be good to have more than one element in the test, so we know that works, too :)

engine/src/test/java/org/hibernate/validator/test/cfg/ProgrammaticConstraintDefinitionsTest.java
```diff
@@ -61,6 +63,19 @@ public void safeHtmlProgrammaticDefinition() {
...
+		tagDescriptor.setValue( "name", "td" );
+		tagDescriptor.setValue( "attributes", new String[]{ "class", "id" } );
+		doProgrammaticTest( new SafeHtmlDef().whitelistType( SafeHtml.WhiteListType.NONE )
+				.additionalTagsWithAttributes( AnnotationFactory.create( tagDescriptor ) ), "<td class='class' id='tableId'>1234qwer</td>", 0 );
+
+		AnnotationDescriptor<SafeHtml.Tag> protocolDescriptor = new AnnotationDescriptor( SafeHtml.Tag.class );
+		protocolDescriptor.setValue( "name", "img" );
+		protocolDescriptor.setValue( "attributes", new String[]{ "src" } );
+		protocolDescriptor.setValue( "protocols", new String[]{ "data" } );
+		doProgrammaticTest( new SafeHtmlDef().whitelistType( SafeHtml.WhiteListType.NONE )
```
gunnarmorling: Oh, I'm just realizing that some unfortunate approach has been chosen for `SafeHtmlDef` in the past. The programmatic API should never require to deal with annotation proxies. Instead there should have been members added on this def class so the properties can be set via the fluent DSL. Can we get this fixed for 6?

marko-bekhta: yes of course :) 
So we should deprecate current `SafeHtmlDef#additionalTagsWithAttributes` and add something like:
```java
public SafeHtmlDef additionalTagWithAttributes(String name, String[] attributes, String[] protocols) {
		// ...
		return this;
	}
```
or
```java
	public SafeHtmlDef additionalTagsWithAttributes(Tag... additionalTagsWithAttributes) {
		// ...
		return this;
	}

	public static class Tag {
		private final String name;

		private final String[] attributes;

		private final String[] protocols;

		public Tag(String name, String[] attributes, String[] protocols) {
			this.name = name;
			this.attributes = attributes;
			this.protocols = protocols;
		}
	}
```
or
```java
public Tag addTag(String tagName){
		return new Tag(tagName); 
	}

	public class Tag {
		private final String name;

		private final Set<String> attributes;

		private final Set<String> protocols;

		private Tag(String name) {
			this.name = name;
			this.attributes = CollectionHelper.newHashSet();
			this.protocols = CollectionHelper.newHashSet();
		}

		public Tag addAttribute(String name) {
			attributes.add( name );
			return this;
		}

		public Tag addProtocol(String name) {
			protocols.add( name );
			return this;
		}
		
		public SafeHtmlDef add(){
			//...
			return SafeHtmlDef.this;
		}
	}
```
instead, right ?

@gunnarmorling which option would you prefer ?

gunnarmorling: Let's think about it from a user's perspective. That's what I'd like to be able to write:

    new SafeHtmlDef().whitelistType( SafeHtml.WhiteListType.NONE )
        // let me add a tag and configure i all the way
        .additionalTag( "img" )
            .attribute( "src" )
                .protocols( "data", "snafu" )
            .attribute( "ref" )
                .protocols( "bla" )
        .additionalTag( "a" )
            .attribute( "href" )
                .protocols( "data" )
            .attribute( "bla" )
            .attribute( "blub" )
        // fast path for adding tags without further config
        .additionalTags( "foo", "bar" )

gsmet:
@marko-bekhta 

I marked it as Changes suggested as there are a couple of minor changes Gunnar suggested. Once they are fixed, we could merge this one.

As for the programmatic API issue, it should be pursued separately (and then the tests should be updated): I created https://hibernate.atlassian.net/browse/HV-1307. I affected it to you as you seemed to be interested in it.

marko-bekhta:
@gsmet ok, great. I was thinking that the programmatic API change should go in this one as well, hence the delay on other changes :)

gsmet:
@marko-bekhta yeah the bottom rule is whenever you can split a task in separate ones that make sense, do it. Obviously, we want to avoid micro tasks that don't make sense so there's a balance to find.

This way, you get the code in earlier and avoid big PR being postponed and postponed.

engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
```diff
@@ -81,6 +81,12 @@
 		 * @return list of tag attributes which are whitelisted.
 		 */
 		String[] attributes() default { };
+
+		/**
+		 * @return list of valid protocols.
+		 * @since 6.0
+		 */
+		String[] protocols() default { };
```
gsmet: Hi @marko-bekhta ,

Sorry for being so late to the party. The BV release drained most of our workforce lately. I think there's a logic error here. Gunnar spotted it in his comment here (https://github.com/hibernate/hibernate-validator/pull/714/commits/9b14baa4d03b88789e7be478dbf6e0c80192df85#r110625190) but without underlining it: the protocols should be attached to the attributes and not to the tag.

The issue I can see with this is that we would break the compatibility if we make attribute its own annotation. I suppose we could follow the same logic as for the tags and add an `additionalAttributesWithProtocols` thingy. WDYT?

gunnarmorling:
@marko-bekhta, should we close this one in the favor of #728?

marko-bekhta:
@gunnarmorling if it's OK to have those two issues combined in on PR - I'd say we can close this one.

gunnarmorling:
Yes, that's ok :) Thanks!

PR #716, opened 2017-04-04
HV-1302: @SafeHtml should support adding additional protocols

oyvindhorneland:


gunnarmorling:
Hi Øyvind, thanks a lot for sending this PR. It seems though Marko bet you doing this change, there is already PR #714. I suggest to continue discussion over there, as it's a bit more complete (doc update, fluent DSL update).

PR #717, opened 2017-04-06
HV-1303 @SafeHtml should support relative URLs

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1303

I've added `baseUri` to `@SafeHtml`. I defaulted it to blank string so it will not change the behaviour of any existing usages.

This PR goes on top of the previous one for `@SafeHtml`.

### DIFF
--- a/documentation/src/main/asciidoc/ch02.asciidoc
+++ b/documentation/src/main/asciidoc/ch02.asciidoc
@@ -598,7 +598,7 @@ With one exception also these constraints apply to the field/property level, onl
 	Supported data types::: `BigDecimal`, `BigInteger`, `CharSequence`, `byte`, `short`, `int`, `long` and the respective wrappers of the primitive types
 	Hibernate metadata impact::: None
 
-`@SafeHtml(whitelistType= , additionalTags=, additionalTagsWithAttributes=)`:: Checks whether the annotated value contains potentially malicious fragments such as `<script/>`. In order to use this constraint, the link:$$http://jsoup.org/$$[jsoup] library must be part of the class path. With the `whitelistType` attribute a predefined whitelist type can be chosen which can be refined via `additionalTags` or `additionalTagsWithAttributes`. The former allows to add tags without any attributes, whereas the latter allows to specify tags and optionally allowed attributes using the annotation `@SafeHtml.Tag`.
+`@SafeHtml(whitelistType= , additionalTags=, additionalTagsWithAttributes=)`:: Checks whether the annotated value contains potentially malicious fragments such as `<script/>`. In order to use this constraint, the link:$$http://jsoup.org/$$[jsoup] library must be part of the class path. With the `whitelistType` attribute a predefined whitelist type can be chosen which can be refined via `additionalTags` or `additionalTagsWithAttributes`. The former allows to add tags without any attributes, whereas the latter allows to specify tags and optionally allowed attributes, as well as additional protocols for parameters of those attributes using the annotation `@SafeHtml.Tag`.
 	Supported data types::: `CharSequence`
 	Hibernate metadata impact::: None
 
--- a/documentation/src/main/asciidoc/ch02.asciidoc
+++ b/documentation/src/main/asciidoc/ch02.asciidoc
@@ -598,7 +598,8 @@ With one exception also these constraints apply to the field/property level, onl
 	Supported data types::: `BigDecimal`, `BigInteger`, `CharSequence`, `byte`, `short`, `int`, `long` and the respective wrappers of the primitive types
 	Hibernate metadata impact::: None
 
-`@SafeHtml(whitelistType= , additionalTags=, additionalTagsWithAttributes=)`:: Checks whether the annotated value contains potentially malicious fragments such as `<script/>`. In order to use this constraint, the link:$$http://jsoup.org/$$[jsoup] library must be part of the class path. With the `whitelistType` attribute a predefined whitelist type can be chosen which can be refined via `additionalTags` or `additionalTagsWithAttributes`. The former allows to add tags without any attributes, whereas the latter allows to specify tags and optionally allowed attributes, as well as additional protocols for parameters of those attributes using the annotation `@SafeHtml.Tag`.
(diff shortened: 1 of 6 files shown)


engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
```diff
@@ -65,6 +65,12 @@
 	Tag[] additionalTagsWithAttributes() default { };
 
 	/**
+	 * @return Base URI to be used in resolving relative links to absolute ones.
```
oyvindhorneland: Should there be a note here stating that relative URLs will fail validation if baseUri is not set?

oyvindhorneland:
Looks good to me - just the one javadoc comment. Thanks!

engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
```diff
@@ -65,6 +65,12 @@
...
+	 * @since 5.4.2
```
gsmet: Let's make it for 6.0 for now.

gsmet:
@oyvindhorneland in the issue you created, you seem to suggest than just setting the baseURI to a dummy one makes the issue go away. Typically is `../../../my/relative/url.txt` valid with a baseURI of `http://localhost/`? I would have expected to not be valid.

I ask this question because, if it's the case, we might consider having the default baseUri set to `http://localhost/`.

documentation/src/main/asciidoc/ch02.asciidoc
```diff
@@ -587,7 +587,8 @@ With one exception also these constraints apply to the field/property level, onl
...
+`@SafeHtml(whitelistType= , additionalTags=, additionalTagsWithAttributes=, baseUri=)`:: Checks whether the annotated value contains potentially malicious fragments such as `<script/>`. In order to use this constraint, the link:$$http://jsoup.org/$$[jsoup] library must be part of the class path. With the `whitelistType` attribute a predefined whitelist type can be chosen which can be refined via `additionalTags` or `additionalTagsWithAttributes`. The former allows to add tags without any attributes, whereas the latter allows to specify tags and optionally allowed attributes, as well as additional protocols for parameters of those attributes using the annotation `@SafeHtml.Tag`.
+In addition `baseUri` allows to specify base URI to be used to resolve relative ones.
```
gsmet: to specify `the` base URI to be used to resolve relative `URIs`

gsmet: I would use `baseURI` to be consistent with `java.net.URI`

engine/src/main/java/org/hibernate/validator/constraints/SafeHtml.java
gsmet: I suggest something like:
`Base URI used to resolve relative URIs to absolute ones. If not set, validation of HTML containing relative URIs will fail.`

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/SafeHtmlValidatorTest.java
```diff
@@ -164,6 +164,20 @@ public void testAdditionalProtocols() {
 		assertNumberOfViolations( validator.validate( new Bar( "<div src='data:image/png;base64,100101' />" ) ), 1 );
 	}
 
+	@Test
+	@TestForIssue(jiraKey = "HV-817")
+	public void testPreserveRelativeLinks() throws Exception {
+		descriptor.setValue( "whitelistType", WhiteListType.RELAXED );
+		descriptor.setValue( "baseUri", "http://127.0.0.1" );
+
+		assertTrue( getSafeHtmlValidator().isValid( "<img src='/some/relative/url/image.png' />", null ) );
```
gsmet: It would be nice to also have a test with a real relative URI: `../../some/relative/url/image.png` (one working and one not working if the URI tries to go back too far).

oyvindhorneland: @gsmet It seems jsoup does not fail validation if the relative path goes back too far. jsoup's `StringUtil.resolve(URL base, String relUrl)` uses `new java.net.URL(base, relUrl)`, which will resolve `http://localhost/` with `../../../../../test.png` as URL containing `http://localhost/../test.png`. It is not examined further. No active issues with jsoup in this area, so we might consider raising one - what do you think?

If the baseURI default value is set by default, the default behaviour of SafeHtml would change, since it currently does not allow relative paths.

gsmet: > If the baseURI default value is set by default, the default behaviour of SafeHtml would change, since it currently does not allow relative paths.

Yeah, I'm wondering if we should consider it a feature or a bug. If, in the future, Jsoup checks the relative URL, I suppose it makes sense to keep it this way.

I was mostly wondering if people would really consider providing the baseURI in the annotation definition and if, maybe, we should just pass `http://localhost/` to Jsoup and consider relative links always valid from our POV.

I don't have a strong opinion about it though.

oyvindhorneland: While it was a surprise to me that I by default got validation failure for both relative images and inline data images (HV-1302), it's hard to know if someone depend on this behaviour to ensure their definition of _valid_ input. 

From my perspective I'd say it's probably both sensible and expected to support relative paths by default, but changing the default is scary.

gsmet: Yeah, let's be safe.

@marko-bekhta can you address the minor comments I made? Then we will merge it.

As for backporting it, as mentioned by Gunnar, we would only do it if we do a new 5.x release. For now we don't really have a lot of material to justify a new 5.x release and the associated maintenance burden but we'll see in the future. Using the next alpha of HV 6 would be an option if you use JDK 8.

marko-bekhta: @gsmet yes, I plan to make changes to both PRs related to `SafeHtml` later today.

marko-bekhta:
hi @gsmet, I forgot to change the labels after I've applied the changes. I didn't add an invalid relative URI test as jsoup is not failing for such cases.

gsmet:
@marko-bekhta I merged the second commit with a few additional (very minor) changes. Thought it was time to merge this one. You will have to rebase your other PR about the protocols.

PR #720, opened 2017-04-14
HV-1310 Add support for container element constraints for nested arrays

gsmet:
* https://hibernate.atlassian.net/browse/HV-1310


engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/NestedTypeArgumentsValueExtractorTest.java
```diff
@@ -306,7 +336,35 @@ private static ArrayOfOptionalsWithAutomaticUnwrapping invalidArray() {
 
 			return baz;
 		}
-
 	}
 
+	@SuppressWarnings({ "unused" })
+	private static class NestedArray {
+
+		private String @Size(min = 2) [] @Email [] array;
```
gunnarmorling: Nice; do we have one for mixing say `List` and arrays, too?

PR #727, opened 2017-04-26
HV-1320 Replace <iterable element> by <list element> in the node path of Lists

gsmet:
In passing, be consistent in the tests: when we test the property path,
we don't use the constants so that it is more readable and we have a
clear vision of the obtained property path.

 * https://hibernate.atlassian.net/browse/HV-1320


gunnarmorling:
Merged; thanks, @gsmet!

PR #728, opened 2017-05-01
HV-1307 SafeHtmlDef programmatic API should not use annotation proxies

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1307

this one goes on top of HV-1302 as the corresponding changes to annotations are there.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/cfg/defs/SafeHtmlDef.java
+++ b/engine/src/main/java/org/hibernate/validator/cfg/defs/SafeHtmlDef.java
@@ -6,16 +6,32 @@
  */
 package org.hibernate.validator.cfg.defs;
 
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import org.hibernate.validator.cfg.ConstraintDef;
 import org.hibernate.validator.constraints.SafeHtml;
+import org.hibernate.validator.internal.util.CollectionHelper;
+import org.hibernate.validator.internal.util.annotationfactory.AnnotationDescriptor;
+import org.hibernate.validator.internal.util.annotationfactory.AnnotationFactory;
 
 /**
  * @author Marko Bekta
  */
 public class SafeHtmlDef extends ConstraintDef<SafeHtmlDef, SafeHtml> {
 
+	private final Set<TagWithAttributes> tags;
+
 	public SafeHtmlDef() {
 		super( SafeHtml.class );
+		tags = CollectionHelper.newHashSet();
+	}
+
+	private SafeHtmlDef(ConstraintDef<?, SafeHtml> original) {
+		super( original );
+		tags = CollectionHelper.newHashSet();
 	}
 
 	public SafeHtmlDef whitelistType(SafeHtml.WhiteListType whitelistType) {
@@ -28,6 +44,10 @@ public class SafeHtmlDef extends ConstraintDef<SafeHtmlDef, SafeHtml> {
 		return this;
 	}
 
+	/**
+	 * @deprecated Use {@link SafeHtmlDef#additionalTag(String)} instead.
+	 */
+	@Deprecated
 	public SafeHtmlDef additionalTagsWithAttributes(SafeHtml.Tag... additionalTagsWithAttributes) {
 		addParameter( "additionalTagsWithAttributes", additionalTagsWithAttributes );
 		return this;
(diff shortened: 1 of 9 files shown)


engine/src/main/java/org/hibernate/validator/cfg/defs/SafeHtmlDef.java
```diff
@@ -6,16 +6,32 @@
...
+import org.hibernate.validator.internal.util.CollectionHelper;
+import org.hibernate.validator.internal.util.annotationfactory.AnnotationDescriptor;
+import org.hibernate.validator.internal.util.annotationfactory.AnnotationFactory;
 
 /**
  * @author Marko Bekta
  */
 public class SafeHtmlDef extends ConstraintDef<SafeHtmlDef, SafeHtml> {
 
+	private final Set<TagWithAttributes> tags;
```
gsmet: We discussed it with Gunnar. The idea would be to have a new ``AnnotationDef`` type (similar to the `ConstraintDef` approach with `annotationType` and a map of `parameters`).

This type would replace `TagWithAttributes` and `Attribute` and avoid the need to "refresh" the configuration after each operation.

The new ``AnnotationDef`` would be transformed to annotation proxies in `ConfiguredConstraint#createAnnotationProxy()` (e.g. you would need to special case if the value is an `AnnotationDef` and recurse into it).

The overall idea is:
 * to avoid unneeded instantiations
 * to keep the creation of the annotation proxies all in the same place

If it's not clear, feel free to ping me for additional discussions.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/SafeHtmlValidator.java
```diff
@@ -56,10 +56,14 @@ public void initialize(SafeHtml safeHtmlAnnotation) {
...
-				for ( String attribute : tag.attributes() ) {
-					whitelist.addProtocols( tag.name(), attribute, tag.protocols() );
+			whitelist.addTags( tag.name() );
+			if ( tag.attributes().length > 0 ) {
+				whitelist.addAttributes( tag.name(), tag.attributes() );
+			}
+			if ( tag.additionalAttributesWithProtocols().length > 0 ) {
+				for ( SafeHtml.Tag.Attribute attribute : tag.additionalAttributesWithProtocols() ) {
+					whitelist.addAttributes( tag.name(), attribute.name() );
+					whitelist.addProtocols( tag.name(), attribute.name(), attribute.protocols() );
```
marko-bekhta: I'll double check this. But as I remember there are different Sets and Maps for each thing (tag, attribute, protocol etc) and validation will fail if you add a protocol for example `data` for `img` -> `src` then validation will fail as `img` tag is absent in tags, and the same for the attribute...
But again - I'll double check if that's really so :)

marko-bekhta:
Hi,

I  followed your advice about `AnnotationDef`. And here's what I've done. I moved some parts from `ConstraintDef` to `AnnotationDef` and extended from it, as `ConstraintDef` is an annotation as well, right? 

I also added `annotationsAsParameters` to  `AnnotationDef`, it contains those annotation parameters which are annotations themselves. I moved `createAnnotationProxy` to `AnnotationDef` as it looked as a better place for it.

The only thing that I'd like to improve is to cover a case when parameter is not an array of annotations but a single annotation. Right now there seems to be no such cases, but I'd like to add it so the functionality is more complete. So if in general the approach is fine - I'll add this additional functionality for single annotation.

gsmet:
Force pushed a rebased version with one additional commit. Have to discuss something with Gunnar before merging.

engine/src/test/java/org/hibernate/validator/test/cfg/ProgrammaticConstraintDefinitionsTest.java
```diff
@@ -66,6 +68,50 @@ public void safeHtmlProgrammaticDefinition() {
...
+				new SafeHtmlDef().whitelistType( SafeHtml.WhiteListType.NONE )
+						.additionalTag( "img" ).attribute( "src" ),
+				"<img src='data:image/png;base64,100101' />", 1
+		);
+		doProgrammaticTest(
+				new SafeHtmlDef().whitelistType( SafeHtml.WhiteListType.NONE )
+						.additionalTag( "img" ).attribute( "src" ).protocols( "data" ),
+				"<img src='data:image/png;base64,100101' />", 0
+		);
+		doProgrammaticTest(
```
gunnarmorling: @marko-bekhta @gsmet Can you take a look at this PR: https://github.com/hibernate/hibernate-validator/pull/739. It's based on your's but takes a slightly different approach for handling the inner types of `@SafeHtml` in the API. Instead of having one fluent invocation on one level this adds `new TagDef()` and new `AttributeDef()` for those. For the common case of no further configuration on the next level, an overloaded version of just taking a string is added.

It's failing the test, i.e. I must have done something wrong within the implementation, but before going further I wanted to share it with you to discuss on the API. I like that it makes clearer, when one element has been fully configured as opposed to one huge fluent invocation. WDYT?

marko-bekhta: Hi @gunnarmorling, your approach looks cleaner without those instantiations of inner classes (inside `SafeHtmlDef`), and as you said - it's easier to see where definition of one attribute starts and where it ends.

gunnarmorling: Ok, I'll check out the test failure then and update my PR. Let's drive home this one :)

gunnarmorling:
Closing this in favor of #739 which is based on this one. Thanks, @marko-bekhta!

PR #729, opened 2017-05-01
HV-1327 Using custom numbering extension for examples and tables in documentation

marko-bekhta:
Used that `CustomNumberingProcessor` for HV documentation. This will fail until 1.0.1.Final version for extensions is not released and available.

After discussion on asciidoctor issue, they suggested to update to 1.6 version of asciidoctor and latest pdf version as well. Otherwise pdf version was not generated and an exception was thrown.

Here's a generated pdf:
[hibernate_validator_reference.pdf](https://github.com/hibernate/hibernate-validator/files/968568/hibernate_validator_reference.pdf)

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -202,6 +202,11 @@
                             <backend>html5</backend>
                             <outputDirectory>${asciidoctor.base-output-dir}/html_single</outputDirectory>
                             <sourceHighlighter>prettify</sourceHighlighter>
+                            <extensions>
+                                <extension>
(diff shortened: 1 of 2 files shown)


gsmet:
Current status: we are waiting for the next release of asciidoctor-ant before merging this (we need the extension to work for HV and BV).

gsmet:
@marko-bekhta thanks! I amended your commit as the extension can be added to the shared configuration part instead of adding it to both executions.

marko-bekhta:
@gsmet ok, cool!

PR #730, opened 2017-05-05
HV-1305 Use HTML5 doclet when generating JavaDoc with Java 9

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1305

Using HTML5 for JavaDoc generation with JDK 9. I've added `-html5` to parameters and also replaced a couple of `<tt>...</tt>` with `{@code}` or `{@link}` as otherwise it will give a next error:

```
error: tag not supported in the generated HTML version: tt
```

https://www.w3schools.com/tags/tag_tt.asp

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -174,7 +174,7 @@
                 <jdk>9</jdk>
             </activation>
             <properties>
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Nice, thanks, @marko-bekhta!

PR #731, opened 2017-05-05
HV-1326 Fail fast when the value is null during nested cascaded validation

gsmet:
* https://hibernate.atlassian.net/browse/HV-1326


gunnarmorling:
Merged; thanks, @gsmet!

PR #732, opened 2017-05-06
HV-1300 Splitting number validators for for each number type

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1300

Hi @gsmet, hope you will not be angry at me that I've took this one :) I asked about it in Jira but you probably haven't seen the comment - so I thought I'll risk it and work on it :)

I've split `Positive`, `Negative`, `Min`, `Max` and their `Decimal*` versions into separate validators per type. I've left `Byte`, `Short`, `Integer` without implementation in some cases as comparing was done the same way as for `Number`. 

What I've done is next:

- extracted initialization and main validation logic into abstract base class
- introduced "helpers' for comparing numbers (`NumberComparatorHelper`, `DecimalNumberComparatorHelper`, `NumberSignumHelper`) and to check for infinity and NaN values (`InfinityNumberComparatorHelper`) 
- moved all new validators into new `number` package
- updated tests for min/max and added test for `Positive`/`Negative`


gsmet:
Hi @marko-bekhta ,

Sorry for not answering earlier, I was busy preparing the Beta1 of HV 6 and didn't have a chance to review this before.

I added a couple of commits, mostly to avoid instantiating new objects and be consistent with the rest of the code.

As for the tests for positive/negative, I added tests in the TCK, that's why there weren't any tests in HV. I'll include them for now but I might remove them later when I will review the TCK tests.

Thanks!

PR #733, opened 2017-05-09
HV-1313 Support group conversion for container element types

gsmet:
* https://hibernate.atlassian.net/browse/HV-1313

I added some tests in the TCK for the BV part (only the annotations support for now though, I wanted to exercise the logic, will see if I have the time to add XML tests).

In the end, I think it was a mistake to make the annotated object the root of the container elements `CascadingMetaData`. It would have worked if we didn't have to support the legacy behavior of `@Valid` on the root element for Maps and Lists, but considering this legacy behavior, it was not a good idea.

So I think I will change it to having:
- a separate `CascadingMetaData` for the annotated object: the container elements metadata will be empty for this one
- a root `CascadingMetaData` (with `cascading` to false) for the container elements. It's really nice to have a root.

Sorry for the big patch but it was the minimum to get it working without compilation issues and existing test failures.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/cfg/context/CascadableConstraintMappingContextImplBase.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/cfg/context/CascadableConstraintMappingContextImplBase.java
@@ -146,7 +146,9 @@ abstract class CascadableConstraintMappingContextImplBase<C extends Cascadable<C
 	}
 
 	protected CascadingTypeParameter getCascadingMetaData() {
-		Map<TypeVariable<?>, CascadingTypeParameter> typeParametersCascadingMetaData = new HashMap<>();
+		Map<TypeVariable<?>, CascadingTypeParameter> typeParametersCascadingMetaData = containerElementContexts.values().stream()
+				.filter( c -> c.getCascadingTypeParameter() != null )
+				.collect( Collectors.toMap( c -> c.getCascadingTypeParameter().getTypeParameter(), c -> c.getCascadingTypeParameter() ) );
 
 		for ( ContainerElementConstraintMappingContextImpl typeArgumentContext : containerElementContexts.values() ) {
 			CascadingTypeParameter cascadingTypeParameter = typeArgumentContext.getCascadingTypeParameter();
@@ -154,6 +156,7 @@ abstract class CascadableConstraintMappingContextImplBase<C extends Cascadable<C
 				typeParametersCascadingMetaData.put( cascadingTypeParameter.getTypeParameter(), cascadingTypeParameter );
 			}
 		}
+
 
 		boolean isArray = TypeHelper.isArray( configuredType );
 		CascadingTypeParameter cascadingMetaData = isArray
--- a/engine/src/main/java/org/hibernate/validator/internal/cfg/context/CascadableConstraintMappingContextImplBase.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/cfg/context/CascadableConstraintMappingContextImplBase.java
@@ -10,11 +10,10 @@ import static org.hibernate.validator.internal.util.CollectionHelper.newHashMap;
 
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.ArrayList;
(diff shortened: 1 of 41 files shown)


engine/src/main/java/org/hibernate/validator/internal/cfg/context/CascadableConstraintMappingContextImplBase.java
```diff
@@ -150,24 +145,22 @@ public boolean isCascading() {
 			.collect( Collectors.toSet() );
 	}
 
-	protected List<CascadingTypeParameter> getCascadedTypeParameters() {
-		List<CascadingTypeParameter> cascadingTypeParameters = new ArrayList<>();
+	protected CascadingTypeParameter getCascadingMetaData() {
+		Map<TypeVariable<?>, CascadingTypeParameter> typeParametersCascadingMetaData = new HashMap<>();
```
gunnarmorling: Can we size it with 3 or 4 perhaps?

gunnarmorling: Or maybe even use a stream:

    containerElementConstraints.stream()
        .filter( c -> c.getCascadingTypeParameter() != null )
        .collectToMap( collectors.toKey(), collectors.toValue() )

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedExecutable.java
```diff
@@ -249,7 +240,8 @@ public boolean isEquallyParameterConstrained(ConstrainedExecutable other) {
 		int i = 0;
 		for ( ConstrainedParameter parameter : parameterMetaData ) {
 			ConstrainedParameter otherParameter = other.getParameterMetaData( i );
-			if ( parameter.isCascading() != otherParameter.isCascading()
+			// FIXME: how to deal with method overriding with type overloading of one of the parameters?
```
gunnarmorling: Should we create a JIRA issue (with an example)?

engine/src/main/java/org/hibernate/validator/internal/xml/ConstrainedExecutableBuilder.java
```diff
@@ -211,7 +213,7 @@ private ConstrainedExecutable parseExecutableType(String defaultPackage,
 		Set<MetaConstraint<?>> returnValueTypeArgumentConstraints = new HashSet<>();
 		List<CascadingTypeParameter> cascadingTypeParameters = new ArrayList<>();
 		Map<Class<?>, Class<?>> groupConversions = new HashMap<>();
```
gunnarmorling: Do we still need this var? (Probably yes, just looking at the GH diff, so lacking the full context).

engine/src/main/java/org/hibernate/validator/internal/cfg/context/CascadableConstraintMappingContextImplBase.java
gsmet: I used streams everywhere and you spot the one place where I didn't :).

gunnarmorling:
Ahaha :)

engine/src/main/java/org/hibernate/validator/internal/xml/ConstrainedExecutableBuilder.java
gsmet: We don't, good catch.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/CascadableConstraintMappingContextImplBase.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedExecutable.java
gsmet: https://hibernate.atlassian.net/browse/HV-1328

gunnarmorling:
Closing; superseded by #736.

PR #735, opened 2017-05-10
HV-1304 Add metadata support for container elements

gsmet:
* https://hibernate.atlassian.net/browse/HV-1304

This one is on top of https://github.com/hibernate/hibernate-validator/pull/733.


gsmet:
Looks like this one was based on an old version of HV-1313. Opening a new one based on the right version.

PR #739, opened 2017-05-16
HV-1307

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1307

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/cfg/ProgrammaticConstraintDefinitionsTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/cfg/ProgrammaticConstraintDefinitionsTest.java
@@ -155,7 +155,7 @@ public class ProgrammaticConstraintDefinitionsTest {
 
 		tagDescriptor = new AnnotationDescriptor( SafeHtml.Tag.class );
 		tagDescriptor.setValue( "name", "img" );
-		tagDescriptor.setValue( "additionalAttributesWithProtocols", new SafeHtml.Attribute[]{ AnnotationFactory.create( attributeDescriptor ) } );
+		tagDescriptor.setValue( "attributesWithProtocols", new SafeHtml.Attribute[]{ AnnotationFactory.create( attributeDescriptor ) } );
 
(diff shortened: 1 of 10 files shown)


engine/src/test/java/org/hibernate/validator/test/cfg/ProgrammaticConstraintDefinitionsTest.java
```diff
@@ -66,6 +69,61 @@ public void safeHtmlProgrammaticDefinition() {
 		doProgrammaticTest( new SafeHtmlDef().whitelistType( SafeHtml.WhiteListType.RELAXED ).baseURI( "http://localhost" ),
 				"<img src='/some/relative/url/image.png' />", 0
 		);
+
+		doProgrammaticTest(
+				new SafeHtmlDef().whitelistType( SafeHtml.WhiteListType.NONE )
+						.additionalTags( new TagDef( "img" )
+								.attributes( "src" )
+						),
+				"<img src='data:image/png;base64,100101' />", 1
```
gunnarmorling: So that's the test failing for me after my changes and I've come to think that this assertion actually is wrong. There should be no violation, as `protocol()` ﻿hasn't been given, hence no restrictions apply for the protocols.

Before my change it would result in a violation as we'd call `addProtocol()` with an empty list of protocols, i.e. all the protocols are disallowed.

gunnarmorling:
@marko-bekhta @gsmet Added some more commits, it should be good now. It's not quite clear to me why we moved the code for creating the proxy. It's for internal usage only and now we've increased the public surface (also if we explicitly exclude the protected stuff in the comments). Anyways, let's move forward with it.

My last question is on the names of the def members; I've used the same names for the simple version (accepting Strings) and the overloaded more powerful version (accepting TagDef / AttributeDef). Whereas in the annotations we have to use different names as we cannot overload the methods there. I think the simpler names are nicer, but if you prefer to use the exactly same ones as in the annotation counterparts, we can do that.

gsmet:
Rebased and merged! Thanks to both of you.

PR #740, opened 2017-05-16
HV-1339 Add support for non parameterized type in value extraction

gsmet:
* https://hibernate.atlassian.net/browse/HV-1339

Requires https://github.com/beanvalidation/beanvalidation-api/pull/107 to compile.


gunnarmorling:
Jenkins, retest this, please.

gunnarmorling:
Merged; thanks, @gsmet! I've added one more commit with minor adjustments, most notably to have the test assert on the paths first and omit the assertion on count which usually isn't as helpful in case of test failures.

PR #742, opened 2017-05-19
HV-1340 Add container element nodes/information to the node builder API

gsmet:
* https://hibernate.atlassian.net/browse/HV-1340

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/constraints/ConstraintValidatorContextImplTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/constraints/ConstraintValidatorContextImplTest.java
@@ -6,25 +6,29 @@
  */
 package org.hibernate.validator.test.constraints;
 
+import static org.hibernate.validator.testutil.ConstraintViolationAssert.assertPathEquals;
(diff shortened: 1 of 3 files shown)


engine/src/test/java/org/hibernate/validator/test/constraints/ConstraintValidatorContextImplTest.java
```diff
@@ -51,7 +57,9 @@ public void testIterableKeyed() {
 				.addConstraintViolation();
 
 		List<ConstraintViolationCreationContext> constraintViolationCreationContextList = context.getConstraintViolationCreationContexts();
-		assertMessageAndPath( constraintViolationCreationContextList.get( 0 ), message, "foo[test]" );
+		assertMessageAndPath( constraintViolationCreationContextList.get( 0 ), message, pathWith()
+				.property( "foo" )
+				.property( null, true, "test", null ) );
```
gunnarmorling: Instead of the gazillion of parameters, should we rather have dedicated methods such as `listProperty()` etc., each taking the required parameters only. Definitely a separate thing, though.

gunnarmorling:
Merged; thanks, @gsmet!

PR #744, opened 2017-05-22
HV-1344 Change in legacy @Valid implementation

gsmet:
@gunnarmorling could you take a look at this?

I think it's the way to go as it guarantees a similar behavior in both cases (legacy and new behavior). Typically, see the tests changed: the behavior looks more sane now.

Without this commit, we have the following behavior:
```java
@Valid
private List<@NotNull String> property;
```
results in nodes of the property paths not containing the `containerClass` and `typeArgumentIndex` information as they are directly added by the `LegacyListValueExtractor` at the `AnnotatedObject` level. I found this issue while converting the TCK to use property path nodes equality instead of relying on the string representation of the path.

If you think it's a good idea, I'll create a JIRA and add specific tests for that (note that there are already tests that exercise this behavior but I'd like to have specific ones).

By the way, we should probably define in the spec the behavior of:
```java
@Valid
@ConvertGroup(blabla)
private List<@Valid @ConvertGroup(blibli)> property;
```
At the moment, I simply merged the 2 maps when creating the metadata for the second `@Valid`. We cannot only take into account the second one as it breaks the legacy behavior.


gsmet:
Closing this one, opening a new one with additional fixes.

PR #745, opened 2017-05-22
HV-1343 Fix equals/hashCode implementation of TypeArgumentConstraintLocation

gsmet:
* https://hibernate.atlassian.net/browse/HV-1343

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
@@ -29,6 +29,7 @@ public class TypeArgumentConstraintLocation implements ConstraintLocation {
 	private final Type typeForValidatorResolution;
 	private final Type containerType;
 	private final ConstraintLocation outerDelegate;
+	private final int hashCode;
 
 	TypeArgumentConstraintLocation(ConstraintLocation delegate, TypeVariable<?> typeParameter, Type typeOfAnnotatedElement) {
 		this.delegate = delegate;
@@ -41,6 +42,7 @@ public class TypeArgumentConstraintLocation implements ConstraintLocation {
 			outerDelegate = ( (TypeArgumentConstraintLocation) outerDelegate ).delegate;
 		}
 		this.outerDelegate = outerDelegate;
+		this.hashCode = buildHashCode( delegate, typeParameter );
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
```diff
@@ -114,6 +117,7 @@ public boolean equals(Object o) {
 	@Override
 	public int hashCode() {
 		int result = delegate != null ? delegate.hashCode() : 0;
+		result = 31 * result + typeParameter.hashCode();
 		result = 31 * result + typeForValidatorResolution.hashCode();
```
gunnarmorling: Do we actually need `typeForValidatorResolution` in here? Shouldn't delegate and type parameter suffice for determining the identity?

gunnarmorling: Also the hash code could be calculated once and be cashed.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
```diff
@@ -101,6 +101,9 @@ public boolean equals(Object o) {
 
 		TypeArgumentConstraintLocation that = (TypeArgumentConstraintLocation) o;
 
+		if ( !typeParameter.equals( that.typeParameter ) ) {
+			return false;
+		}
 		if ( delegate != null ? !delegate.equals( that.delegate ) : that.delegate != null ) {
```
gunnarmorling: I don't think it can be null?

gsmet: No, it can't. It's either a valid TypeVariable or the ones we forge.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
gsmet: Force pushed an update.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/TypeArgumentConstraintLocation.java
```diff
@@ -113,8 +115,12 @@ public boolean equals(Object o) {
 
 	@Override
 	public int hashCode() {
+		return hashCode;
+	}
+
+	private static int buildHashCode(ConstraintLocation delegate, TypeVariable<?> typeParameter) {
 		int result = delegate != null ? delegate.hashCode() : 0;
```
gunnarmorling: Didn't we agree that the null check on delegate isn't needed (sambe above)?

gsmet: Oh, I thought you were checking with me that typeArgument cannot be null :). Yeah agreed, will fix.

gsmet:
@gunnarmorling fixed.

PR #747, opened 2017-05-23
HV-1344 Change in legacy @Valid implementation

gsmet:
* https://hibernate.atlassian.net/browse/HV-1344

In the last commit, I fixed the last issue regarding:
```java
@Valid
private MyList myList;
```
with `MyList` extending `List<MyBean>`
    
In this case, the `containerClass` will be `MyList` and the `typeArgumentIndex` will be `null`.


engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaData.java
```diff
@@ -40,6 +40,16 @@
 	private final TypeVariable<?> typeParameter;
 
 	/**
+	 * The declared container class: it is the one used in the node of the property path.
+	 */
+	private final Class<?> declaredContainerClass;
```
gunnarmorling: You'll still send this mail, right (on declared vs. runtime)? Or is it settled?

gunnarmorling:
Looks good overall.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaData.java
gsmet: Well, I think it's settled considering the last example I gave you? That's what I put in the BVAL issue I created here: https://hibernate.atlassian.net/projects/BVAL/issues/BVAL-631 .

Moreover the last commit fixes the oddity that was bothering me and triggered the question.

(I have the email mostly prepared if needed)

gunnarmorling: Ok.

PR #749, opened 2017-05-29
HV-1349 Add @CandidateForTck marker

gsmet:
* https://hibernate.atlassian.net/browse/HV-1349

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/ContainerElementPropertyPathTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/ContainerElementPropertyPathTest.java
@@ -34,9 +34,11 @@ import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
 
 import org.hibernate.validator.internal.engine.path.NodeImpl;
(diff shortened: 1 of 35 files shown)


engine/src/test/java/org/hibernate/validator/test/internal/engine/valuehandling/ContainerElementPropertyPathTest.java
```diff
@@ -54,7 +55,6 @@ public void testMapInvalidKeyTypeArgument() {
 
 		Set<ConstraintViolation<DemographicStatistics>> constraintViolations = validator.validate( statictics );
 
-		assertCorrectPropertyPaths( constraintViolations, "inhabitantsPerAddress<K>[].<map key>" ); // the key is null, thus the '[]'
```
gunnarmorling: Ah, so we were essentially asserting on the path twice?

gsmet: Not really. We check the path and the string representation. Reason why I divided the tests in 2 parts: the path part can go in the TCK, the string representation part is implementation specific.

PR #750, opened 2017-06-01
HV-1337 Test framework adjustments

gsmet:
* https://hibernate.atlassian.net/browse/HV-1337

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/cfg/ConstraintMappingTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/cfg/ConstraintMappingTest.java
@@ -28,6 +28,8 @@ import javax.validation.GroupSequence;
 import javax.validation.ValidationException;
 import javax.validation.Validator;
 import javax.validation.ValidatorFactory;
+import javax.validation.constraints.AssertTrue;
+import javax.validation.constraints.Future;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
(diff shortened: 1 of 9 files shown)


engine/src/test/java/org/hibernate/validator/test/cfg/ConstraintMappingTest.java
```diff
@@ -145,7 +149,7 @@ public void testSingleConstraint() {
 
 		Set<ConstraintViolation<Marathon>> violations = validator.validate( new Marathon() );
 		assertNumberOfViolations( violations, 1 );
-		assertConstraintViolation( violations.iterator().next(), "may not be null" );
+		assertConstraintViolation( violations.iterator().next(), NotNull.class, "may not be null" );
```
gunnarmorling: (Independently of this PR) It'd be great to have this:

    assertSingleConstraintViolation( violations, NotNull.class, "may not be null" );

It avoids the - often not so helpful - assertion on the number of constraints - on the line before.

gsmet: Yeah, to be honest, I did the minimum here to have something more consistent and a little better with as little effort as possible.

But I think we need something similar to what we do with containsOnlyPaths but more general:
```java
assertThat( constraintViolations ).containsOnlyViolations(
     violationWith().constraint( NotNull.class ).message( "may not be null" ).path( ... )
);
```

(and it would be possible to only test for the constraint type by using only `constraint()`)

It might just look like a cute thing but I think we have an important issue in our test suite: we test the constraint type and the path independently. We could have a mix-up between the constraint type and the path and it would not be reported as an error.

And it would also solve your issue with `assertNumberOfViolations`.

gunnarmorling: > we test the constraint type and the path independently. We could have a mix-up between the constraint type and the path and it would not be reported as an error.

Indeed, that's nagging me, too. I've filed HV-1350, please feel free to generalize this one for related improvements around assertions.

PR #752, opened 2017-06-12
HV-1356 Enable CDI for value extractors added via the ServiceLoader

gsmet:
* https://hibernate.atlassian.net/browse/HV-1356
 * https://hibernate.atlassian.net/browse/HV-1358

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
@@ -15,10 +15,8 @@ import java.io.InputStream;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -48,6 +46,7 @@ import org.hibernate.validator.internal.util.Version;
 import org.hibernate.validator.internal.util.logging.Log;
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
 import org.hibernate.validator.internal.util.privilegedactions.GetClassLoader;
+import org.hibernate.validator.internal.util.privilegedactions.GetInstancesFromServiceLoader;
 import org.hibernate.validator.internal.util.privilegedactions.SetContextClassLoader;
 import org.hibernate.validator.internal.xml.ValidationBootstrapParameters;
 import org.hibernate.validator.internal.xml.ValidationXmlParser;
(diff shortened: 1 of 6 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -494,23 +498,28 @@ private void parseValidationXml() {
...
 	private void loadValueExtractorsFromServiceLoader() {
-		@SuppressWarnings("rawtypes")
-		ServiceLoader<ValueExtractor> loader = ServiceLoader.load(
-				ValueExtractor.class,
-				externalClassLoader != null ? externalClassLoader : getClass().getClassLoader()
-		);
+		List<ValueExtractor<?>> valueExtractors;
+
+		if ( externalClassLoader != null ) {
+			final ClassLoader originalContextClassLoader = run( GetClassLoader.fromContext() );
```
gunnarmorling: Instead of setting the TCCL, I'd suggest to pass the CL as a parameter to GetValueExtractorsFromServiceLoader#action()`. We'd either pass the external CL or the current TCCL. Behavior should be the same, but a bit less convoluted.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetValueExtractorsFromServiceLoader.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+import java.util.ServiceConfigurationError;
+import java.util.ServiceLoader;
+
+import javax.validation.valueextraction.ValueExtractor;
+
+/**
+ * @author Guillaume Smet
+ */
+public class GetValueExtractorsFromServiceLoader implements PrivilegedAction<List<ValueExtractor<?>>> {
```
gunnarmorling: This looks quite the same as `GetConstraintValidatorsFromServiceLoader`. Couldn't it be one class taking the type to load as a parameter?

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
gsmet: Mmmmh, OK, I must admit I wanted to be on par with the patterns we used elsewhere.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetValueExtractorsFromServiceLoader.java
gsmet: Yes, good idea.

gunnarmorling:
I don't hope we set the TCCL when calling our own code. We should do it
only when calling into other libs which don't take a classloader as a
parameter.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetValueExtractorsFromServiceLoader.java
gsmet: I created the generic class.

Getting rid of `GetConstraintValidatorsFromServiceLoader` will be part of https://hibernate.atlassian.net/browse/HV-1359.

gsmet:
@gunnarmorling addressed your comments.

gunnarmorling:

PR #754, opened 2017-06-13
HV-1350 Add assertSingleConstraintViolation() test method

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1350

Hi!

I've actually added two methods `assertSingleConstraintViolation` and `assertNoViolations`. Second one replaces usages of `assertNumberOfViolations( constraintViolations, 0 );` as there's a lot of such cases. 

Also I did put static imports on top for those test classes affected by this change.


gunnarmorling:
Yeah, @marko-bekhta is back! :champagne: 🎉 🎊

gsmet:
Hi @marko-bekhta ,

Glad to see you back.

+1 for assertNoViolations. I had this idea too.

About `assertSingleConstraintViolation`, I have introduced something better in the TCK recently. You can see it in action here: https://github.com/beanvalidation/beanvalidation-tck/pull/111/files#diff-0c826c4408ceaa1381e0e082b153a312R75 .

I have a branch where I backported it to HV. I thought I had already pushed it but it's not the case. I will push it tomorrow morning.

Basically, the idea would be to use this infrastructure from now on for the new tests and fix the existing tests along the way.

gsmet:
@marko-bekhta I pushed my branch here: https://github.com/hibernate/hibernate-validator/pull/755

I remember that I refrained from pushing it because I was waiting to use it really in a test and currently I'm only working on the TCK so it didn't feel urgent. That being said, it should work OK as it's now heavily used in the TCK as you can see in the examples I posted below.

I think you should be able to cherry-pick my commit and replace `assertSingleConstraintViolation` with this. It's more readable IMHO. It would be nice to follow the indentation convention I used in the TCK for that. I use a short form with the whole `violationWith()...` on one line if there is only one thing tested and the form you can see in the example I pointed when we test at least 2 elements.

I didn't expect you to work on this so I'm very sorry for the work you did :/.

HTH.

gsmet:
@marko-bekhta btw, I'm working on some changes after discussing the API with Gunnar. Will push an update soon.

marko-bekhta:
Hi @gsmet, 
no worries :) it'll be my payment for my long absence :)
This API looks much more powerful than simple `assertSingleConstraintViolation`. So maybe I'll just drop this `assertSingleConstraintViolation` and leave only `assertNoViolations` ? And after you merge in that other PR I'll just go through the tests and update them to use this new API. WDYT ?

gunnarmorling:
@marko-bekhta I've just merged https://github.com/hibernate/hibernate-validator/pull/755 which puts the discussed changes to the assertion API in place. So you could rework this PR here to make use of that. Thanks a lot!

gunnarmorling:
Note it's `violationOf( SomeConstraint.class )` now, and there's a dedicated `property()` method taking a single property name, avoiding the usage of the more powerful but also more chatty `pathWith()` API if not needed.

gsmet:
What I suggest is the following plan:
* we remove `assertNumberOfViolations()` and you introduce your `assertNoViolations()`
* we remove `assertCorrectConstraintViolationMessages()`: we use `containsOnlyViolations()` instead
* we remove `assertConstraintViolation()` variations: we use `containsOnlyViolations()` instead
* we use the new API for the rest: the idea is that we should at least test the constraint type each time so you now have `violationOf( constraintType )` as the entry point
* it allows to test the path in conjunction with the constraint type whereas before they were tested in 2 unrelated assertions. Most of the `containsOnlyPaths()` calls should go away and be replaced with `containsOnlyViolations` which will test the constraint type at the same time. There might be valid usage of `containsOnlyPaths()` so don't over do it if you have some doubts on some occurrences.

I already changed the TCK here so you have a couple of examples of the new API: https://github.com/beanvalidation/beanvalidation-tck/pull/117 .

If you have any question or want to push something not finished so that I confirm you're on the right track, feel free to ping me.

(btw, just to be clear: we will probably do the same "improve the consistency" work for the TCK at one point but now is not the right time)

gsmet:
@marko-bekhta I saw you pushed some commits. I let you change the label when you're ready. Thanks!

marko-bekhta:
@gsmet, there's a few more `containsOnlyPaths` left so I thought I'll replace those and then update the label.

gsmet:
OK, cool.

marko-bekhta:
rebased and added a change for the `Email` constraint test. It started to fail as the `EmailDef` was changed to use the new annotation while the test still used the previous one in `EmailContainerAnnotated`.

gsmet:
@marko-bekhta Very. Nice. Work. I merged it.

I added some more commits for very minor things. Thanks for not making the review completely useless ;).

I also created https://hibernate.atlassian.net/browse/HV-1387 as we will have to add some tests to prevent regressions on the string representation.

marko-bekhta:
nice ! :)

PR #755, opened 2017-06-13
HV-1350 Introduce the containsOnlyViolations utility previously introduced in the TCK

gsmet:


test-utils/src/main/java/org/hibernate/validator/testutil/ConstraintViolationAssert.java
```diff
@@ -371,6 +401,254 @@ public void containsPaths(PathExpectation... expectedPaths) {
...
+
+		private Class<?> rootBeanClass;
+
+		private String message;
+
+		private Object invalidValue;
+
+		private PathExpectation propertyPath;
+
+		public ViolationExpectation(Class<? extends Annotation> constraintType) {
```
gunnarmorling: @gsmet Can we make the constructor private? Only `violationOf()` should be used in tests.

gsmet: Done and pushed.

PR #757, opened 2017-06-14
HV-1365 Creating common classes of property path example;

gunnarmorling:
Adding examples for containerClass and getTypeArgumentIndex()

@gsmet This adds executable examples for the newly added cases. I suggest to open a follow up issue to add the missing ones for the existing cases, too.

https://hibernate.atlassian.net/browse/HV-1365

Btw. one should be careful when taking the output from the RI because it could be erroneous or not what's intended in the spec actually. So what I did was to add the expected output manually and then verified by replacing it with the actual output from the RI.


gsmet:
Merged, thanks!

PR #758, opened 2017-06-15
HV-1367 Being unable to instantiate a value extractor should throw a ValueExtractorDeclarationException

gsmet:
Test is in the TCK.

 * https://hibernate.atlassian.net/browse/HV-1367


gsmet:
After some discussion, we decided to keep the current behavior.

PR #765, opened 2017-06-20
HV-1377 StackOverflowError during compilation with annotation-processor enabled

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1377

I've added a check if the type was already processed, and a test case that reflects the problem similar to the one in the test case project provided in the issue. In that case the loop was caused by the interface `com.vaadin.data.Property` : 
```java
public interface Property<T> extends Serializable {
   // .....
   public interface Transactional<T> extends Property<T> {
      // .....
   }
   // .....
}
```

The problem was that when a type is passed to `org.hibernate.validator.ap.ClassVisitor#visitAllMyElements`  there's a loop over all members -
 `elementUtils.getAllMembers( typeElement )` returns inner classes and interfaces from all extended classes and implemented interfaces. As a result the loop occurs:
```
    Field -(contains)> Property -(contains)> Transactional -(contains)> Property -(contains)> Transactional ....
```


gsmet:
I made a few minor changes that I squashed into your commit and merged.

Thanks!

PR #766, opened 2017-06-21
HV-1376 Check that @Positive/@Negative and OrZero versions are taken into account in the annotation processor

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1376

`@Positive` and `@Negative` were already added some time ago. So I've added support for `*OrZero` ones only.


gsmet:
Merged, thanks!

PR #768, opened 2017-06-21
HV-1382 Implement @PastOrPresent and @FutureOrPresent constraints

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1382

Added validators for `@FutureOrPresent`/`@PastOrPresent` and also Defs + added changes for annotation processor.

it seems that there are also changes to `ElementDescriptor` interface which fails all the builds for now.

### DIFF
--- a/annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
+++ b/annotation-processor/src/main/java/org/hibernate/validator/ap/util/ConstraintHelper.java
@@ -175,7 +175,7 @@ public class ConstraintHelper {
 	}
 
 	/**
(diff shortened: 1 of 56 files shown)


gsmet:
@marko-bekhta yeah, the metadata issue is expected. I have a patch in the works but it's not an easy issue to solve :)

Your patch looks very good. I just moved the validators to specific `pastorpresent` and `futureorpresent` to avoid stacking unrelated things. Other than that, it's good to go and i'm going to merge it right now.

Thanks!

marko-bekhta:
@gsmet ok, cool.

PR #769, opened 2017-06-22
HV-1352 Incorrect value handler on trace logging

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1352

`ValidatedValueHandler` was left in the ValueContext after validation of one constraint was finished and then when next constraint was validated in case of `TRACE` logging it was trying to use that handler even though it was not needed. So I've just set it to `null` at the same time when `UnwrapMode` was reseted. I believe that this one is only for HV 5.* as in 6 this part is implemented differently.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -582,6 +582,8 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 
 		// reset the unwrapping mode
 		valueContext.setUnwrapMode( UnwrapMode.AUTOMATIC );
+		// reset ValidatedValueHandler to prevent HV-1352
+		valueContext.setValidatedValueHandler( null );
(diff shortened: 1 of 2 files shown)


gunnarmorling:
Hey @marko-bekhta, thanks for this fix! While it works, I'm thinking so that the issue should better be fixed by just pushing the logging after the call to `getInitializedConstraintValidator()`. With your fix we get these log messages:

> Validating value 1 ...
> Validating value Optional[1] ...

But indeed it should say 

> Validating value 1 ...
> Validating value 1 ...

As the unwrapped value is the target of the validation.

Also it keeps the management of the value unwrapper limited to `ConstraintTree` which is a good thing. I'll close this PR and send another one (as I've got the fix already done while reviewing your's). Thanks!

marko-bekhta:
Hi @gunnarmorling, OK great. I was actually thinking that the message should say that the optional is validated ... But yeah, the constraint is applied to what is inside of that Optional :)

gunnarmorling:
Right, that's my thinking, too.

PR #770, opened 2017-06-22
HV-1385 Adjustments on the metadata API following BVAL-655

gsmet:
This will allow HV to be compiled against CR1 of the API.

 * https://hibernate.atlassian.net/browse/HV-1383
 * https://hibernate.atlassian.net/browse/HV-1385

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/AbstractConstraintMetaData.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/AbstractConstraintMetaData.java
@@ -8,8 +8,6 @@ package org.hibernate.validator.internal.metadata.aggregated;
 
 import static org.hibernate.validator.internal.util.CollectionHelper.newHashSet;
 
-import java.lang.reflect.GenericArrayType;
-import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.util.ArrayList;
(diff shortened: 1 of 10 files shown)


gsmet:
Jenkins, retest this please

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/AbstractConstraintMetaData.java
```diff
@@ -168,79 +166,56 @@ else if ( !name.equals( other.name ) ) {
...
-			// TODO Container element constraints are not supported for arrays in the Bean Validation specification.
-			// Let's ignore them for now.
-			return Collections.emptyList();
-		}
-		else {
-			return Collections.emptyList();
-		}
-	}
-
-	private static class ContainerElementMetaConstraintTree {
```
gunnarmorling: Is this type only short-living during initialisation or are instances kept at runtime? In case of the latter I'm a bit concerned about the non-final fields in here. A short class comment would be nice, too.

gsmet: As mentioned on HipChat, it's just a temporary data structure to join the the cascading metadata with the constraint violations. I'll add a comment.

gsmet: @gunnarmorling added a comment here: https://github.com/hibernate/hibernate-validator/pull/770/files#diff-28ddc9ae996d630baa95de971c31923fR203

gunnarmorling:
Rebased and merged. Many thanks, @gsmet!

PR #771, opened 2017-06-23
HV-1352 Only accessing validated value after a value unwrapper has se…

gunnarmorling:
…t (if present)


gunnarmorling:
@marko-bekhta That's the new PR for HV-1352. WDYT?

marko-bekhta:
Looks good !

gsmet:
Merged to 5.4.

PR #772, opened 2017-06-23
HV-1366 Declaring the exact same value extractor twice in XML should throw an exception

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1366


gsmet:
Merged with some minor adjustments!

Thanks!

PR #773, opened 2017-06-23
HV-1151 Update the XML schema images of the documentation for BV 2.0

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1151

So here's one of those CR1 items :) 

I tried to open the same kind of boxes as in the existing images. If something else is needed to be opened - let me know. 

Also I did find an example 11.2 using `validation-configuration-1.0.xsd` should it maybe be updated to 2.0 as well ?


gsmet:
Merged, thanks! I also removed the old ones.


> Also I did find an example 11.2 using validation-configuration-1.0.xsd should it maybe be updated to 2.0 as well ?

@marko-bekhta could you file an issue for it? Thanks!

marko-bekhta:
> @marko-bekhta could you file an issue for it? Thanks!

@gsmet done - https://hibernate.atlassian.net/browse/HV-1393

PR #774, opened 2017-06-23
HV-1392 Annotation Processor should check that constraint payload does not contains both Unwrap and Skip at the same time

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1392

Even though it's not a CR1 item I just thought that we should have something for AP added as well, cannot just leave it as is :)


marko-bekhta:
Also another thing related to AP.... I was looking through Jira issues and I saw this example:
```java
private class BeanWithWrapperWithTwoTypeArguments {
	@Min(value = 10, payload = { Unwrapping.Unwrap.class })
	private final WrapperWithTwoTypeArguments<Long, String> wrapper = new WrapperWithTwoTypeArguments<>( 5L, "value" );
}
```
AP will fail in such cases as `@Min` is not allowed for such type... So I think type check in AP should be improved a bit.

gsmet:
@marko-bekhta yeah, I don't think we will be able to make this work consistently. Typically, what about types with an `@UnwrapByDefault` value extractor?

Or in your case, how would we know if the Long or the String is the type we target...

Pinging @gunnarmorling so he's aware of this limitation.

gsmet:
Merged, thanks!

PR #775, opened 2017-06-24
HV-1334 Follow-up to-dos around value extractor implementation

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1334

This one only addresses the part about node keys:

> Revisit formats for node keys (from Emmanuel Bernard: I would have preferred something like "addressByType[key('too short')].value", At least for strings and any type that we don't know and will be toStringed. We make exceptions for numbers, booleans, dates

There's no exclusion for `java.util.Date` and `java.util.Calendar` at this moment. Let me know if that's needed and I'll make the updates.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
@@ -7,6 +7,7 @@
 package org.hibernate.validator.internal.engine.path;
 
 import java.io.Serializable;
+import java.time.temporal.TemporalAccessor;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


gunnarmorling:
Hey @marko-bekhta, after discussing with @gsmet a bit, we've concluded that the current representation is fine. Specifically, we'll add "&lt;K&gt;" if the node is about a key of a map, so it already is distinguishable from map values (where we don't add "&ltV&gt;" to produce the same string as in HV 5). More generally, we add the type parameter for all container elements unless the "known ones" from BV 1 (map values, iterable/collection elements). So I hope you don't mind if I'm going to close this one. Thanks nevertheless for your effort!

marko-bekhta:
@gunnarmorling  that's totally fine :)

PR #776, opened 2017-06-24
HV-1387 Add tests for the string representation of the property path

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1387

I looked through the tests that we already have. And I didn't want to duplicate beans and stuff to just validate a string representation of the path. So here's what I came up with...

I've added `ConstraintViolationAssert#violationOfAnything` which allows to ignore the check of actual constraint but you can still do all the other stuff like path, message etc. I also added `String propertyPathString` to `ViolationExpectation` so string representations can be tested using the same approach. And as a result I replaced all usages of `ConstraintViolationAssert#assertCorrectPropertyPaths`.

Looking at the changed tests it seems that string representation of paths covers the cases that you mentioned. Or should we have a set of dedicated tests just for this string representation?

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/cfg/ScriptAssertDefTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/cfg/ScriptAssertDefTest.java
@@ -6,7 +6,7 @@
  */
 package org.hibernate.validator.test.cfg;
 
-import static org.hibernate.validator.testutil.ConstraintViolationAssert.assertCorrectPropertyPaths;
+import static org.hibernate.validator.testutil.ConstraintViolationAssert.assertCorrectPropertyPathStringRepresentations;
 import static org.testng.Assert.assertTrue;
 
 import java.time.Instant;
@@ -88,7 +88,7 @@ public class ScriptAssertDefTest {
 				)
 		);
 
-		assertCorrectPropertyPaths( violations, propertyPath );
+		assertCorrectPropertyPathStringRepresentations( violations, propertyPath );
 	}
 
 	/**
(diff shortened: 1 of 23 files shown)


gsmet:
Hmmm, I'm not sold on this one.

I really think it would be better to have a separate test for the string representation, similar to what I did for container elements' string representation. I like the idea of being able to see the behavior of the string representation at a quick glance and having a quick overview of all the cases. Having it widespread everywhere is not a very good idea IMHO.

The other thing that makes me lean this way is that I don't like very much the property path being handled by the `containsOnlyViolations` facility: I really would like to have people compare the paths instead.

Basically, `assertCorrectPropertyPaths` should only be used in the string representation tests.

marko-bekhta:
yeah ... I was 50-50 on this one. Having dedicated test for string representation sounds good. But on the other hand we would write similar things that are already in the code to test that path...

So if you think that having dedicated tests is the way to go for this one - I'll agree with you :)

I'll rework this PR, the only question would be - do we want that part about `ConstraintViolationAssert#violationOfAnything` and string representation validation removed from `containsOnlyViolations` as well ?

gsmet:
> So if you think that having dedicated tests is the way to go for this one - I'll agree with you :)

Yes, I think it's the right path. There will be some duplication for the validated beans but I don't think it's an issue.

Maybe we could create a stringrepresentation package in `internal.engine.path`? And move ContainerElementStringRepresentationTest there too?

> I'll rework this PR, the only question would be - do we want that part about ConstraintViolationAssert#violationOfAnything and string representation validation removed from containsOnlyViolations as well ?

Yes. In these tests, we will use `assertCorrectPropertyPaths` that we might rename `assertCorrectPropertyPathStringRepresentations`.

gsmet:
@marko-bekhta FYI, we plan to release CR1 tomorrow.

It would be nice if we could have these tests in by then! If not, we'll push them in the next CR.

marko-bekhta:
@gsmet I see ... I did start working on those, but the changes are only on my laptop at home :)
I think I'll have something for this release, but closer to today's evening.

gsmet:
@marko-bekhta OK, perfect. I plan to wrap the release on tomorrow late afternoon.

marko-bekhta:
@gsmet here's what I have done atm. I changed the function name in a separate commit so it'll be easier to read the tests, but it can be probably squashed later.

gsmet:
Merged, thanks!

PR #777, opened 2017-06-24
HV-1325 Add an Automatic-Module-Name entry to the jar manifests

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1325

I've added this new entry to manifests and also rearranged packages in cdi module. I'm not sure about the last part though ... I've done it so that CDI module has one base package `org.hibernate.validator.cdi` and then there's an `internal` subpackage.

And you should probably check the names that I've used as those would probably stay for a long time if released :)


documentation/pom.xml
```diff
@@ -42,6 +42,7 @@
 
         <forbiddenapis-junit.path>forbidden-allow-junit.txt</forbiddenapis-junit.path>
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
+        <automatic.module.name>org.hibernate.validator.referenceguide</automatic.module.name>
```
gsmet: I don't think this one is necessary: we don't publish anything and it's not supposed to be used.

osgi/integrationtest/pom.xml
```diff
@@ -26,6 +26,7 @@
 
         <forbiddenapis-junit.path>forbidden-allow-junit.txt</forbiddenapis-junit.path>
         <hibernate-validator-parent.path>../..</hibernate-validator-parent.path>
+        <automatic.module.name>org.hibernate.validator.osgi.integrationtest</automatic.module.name>
```
gsmet: I don't think this one is necessary: we don't publish anything and it's not supposed to be used.

integration/pom.xml
```diff
@@ -24,6 +24,7 @@
         <wildflyTargetDir>${basedir}/target/wildfly-${wildfly.version}</wildflyTargetDir>
         <wildflyModulesDir>${wildflyTargetDir}/modules/system/layers/base</wildflyModulesDir>
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
+        <automatic.module.name>org.hibernate.validator.integration</automatic.module.name>
```
gsmet: I don't think this one is necessary: we don't publish anything and it's not supposed to be used.

performance/pom.xml
```diff
@@ -34,6 +34,7 @@
         <beanvalidation-impl.version>none</beanvalidation-impl.version>
 
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
+        <automatic.module.name>org.hibernate.validator.performance</automatic.module.name>
```
gsmet: I don't think this one is necessary: we don't publish anything and it's not supposed to be used.

test-utils/pom.xml
```diff
@@ -18,6 +18,7 @@
 
     <properties>
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
+        <automatic.module.name>org.hibernate.validator.testutil</automatic.module.name>
```
gsmet: I would choose to be consistent with the artifact name instead of the package name and use testutils with a trailing s.

annotation-processor/pom.xml
```diff
@@ -21,6 +21,7 @@
 
     <properties>
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
+        <automatic.module.name>org.hibernate.validator.ap</automatic.module.name>
```
gsmet: Let's stick to the artifact name convention: this one would be annotationprocessor.

marko-bekhta:
@gsmet I've removed a property from the poms that you've mentioned. As for the CDI - I just saw that it was "sort of inverted" and that was a bit odd for me that's why I said that I'm not sure about that change ...  as for the output of `jar --describe-module` here's the one for CDI before changes:

```
hibernate.validator.cdi@6.0.0-SNAPSHOT automatic
requires java.base mandated
provides javax.enterprise.inject.spi.Extension with org.hibernate.validator.internal.cdi.validationextension
contains org.hibernate.validator.cdi
contains org.hibernate.validator.internal.cdi
contains org.hibernate.validator.internal.cdi.interceptor
```
and the one after changes:
```
org.hibernate.validator.cdi@6.0.0-SNAPSHOT automatic
requires java.base mandated
provides javax.enterprise.inject.spi.Extension with org.hibernate.validator.cdi.internal.validationextension
contains org.hibernate.validator.cdi
contains org.hibernate.validator.cdi.internal
contains org.hibernate.validator.cdi.internal.interceptor
```

Had run it with:
```
$ java -version
java version "9"
Java(TM) SE Runtime Environment (build 9+175)
Java HotSpot(TM) 64-Bit Server VM (build 9+175, mixed mode)
```

gsmet:
@gunnarmorling do you agree with the naming convention? If so, I'll proceed with this one.

tck-runner/pom.xml
```diff
@@ -28,6 +28,7 @@
         <remote.debug />
 
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
+        <automatic.module.name>org.hibernate.validator.tckrunner</automatic.module.name>
```
gunnarmorling: That shouldn't be needed, this module isn't distributed.

gunnarmorling:
I'll remove the superfluous header for the tck-runner, add something to the docs and merge.

marko-bekhta:
@gunnarmorling ok great! I was just thinking to update the PR. So I will not  :)
And one more thing that I've noticed. for those jars without `automatic.module.name` property set if you do `jar --describe-module` it'll fail with something like :
```
Unable to derive module descriptor for: some.jar
invalid header field
```
I have a change where the new manifest entry is added only to the jars that need it. So if you'd like I can force push an update :)

gunnarmorling:
Yes, that'd be great. I was wondering about that very behaviour for modules
with an empty name. I'll limit myself to the doc update then. Thanks!

marko-bekhta:
done. Updated that first commit and forced pushed it all.

gunnarmorling:
Rebased and applied. Thanks a lot, @marko-bekhta. I've pushed one more commit for adding the module names to the docs, too.

PR #778, opened 2017-06-26
HV-1394 Disable the access log of the security manager by default

gsmet:
* https://hibernate.atlassian.net/browse/HV-1394

### DIFF
--- a/tck-runner/pom.xml
+++ b/tck-runner/pom.xml
@@ -177,7 +177,9 @@
                 <!-- Use Local for testing without security manager -->
                 <arquillian.protocol>LocalSecurityManagerTesting</arquillian.protocol>
                 <!-- Remove for testing without security manager -->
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/hibernate/validator/internal/util/HibernateValidatorPermission.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.util;
```
gunnarmorling: It should be public as users will need to specify it in their policies.

PR #780, opened 2017-06-26
HV-1381 Fix value extraction resolution logic for forced and implicit unwrapping

gsmet:
* https://hibernate.atlassian.net/browse/HV-1381

Test cases added in the TCK as part as: https://github.com/beanvalidation/beanvalidation-tck/pull/132


gsmet:
Added one more commit to introduce the helper class we talked about the other day as I found another method to put in it.

gunnarmorling:
Rebased and merged; thanks, @gsmet!

PR #782, opened 2017-06-27
HV-1396 Move non-public classes of annotation processor to org.hibern…

gsmet:
…ate.validator.ap.internal

 * https://hibernate.atlassian.net/browse/HV-1396


gunnarmorling:
Rebased and applied. Thanks, @gsmet! I've added a note to the [migration guide](https://developer.jboss.org/wiki/HibernateValidatorMigrationGuide#jive_content_id_600CR1), altough I don't expect any outside usage of these classes anyways. Specifically, the AP isn't part of WF.

PR #788, opened 2017-06-30
HV-1409 Fix ContainerClassTypeParameterAndExtractor toString()

gsmet:
* https://hibernate.atlassian.net/browse/HV-1409


gunnarmorling:
Good spot :)

PR #789, opened 2017-07-03
HV-1408 Add small usage tests for each HV annotation

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1408

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/constraints/annotations/AbstractConstrainedTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/constraints/annotations/AbstractConstrainedTest.java
@@ -1,0 +1,26 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.test.constraints.annotations;
+
+import static org.hibernate.validator.testutils.ValidatorUtil.getValidator;
+
+import javax.validation.Validator;
+
+import org.testng.annotations.BeforeMethod;
+
+/**
+ * @author Marko Bekhta
+ */
+public abstract class AbstractConstrainedTest {
+
+	protected Validator validator;
+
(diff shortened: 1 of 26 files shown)


engine/src/test/java/org/hibernate/validator/test/constraints/annotations/AbstractConstrainedTest.java
```diff
@@ -0,0 +1,20 @@
+package org.hibernate.validator.test.constraints.annotations;
```
gunnarmorling: The files are missing the right license hearder. You can add them by running `mvn license:format -pl engine`.

engine/src/test/java/org/hibernate/validator/test/constraints/annotations/hv/EmailConstrainedTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.hibernate.validator.constraints.Email;
+import org.hibernate.validator.test.constraints.annotations.AbstractConstrainedTest;
+
+import org.testng.annotations.Test;
+
+/**
+ * @author Marko Bekhta
+ */
+@SuppressWarnings("depricated")
```
gunnarmorling: s/deprecited/deprecated/

marko-bekhta:
Thanks @gunnarmorling, just got to this. It's strange that headers are missing as I remember adding them (at least I think I do:) ). 
As for the removed tests - I actually just moved them between packages. As for the tests themselves - I tried to make them as simple as possible to just test positive/negative scenario to make sure that all implemented validators are picked up. 

I was also thinking to maybe rewrite the country specific constraints tests - to have a test for the validator and in the tests of this new package just do a simple verification that annotations work. WDYT?

gsmet:
> I was also thinking to maybe rewrite the country specific constraints tests - to have a test for the validator and in the tests of this new package just do a simple verification that annotations work. WDYT?

I don't think it's worth it.

Btw, I removed the package-info file as I don't think it's time well spent to add them for the tests. Let's only add them for the main packages.

Merged. Thanks!

PR #793, opened 2017-07-04
HV-1412 Take into account value unwrapping API changes

gsmet:
* https://hibernate.atlassian.net/browse/HV-1412

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraints.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraints.java
@@ -74,7 +74,7 @@ public class MetaConstraints {
 	private static <A extends Annotation> Type addValueExtractorDescriptorForWrappedValue(TypeResolutionHelper typeResolutionHelper,
 			ValueExtractorManager valueExtractorManager, ConstraintDescriptorImpl<A> constraintDescriptor,
 			List<ContainerClassTypeParameterAndExtractor> valueExtractionPath, ConstraintLocation location) {
-		if ( ValidateUnwrappedValue.NO.equals( constraintDescriptor.validateUnwrappedValue() ) ) {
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/validator/metadata/descriptor/HibernateConstraintDescriptor.java
```diff
@@ -0,0 +1,22 @@
...
+import javax.validation.metadata.ConstraintDescriptor;
+
+/**
+ * Extension to {@code ConstraintDescriptor} which provides functionality
+ * specific to Hibernate Validator.
+ *
+ * @author Guillaume Smet
+ * @since 6.0
+ */
+public interface HibernateConstraintDescriptor<T extends Annotation> extends ConstraintDescriptor<T> {
```
gunnarmorling: What's the motivation for adding it already? Is it just for the sake of testing? I'd rather only add it if we actually need it.

gsmet: No, it's not only for testing. We also need it for the implementation as we need a reference type and I don't think using the `javax.validation` one makes sense.

I don't think it costs us much to have it prepared.

gunnarmorling: As discussed on HipChat I'm not in favor of adding it prematurely. People will only will wonder what the empty interface is about. For the `unwrap()` implementation, we can simply raise an exception at this point, as no type to unwrap to currently is supported. The alternative is to allow to unwrap into `ConstraintDescriptor` and super-types. I don't mind either way.

gsmet: OK, force pushed an update.

gunnarmorling:
Thx!

PR #794, opened 2017-07-04
HV-1423 from of @ConvertGroup and <convert-group /> now has a default…

gsmet:
… value

 * https://hibernate.atlassian.net/browse/HV-1423


gunnarmorling:
Rebased, applied and added one more commit to restore the HV license file header for the XSD files. Seems it got replaced with the BV one. Thanks, @gsmet!

PR #801, opened 2017-07-07
HV-1426 Add entries for com.sun.el to the MANIFEST.MF

gsmet:
@gunnarmorling maybe I missed something obvious?

I have the following warning during the build:
```
[INFO] --- maven-bundle-plugin:3.3.0:manifest (bundle-manifest) @ hibernate-validator ---
[WARNING] Manifest org.hibernate.validator:hibernate-validator:jar:6.0.0-SNAPSHOT : Unused Import-Package instructions: [com.sun.el.*, com.sun.el.lang.*, com.sun.el.util.*]
```
I got them into the MANIFEST by removing the `.*` at the end but it does not work either. I upgraded the maven-bundle-plugin hoping for a more descriptive message as I read they improve this from version to version but no luck.

Tried your trick of using the `ELManager` class loader but it didn't work either.

Note: I commented all the other tests as the expression factory is built statically so as soon as a test loading it succeeded, it doesn't load the `ExpressionFactory` anymore.


gunnarmorling:
@gsmet I've filed https://github.com/hibernate/hibernate-validator/pull/802, it should eventually work.

> Note: I commented all the other tests as the expression factory is built statically so as soon as a test loading it succeeded, it doesn't load the ExpressionFactory anymore.

Ah, that's a good point. I've changed the reactor strategy to "per method" to prevent these caching issues. I'm going to close this PR for the sake of mine. Thanks!

PR #802, opened 2017-07-07
HV-1426 Loading expression factory via CL of the EL module as a fallb…

gunnarmorling:
…ack;

This makes the expression factory usable under OSGi without any further configuration.

https://hibernate.atlassian.net/browse/HV-1426


gsmet:
Rebased and merged, thanks!

PR #805, opened 2017-07-10
HV-1431 Clean up for constraint locations

gsmet:
* https://hibernate.atlassian.net/browse/HV-1431

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/HibernateValidatorPermission.java
+++ b/engine/src/main/java/org/hibernate/validator/HibernateValidatorPermission.java
@@ -1,0 +1,29 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator;
+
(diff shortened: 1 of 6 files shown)


gunnarmorling:
Thanks, @gsmet!

nmittles:
Hi.  The Open Liberty team is looking into using the Bean Validation 2.0 RI, and wants to better understand the motivation for this change that adds accessPrivateMembers?  The permission seems redundant to requiring accessDeclaredMembers, but we're thinking there must be more to it?  Are there specific scenarios where a user would want to have a security manager enabled, but not grant accessPrivateMembers to the Bean Validation 2.0 RI jar?

gunnarmorling:
Hey @nmittles, the reason is that we want to avoid that direct instantiation of classes like `FieldConstraintLocation` gives private access to the represented member to outside callers.

One way for doing this would have been to not use a privileged action for the `setAccessible()` call, so that the caller would be required to have that permission (and use a privileged action themselves, or all frames on the stack above have that permission). But we concluded it'd be better to check for a separate permission which the caller must have. That way the permission for validating private fields can be assigned to the calling code base while not granting general private access at the same time.

Just seeing though that we need to update the documentation, it's currently not described 100% correctly.

PR #806, opened 2017-07-10
HV-1422 Extract properties for base URLs commonly used in the reference guide

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1422

I haven't found any mention of Asciidoctor supporting property files (I saw one suggestion to write a custom processor which can read *.properties file and use values from it while processing). So I thought that this approach might work - having urls extracted to attributes. I've added a separate file just for such attributes and extracted a few urls from first chapter just to test and see if it works. 

If you think such approach is ok I can look for and extract those commonly used ones in such a way.

### DIFF
--- a/annotation-processor/pom.xml
+++ b/annotation-processor/pom.xml
@@ -93,8 +93,8 @@
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <configuration>
                     <links>
-                        <link>http://docs.oracle.com/javase/8/docs/api/</link>
-                        <link>http://docs.jboss.org/hibernate/stable/beanvalidation/api/</link>
+                        <link>${java.doc.base.url}api/</link>
+                        <link>${bv.doc.base.url}</link>
                     </links>
                     <packagesheader>Hibernate Validator Annotation Processor Packages</packagesheader>
                     <doctitle>Hibernate Validator Annotation Processor ${project.version}</doctitle>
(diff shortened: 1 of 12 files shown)


gunnarmorling:
Hey, thanks for taking a look at this one, @marko-bekhta!

Some thoughts:

* Instead of the separate AsciiDoc file, I'd rather declare properties in _documentation/pom.xml_ or even the parent POM, and pass them as attributes to the Maven plug-in; we already do it for [some other values](https://github.com/hibernate/hibernate-validator/blob/master/documentation/pom.xml#L230).

* Instead of extracting the entire URLs, I'd rather extract common bases, most notably

        java_se_docs=http://docs.oracle.com/javase/8/docs/

    This property should also be referenced for the JavaDoc API links in the _distribution_ module.

* I wouldn't bother extracting the JCP URLs, they seem unlikely to ever change. I think I'd go for `java_se_docs`, `bv_api_docs`, and that's about it.
* We still got several links using that weird $$ syntax; so if you feel like it, you could update those to the simpler one used in most places.

marko-bekhta:
Hi @gunnarmorling, I was thinking that there will be much more URLs to extract, hence the additional file :)  But if it's just a few base ones then the attributes in pom looks better. I'll make the corresponding changes and update the PR.

gsmet:
Merged with one additional commit. Thanks!

PR #810, opened 2017-07-11
HV-1434 Make order of <constraint> and <container-element-type> consistent

gsmet:
* https://hibernate.atlassian.net/browse/HV-1434


gunnarmorling:
Merged; thanks, @gsmet!

PR #812, opened 2017-07-12
HV-1405 Review assertCorrectPropertyPathStringRepresentations usage

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1405

I found only that "container array thing" and set examples were missing, based on other usages of `assertCorrectPropertyPathStringRepresentations`. Also I changed the way this method works, because it was not giving expected result in case of Set. 

That `Set` example also got me thinking that maybe the string representation for it should be different ? Maybe for example something like this:
```
"addresses[hashcode_goes_here].street",
```

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/cfg/ScriptAssertDefTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/cfg/ScriptAssertDefTest.java
@@ -6,7 +6,9 @@
  */
 package org.hibernate.validator.test.cfg;
 
-import static org.hibernate.validator.testutil.ConstraintViolationAssert.assertCorrectPropertyPathStringRepresentations;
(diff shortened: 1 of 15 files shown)


gsmet:
@marko-bekhta thanks, looks good! Will merge as soon as we released 6.0.0.Final as we want to limit the changes between CR3 and Final to the minimum as it's the version currently being voted for the spec.

As for your question on `Set`, the first thing is that we're kinda stuck with the representation we have now for backward compatibility reason. Moreover, the hashcode is not that useful as you can extract the value from the `Node` by casting it to the HV specific `PropertyNode` or `ContainerElementNode` which have a `getValue()` method.

gsmet:
Rebased and merged, thanks!

(master is now 6.0.1-SNAPSHOT, I created a 6.0.0-build-branch for 6.0.0.Final)

PR #814, opened 2017-07-19
HV 6 Memory and performance improvements

gsmet:
* https://hibernate.atlassian.net/browse/HV-1437
 * https://hibernate.atlassian.net/browse/HV-1438
 * https://hibernate.atlassian.net/browse/HV-1439
 * https://hibernate.atlassian.net/browse/HV-1441
 * https://hibernate.atlassian.net/browse/HV-1443

Note 2: the order of the commits displayed below is not exactly the one I have in my branch - the order of the HV-1439 commits is more consistent in my branch.

It's a big one. I started working on the different subjects separately but as I touched the same code and had the same issues, I had to cherry-pick some of the commits in other branches and finally ended up joining the efforts in one branch. The commits are well isolated and should be easily understandable by reading the commit comment and the code.

## HV-1438

There are several things here:

 * I removed completely the `ObjectValueExtractor`. It's not a good idea to rely on the extractor infrastructure to validate the annotated object itself. Funnily enough, removing it revealed a few issues (these are the commits before the "Remove the ObjectValueExtractor" commit);
 * Narrowing down the value extractors considered is a necessity as cascading is real slow now, mostly due to the calls to `TypeVariableBindings`. What I did work, but I think I found an issue in the TCK: https://github.com/beanvalidation/beanvalidation-tck/pull/149 .

## HV-1437

These are memory footprint improvements. The idea was mostly to reduce the size of the metadata. The example reported by Stuart is kinda extreme: see https://github.com/keycloak/keycloak/blob/master/model/jpa/src/main/java/org/keycloak/models/jpa/entities/RealmEntity.java .

6.0 was already better than 5.x as I introduce the `CollectionHelper.toImmutable*` utilities: they trim down the empty and one element collections to the minimum but there was still some room for improvements.

All in all, we reduced the `BeanMetaDataImpl` of `RealmEntity` from ~177 kB in the original snapshot of Stuart (using 5.x) to ~60 kB (using this branch).

I'll post something on the list about some other things I'd like to discuss.

### Benchmark improvements

The benchmarks were a bit painful to use when you wanted to compare closely 2 versions and switch from one to the other.

I pushed some clean ups to them.

I also removed the iterative benchmarks which were useless.

### Some runtime performance improvements

Nothing big but a few things I noticed while benchmarking and it definitely improved the performances.

### Quite a lot of parsing/metadata building improvements

The situation was quite bad here so I made quite a lot of improvements to reduce the startup cost of HV.

## HV-1443

A small fix required by the other changes.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -189,22 +189,22 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		sanityCheckPropertyPath( propertyName );
 		sanityCheckGroups( groups );
 
-		ValidationContext<T> context = getValidationContextBuilder().forValidateProperty( object );
+		ValidationContext<T> validationContext = getValidationContextBuilder().forValidateProperty( object );
 
-		if ( !context.getRootBeanMetaData().hasConstraints() ) {
+		if ( !validationContext.getRootBeanMetaData().hasConstraints() ) {
 			return Collections.emptySet();
 		}
 
 		PathImpl propertyPath = PathImpl.createPathFromString( propertyName );
-		ValueContext<?, Object> valueContext = getValueContextForPropertyValidation( context, propertyPath );
+		ValueContext<?, Object> valueContext = getValueContextForPropertyValidation( validationContext, propertyPath );
 
 		if ( valueContext.getCurrentBean() == null ) {
-			throw log.getUnableToReachPropertyToValidateException( context.getRootBean(), propertyPath );
+			throw log.getUnableToReachPropertyToValidateException( validationContext.getRootBean(), propertyPath );
 		}
 
 		ValidationOrder validationOrder = determineGroupValidationOrder( groups );
 
-		return validateInContext( context, valueContext, validationOrder );
+		return validateInContext( validationContext, valueContext, validationOrder );
 	}
 
 	@Override
@@ -213,16 +213,16 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		sanityCheckPropertyPath( propertyName );
 		sanityCheckGroups( groups );
 
-		ValidationContext<T> context = getValidationContextBuilder().forValidateValue( beanType );
+		ValidationContext<T> validationContext = getValidationContextBuilder().forValidateValue( beanType );
 
-		if ( !context.getRootBeanMetaData().hasConstraints() ) {
+		if ( !validationContext.getRootBeanMetaData().hasConstraints() ) {
 			return Collections.emptySet();
 		}
 
 		ValidationOrder validationOrder = determineGroupValidationOrder( groups );
 
 		return validateValueInContext(
-				context,
+				validationContext,
 				value,
 				PathImpl.createPathFromString( propertyName ),
 				validationOrder
@@ -265,42 +265,42 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 	private <T> Set<ConstraintViolation<T>> validateParameters(T object, Executable executable, Object[] parameterValues, Class<?>... groups) {
 		sanityCheckGroups( groups );
 
-		ValidationContext<T> context = getValidationContextBuilder().forValidateParameters(
+		ValidationContext<T> validationContext = getValidationContextBuilder().forValidateParameters(
 				parameterNameProvider,
 				object,
 				executable,
 				parameterValues
 		);
 
-		if ( !context.getRootBeanMetaData().hasConstraints() ) {
+		if ( !validationContext.getRootBeanMetaData().hasConstraints() ) {
 			return Collections.emptySet();
 		}
 
 		ValidationOrder validationOrder = determineGroupValidationOrder( groups );
 
-		validateParametersInContext( context, parameterValues, validationOrder );
+		validateParametersInContext( validationContext, parameterValues, validationOrder );
 
-		return context.getFailingConstraints();
+		return validationContext.getFailingConstraints();
 	}
 
 	private <T> Set<ConstraintViolation<T>> validateReturnValue(T object, Executable executable, Object returnValue, Class<?>... groups) {
 		sanityCheckGroups( groups );
 
-		ValidationContext<T> context = getValidationContextBuilder().forValidateReturnValue(
+		ValidationContext<T> validationContext = getValidationContextBuilder().forValidateReturnValue(
 				object,
 				executable,
 				returnValue
 		);
 
-		if ( !context.getRootBeanMetaData().hasConstraints() ) {
+		if ( !validationContext.getRootBeanMetaData().hasConstraints() ) {
 			return Collections.emptySet();
 		}
 
 		ValidationOrder validationOrder = determineGroupValidationOrder( groups );
 
-		validateReturnValueInContext( context, object, returnValue, validationOrder );
+		validateReturnValueInContext( validationContext, object, returnValue, validationOrder );
 
-		return context.getFailingConstraints();
+		return validationContext.getFailingConstraints();
 	}
 
 	@Override
@@ -366,7 +366,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 
 	/**
 	 * Validates the given object using the available context information.
-	 * @param context the global validation context
+	 * @param validationContext the global validation context
 	 * @param valueContext the current validation context
 	 * @param validationOrder Contains the information which and in which order groups have to be executed
 	 *
@@ -374,7 +374,8 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 	 *
 	 * @return Set of constraint violations or the empty set if there were no violations.
 	 */
-	private <T, U> Set<ConstraintViolation<T>> validateInContext(ValidationContext<T> context, ValueContext<U, Object> valueContext, ValidationOrder validationOrder) {
+	private <T, U> Set<ConstraintViolation<T>> validateInContext(ValidationContext<T> validationContext, ValueContext<U, Object> valueContext,
+			ValidationOrder validationOrder) {
 		if ( valueContext.getCurrentBean() == null ) {
 			return Collections.emptySet();
 		}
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/hibernate/validator/internal/util/annotationfactory/AnnotationProxy.java
```diff
@@ -176,6 +182,18 @@ else if ( m.getDefaultValue() != null ) {
 		return result;
 	}
 
+	private Map<String, Object> getAnnotationValues(Annotation annotation) {
+		if ( Proxy.isProxyClass( annotation.getClass() ) ) {
+			InvocationHandler invocationHandler = Proxy.getInvocationHandler( annotation );
+			if ( invocationHandler instanceof AnnotationProxy ) {
```
gunnarmorling: The commit message is a bit mis-leading IMHO, as it's only about proxies created by ourselves. I first thought you'd access the parameter map of the actual `sun.reflect.annotation.AnnotationInvocationHandler`.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetAnnotationParameters.java
```diff
@@ -0,0 +1,100 @@
...
+			catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
+				throw LOG.getUnableToGetAnnotationParameterException( parameterName, annotation.getClass(), e );
+			}
+		}
+		return new AnnotationParameters( parameters );
+	}
+
+	public static class AnnotationParameters implements Serializable {
+
+		private final Map<String, Object> parameters;
```
gunnarmorling: Would be nice to have the `@Immutable` tag here.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
```diff
@@ -296,9 +297,13 @@
 	private static final String JODA_TIME_CLASS_NAME = "org.joda.time.ReadableInstant";
 	private static final String JAVA_MONEY_CLASS_NAME = "javax.money.MonetaryAmount";
 
-	// immutable
+	@Immutable
 	private final Map<Class<? extends Annotation>, List<? extends ConstraintValidatorDescriptor<?>>> builtinConstraints;
 
+	private final ConcurrentHashMap<Class<? extends Annotation>, Boolean> externalConstraints = new ConcurrentHashMap<>();
```
gunnarmorling: Minor nitpick -- could be declared as `ConcurrentMap`.

gunnarmorling: More generally, though, it seems to go a bit against the goal of reducing memory usage.

engine/src/main/java/org/hibernate/validator/internal/util/TypeVariableBindings.java
```diff
@@ -69,7 +69,7 @@ else if ( genericSuperType instanceof ParameterizedType ) {
 
 				boolean typeParameterFoundInSubType = false;
 				for ( Entry<TypeVariable<?>, TypeVariable<?>> subTypeParameter : bindings.entrySet() ) {
-					if ( typeArgument.equals( subTypeParameter.getValue() ) ) {
+					if ( typeArgument == subTypeParameter.getValue() ) {
```
gunnarmorling: Just to confirm: are we totally sure we get the same instance when asking for the same argument multiple times? Admittedly, it'd feel safer to stick to equals.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -164,25 +223,58 @@ public ValueExtractorDescriptor getMaximallySpecificAndContainerElementCompliant
...
+	 */
+	private Set<ValueExtractorDescriptor> getPotentiallyRuntimeTypeCompliantAndContainerElementCompliantValueExtractors(Type declaredType,
+			TypeVariable<?> typeParameter) {
+		boolean isInternal = TypeVariables.isInternal( typeParameter );
+		Type erasedDeclaredType = TypeHelper.getErasedReferenceType( declaredType );
+
+		Set<ValueExtractorDescriptor> typeCompatibleExtractors = valueExtractors.values()
+				.stream()
+				.filter( e -> TypeHelper.isAssignable( erasedDeclaredType, e.getContainerType() ) )
+				.collect( Collectors.toSet() );
```
gunnarmorling: Could we collect into our immutable helper right away?

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -786,4 +786,11 @@ ValueExtractorDefinitionException getOnlyUnboundWildcardTypeArgumentsSupportedFo
 
 	@Message(id = 226, value = "Container element constraints and cascading validation are not supported on arrays: %1$s")
 	ValidationException getContainerElementConstraintsAndCascadedValidationNotSupportedOnArraysException(@FormatWith(TypeFormatter.class) Type type);
+
+	@Message(id = 227, value = "The value extractor candidate set may not be empty for type: %1$s.")
+	IllegalArgumentException getValueExtractorCandidateSetMayNotBeEmptyException( @FormatWith(ClassObjectFormatter.class) Class<?> valueType );
```
gunnarmorling: Not a specific BV exception?

engine/src/main/java/org/hibernate/validator/internal/xml/ConstrainedExecutableBuilder.java
```diff
@@ -322,17 +322,14 @@ private CascadingTypeParameter parseReturnValueType(ReturnValueType returnValueT
...
-		boolean isArray = executable instanceof Method && ( (Method) executable ).getReturnType().isArray();
 		Type type = ReflectionHelper.typeOf( executable );
 		boolean isCascaded = returnValueType.getValid() != null;
 		Map<Class<?>, Class<?>> groupConversions = groupConversionBuilder.buildGroupConversionMap(
 				returnValueType.getConvertGroup(),
 				defaultPackage
 		);
 
-		return isArray
-				? CascadingTypeParameter.arrayElement( type, isCascaded, containerElementTypesCascadingMetaData, groupConversions )
```
gunnarmorling: Is that method `array` element obsolete wit that? I.e. can it be removed?

gsmet: It's not removed because we still use it in the code I kept if we want to revive array support. It makes sense for the array elements but here we are talking about the whole array so it should be an annotated object.

engine/src/main/java/org/hibernate/validator/internal/util/TypeVariableBindings.java
gsmet: I think so. You compare the type arguments of the same class so they should be identical.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
gsmet: Yes, I know. But I think it's sufficiently beneficial and we improved a lot the memory footprint so we can afford it.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetAnnotationParameters.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetAnnotationParameters.java
```diff
@@ -0,0 +1,100 @@
...
+			this.parameters = CollectionHelper.toImmutableMap( parameters );
+		}
+
+		public Map<String, Object> getParameters() {
+			return parameters;
+		}
+
+		@SuppressWarnings("unchecked")
+		public <T> T getMandatoryParameter(String parameterName, Class<T> type) {
+			if ( !parameters.containsKey( parameterName ) ) {
```
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/util/annotationfactory/AnnotationProxy.java
gsmet: Not sure what is unclear. I replaced `is a proxy` by `is an AnnotationProxy` in the commit message.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
gsmet: This is a transient set, it's not reused after that so I don't think it makes sense to make it immutable here.

We usually only make collections immutable when we store them in a field.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gsmet: Changed it to use `Contracts`.

engine/src/main/java/org/hibernate/validator/internal/util/TypeVariableBindings.java
gsmet: OK, considering this part is not a hot spot anymore, I think we can go back to using equals. It made a big difference when `TypeVariableBindings` was called at runtime but it's probably negligible now that it's only a startup cost.

I don't see a case where `==` would fail but as it's deep into Java, let's be on the safe side.

gsmet:
@gunnarmorling force pushed an update fixing your remarks.

I answered to the others, let me know if there's something else you want me to fix.

gsmet:
@gunnarmorling added 2 more commits to remove the metadata of the unconstrained methods (one is purely cosmetic).

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
```diff
@@ -280,8 +306,24 @@ public PropertyMetaData getMetaDataFor(String propertyName) {
 	}
 
 	@Override
-	public ExecutableMetaData getMetaDataFor(Executable executable) {
-		return executableMetaDataMap.get( ExecutableHelper.getSignature( executable ) );
+	public Optional<ExecutableMetaData> getMetaDataFor(Executable executable) {
+		String signature = ExecutableHelper.getSignature( executable );
```
gsmet: First I thought about having 2 separate methods but the call to `ExecutableHelper.getSignature( executable )` has a higher cost than expected. Thus, it's better to have only one method.

I think the `Optional` makes it clear that the behavior is a bit different. We might even rename the method to differentiate it better? Maybe `getConstraintMetaDataFor()`?

engine/src/main/java/org/hibernate/validator/internal/util/annotationfactory/AnnotationProxy.java
gunnarmorling: Ok, that was what I meant.

engine/src/main/java/org/hibernate/validator/internal/util/TypeVariableBindings.java
gunnarmorling: Ok, cool. My concern is that it's relying on an implementation detail of the class library whether "==" is fine or not.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
gunnarmorling: No strong opinion on the rename.

PR #817, opened 2017-07-30
Update ValidationMessages.properties

Hilmerc:
Adding 2 properties javax.validation.constraints.PastOrPresent.message and javax.validation.constraints.FutureOrPresent.message  
These are using from javax/validation/constraints


gsmet:
@Hilmerc good catch.

I won't use your error messages as they are defined in the spec so I'm gonna use the ones from the spec but thanks for the report!

PR #824, opened 2017-08-01
HV-1455 Be more consistent about when sanity checks are applied

gsmet:
* https://hibernate.atlassian.net/browse/HV-1455

Quick explanation: right now, we are a bit inconsistent about when the sanity checks are applied: sometimes it's after testing that there are no constraints, sometimes it's after. We should be consistent in 2 ways:
 * for all the different methods
 * for all the different types of sanity checks


gunnarmorling:
LGTM. Thanks, @gsmet!

PR #826, opened 2017-08-01
HV-1445 Remove the metadata cache in AnnotationMetaDataProvider

gsmet:
We only keep the Object metadata handy in a SoftReference.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
@@ -9,7 +9,6 @@ package org.hibernate.validator.internal.metadata.provider;
 import static org.hibernate.validator.internal.util.CollectionHelper.newArrayList;
 import static org.hibernate.validator.internal.util.CollectionHelper.newHashMap;
 import static org.hibernate.validator.internal.util.CollectionHelper.newHashSet;
-import static org.hibernate.validator.internal.util.ConcurrentReferenceHashMap.ReferenceType.SOFT;
 import static org.hibernate.validator.internal.util.logging.Messages.MESSAGES;
 
 import java.lang.annotation.Annotation;
@@ -63,7 +62,6 @@ import org.hibernate.validator.internal.metadata.raw.ConstrainedField;
 import org.hibernate.validator.internal.metadata.raw.ConstrainedParameter;
 import org.hibernate.validator.internal.metadata.raw.ConstrainedType;
 import org.hibernate.validator.internal.util.CollectionHelper;
-import org.hibernate.validator.internal.util.ConcurrentReferenceHashMap;
 import org.hibernate.validator.internal.util.ExecutableHelper;
 import org.hibernate.validator.internal.util.ReflectionHelper;
 import org.hibernate.validator.internal.util.TypeResolutionHelper;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -103,11 +104,8 @@ public AnnotationMetaDataProvider(ConstraintHelper constraintHelper,
 		this.typeResolutionHelper = typeResolutionHelper;
 		this.valueExtractorManager = valueExtractorManager;
 		this.annotationProcessingOptions = annotationProcessingOptions;
-		this.configuredBeans = new ConcurrentReferenceHashMap<>(
-				DEFAULT_INITIAL_CAPACITY,
-				SOFT,
-				SOFT
-		);
+
+		this.objectBeanConfigurationReference = new AtomicReference<>( new SoftReference<>( retrieveBeanConfiguration( Object.class ) ) );
```
gsmet: So the object metadata is now always initialized. We can move it to a DCL but I thought with the SoftReference, it would be a bit complicated to follow. Opinions welcome.

Sanne:
I suspect using SOFT references is going to give you some trouble.

I don't know all details of the problem but I suspect a WEAK reference would be less troublesome.

On several implementations a Soft reference is only collected as last resort, i.e. on "full-gc" events.
When a full gc is triggered you probably are in big trouble already especially on the CPU usage, and at that point it will release _all of them_! Validations in progress will have to rebuild metadata on the fly, aggravating the situation as your CPU is already on tipping point.

Also with this strategy a container like WildFly would only release references to the classes - which have possibly been undeployed - on full gc events. So it would look like there's a memory leak, always growing, until finally a full gc event is triggered; for some use cases that's not considered acceptable.

With a weak reference on the other hand metadata would vanish periodically (each time gc notices it could remove them). I don't think that's too bad as you'd at least prevent _some_  metadata rebuilds: all those validations which happen approxximately in parallel or close enough would benefit from the same cached metadata. This is generally nicer as the computation cost is spread and overall performance more predictable.

You'd not want to use an exclusively synchronized block though; allow some more optimistic locking: in worst case some threads will have re-computed the same metadata. Or reduce the granularity of the lock?

To have your cake and eat it too you probably want to change the structure a bit, to use a weak reference on the validated Class. The Class will not be periodically reclaimed, which is good, but it will be reclaimed on undeployment of the class so that gives you the right level.

BTW by synchronizing on `AnnotationMetaDataProvider.class` that would kill any parallelism even across independent deployments, so you might want to rethink the caching strategy anyway. If you really need such a coarse lock, use a non static instance. (I didn't check the lock design! trust you on that)

gsmet:
@Sanne thanks for taking a look at it.

First, it's important to know that this is only used when building the metadata of a bean, not at runtime. So it's basically used only at the startup of the application when the first bean of a given type is validated. It might be used later if we only validate a bean of a given type at a later point but anyway, it won't add any pressure when doing the full gc as we probably wouldn't use it anymore.

It was using SoftReference in the ConcurrentReferenceHashMap so I kept that. Maybe it can be moved to a WeakReference, I don't have a strong opinion about it. @gunnarmorling WDYT?

As for the lock granularity, you're right, I reduced it.

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -81,19 +80,18 @@
...
-	protected final ConcurrentReferenceHashMap<Class<?>, BeanConfiguration<?>> configuredBeans;
-	protected final AnnotationProcessingOptions annotationProcessingOptions;
-	protected final ValueExtractorManager valueExtractorManager;
+	private final ConstraintHelper constraintHelper;
+	private final TypeResolutionHelper typeResolutionHelper;
+	private final AnnotationProcessingOptions annotationProcessingOptions;
+	private final ValueExtractorManager valueExtractorManager;
+
+	private volatile SoftReference<BeanConfiguration<Object>> objectBeanConfigurationReference;
+	private final Object objectBeanConfigurationReferenceLock = new Object();
```
Sanne: you can avoid creating this object by using synchronized on this AnnotationMetaDataProvider ?

gsmet: Yeah, you're right. I don't expect other synchronized blocks in this class. I'll push something along this way.

gunnarmorling:
@Sanne I don't quite follow on your concerns about an ever growing data structure. This is a single meta-data (representing java.lang.Object), one instance of it will exist per validator (or validator factory). I wouldn't even bother making this an `*Reference` at all, just a plain `volatile BeanConfiguration<Object>`, it seems not worth to make this single reference collectable.

If we indeed wanted to use a `Reference`, I think `SoftReference` is more appropriate, as metadata for `Object` sub-types is calculated lazily, so a weak reference would very likely never be of any benefit.

+1 on the lock change.

gsmet:
Merged!

PR #828, opened 2017-08-03
HV-1363 : custom property selector

asereda-gs:
Allows to validate entities not following standard java bean convention (see [HV-1363](https://hibernate.atlassian.net/browse/HV-1363)) via pluggable interface.

This PR is intended to start a discussion and review. Work is not yet finished (so please don't merge).

So far I have the following questions  : 
1. `PropertyAccessorSelector` interface. Does it have too many / too little methods ? 
What about reverse lookup ( string -> method ) : 
```java
Member find(Class klass, String property)
```
which is used in [TypeConstraintMappingContextImpl](https://github.com/hibernate/hibernate-validator/blob/70c879b776d474f174afdc6c8effb8da302f0f2f/engine/src/main/java/org/hibernate/validator/internal/cfg/context/TypeConstraintMappingContextImpl.java#L266) ?

2. Not really sure how to pass `ExecutableHelper` to `DefaultConstraintMapping` (see [ConfigurationImpl](https://github.com/hibernate/hibernate-validator/blob/5b9192e6cfd39852b6073ac247c4169876041992/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java#L264)) since the former is instantiated in `ValidatorFactoryImpl` but the later is needed to construct metadata. Obviously both need to share same instance of `ExecutableHelper`.

I have purposefully omitted docs and unit tests until we agree on API and implementation.  

Please forgive any inaccuracies I'm not, yet, fully familiar with whole codebase  .


gunnarmorling:
ok to test

gunnarmorling:
Hey @asereda-gs, thanks a lot! We're going to look into this ASAP.

asereda-gs:
As you will notice, most of the code is passing `ExecutableHelper` as dependency. I'm not sure if there is a less verbose way.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/ConfiguredConstraint.java
```diff
@@ -41,7 +41,7 @@ private ConfiguredConstraint(ConstraintDef<?, A> constraint, ConstraintLocation
 		return new ConfiguredConstraint<>( constraint, ConstraintLocation.forClass( beanType ), ElementType.TYPE );
 	}
 
-	static <A extends Annotation> ConfiguredConstraint<A> forProperty(ConstraintDef<?, A> constraint, Member member) {
+	static <A extends Annotation> ConfiguredConstraint<A> forProperty(ConstraintDef<?, A> constraint, Member member, ExecutableHelper executableHelper) {
```
gsmet: Usually, we add the helpers as the first parameters of the methods.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/ConstructorConstraintMappingContextImpl.java
```diff
@@ -17,8 +18,8 @@
  */
 class ConstructorConstraintMappingContextImpl extends ExecutableConstraintMappingContextImpl implements ConstructorConstraintMappingContext {
 
-	<T> ConstructorConstraintMappingContextImpl(TypeConstraintMappingContextImpl<T> typeContext, Constructor<T> constructor) {
-		super( typeContext, constructor );
+	<T> ConstructorConstraintMappingContextImpl( TypeConstraintMappingContextImpl<T> typeContext, Constructor<T> constructor, ExecutableHelper executableHelper) {
```
gsmet: Please don't add a space here. We have https://github.com/hibernate/hibernate-ide-codestyles that might be of help.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/TypeConstraintMappingContextImpl.java
```diff
@@ -263,6 +267,7 @@ private Member getMember(Class<?> clazz, String property, ElementType elementTyp
 			member = run( GetDeclaredField.action( clazz, property ) );
 		}
 		else {
+			// TODO add find method to PropertyAccessorSelector ?
```
gsmet: Yup, looks like we will need one.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -262,7 +277,8 @@ public MethodValidationConfiguration getMethodValidationConfiguration() {
 
 	@Override
 	public final DefaultConstraintMapping createConstraintMapping() {
-		return new DefaultConstraintMapping();
+		// TODO should be single instance of ExecutableHelper per validator
```
gsmet: +1

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -168,14 +168,14 @@ public ValidatorFactoryImpl(ConfigurationState configurationState) {
 			tmpFailFast = hibernateSpecificConfig.getFailFast();
 
 			tmpAllowOverridingMethodAlterParameterConstraint =
-					hibernateSpecificConfig.getMethodValidationConfiguration()
-							.isAllowOverridingMethodAlterParameterConstraint();
+							hibernateSpecificConfig.getMethodValidationConfiguration()
+											.isAllowOverridingMethodAlterParameterConstraint();
```
gsmet: Formatting issue. It would be nice if you could review your patch with Eclipse or something and revert all the formatting noise.

engine/src/main/java/org/hibernate/validator/spi/property/JavaBeanPropertySelector.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.reflect.Method;
+
+/**
+ * Default property accessor which follows java bean naming convention.
+ */
+public class JavaBeanPropertySelector implements PropertyAccessorSelector {
+
+	@Override
+	public MethodType methodType( Method method ) {
+		return ReflectionHelper.isGetterMethod( method ) ?  MethodType.PROPERTY_READ_ACCESSOR : MethodType.OTHER;
```
gsmet: Let's call them `getMethodType()` and `getPropertyName()`.

engine/src/main/java/org/hibernate/validator/spi/property/JavaBeanPropertySelector.java
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	@Override
+	public String propertyName( Method method ) {
+		return ReflectionHelper.getPropertyName( method );
+	}
+
+	@Override
+	public boolean supports( Method method ) {
```
gsmet: This method doesn't look to be used?

engine/src/main/java/org/hibernate/validator/spi/property/JavaBeanPropertySelector.java
gsmet: Btw, it looks like we don't need this method at all. Let's simply have a `isGetterMethod()` declared and remove the `methodType()` method altogether (and we don't need the default method in the interface then).

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -171,6 +172,20 @@
...
+
+	/**
+	 * Registers custom property accessor which allows extracting bean properties with different naming convention than
+	 * standard Java Beans (eg. {@code foo()} vs {@code getFoo()}).
+	 *
+	 * @param selector custom property selector
+	 *
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 6.0
```
gunnarmorling: It's an interesting question really.

Strictly speaking, it'd be 6.1 as it's a new feature. But then, the cadence of minor releases has been slow for us traditionally, so should we wait with shipping this for 6 months (arbitrary number just for the sake of discussing it)?

Or should we move to a Chrome/Firefox style and simply release minor versions as soon as there's one new feature, in which case we may arrive at HV 6.42 quickly? Or should we allow ourselves to put out new features like this in micro-versions, too? It's an open question which we probably should take to hibernate-dev, I haven't made up my mind. Only that I wouldn't like holding back finished features for a long time.

engine/src/main/java/org/hibernate/validator/spi/property/PropertyAccessorSelector.java
```diff
@@ -0,0 +1,33 @@
...
+
+	MethodType methodType( Method method );
+
+	String propertyName( Method method );
+
+	default boolean isGetterMethod( Method method ) {
+		return methodType( method ) == MethodType.PROPERTY_READ_ACCESSOR;
+	}
+
+	boolean supports( Method method );
```
gunnarmorling: What's that method about?

engine/src/main/java/org/hibernate/validator/spi/property/PropertyAccessorSelector.java
```diff
@@ -0,0 +1,64 @@
...
+	 *
+	 * TODO: should this support non-getters methods as well (like setters, void return type etc.) ?
+	 *
+	 * <p>Example for classic java bean convention: {@code getFoo()} -> {@code foo}</p>
+	 *
+	 * @param method method to be analyzed
+	 * @return property name which is exposed by this method, {@code null} if method is unknown to current
+	 * selector instance.
+	 */
+	String getPropertyName(Method method);
```
gunnarmorling: Hi @asereda-gs, thanks a lot for moving forward with this work!

The contract looks ok for Java-based models, but one related future work I'd like us to keep in mind is [HV-1000](https://hibernate.atlassian.net/browse/HV-1000) which is about supporting other models than JavaBeans, e.g. map-based models (with key/value pairs as properties), or JSON-based models etc. This will not let us with `Method` objects, though. Instead we'd have to work with a more abstract representation of "types" and "properties". E.g. something like this:

    public interface PropertySelector {

        Set<Property> getProperties(Type type);
    }

    interface Property {
        String getName();
        Object getValueFrom(Object bean);
    }

    class JavaBeansField implements Property {
        ...
    }

    class JavaBeansGetter implements Property {
        ...
    }

    interface Type { ... }

    /* Implementation based on classes */
    class JavaBean implements Type { ... }

Then, your selector would return `JavaBeansGetter` based on your more relaxed naming pattern.

Do you think you could explore this approach a bit? I realize it's asking a lot, the vision is blurry and it's quite an expansion of what's needed for your very problem. But it'd be great if we could avoid introducing an SPI contract which gets obsolete soon again. Any thoughts?

asereda-gs: Hi Gunnar,

I'll take a look. 

How `Property` is different from `ConstrainedElement` in this case ?

gunnarmorling: Good question, there'd be similarities for sure, i.e. probably a corresponding `ConstrainedElement` would be derived from such `Property`. In general, I'd try to make the SPI-level contract as small as possible, i.e. we shouldn't expose `ConstrainedElement` itself.

gunnarmorling:
Hey @asereda-gs, @gsmet, where are we with this one? Did you have a chance to explore the alternative SPI approach discussed above?

asereda-gs:
Sorry I got side-tracked with other stuff. Should resume work on Property model shortly.

gsmet:
@asereda-gs cool, thanks for the heads up. Feel free to ping us to discuss the design more in depth if you need any help!

gsmet:
Closing this one. @marko-bekhta is working on the new approach suggested by Gunnar.

PR #829, opened 2017-08-03
HV-1446 Remove UnconstrainedEntityMetaDataSingleton

gsmet:
We can't use it as it prevents to return proper errors when a property
or a method doesn't exist in a bean.

 * https://hibernate.atlassian.net/browse/HV-1446


gunnarmorling:
Merged; thanks, @gsmet!

PR #831, opened 2017-08-03
HV-1459 HV-1460 Build improvements

gsmet:
@gunnarmorling here are 2 small commits to clear a Maven warning when we launch a build with Maven 3.5.0 (maybe it did exist before but it's more visible with the colors) and integrate the sigtest testing in a profile.

Note that I did not include the sigtest directly in the build as I thought we might not want to execute it  when working on new APIs and having the TCK lagging.

I'll create a CI job though once it's integrated so that we can be warned when there is an issue.

 * https://hibernate.atlassian.net/browse/HV-1459
 * https://hibernate.atlassian.net/browse/HV-1460

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -386,5 +386,57 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>sigtest</id>
+            <build>
+                <plugins>
+                    <plugin>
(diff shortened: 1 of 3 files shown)


gunnarmorling:
Merged; thanks, @gsmet!

PR #833, opened 2017-08-07
HV-1395 Support OptionalInt/OptionalDouble/OptionalLong in the annotation processor

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1395

### DIFF
--- a/annotation-processor/src/main/java/org/hibernate/validator/ap/internal/checks/TypeCheck.java
+++ b/annotation-processor/src/main/java/org/hibernate/validator/ap/internal/checks/TypeCheck.java
@@ -7,15 +7,21 @@
 package org.hibernate.validator.ap.internal.checks;
 
 import java.util.Collections;
+import java.util.Optional;
 import java.util.Set;
 
 import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Name;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Types;
 
+import org.hibernate.validator.ap.internal.util.AnnotationApiHelper;
 import org.hibernate.validator.ap.internal.util.CollectionHelper;
 import org.hibernate.validator.ap.internal.util.ConstraintHelper;
 import org.hibernate.validator.ap.internal.util.ConstraintHelper.AnnotationProcessorValidationTarget;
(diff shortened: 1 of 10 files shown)


gsmet:
@marko-bekhta I force pushed some very minor changes and will add a few comments inline.

annotation-processor/src/main/java/org/hibernate/validator/ap/internal/checks/TypeCheck.java
```diff
@@ -60,25 +69,60 @@ public TypeCheck(ConstraintHelper constraintHelper) {
...
 	}
 
+	private boolean isAnnotationAllowedForType(AnnotationMirror annotation, TypeMirror type) {
+		return ConstraintCheckResult.ALLOWED.equals( constraintHelper.checkConstraint( annotation.getAnnotationType(), type ) );
+	}
+
+	private boolean usesUnwrapping(AnnotationMirror annotationMirror, TypeMirror typeMirror) {
+		//need to check if this annotation is not on one of the types that are automatically unwrapped:
+		if ( constraintHelper.isSupportedForUnwrappingByDefault( getQualifiedName( typeMirror ) ) ) {
+			return true;
```
gsmet: We have to take into account the case where it's unwrapped by default BUT we have the `Skip` payload. In this case, it doesn't use unwrapping.

annotation-processor/src/main/java/org/hibernate/validator/ap/internal/checks/TypeCheck.java
```diff
@@ -60,25 +69,60 @@ public TypeCheck(ConstraintHelper constraintHelper) {
...
+	}
+
+	private TypeMirror getUnwrappedType(TypeMirror type) {
+		Optional<TypeMirror> optional = constraintHelper.getUnwrappedToByDefault( getQualifiedName( type ) );
+		if ( optional.isPresent() ) {
+			return optional.get();
+		}
+		else {
+			//TODO: need to find a way to check for unwrapping
+			return type;
```
gsmet: Thinking out loud, could we allow to declare the types in the annotation processor configuration (thinking about Maven configuration here)?

Something like:
```xml
<containerTypesUnwrappedByDefault>
    <containerTypeUnwrappedByDefault>
        <containerType>...</containerType>
        <unwrappedType>...</unwrappedType>
    </containerTypeUnwrappedByDefault>
</containerTypesUnwrappedByDefault>
```

I mean, one way or another, we need a way to declare them.

Note: it's not blocking for merging this PR, just thinking out loud. We can do it in a second step (if even feasible).

gsmet: By the way, in the case we don't have the unwrapped type, I think we should throw a warning and avoid validating anything after that. Don't know if it's complicated to do that in the current execution flow but I think it's the best we can do.

Note that even if we allow to declare the types, I think this warning thing would still be useful.

marko-bekhta: Yes, I think something like that would be nice to have. That way it'll be possible to register custom extractors. And we can add the ones that we already have somewhere in the code, so they will not be needed in the properties.

gsmet:
Rebased and merged, thanks!

@marko-bekhta could you create an issue about this configuration idea so that we don't forget about it? Thanks!

marko-bekhta:
@gsmet - https://hibernate.atlassian.net/browse/HV-1465 - done :)

PR #835, opened 2017-08-08
HV-1462 Provide patch file for WildFly 11 Beta1

gsmet:
* https://hibernate.atlassian.net/browse/HV-1462


gunnarmorling:
LGTM; the only thing would be to re-word the [reference guide section](http://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#_updating_hibernate_validator_in_wildfly) on the WF modules so it covers 10 and 11.

gsmet:
Yeah, tbh, it was intentional as I didn't want to add information about the development version of WF. I'll propose something.

gsmet:
@gunnarmorling done

gunnarmorling:
Thanks!

PR #837, opened 2017-08-11
Update DomainNameUtil.java

Hilmerc:
The constant DOMAIN = DOMAIN_LABEL +"\\."+ DOMAIN_LABEL;  validates that the e-mail address must be composed for an username and a domain name with a dot sign between two or more words

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
@@ -26,7 +26,7 @@ public final class DomainNameUtil {
 	private static final String DOMAIN_CHARS_WITHOUT_DASH = "[a-z\u0080-\uFFFF0-9!#$%&'*+/=?^_`{|}~]";
 	private static final String DOMAIN_LABEL = "(" + DOMAIN_CHARS_WITHOUT_DASH + "-?)*" + DOMAIN_CHARS_WITHOUT_DASH + "+";
 
-	private static final String DOMAIN = DOMAIN_LABEL + "+(\\." + DOMAIN_LABEL + "+)*";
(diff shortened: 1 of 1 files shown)


gsmet:
Hi @Hilmerc 

Can you explain what you're trying to fix? I don't follow you.

-- 
Guillaume

marko-bekhta:
we have these kind of tests for the `@Email` validator:
```java
isValidEmail( "emmanuel@hibernate" );
isValidEmail( "emma-n_uel@hibernate" );
```
Which I assume will fail after such change. Also it looks that it'll allow only a two parts domains as `*` is removed from expression.

Hilmerc:
Marko you got it, before the change e-mail address like emmanuel@hibernate pass the validation. With my propose the domain must be composed for at least two parts separated by a dot: @hibernate.com @hibernate.net.fr hibernate.fr

marko-bekhta:
@Hilmerc thanks for clarification. I think, if I remember it correctly, some time ago we changed implementation of the rules for email validation when we added more case from [Wiki](https://en.wikipedia.org/wiki/Email_address#Domain) examples. And there's this example in there:

> admin@mailserver1 (local domain name with no TLD, although ICANN highly discourages dotless email addresses)

Also we would not match `localhost` in this case as well:

> user@localserver

So maybe it's better to leave it as it was before? @gsmet @Hilmerc WDYT?

gunnarmorling:
> before the change e-mail address like emmanuel@hibernate pass the validation

We conciously decided to consider this valid, `name@localhost` being the canonical example. If you'd like to exclude this kind of address, you can do so using the `Email#regexp()` option. I'm thus going to close this PR, thanks for your efforts nevertheless, @Hilmerc.

PR #838, opened 2017-08-15
HV-1463 Add test for using @ScriptAssert under OSGi

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1463

I've added a simple test for `ScriptAssert`. The problem described in the issue is caused by how OSGi works with script engines. What I found while reading through similar issues - is that engines are not picked up correctly by OSGi and as a result an exception from `javax.script` is thrown. So to fix

> java.lang.ClassNotFoundException: javax.script.ScriptException ....

I just added one more import for `javax.script.*`. But the problem remains - engines are not picked up. I'm not very familiar with the OSGi and from what I found on OSGi+JSR-233 topic is that people recommend to manually create engines rather than request them like this `new ScriptEngineManager().getEngineByName( languageName )`. Which I don't think what we should do...

So maybe you have an  idea how this should be fixed correctly? If needed current exception is next:
```
javax.script.ScriptException: No JSR 223 script engine found for language "groovy".

	at org.hibernate.validator.internal.util.scriptengine.ScriptEvaluatorFactory.createNewScriptEvaluator(ScriptEvaluatorFactory.java:90)
	at org.hibernate.validator.internal.util.scriptengine.ScriptEvaluatorFactory.getScriptEvaluatorByLanguageName(ScriptEvaluatorFactory.java:70)
	at org.hibernate.validator.internal.constraintvalidators.hv.ScriptAssertContext.getScriptEvaluator(ScriptAssertContext.java:60)
	at org.hibernate.validator.internal.constraintvalidators.hv.ScriptAssertContext.<init>(ScriptAssertContext.java:34)
	at org.hibernate.validator.internal.constraintvalidators.hv.ScriptAssertValidator.initialize(ScriptAssertValidator.java:43)
	at org.hibernate.validator.internal.constraintvalidators.hv.ScriptAssertValidator.initialize(ScriptAssertValidator.java:28)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager.initializeValidator(ConstraintValidatorManager.java:225)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager.createAndInitializeValidator(ConstraintValidatorManager.java:151)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager.getInitializedValidator(ConstraintValidatorManager.java:109)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.getInitializedConstraintValidator(ConstraintTree.java:174)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:111)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:82)
	at org.hibernate.validator.internal.metadata.core.MetaConstraint.doValidateConstraint(MetaConstraint.java:117)
	at org.hibernate.validator.internal.metadata.core.MetaConstraint.validateConstraint(MetaConstraint.java:110)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateMetaConstraint(ValidatorImpl.java:552)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForSingleDefaultGroupElement(ValidatorImpl.java:510)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForDefaultGroup(ValidatorImpl.java:479)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForCurrentGroup(ValidatorImpl.java:444)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateInContext(ValidatorImpl.java:394)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validate(ValidatorImpl.java:183)
	at org.hibernate.validator.osgi.integrationtest.OsgiIntegrationTest.canUseScripAssertConstraint(OsgiIntegrationTest.java:269)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.ops4j.pax.exam.invoker.junit.internal.ContainerTestRunner.runLeafWithRetry(ContainerTestRunner.java:97)
	at org.ops4j.pax.exam.invoker.junit.internal.ContainerTestRunner.runChildWithRetry(ContainerTestRunner.java:84)
	at org.ops4j.pax.exam.invoker.junit.internal.ContainerTestRunner.runChild(ContainerTestRunner.java:75)
	at org.ops4j.pax.exam.invoker.junit.internal.ContainerTestRunner.runChild(ContainerTestRunner.java:44)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
	at org.ops4j.pax.exam.invoker.junit.internal.JUnitProbeInvoker.invokeViaJUnit(JUnitProbeInvoker.java:124)
	at org.ops4j.pax.exam.invoker.junit.internal.JUnitProbeInvoker.findAndInvoke(JUnitProbeInvoker.java:97)
	at org.ops4j.pax.exam.invoker.junit.internal.JUnitProbeInvoker.call(JUnitProbeInvoker.java:73)
	at org.ops4j.pax.exam.rbc.internal.RemoteBundleContextImpl.remoteCall(RemoteBundleContextImpl.java:83)
	at sun.rmi.server.UnicastServerRef.dispatch(UnicastServerRef.java:324)
	at sun.rmi.transport.Transport$1.run(Transport.java:200)
	at sun.rmi.transport.Transport$1.run(Transport.java:197)
	at sun.rmi.transport.Transport.serviceCall(Transport.java:196)
	at sun.rmi.transport.tcp.TCPTransport.handleMessages(TCPTransport.java:568)
	at sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run0(TCPTransport.java:826)
	at sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.lambda$run$0(TCPTransport.java:683)
	at sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run(TCPTransport.java:682)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```


gsmet:
Getting Groovy to work is a nice second step but can you try to get it to work with Nashorn first? It's included in the JDK so it should be easier. AFAICS, they fixed issues with it in Karaf 4 so it's supposed to work.

As for getting Groovy to work, the ScriptManager is using the TCCL by default so we probably need the same sort of magic we do for the EL factory. See https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/messageinterpolation/ResourceBundleMessageInterpolator.java for how we initialize the EL trying different CL until we find one working.

We should also take into account the `externalClassLoader`. See what we do here: https://github.com/hibernate/hibernate-validator/blob/ff2ff96b1f30f198c37109f3b7825bc9ff8c4779/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java#L546 . The issue is that you will probably need to pass it to the `HibernateConstraintValidatorContext` through the `ValidationContext`. That shouldn't be too hard but that would put some pressure on the `ScriptEvalutor`map as we wouldn't be able to interrogate it only once at the validator initialization.

@gunnarmorling what's your thoughts on this?

marko-bekhta:
@gsmet thanks for the info! I did find one more thing, which I'm half-way finished implementing. The idea is to try and get ScriptEngine the way it is right now, and if nothing is found then try to search for it through OSGi bundles. So I think it's similar to what you were saying about the EL. The only difference would be that we will not be able to use something like `GroovyScriptEngine.class.getClassLoader` as we don't know which scripting engine is requested by the user ...

I think I'll have something working later today.

marko-bekhta:
I used a couple of classes from https://svn.apache.org/repos/asf/felix/trunk/mishell/src/main/java/org/apache/felix/mishell/  as they seem to do what we need, and I thought "why reinvent the wheel"? I added the classes and in the next commit I slightly modified `findFactoryCandidates` method as it was not handling files correctly (adding comments and empty lines as candidates) + reformatted to match HV formatting. 

Right now it's not very clean but if we agree that using these is OK - I'll clean it all up (move the dependencies to dependency management section, remove those retries of `getEngineByName` in `ScriptEvaluatorFactory` and hide it all behind selection of correct `ScriptEngineManager`; also those `getEngineBy***` methods should probably be using something like `run( SetContextClassLoader.action( ResourceBundleMessageInterpolator.class.getClassLoader() ) );` as well ?)

gsmet:
@marko-bekhta I'm not very convinced about this approach even if it solves the issue (and kudos for finding a way to do it).

Even if we clean it up, it's still quite a hack so I wouldn't like to have it in plain HV.

Another approach might be to provide a way to inject a `ScriptEvaluatorFactory` (not exactly in this form, we would need to make an API out of it) from the user space to the `HibernateConstraintValidatorContext`.

That way, people would have the ability to extend it the way they want (and you could clean up your impl and provide a jar for it).

WDYT? Does it make sense to you?

gunnarmorling:
On a related matter, I've just filed [HV-1469](https://hibernate.atlassian.net/browse/HV-1469) for managing the lifecycle of SEF in the scope of the validator factory instead of the static singleton it is currently.

gunnarmorling:
Btw. `ScriptEngineManager` can be instantiated with a dedicated classloader, so we could try to pass the `externalClassLoader` for that. I'm doubting it helps much for the OSGi issue, but it likely is a good idea nevertheless, as we should consult that CL for all resources we try to look up.

marko-bekhta:
hmm ... You have good points. 
It got me thinking ... I kind of like the idea of passing the `ScriptEngineFactory` as in this case we don't need lookups at all, but the users should preconfigure HV anytime they want to use `@ScriptAssert`.
If we pass the `ScriptEvaluatorFactory` then we can use the current implementation by default (adding an `externalClassLoader` to it as Gunnar mentioned), and if somebody needs an OSGi version we can give them a "cleaned up version" from this PR... 

In case of passing engine factories I think it'll be enough to have an interface like:
```java
public interface HVScriptEngineFactory {
    public String getLanguageName();
    public  ScriptEngine getScriptEngine(); // or maybe return ScriptEvaluator right away rather then ScriptEngine
}
```
Which is a subset of methods from `ScriptEngineFactory`. As we just need an `ScriptEngine` itself, and a language name - so we can identify and cache them by name. 

If the idea with `ScriptEngineFactory`s is not a good solution :) I'll proceed with `ScriptEvaluatorFactory` then. And in this case I think the only method that we need to expose is `ScriptEvaluatorFactory#createNewScriptEvaluator`.

gsmet:
@marko-bekhta not sure I follow you. The point would be to pass an API equivalent of `ScriptEvaluatorFactory` from the `ValidationFactory` and you would be able to override it at bootstrap. So you wouldn't have to preconfigure anything as it would use the default one if not overridden.

Maybe I missed something?

marko-bekhta:
@gsmet right that's one option and in case of overriding it looks that the only method needed is similar to `ScriptEvaluatorFactory#createNewScriptEvaluator`.

And as another option we make `ScriptEvaluatorFactory` just a "storage" of `HVScriptEngineFactory` something like:
```java
class ScriptEvaluatorFactory {
   Map<String, HVScriptEngineFactory> mapOfEngineFactories;
   //...

  public ScriptEvaluator getScriptEvaluatrorForLanguage(String name){
      return mapOfEngineFactories.get(name). getScriptEngine();
  }
}
```
And these `HVScriptEngineFactory` will be passed by user at bootstrap:
```java
Validator validator = Validation.byProvider( HibernateValidator.class )
        .configure()
        .scripEngineFactories(
            Arrays.asList( new HVGroovyScriptEngineFactory(), new HVRubyScriptEngineFactory() )
        ).buildValidatorFactory()
        .getValidator();
```

gsmet:
It kinda defeats the whole point of the ScriptEngineManager dynamically finding the script implementations.

The advantage of my proposition is that people not running OSGi wouldn't need any configuration as the default implementation would rely on the JDK `ScriptEngineManager` (probably passing the externalClassLoader as the class loader if defined).

And if you want something like your proposal, you could still have something like:
```java
Validator validator = Validation.byProvider( HibernateValidator.class )
        .configure()
        .scripEngineFactory(
            new DeclarativeScriptEngineFactory( new HVGroovyScriptEngineFactory(), new HVRubyScriptEngineFactory() )
        ).buildValidatorFactory()
        .getValidator();
```
(either provided by HV or not depending on how clean we can make it)

marko-bekhta:
I've changed the approach as it was suggested. It still needs some things here and there (missing docs, OSGi factory which takes bundle (similar to what was in the original version), finishing implementation in xml (need to confirm if it's ok to "extend" the schema and add `ScriptEvaluatorFactory` to it, I assume it'll be HV specific schema in this case, right?).

Also I'm not very happy about the fact that it seems that we need to add package imports for each scripting language we want to be used (that list should be maintained and what about some custom user implementations, which we know nothing about ?)

But in general - that's what we'd like to have, right ?

engine/src/main/xsd/hibernate-validation-configuration-2.0.xsd
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gsmet: This is a bold move :).

We'll have to wait for Gunnar for that (he's on vacation). I'm not sure we want to extend the schema for Hibernate specific features.

Wondering if it wouldn't be better to use a `<property>`, especially considering it's such a niche feature and it fits well.

pom.xml
```diff
@@ -167,7 +167,7 @@
 
         <puppycrawl.checkstyle.version>7.6</puppycrawl.checkstyle.version>
 
-        <groovy.version>2.4.8</groovy.version>
+        <groovy.version>2.4.12</groovy.version>
```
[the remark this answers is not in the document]
marko-bekhta: OK we can do that. I just saw that there's a newer version and updated it as I was using that dependency. I'll create a ticket for it.

pom.xml
```diff
@@ -292,6 +292,11 @@
                 <version>${moneta.version}</version>
             </dependency>
             <dependency>
+                <groupId>org.osgi</groupId>
+                <artifactId>org.osgi.core</artifactId>
+                <version>6.0.0</version>
```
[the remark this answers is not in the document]
marko-bekhta: OK :)  I just wasn't 100% sure if we'd keep it or not because of that OSGiFactory thing. So I left it "in a bad sate" so we'd see it :)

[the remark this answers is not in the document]
gsmet: Yes, tbh, I'm not sure either. But let's make it clean, it's not as if it was a lot of work.

engine/src/main/xsd/hibernate-validation-configuration-2.0.xsd
marko-bekhta: :) yeah ... I wasn't sure about this as well - if it's a right way to do it. And also about that comment on `validation-custom-config.xml` it's commented out as the test failed and I didn't have time to figure out what happened...

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -150,6 +156,7 @@ public ValidatorFactoryImpl(ConfigurationState configurationState) {
 		this.traversableResolver = configurationState.getTraversableResolver();
 		this.parameterNameProvider = new ExecutableParameterNameProvider( configurationState.getParameterNameProvider() );
 		this.clockProvider = configurationState.getClockProvider();
+		this.scriptEvaluatorFactory = ( (ConfigurationImpl) configurationState ).getScriptEvaluatorFactory();
```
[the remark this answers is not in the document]
marko-bekhta: @gsmet I didn't extract the getter to the in interface. Extract it and then change the casting ?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ScriptAssertContext.java
```diff
@@ -27,25 +29,26 @@
...
+	public boolean evaluateScriptAssertExpression(Object object, String alias, ConstraintValidatorContext context) {
 		Map<String, Object> bindings = newHashMap();
 		bindings.put( alias, object );
 
-		return evaluateScriptAssertExpression( bindings );
+		return evaluateScriptAssertExpression( bindings, context );
 	}
 
-	public boolean evaluateScriptAssertExpression(Map<String, Object> bindings) {
+	public boolean evaluateScriptAssertExpression(Map<String, Object> bindings, ConstraintValidatorContext context) {
```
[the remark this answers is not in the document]
marko-bekhta: Right ... But it's either here or via the constructor right? And in case of constructor we'd need to get it in the `initailize` method:

```java
public void initialize(ParameterScriptAssert constraintAnnotation) {
   // get the Context somehow
}
```

Or did you have something different on your mind?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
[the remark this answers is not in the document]
gsmet: +1

engine/src/main/xsd/hibernate-validation-configuration-2.0.xsd
marko-bekhta: @gsmet I've removed the schema, so hope that Gunnar won't see that :). I changed the code to use the property and if the name of the property is OK, I think I could add some paragraph to documentation on how to set the `ScriptEvaluatorFactory` why it's needed and the ways it can be done.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ScriptAssertContext.java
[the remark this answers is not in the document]
gsmet: No it was more about building what's needed in `ParameterScriptAssertValidator` and pass the `ScriptEvaluator` directly.

About the `initialize()`, that's what I posted in another message, it's too bad we don't have a `Context` object here, it would have been useful.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ScriptAssertContext.java
```diff
@@ -54,14 +58,16 @@ public boolean evaluateScriptAssertExpression(Map<String, Object> bindings) {
...
 	}
 
-	private ScriptEvaluator getScriptEvaluator(String languageName) {
-		try {
-			ScriptEvaluatorFactory evaluatorFactory = ScriptEvaluatorFactory.getInstance();
-			return evaluatorFactory.getScriptEvaluatorByLanguageName( languageName );
-		}
-		catch (ScriptException e) {
-			throw log.getCreationOfScriptExecutorFailedException( languageName, e );
+	private synchronized ScriptEvaluator getScriptEvaluator( ConstraintValidatorContext context) {
```
gsmet: So if we move that to the Validator, please don't use a plain `synchronized` method but take a look at http://www.javamex.com/tutorials/double_checked_locking_fixing.shtml , especially the last paragraph: Use DCL plus volatile (except the synchronized should be on (this) not on the class).

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -62,6 +63,16 @@
...
 
 	/**
+	 * Property for configuring script evaluator factory, allowing to set up which factory will be used to create
+	 * {@link org.hibernate.validator.cfg.scriptengine.ScriptEvaluator} for evaluation of script expressions in
+	 * {@link org.hibernate.validator.constraints.ScriptAssert} and {@link org.hibernate.validator.constraints.ParameterScriptAssert}
+	 * constraints. A fully qualified name of a class implementing  {@link ScriptEvaluatorFactory} is expected as a value.
+	 *
+	 * @since 6.1
+	 */
+	String SCRIPT_EVALUATOR_CLASSNAME = "hibernate.validator.script_evaluator_factory_classname";
```
gsmet: +1 but you can remove the _classname suffix, it's useless.

marko-bekhta:
@gsmet Thanks! I'm writing the doc section about how this Script Factory can be configured and I think I'll finish it all till the end of today (doc + some clean ups that left).

gsmet:
OK, nice! No pressure!

marko-bekhta:
@gsmet I've forgot to push the changes :) 

I've added documentation section and also I've changed how the `ScriptAssertContext` is created. I've used the double checking as you've suggested and also removed the `HibernateConstraintValidatorContext` parameter from the evaluation method and created an abstract validator for both `Script` constraints and placed the code for evaluator creation there.  I've also added that `externalClassLoader` to the default factory.

engine/src/main/java/org/hibernate/validator/internal/util/scriptengine/DefaultLookupScriptEvaluatorFactory.java
```diff
@@ -29,11 +26,6 @@
  */
 public class DefaultLookupScriptEvaluatorFactory extends AbstractCacheableScriptEvaluatorFactory {
 
-	/**
```
gsmet: @marko-bekhta ah have a misunderstanding here. I was asking to remove the `SoftReference` but keep the singleton instance, just not with a `SoftReference` :).

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,100 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
 Refer to <<section-fail-fast>> and <<section-programmatic-api>> to learn more about the fail fast
 mode and the constraint declaration API.
 
+[[section-script-evaluator-factory]]
+==== Configuring script evaluator factory (`ScriptEvaluatorFactory`)
+
+For constrains like `@ScriptAssert` and `@ParameterScriptAssert` it might be useful to configure
```
gsmet: constrain**t**s
add a comma before "it might", it's more readable.

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,100 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+how scripts are evaluated, and which scripting engines are used. This can be done by setting
```
gsmet: It's more about configuring how the script engines are built?

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,100 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+a custom implementation of `ScriptEvaluatorFactory`. In particular this is important for
+OSGi environments (or similar to it) where user might face issues with dynamic class loading
```
gsmet: s/dynamic/modular/ ?

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,100 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+and https://www.jcp.org/en/jsr/detail?id=223[JSR-223]. It also would allow to use any custom
+script evaluation, not necessarily based on the https://www.jcp.org/en/jsr/detail?id=223[JSR-223].
+
+===== Programmatic configuration
+There are two ways to change the default `ScriptEvaluatorFactory` programmatically.
+First one is to use `hibernate.validator.script_evaluator_factory` property as described
```
gsmet: I don't think I would advertise the property configuration in the programmatic API. I would only use it for the XML configuration.

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,100 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/BootstrappingTest.java[tags=scriptEvaluatorFactoryViaAddProperty]
+----
+====
+
+Another option is to pass an instance of `ScriptEvaluatorFactory` when building `ValidatorFactory`.
+This gives more flexibility in configuration of the `ScriptEvaluatorFactory`. <<example-script-evaluator-factory-programmatically>>
+shows how this can be done.
```
gsmet: Should be reworked because of the remark above. But just as I think this part will be kept: "when building **the** `ValidatorFactory`"

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,100 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/BootstrappingTest.java[tags=scriptEvaluatorFactoryProgrammatically]
+----
+====
+
+===== XML configuration
+To specify `ScriptEvaluatorFactory` via XML the same `hibernate.validator.script_evaluator_factory`
+property can be used.
```
gsmet: This should be changed accordingly.

+ To specify **the** `ScriptEvaluatorFactory`

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,100 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+.Method constraints configured via XML
+====
+[source, XML, indent=0]
+----
+include::{resourcesdir}/org/hibernate/validator/referenceguide/chapter09/script-evaluator-factory-validation.xml[]
+----
+====
+
+In case of using property to configure a factory (either in XML configuration, or by adding it programmatically)
+user is limited to use factories that has constructors without parameters.
```
gsmet: I would simply have this:
In this case, the specified `ScriptEvaluatorFactory` must have a no-arg constructor.

(this is the wording we use in the spec so let's be consistent)

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,100 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+
+===== Predefined `ScriptEvaluatorFactory` implementations
+Hibernate Validator provides multiple implementations of `ScriptEvaluatorFactory` that user
```
gsmet: OK, I have a couple more comments for the rest of the wording but it's probably easier if I fix it on merging.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/AbsctractScriptAssertValidator.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.hibernate.validator.cfg.scriptengine.ScriptEvaluator;
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorContext;
+import org.hibernate.validator.internal.util.logging.Log;
+import org.hibernate.validator.internal.util.logging.LoggerFactory;
+
+/**
+ * @author Marko Bekhta
+ */
+public abstract class AbsctractScriptAssertValidator<A extends Annotation, T> implements ConstraintValidator<A, T> {
```
gsmet: **Abstract**

engine/src/main/java/org/hibernate/validator/internal/util/scriptengine/DefaultLookupScriptEvaluatorFactory.java
marko-bekhta: right ... It looks like that that's how I addressed this ticket - https://hibernate.atlassian.net/browse/HV-1469 
And then when I've added that part with the external classloader we cannot cache the instance unless it's just one external loader (and it's passed to `getInstance()`), right?

marko-bekhta:
@gsmet thanks for reviewing. I addressed your suggestions and switched places of XML and programmatic configuration so the part about programmatic one having more flexibility doesn't look out of place.

engine/src/main/java/org/hibernate/validator/HibernateValidatorBootstrapConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+public interface HibernateValidatorBootstrapConfiguration extends BootstrapConfiguration {
+	/**
+	 * Class name of the {@link org.hibernate.validator.cfg.scriptengine.ScriptEvaluatorFactory} implementation or
+	 * {@code null} if none is specified.
+	 *
+	 * @return script evaluator factory class name or {@code null}
+	 *
+	 * @since 6.1
+	 */
+	default String getScriptEvaluatorFactoryClassName() {
```
gunnarmorling: I don't think we should add this method. One can get the configured value via `BootstrapConfiguration#getProperties()`. And `BootstrapConfiguration` is not meant to expose the implementation-specific class names, instead there should be `HibernateValidatorConfiguration#getDefaultScriptEvaluatorFactory()` which returns an _instance_.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -62,6 +63,16 @@
 	String CONSTRAINT_MAPPING_CONTRIBUTORS = "hibernate.validator.constraint_mapping_contributors";
 
 	/**
+	 * Property for configuring script evaluator factory, allowing to set up which factory will be used to create
+	 * {@link org.hibernate.validator.cfg.scriptengine.ScriptEvaluator} for evaluation of script expressions in
```
gunnarmorling: Let's use unqualified names (via imports) in JavaDocs for the sake of readability of rendered docs.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -210,4 +221,26 @@
...
+	 * used to evaluate script expressions for {@link org.hibernate.validator.constraints.ScriptAssert} and
+	 * {@link org.hibernate.validator.constraints.ParameterScriptAssert} constraints.
+	 *
+	 * @param scripEngineFactory The {@link ScriptEvaluatorFactory} to be used by validator
+	 *
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 6.1
+	 */
+	HibernateValidatorConfiguration scriptEngineFactory(ScriptEvaluatorFactory scripEngineFactory);
```
gunnarmorling: s/scriptEngineFactory/scriptEvaluatorFactory/

engine/src/main/java/org/hibernate/validator/HibernateValidatorFactory.java
```diff
@@ -18,6 +20,16 @@
 public interface HibernateValidatorFactory extends ValidatorFactory {
 
 	/**
+	 * Returns a factory which is used to create {@link org.hibernate.validator.cfg.scriptengine.ScriptEvaluator}s for
```
gunnarmorling: We can use non-qualified names.

gunnarmorling: Why is btw. that this method is exposed here. What's is use case?

engine/src/main/java/org/hibernate/validator/cfg/scriptengine/ScriptEvaluator.java
```diff
@@ -0,0 +1,35 @@
...
+	/**
+	 * Evaluates a {@code script} expression and returns a result of this evaluation.
+	 *
+	 * @param script A script to evaluate
+	 * @param bindings The bindings to be used
+	 * @return The result of script evaluation
+	 *
+	 * @throws ScriptException in case of any errors during script evaluation
+	 */
+	Object evaluate(String script, Map<String, Object> bindings) throws ScriptException;
```
gunnarmorling: I didn't have an abstraction as such in mind, but I think it's a good idea. But if we want to go for it, the exception in the signature shouldn't be from `javax.script` ;)

engine/src/main/java/org/hibernate/validator/cfg/scriptengine/impl/AbstractCacheableScriptEvaluatorFactory.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Basic cacheable factory responsible for the creation of {@link ScriptEvaluator}s. This
+ * class is thread-safe. Caches {@code ScriptEvaluator} when they are requested.
+ *
+ * @author Gunnar Morling
+ * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
+ * @author Marko Bekhta
+ */
+public abstract class AbstractCacheableScriptEvaluatorFactory implements ScriptEvaluatorFactory {
```
gunnarmorling: I don't think we should provide any user-facing implementations (assuming that this is what you intended to do with that `impl` package). We can add a sketch for OSGi to the docs, but I don't feel it's something I'd like to maintain as part of the library proper.

engine/src/main/java/org/hibernate/validator/cfg/scriptengine/impl/AbstractCacheableScriptEvaluatorFactory.java
```diff
@@ -0,0 +1,69 @@
...
+	 * Retrieves a script executor for the given language.
+	 *
+	 * @param languageName The name of a scripting language as expected by {@link ScriptEngineManager#getEngineByName(String)}.
+	 * @return A script executor for the given language. Never null.
+	 *
+	 * @throws ValidationException In case no compatible evaluator for the given language could be found.
+	 */
+	@Override
+	public ScriptEvaluator getScriptEvaluatorByLanguageName(String languageName) {
+		if ( !scriptExecutorCache.containsKey( languageName ) ) {
```
gunnarmorling: We could use calculateIfAbsent() now that we are on Java 8.

engine/src/main/java/org/hibernate/validator/cfg/scriptengine/impl/OSGiScriptEvaluatorFactory.java
```diff
@@ -0,0 +1,111 @@
...
+				.findFirst()
+				.orElseThrow( () -> new ValidationException( String.format( "Wasn't able to find script evaluator for '%s'.", languageName ) ) );
+	}
+
+
+	private List<ScriptEngineManager> findManagers(BundleContext context) {
+		return findFactoryCandidates( context ).stream()
+				.map( className -> {
+					try {
+						return new ScriptEngineManager( Class.forName( className ).getClassLoader() );
```
gunnarmorling: I don't think using `Class#forName()` is a good idea in OSGi. I'd expect people will rather plug-in just a very simple implementation of `ScriptEvaluatorFactory` which just returns an implementation they statically passed via the constructor.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
```diff
@@ -6,8 +6,11 @@
  */
 package org.hibernate.validator.constraintvalidation;
 
+import javax.script.ScriptException;
 import javax.validation.ConstraintValidatorContext;
 
+import org.hibernate.validator.cfg.scriptengine.ScriptEvaluator;
```
gunnarmorling: That package seems a bit odd, how about `org.hibernate.validator.script(ing)`?

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
```diff
@@ -106,4 +109,19 @@
...
+	 *
+	 * @param languageName The name of a scripting language
+	 * @return A script executor for the given language. Never null.
+	 *
+	 * @throws ScriptException in case when no {@link ScriptEvaluator} was
+	 * found for a given {@code languageName}.
+	 *
+	 * @since 6.1
+	 */
+	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName) throws ScriptException;
```
gunnarmorling: Nice, that's much better than before.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/ScriptAssertValidatorTest.java
```diff
@@ -93,34 +105,52 @@ public void emptyAliasRaisesException() throws Exception {
 		getInitializedValidator( "lang", "script", "" );
 	}
 
-	@Test(expectedExceptions = ConstraintDeclarationException.class)
```
gunnarmorling: What's the reason to change that? I find that much more expressive personally.

engine/src/main/java/org/hibernate/validator/HibernateValidatorFactory.java
marko-bekhta: It was done by analogy with the `ClockProvider` (for which there's `ValidatorFactory#getClockProvider()` ), so that I don't miss any places where this new ScriptEvaluator is needed. But looking at this method - it doesn't seems to be necessary. I'll remove it.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/ScriptAssertValidatorTest.java
marko-bekhta: I do like the `expectedExceptions` more compared to this change. But as
now that we only try to create a `ScriptEvaluator` when `isValid` is called. We are not able to check for things like "is there an evaluator for specified language?" during initialization. As a result the exception (`ConstraintDeclarationException`) is only thrown from isValid method and it is wrapped with `ValidationException` ...

engine/src/main/java/org/hibernate/validator/cfg/scriptengine/impl/OSGiScriptEvaluatorFactory.java
marko-bekhta: this factory is based on [1] and [2], and how script engines are handled there. But if as you said earlier - we will not support factory implementations for users to choose from - we probably wouldn't need this `OSGiScriptEvaluatorFactory` anyway :)

[1]https://svn.apache.org/repos/asf/felix/trunk/mishell/src/main/java/org/apache/felix/mishell/OSGiScriptEngineManager.java
[2]https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/osgi/impl/OSGiScriptEngineManager.java

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/ScriptAssertValidatorTest.java
gunnarmorling: Ok.

marko-bekhta:
So I finally got back :)
I think I've addressed all the comments. 
I've also added a test for SpEL (see `CustomScriptEvaluatorFactoryForSpringELTest`) as @gunnarmorling suggested . It seems to be working with the current interfaces for factory and evaluator, but personally I'd go for something like 
```java
public interface ScriptEvaluator {

	/**
	 * Evaluates a {@code script} expression and returns a result of this evaluation.
	 *
	 * @param script A script to evaluate
   * @param alias The alias for the object used in script
   * @param object The object against which to evaluate the script  
	 *
	 * @return The result of script evaluation
	 *
	 * @throws ScriptEvaluationException in case of any errors during script evaluation
	 */
	Object evaluate(String script, String alias, Object object) throws ScriptEvaluationException;

}
```
WDYT? Should we go for such change or keep it as is?

I've also left the documentation part unchanged for now. I wanted to confirm first  if we should remove those factory implementations completely or just move them from engine to examples and show them in the documentation ?

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,86 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
 Refer to <<section-fail-fast>> and <<section-programmatic-api>> to learn more about the fail fast
 mode and the constraint declaration API.
 
+[[section-script-evaluator-factory]]
+==== Configuring script evaluator factory (`ScriptEvaluatorFactory`)
+
+For constraints like `@ScriptAssert` and `@ParameterScriptAssert`, it might be useful to configure
+how script evaluators (`ScriptEvaluator`) are build, and which scripting engines are used to build
```
gsmet: s/build/built/

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,86 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+them. This can be done by setting a custom implementation of `ScriptEvaluatorFactory`. In
+particular this is important for OSGi environments (or similar to it) where user might face issues
```
gsmet: "this is important for modular environments (e.g. OSGi)" instead of "this is important for OSGi environments (or similar to it)"

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,86 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+[source, XML, indent=0]
+----
+include::{resourcesdir}/org/hibernate/validator/referenceguide/chapter09/script-evaluator-factory-validation.xml[]
+----
+====
+
+In this case, the specified `ScriptEvaluatorFactory` must have a no-arg constructor.
+
+===== Programmatic configuration
+To do the change programmatically an instance of `ScriptEvaluatorFactory` should be passed when building the `ValidatorFactory`.
```
gsmet: "To do the change programmatically" -> "To configure it programmatically, "

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -370,6 +370,86 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/Bootstrapp
...
+.Changing `ScriptEvaluatorFactory` by configuring `ValidatorFactory`
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/BootstrappingTest.java[tags=scriptEvaluatorFactoryProgrammatically]
+----
+====
+
+===== Predefined `ScriptEvaluatorFactory` implementations
+Hibernate Validator provides multiple implementations of `ScriptEvaluatorFactory` that user
```
gsmet: OK, I'll let you change that, considering we won't include both by default.

documentation/src/test/resources/org/hibernate/validator/referenceguide/chapter09/script-evaluator-factory-validation.xml
```diff
@@ -0,0 +1,10 @@
+<validation-config
+        xmlns="http://xmlns.jcp.org/xml/ns/validation/configuration"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/validation/configuration
+            http://xmlns.jcp.org/xml/ns/validation/configuration/validation-configuration-2.0.xsd"
+        version="2.0">
+
+    <property name="hibernate.validator.script_evaluator_factory">org.hibernate.validator.referenceguide.chapter09.CustomScriptEvaluatorFactory</property>
+
```
gsmet: Add a new line at the end of file.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
```diff
@@ -106,4 +110,19 @@
...
```
gsmet: Mmmmh, should it be a `ScriptException`? I thought we were trying to get away from `javax.script` specific exceptions?

engine/src/main/java/org/hibernate/validator/scripting/impl/OSGiScriptEvaluatorFactory.java
```diff
@@ -0,0 +1,110 @@
...
+	private List<String> toListOfFactoryCandidates(URL url) {
+		try ( BufferedReader reader = new BufferedReader( new InputStreamReader( url.openStream(), "UTF-8" ) ) ) {
+			return reader.lines()
+					.map( String::trim )
+					.filter( line -> !line.isEmpty() )
+					.filter( line -> !line.startsWith( "#" ) )
+					.collect( Collectors.toList() );
+		}
+		catch (IOException e) {
+			throw new ValidationException( "Wasn't able to read a ScripFactory candidate resource file", e );
```
gsmet: s/ScripFactory/ScriptEvaluatorFactory/

engine/src/main/java/org/hibernate/validator/scripting/impl/AbstractCacheableScriptEvaluatorFactory.java
```diff
@@ -0,0 +1,66 @@
...
+	public ScriptEvaluator getScriptEvaluatorByLanguageName(String languageName) {
+		return scriptExecutorCache.computeIfAbsent( languageName, this::createScriptEvaluator );
+	}
+
+	private ScriptEvaluator createScriptEvaluator(String languageName) {
+		try {
+			return createNewScriptEvaluator( languageName );
+		}
+		catch (ScriptEvaluationException e) {
+			throw new ConstraintDeclarationException( "Wasn't able to find a script evaluator for '" + languageName + "' language.", e );
```
gsmet: It would probably be better to move the exception creation in the logger for the "core" script evaluator factory.

engine/src/main/java/org/hibernate/validator/scripting/ScriptEvaluatorFactory.java
```diff
@@ -0,0 +1,35 @@
...
+ * used to evaluate script expressions for {@link ScriptAssert}
+ * and {@link ParameterScriptAssert} constraints. An instance of
+ * such factory should be passed during bootstrapping a validator.
+ *
+ * @author Marko Bekhta
+ */
+public interface ScriptEvaluatorFactory {
+
+	/**
+	 * Retrieves a script executor {@link ScriptEvaluator} for the given language.
```
gsmet: s/executor/evaluator/

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationBootstrapParameters.java
```diff
@@ -243,6 +258,22 @@ private void setClockProvider(String clockProviderFqcn, ClassLoader externalClas
...
+		if ( scriptEvaluatorFactoryFqcn != null ) {
+			try {
+				@SuppressWarnings("unchecked")
+				Class<? extends ScriptEvaluatorFactory> clazz = (Class<? extends ScriptEvaluatorFactory>) run(
+						LoadClass.action( scriptEvaluatorFactoryFqcn, externalClassLoader )
+				);
+				scriptEvaluatorFactory = run( NewInstance.action( clazz, "script evaluator factory class" ) );
+				log.usingScriptEvaluatorFactory( clazz );
+			}
+			catch (ValidationException e) {
```
gsmet: I think we'd better catch any `Exception` here.

engine/src/main/java/org/hibernate/validator/internal/util/scriptengine/ScriptEvaluatorImpl.java
```diff
@@ -52,8 +57,13 @@ public Object evaluate(String script, Map<String, Object> bindings) throws Scrip
 		}
 	}
 
-	private Object doEvaluate(String script, Map<String, Object> bindings) throws ScriptException {
-		return engine.eval( script, new SimpleBindings( bindings ) );
+	private Object doEvaluate(String script, Map<String, Object> bindings) throws ScriptEvaluationException {
+		try {
+			return engine.eval( script, new SimpleBindings( bindings ) );
+		}
+		catch (ScriptException e) {
```
gsmet: Let's catch any exception here to be on the safe side.

marko-bekhta:
@gsmet no worries :) I kind of like these tasks "hmm it's just a 10 minutes probably to add a test" which turn out into quite some changes. I'll update the doc and let you know when it's ready for the next review :).

marko-bekhta:
Hi @gsmet 
I've done some more cleanup around exceptions and updated the documentation. As it was suggested I moved out factory impls out of engine. But what I don't like about it is that there are now two copies of OSGi and MultiClassloader factories as one pair is used in documentation and another - in osgi module to run actual tests in OSGi env....

gsmet:
@marko-bekhta I rebased and pushed a few other commits. I think we're in pretty good shape now. Interested in your feedback on my commits!

@gunnarmorling there are 2 things I wanted to check with you:
* should we commit this in the 6.0 branch or do we keep it for 6.1. I know you were a bit reluctant about creating a 6.1. The current patch does not break any API and fixes an issue with the ScriptEvaluatorFactory currently being a singleton
* I made a few implementations visible in the API (namely `AbstractCachingScriptEvaluatorFactory` and `ScriptEngineScriptEvaluator`): we want them to be accessible so that users can base their own implementations on it. The 2 implementations Marko created for OSGi now use both classes (Marko initially didn't use them but I don't think it's such a good idea).

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -445,21 +452,22 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/OSGiScript
...
+It is designed specifically for OSGi environments and allows you to pass the `BundleContext` as a
 parameter which will be used to search for `ScriptEngineFactory`.
 
-As it was already said, it is also possible to use other than https://www.jcp.org/en/jsr/detail?id=223[JSR-223]
-based script evaluation (e.g. https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#expressions[Spring Expression Language]).
-`SpringELEvaluatorFactory` can be written like:
+As already mentioned, you can also use script engines that are not based on https://www.jcp.org/en/jsr/detail?id=223[JSR 223].
+
+For instance, to use the https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#expressions[Spring Expression Language]),
+you can define a `SpringELEvaluatorFactory` as:
```
marko-bekhta: Probably `SpringELEvaluatorFactory` -> `SpringELScriptEvaluatorFactory` as the name changed

engine/pom.xml
```diff
@@ -229,9 +224,7 @@
                             org.joda.time.*;version="[2.0.0,3.0.0)";resolution:=optional,
                             org.jsoup.*;version="[1.5.2,2.0.0)";resolution:=optional,
                             javax.money;version="[1.0.0,2.0.0)";resolution:=optional,
-                            com.thoughtworks.paranamer.*;version="[2.5.5,3.0.0)";resolution:=optional,
-                            org.codehaus.groovy.jsr223;version="[2.4.0,3.0.0)";resolution:=optional,
```
marko-bekhta: hmm so now that those ScriptEvaluatorFactories are not inside of HV engine there's no need to import scripting packages, that's really nice!

documentation/src/main/asciidoc/ch09.asciidoc
gsmet: Good catch, fixed.

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter09/CustomScriptEvaluatorFactoryForSpringELTest.java
```diff
@@ -0,0 +1,42 @@
...
+				.configure()
+				.scriptEvaluatorFactory( new SpringELScriptEvaluatorFactory() )
+				.buildValidatorFactory()
+				.getValidator();
+		assertTrue( validator.validate( new Foo( 1 ) ).isEmpty() );
+		assertEquals( 1, validator.validate( new Foo( -1 ) ).size() );
+	}
+
+	//tag::include[]
+	@ScriptAssert(script = "value > 0", lang = "spring")
```
gunnarmorling: Very nice to see this example for using SpEL! It doesn't seem to work, though, to use the "_this" alias (see `@ScriptAssert#alias()`). Unlike JSR 223, SpEL resolves the properties against a root object, I'm wondering whether we could resolve references against `_this` (or whatever is chosen as name), too, against that root object. It's not mandatory for sure, but it'd be nice to show it for the sake of completeness.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -62,6 +66,16 @@
 	String CONSTRAINT_MAPPING_CONTRIBUTORS = "hibernate.validator.constraint_mapping_contributors";
 
 	/**
+	 * Property for configuring script evaluator factory, allowing to set up which factory will be used to create
+	 * {@link ScriptEvaluator} for evaluation of script expressions in
+	 * {@link ScriptAssert} and {@link ParameterScriptAssert}
+	 * constraints. A fully qualified name of a class implementing  {@link ScriptEvaluatorFactory} is expected as a value.
+	 *
+	 * @since 6.1
```
gunnarmorling: On the question of the version to introduce this, my proposal is this:

* Let's add it in 6.0.3, marking all the members as `@Experimental` or `@Incubating`
* Promote to non-experimental in the future, e.g. 6.1 or a bit later, based on experiences
* Alter `@since` to 6.0.3 :)

I.e. let's bend the semantic versioning scheme a bit and take the liberty to introduce incubating features in a micro.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/AbstractScriptAssertValidator.java
```diff
@@ -0,0 +1,52 @@
...
+
+	private static final Log log = LoggerFactory.make();
+
+	protected String languageName;
+	protected String script;
+	protected String escapedScript;
+	protected volatile ScriptAssertContext scriptAssertContext;
+
+	protected ScriptAssertContext getScriptAssertContext(ConstraintValidatorContext constraintValidatorContext) {
+		if ( scriptAssertContext == null ) {
```
gunnarmorling: Why is it done lazily upon `isValid()` and not via `initialize()`?

engine/src/main/java/org/hibernate/validator/scripting/AbstractCachingScriptEvaluatorFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Basic cacheable factory responsible for the creation of {@link ScriptEvaluator}s. This
+ * class is thread-safe. Caches {@code ScriptEvaluator} when they are requested.
+ *
+ * @author Gunnar Morling
+ * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
+ * @author Marko Bekhta
+ */
+public abstract class AbstractCachingScriptEvaluatorFactory implements ScriptEvaluatorFactory {
```
gunnarmorling: `@since` and `@Incubating` should be added.

engine/src/main/java/org/hibernate/validator/scripting/AbstractCachingScriptEvaluatorFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+	/**
+	 * A cache of script evaluators (keyed by language name).
+	 */
+	private final ConcurrentMap<String, ScriptEvaluator> scriptEvaluatorCache = new ConcurrentHashMap<>();
+
+	/**
+	 * Retrieves a script executor for the given language.
+	 *
+	 * @param languageName The name of a scripting language as expected by {@link ScriptEngineManager#getEngineByName(String)}.
```
gunnarmorling: Is the reference to `ScriptEngineManager` correct here, as we abstract from JSR 223 now?

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/ScriptAssertValidatorTest.java
```diff
@@ -37,45 +38,55 @@
  * @author Gunnar Morling
  * @author Hardy Ferentschik
  */
-public class ScriptAssertValidatorTest {
+public class ScriptAssertValidatorTest extends AbstractConstrainedTest {
 
 	@Test
 	public void scriptEvaluatesToTrue() throws Exception {
-		ConstraintValidator<ScriptAssert, Object> validator = getInitializedValidator( "groovy", "true" );
+		@ScriptAssert(lang = "groovy", script = "true") class TmpType { }
```
gunnarmorling: Nice usage of local classes :)

gunnarmorling:
Ah, one more question: shouldn't this all go under `o.h.v.spi.scripting`? It's an extension point essentially which we mean to be implemented by advanced users.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/AbstractScriptAssertValidator.java
marko-bekhta: That's actually that problem that @gsmet mentioned in another PR about having some sort of context in initialize method... As the factory is accessible only via `constraintValidatorContext` and there's no way of getting it during init, as we only have an annotation there, there's this lazy initialization ...

marko-bekhta:
@gunnarmorling thanks for reviewing this once more :) if @gsmet don't mind - I can apply the suggested changes and rebase it to current master, as I see there are a couple of conflicts now. 
Also as I understand - this should go with 6.0.3, so I assume it should be ready for tomorrow's morning right?

gsmet:
@marko-bekhta don't worry, I'll do it :)

gunnarmorling:
> That's actually that problem that [...]

Ah, yes, got it. We should look into adding such extended contract indeed
then.

> so I assume it should be ready for tomorrow's morning right?

That'd be perfect, but please don't feel stressed about it. When it's done,
it's done, if not, then not :)

gunnarmorling:
>
> @marko-bekhta <https://github.com/marko-bekhta> don't worry, I'll do it :)
>

Ha, or that :) Thx!

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter09/CustomScriptEvaluatorFactoryForSpringELTest.java
gsmet: We decided against this one as with the Spring syntax, we would end up with `#_this`. I modified the Spring ScriptEvaluatorFactory to inject all the bindings anyway as it seems like a good thing to do.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
gsmet: Done.

gsmet:
And... merged! I rebased the PR and added 2 more commits to polish the documentation and JavaDoc.

@marko-bekhta thanks for your hard work!

PR #840, opened 2017-08-18
HV-1470 ConstraintValidatorCheck not compatible with the programmatic API

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1470


gsmet:
Merged, thanks! (I adjusted the message a bit, I had second thoughts about the one I proposed!)

PR #841, opened 2017-08-28
HV-1474 Fix incorrect equals()/hashCode() contract in NodeImpl

gsmet:
* https://hibernate.atlassian.net/browse/HV-1474

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
@@ -439,7 +439,7 @@ public class NodeImpl
 		result = prime * result + ( ( kind == null ) ? 0 : kind.hashCode() );
 		result = prime * result + ( ( name == null ) ? 0 : name.hashCode() );
 		result = prime * result + ( ( parameterIndex == null ) ? 0 : parameterIndex.hashCode() );
-		result = prime * result + ( ( parameterTypes == null ) ? 0 : parameterTypes.hashCode() );
+		result = prime * result + ( ( parameterTypes == null ) ? 0 : Arrays.hashCode( parameterTypes ) );
 		result = prime * result + ( ( parent == null ) ? 0 : parent.hashCode() );
 		result = prime * result + ( ( containerClass == null ) ? 0 : containerClass.hashCode() );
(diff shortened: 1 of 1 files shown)


marko-bekhta:
@gsmet the changes looks good :) I'm just kind of curious wouldn't it be simpler to use something like:

```java
public final int buildHashCode() {
	return Objects.hash( name, parent, isIterable, index, key, kind, parameterTypes, parameterIndex, containerClass, typeArgumentIndex );
}
```
for the HashCode and then for equals:
```java
@Override 
	if ( this == o ) {
		return true;
	}
	if ( o == null || getClass() != o.getClass() ) {
		return false;
	}
	NodeImpl node = (NodeImpl) o;
	return isIterable == node.isIterable &&
			Objects.equals( name, node.name ) &&
			Objects.equals( parent, node.parent ) &&
			Objects.equals( index, node.index ) &&
			Objects.equals( key, node.key ) &&
			kind == node.kind &&
			Arrays.equals( parameterTypes, node.parameterTypes ) &&
			Objects.equals( parameterIndex, node.parameterIndex ) &&
			Objects.equals( containerClass, node.containerClass ) &&
			Objects.equals( typeArgumentIndex, node.typeArgumentIndex );
}
```

gsmet:
Yes, agreed. It's just that we are using the pre JDK 8 style for now.

I'll ask Gunnar when he's back if he has something against migrating to the new style when we make updates.

marko-bekhta:
OK, cool. If you decide to move to this newer style - let me know, if you'd like me to do the changes.

gsmet:
There is a conflict with my other PR so let's wait for everything to be included before doing so.

gunnarmorling:
This change looks good.

> I'll ask Gunnar when he's back if he has something against migrating to the new style when we make updates.

Nice, didn't know about `Objects#hash()`. It seems convinient, but I think we should only use it for pre-calculated and stored hashes of immutable objects as otherwise its array allocation may cause perf issues in hot code paths.

gsmet:
Good remark.

I took some more time to take a look at the implementation of `Objects#hash(...)`. There is one more downside: it doesn't take into account if we have to include an array in the hash. So I'm a bit worried it could lead to us having invalid hash code (typically, I wouldn't have noticed this issue).

gsmet:
Thanks for the review. Merged.

PR #842, opened 2017-08-30
HV-1475 Update Maven dependencies

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1475

@gsmet I found that some more dependencies could be updated besides groovy, so I've updated them as well.

Also a question about `log4j` should we upgrade to log4j2 ?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -155,7 +155,7 @@
         <arquillian.version>1.1.11.Final</arquillian.version>
         <testng.version>6.8</testng.version>
         <!-- it must be the exact same version than the one used in the Bean Validation TCK -->
-        <assertj-core.version>3.7.0</assertj-core.version>
+        <assertj-core.version>3.8.0</assertj-core.version>
         <junit.version>4.12</junit.version>
         <easymock.version>3.4</easymock.version>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -125,9 +125,9 @@
             These dependencies should be aligned with the ones from the WildFly version we support
             See http://search.maven.org/#search|gav|1|g%3A"org.wildfly"%20AND%20a%3A"wildfly-parent"
         -->
-        <classmate.version>1.3.1</classmate.version>
-        <jsoup.version>1.8.3</jsoup.version>
-        <joda-time.version>2.9.5</joda-time.version>
+        <classmate.version>1.3.3</classmate.version>
+        <jsoup.version>1.10.3</jsoup.version>
+        <joda-time.version>2.9.9</joda-time.version>
```
gsmet: As for these ones, they are aligned on WF 10.1 versions. We will update them once we officially switch to WF 11. See the comment above this group of versions :)

marko-bekhta: how did I missed that ?! :) force-pushed the update without these

gsmet:
Rebased and merged, thanks!

PR #844, opened 2017-08-30
HV-1477 Improve IPv4 validation in email addresses

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1477

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
@@ -28,7 +28,8 @@ public final class DomainNameUtil {
 
 	private static final String DOMAIN = DOMAIN_LABEL + "+(\\." + DOMAIN_LABEL + "+)*";
 
-	private static final String IP_DOMAIN = "[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}";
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
```diff
@@ -28,7 +28,8 @@
 
 	private static final String DOMAIN = DOMAIN_LABEL + "+(\\." + DOMAIN_LABEL + "+)*";
 
-	private static final String IP_DOMAIN = "[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}";
+	private static final String IP_V4_PART = "(25[0-5]|2[0-4][0-9]|[1][0-9][0-9]|[1-9][0-9]?|0)";
+	private static final String IP_DOMAIN = "(" + IP_V4_PART + "\\.){3}" + IP_V4_PART;
```
gsmet: Wondering if it's worth it to make the regexp more complicated for a such narrow use case. I don't have the answer, just thinking out loud :).

marko-bekhta: yeah ... I remember that we try not to over-complicate the rules for emails. And I don't know if the email with IP is used that often :) But as I stumbled upon this I thought I'll suggest the change and then it can be decided if it should be included or not :)

gsmet: After giving it some thoughts, I'm inclined to refuse it. Email with IP addresses are not widely used so I don't think we need to be strict on this.

Thanks anyway!

PR #845, opened 2017-09-05
HV-1480 Performance and memory allocation improvements

gsmet:
* https://hibernate.atlassian.net/browse/HV-1480

Note that, even if this patch improves the situation a bit, we are still allocating a lot of memory for the `PathImpl`/`NodeImpl` construction. It would be nice if we could improve the situation. Both the number of `NodeImpl` creations and the copy of the node list in `PathImpl` are really concerning.

Results for each version for the 3 benchmarks (~ a 20% speed up):

**HV 6.0.2.Final**

  960.500 ±(99.9%) 10.939 ops/ms [Average]
  (min, avg, max) = (945.886, 960.500, 989.234), stdev = 12.597
  CI (99.9%): [949.561, 971.439] (assumes normal distribution)

Result "org.hibernate.validator.performance.cascaded.CascadedValidation.testCascadedValidation":
  452.387 ±(99.9%) 14.470 ops/ms [Average]
  (min, avg, max) = (426.984, 452.387, 496.972), stdev = 16.664
  CI (99.9%): [437.917, 466.857] (assumes normal distribution)

Result "org.hibernate.validator.performance.cascaded.CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems":
  1100.356 ±(99.9%) 15.799 ops/s [Average]
  (min, avg, max) = (1067.990, 1100.356, 1124.266), stdev = 18.194
  CI (99.9%): [1084.557, 1116.155] (assumes normal distribution)

**HV 6.0.3-SNAPSHOT**

  1159.744 ±(99.9%) 9.638 ops/ms [Average]
  (min, avg, max) = (1149.546, 1159.744, 1191.913), stdev = 11.099
  CI (99.9%): [1150.107, 1169.382] (assumes normal distribution)

Result "org.hibernate.validator.performance.cascaded.CascadedValidation.testCascadedValidation":
  542.513 ±(99.9%) 15.477 ops/ms [Average]
  (min, avg, max) = (514.498, 542.513, 581.028), stdev = 17.823
  CI (99.9%): [527.036, 557.990] (assumes normal distribution)

Result "org.hibernate.validator.performance.cascaded.CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems":
  1393.947 ±(99.9%) 22.632 ops/s [Average]
  (min, avg, max) = (1348.617, 1393.947, 1429.664), stdev = 26.063
  CI (99.9%): [1371.315, 1416.580] (assumes normal distribution)

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
@@ -91,7 +91,7 @@ public class ValidationContext<T> {
 	/**
 	 * The set of already processed unit of works. See {@link ProcessedUnit}.
 	 */
-	private final Set<ProcessedUnit> processedUnits;
+	private final Set<Object> processedUnits;
 
 	/**
 	 * Maps an object to a list of paths in which it has been validated. The objects are the bean instances.
@@ -573,14 +573,12 @@ public class ValidationContext<T> {
 		}
 	}
 
-	private interface ProcessedUnit {
-	}
+	private static final class BeanGroupProcessedUnit {
 
-	private static class BeanGroupProcessedUnit implements ProcessedUnit {
-
-		private final Object bean;
-		private final Class<?> group;
-		private final int hashCode;
+		// these fields are final but we don't mark them as final as an optimization
+		private Object bean;
+		private Class<?> group;
+		private int hashCode;
 
 		private BeanGroupProcessedUnit(Object bean, Class<?> group) {
 			this.bean = bean;
@@ -593,7 +591,7 @@ public class ValidationContext<T> {
 			if ( this == o ) {
 				return true;
 			}
-			if ( o == null || getClass() != o.getClass() ) {
+			if ( o == null || getClass() != BeanGroupProcessedUnit.class ) {
 				return false;
 			}
 
@@ -621,12 +619,13 @@ public class ValidationContext<T> {
 		}
 	}
 
-	private static class BeanPathMetaConstraintProcessedUnit implements ProcessedUnit {
+	private static final class BeanPathMetaConstraintProcessedUnit {
 
-		private final Object bean;
-		private final Path path;
-		private final MetaConstraint<?> metaConstraint;
-		private final int hashCode;
+		// these fields are final but we don't mark them as final as an optimization
+		private Object bean;
+		private Path path;
+		private MetaConstraint<?> metaConstraint;
+		private int hashCode;
 
 		private BeanPathMetaConstraintProcessedUnit(Object bean, Path path, MetaConstraint<?> metaConstraint) {
 			this.bean = bean;
@@ -640,7 +639,7 @@ public class ValidationContext<T> {
 			if ( this == o ) {
 				return true;
 			}
-			if ( o == null || getClass() != o.getClass() ) {
+			if ( o == null || getClass() != BeanPathMetaConstraintProcessedUnit.class ) {
 				return false;
 			}
 
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
@@ -6,7 +6,6 @@
  */
 package org.hibernate.validator.internal.engine;
 
-import static org.hibernate.validator.internal.util.CollectionHelper.newHashMap;
 import static org.hibernate.validator.internal.util.CollectionHelper.newHashSet;
 
 import java.lang.reflect.Executable;
@@ -36,7 +35,6 @@ import org.hibernate.validator.internal.metadata.BeanMetaDataManager;
 import org.hibernate.validator.internal.metadata.aggregated.BeanMetaData;
 import org.hibernate.validator.internal.metadata.core.MetaConstraint;
 import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
-import org.hibernate.validator.internal.util.IdentitySet;
 import org.hibernate.validator.internal.util.logging.Log;
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
 
@@ -91,20 +89,14 @@ public class ValidationContext<T> {
 	private final Object executableReturnValue;
 
 	/**
-	 * Maps a group to an identity set to keep track of already validated objects. We have to make sure
-	 * that each object gets only validated once per group and property path.
+	 * The set of already processed unit of works. See {@link ProcessedUnit}.
 	 */
-	private final Map<Class<?>, IdentitySet> processedBeansPerGroup;
+	private final Set<ProcessedUnit> processedUnits;
 
 	/**
 	 * Maps an object to a list of paths in which it has been validated. The objects are the bean instances.
 	 */
 	private final Map<Object, Set<PathImpl>> processedPathsPerBean;
-
-	/**
-	 * Maps processed constraints to the bean and path for which they have been processed.
-	 */
-	private final Map<BeanAndPath, IdentitySet> processedMetaConstraints;
 
 	/**
 	 * Contains all failing constraints so far.
@@ -174,9 +166,8 @@ public class ValidationContext<T> {
 		this.executableParameters = executableParameters;
 		this.executableReturnValue = executableReturnValue;
 
-		this.processedBeansPerGroup = newHashMap();
+		this.processedUnits = new HashSet<>();
 		this.processedPathsPerBean = new IdentityHashMap<>();
-		this.processedMetaConstraints = newHashMap();
 		this.failingConstraintViolations = newHashSet();
 	}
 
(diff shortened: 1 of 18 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -349,22 +340,11 @@ else if ( executableReturnValue != null ) {
 	}
 
 	public boolean hasMetaConstraintBeenProcessed(Object bean, Path path, MetaConstraint<?> metaConstraint) {
-		// TODO switch to proper multi key map (HF)
-		IdentitySet processedConstraints = processedMetaConstraints.get( new BeanAndPath( bean, path ) );
-		return processedConstraints != null && processedConstraints.contains( metaConstraint );
+		return processedUnits.contains( new BeanPathMetaConstraintProcessedUnit( bean, path, metaConstraint ) );
```
Sanne: this line bothers me a bit. From chat conversations we know the contains() is problematic from a performance perspective and I'm inclined to suspect the allocation of `BeanPathMetaConstraintProcessedUnit`.

Would be interesting to see if the performance improves more by having nested maps rather than a Set with composite keys.

gsmet: That's basically what was done before (at least, it was done partly). And the creation/allocations/resizing of the nested structures made it worse.

I can try it again nesting everything though.

gsmet:
Rebased to fix a conflict.

engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
```diff
@@ -69,6 +68,7 @@
 	private final Class<?> containerClass;
 	private final Integer typeArgumentIndex;
 
+	private int hashCode = -1;
```
gunnarmorling: I'm slightly concerned about this change, doesn't it potentially impact safe publication to other threads (it's part of of the published `ConstraintViolation`)?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -609,15 +573,64 @@ private ValidationContextBuilder(
...
+		private int createHashCode() {
+			int result = System.identityHashCode( bean );
+			result = 31 * result + group.hashCode();
+			return result;
+		}
+	}
+
+	private static final class BeanPathMetaConstraintProcessedUnit {
+
+		// these fields are final but we don't mark them as final as an optimization
```
gunnarmorling: Can you shed some light on this optimization?

gsmet: To be honest, it's some dark magic suggested by Sanne. It did improve one of the benchmarks significantly.

engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
gsmet: I understand your concern but it's already done like that in `PathImpl` for a long time. So, either we need to get rid of that optimization in both or we decide it's safe in both.

gunnarmorling: I'd prefer to stay on the safer side in both cases

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
gunnarmorling: Hum, would be nice to have some sort of explanation...

engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
narbsy: Does it matter that we impact safe publication here? The failure mode is to re-calculate the hash in a separate thread, so at worst we're duplicating some effort. 

The String class in the JDK does the same optimization.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
narbsy: Based on the JMM, `final` fields are safely published, which ends up incurring a memory fence.

JLS Specification of final field behavior: https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5
SO post about the memory fence: https://stackoverflow.com/questions/23527252/cost-of-using-final-fields

narbsy:
Playing locally, I was able to get a nice speedup from removing the `CachingTraversableResolver` in `ValidatorImpl`. It seems that when not using JPA, it adds a lot of overhead for little benefit. Is it necessary for correctness or guarantees of the spec? Can we make it configurable?

    1480 + no caching

    Benchmark                                                                                                 Mode  Cnt     Score     Error   Units
    o.h.v.p.cascaded.CascadedValidation.testCascadedValidation                                               thrpt   11   273.248 ± 130.578  ops/ms
    o.h.v.p.cascaded.CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                 thrpt   20  1280.523 ±  50.736   ops/s
    o.h.v.p.multilevel.MultiLevelContainerValidation.testMultiLevelPreGeneratedValidation                    thrpt   50     0.062 ±   0.006  ops/ms
    o.h.v.p.simple.SimpleValidation.testSimpleBeanValidation                                                 thrpt   20  1074.496 ± 198.158  ops/ms
    o.h.v.p.statistical.StatisticalValidation.testValidationWithStatisticalGraphDepthAndConstraintValidator  thrpt   10     0.302 ±   0.014  ops/ms

    6.0.2-Final

    Benchmark                                                                                                 Mode  Cnt     Score    Error   Units
    o.h.v.p.cascaded.CascadedValidation.testCascadedValidation                                               thrpt   20   336.993 ± 80.243  ops/ms
    o.h.v.p.cascaded.CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                 thrpt   20   996.199 ± 90.029   ops/s
    o.h.v.p.multilevel.MultiLevelContainerValidation.testMultiLevelPreGeneratedValidation                    thrpt   50     0.060 ±  0.006  ops/ms
    o.h.v.p.simple.SimpleValidation.testSimpleBeanValidation                                                 thrpt   20  1000.653 ± 33.639  ops/ms
    o.h.v.p.statistical.StatisticalValidation.testValidationWithStatisticalGraphDepthAndConstraintValidator  thrpt   10     0.272 ±  0.013  ops/ms

gsmet:
@narbsy good observation. I noticed that too the other day but I didn't dig further. I'll take a look today.

gsmet:
@narbsy done in my latest commit, thanks for the suggestion.

@gunnarmorling 

So, I included one of your commits (the simple one changing the way we get a reference to the parent node) and added one commit to follow @narbsy 's suggestion.

After giving it a lot of thoughts, I didn't work any further on the PathImpl/NodeImpl front. What I did doesn't make it less fragile that it was before. And changing the way it works to safe publish the PathImpl/NodeImpl would be a lot of work (I can envision a lot more failing cases that you did in your testing, especially if we try to introduce a NodeBuilder too) for a very narrow benefit (I don't see how using the resulting Path in several separate threads might make sense). As for reducing the number of allocations, it sure is something we should pursue but what you did is also a trade-off so I think it's going to be a long and bumpy road to get there.

So for now, I would like us to consider merging this PR as is (except if you have some comment on the last commit of course) and keep the PathImpl/NodeImpl issue for later, considering it's orthogonal to the work I did (the NodeImpl hashCode change does not change anything as it now works as the PathImpl hashCode always did).

Once this is in, I plan to release 6.0.3.Final.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -1277,12 +1278,20 @@ else if ( propertyPathNode.getKey() != null ) {
...
+	 * returned directly.
+	 * <p>
+	 * If it is not, we wrap the resolver for caching. In this case, a new instance is returned each time and it should
+	 * be used only for the duration of a validation call.
 	 *
 	 * @return The resolver for the duration of a full validation.
 	 */
-	private TraversableResolver getCachingTraversableResolver() {
+	private static TraversableResolver wrapTraversableResolverForCachingIfRequired(TraversableResolver traversableResolver) {
+		if ( traversableResolver == TraverseAllTraversableResolver.INSTANCE ) {
```
narbsy: this doesn't help where one registers a custom traversable resolver to use, unfortunately -- so does not help my use case at all :(

any way we can either make wrapping configurable, or allow a user to opt-out of the caching behavior?

gsmet: Sure, it does not help if you're using a custom traversable resolver. Out of curiosity, what does your custom traversable resolver do?

I'll see what I can do about providing a way to disable the caching.

narbsy: I have effectively a feature toggle around a subset of configuration; if the feature that uses that configuration is not active, I don't cascade into that subset of configuration. Removing it would cause quite some user frustration :(

Perhaps something like:

```
        Validation.byDefaultProvider().configure()
            .traversableResolver(new CustomTraversableResolverThatIsReallyFast())
            .disableTraversableResolverCaching()
            .buildValidatorFactory().getValidator();
```

to ensure it is opt-in?

Appreciate your efforts on this project!

gsmet: Yes, that's what I had in mind. I'm gonna remove this commit from the PR and create another PR.

gsmet: @narbsy I created https://github.com/hibernate/hibernate-validator/pull/856

gunnarmorling:
@narbsy, as a heads-up, an option for disabling the traversable resolver cache is provided via [HV-1487](https://hibernate.atlassian.net/browse/HV-1487).

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
gunnarmorling: Ok, it was news to me that there's a prize to pay also if this object is only used from within one thread only. Good to know :)

engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
gunnarmorling: Yes, that's good reasoning.

gunnarmorling:
Rebased and applied. Thanks, @gsmet and @narbsy!

narbsy:
Thank you @gunnarmorling + @gsmet !! Really appreciate your efforts!

PR #846, opened 2017-09-06
HV-1478 OSGi fixes for the cdi module

gsmet:
* https://hibernate.atlassian.net/browse/HV-1478


cdi/pom.xml
```diff
@@ -167,7 +166,8 @@
                 <artifactId>maven-bundle-plugin</artifactId>
                 <configuration>
                     <instructions>
-                        <Fragment-Host>org.hibernate.validator</Fragment-Host>
+                        <Bundle-SymbolicName>${hibernate-validator-cdi.module-name}</Bundle-SymbolicName>
```
gsmet: @gunnarmorling you mentioned making the bundle names consistent with the Java 9 module names in another issue so that's what I did. I'm not sure if it can cause compatibility issues though?

gunnarmorling: Why is the `Fragment-Host` header removed? Do you not want this to be a fragment any longer? On the name, did we actually define one for CDI before? If not, it should be fine, but if we did define it, let's wait with changing it at least until the next minor release (6.1) (same for the engine bundle name).

cdi/pom.xml
```diff
@@ -167,12 +166,13 @@
...
+                        <Fragment-Host>${hibernate-validator.module-name}</Fragment-Host>
                         <Import-Package>
-                            org.hibernate.validator.*;version="[${project.version},${project.version}]",
```
gunnarmorling: Note that this was done on purpose, but reading [this explanation](https://felix.apache.org/documentation/tutorials-examples-and-presentations/apache-felix-osgi-faq.html#should-a-bundle-import-its-own-exported-packages) again, it seems ok to not do it in our case.

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidationExtension.java
```diff
@@ -6,416 +6,9 @@
...
-					throw log.getValidateOnExecutionOnOverriddenOrInterfaceMethodException( method );
-				}
-				return overriddenOrInterfaceMethod;
-			}
-		}
-
-		return method;
-	}
+@Deprecated
+public class ValidationExtension extends org.hibernate.validator.cdi.ValidationExtension {
```
gunnarmorling: I'm wondering why this needs exporting at all? I've asked Arjan on the issue, let's see what he says.

pom.xml
```diff
@@ -99,6 +99,10 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
 
+        <!-- Module names used for Java 9 modules and OSGi bundles -->
+        <hibernate-validator.module-name>org.hibernate.validator</hibernate-validator.module-name>
```
gunnarmorling: I'd keep those properties local to the modules using them.

cdi/pom.xml
gsmet: Mmmh, yeah, the fragment host shouldn't have been removed. Don't know how it went away. I'll reintroduce it.

About the name, we did not define one explicitly but the plugin defined one based on the Maven coordinates. So we're definitely changing something if we do so. Thus my question about compatibility.

gsmet: Ah ah, it didn't go away, it's just there vvvvvv

cdi/pom.xml
```diff
@@ -167,12 +166,13 @@
...
+                            javax.validation.*;version="[2.0,3.0)",
```
gunnarmorling: I think we should inherit this from the fragment host. On the `bootstrap` package I've asked Arjan on the issue why it is that it's needed. It seems odd as we don't reference that class.

cdi/pom.xml
gunnarmorling: Ah yeah. That for superfluous changes :)

pom.xml
gsmet: Yeah, but cdi is referencing the engine module name. And I want to be sure they use the same name, avoiding the issue we have now.

That's why I put both here.

gunnarmorling: Ok.

gsmet:
I force pushed an update keeping the old names for the bundles but fixing the `Fragment-Host` header. I kept the Java 9 module names properties as it's already part of the work done for 6.1.

pom.xml
```diff
@@ -99,6 +99,13 @@
...
+        <hibernate-validator-cdi.module-name>org.hibernate.validator.cdi</hibernate-validator-cdi.module-name>
+        <!-- TODO in 6.1: make the names consistent by removing the bundle properties, see HV-1484 -->
+        <hibernate-validator.bundle-name>org.hibernate.validator.hibernate-validator</hibernate-validator.bundle-name>
+        <hibernate-validator-cdi.bundle-name>org.hibernate.validator.hibernate-validator-cdi</hibernate-validator-cdi.bundle-name>
```
gunnarmorling: We already could use the right one for the CDI module already, no?

gsmet: Why? It already had a name too.

gunnarmorling:
Ah, ok, didn't see that.

pom.xml
gsmet: The thing is that if we don't explicitly define one, a name is forged from the Maven coordinates. So even, if not specified, each bundle already has a name. It's just that, now, the name is specified and clear. And we will change it in 6.1.

gunnarmorling:
Ok, makes sense. I don't expect the name of the fragment to be of much
significance, but better safe than sorry.

The only remaining question then is whether to do the relocation of the
extension class or not. Hoping Arjan provides the stack trace so to decide
whether it must be or not.

pom.xml
gunnarmorling: Makes sense.

gunnarmorling:
Rebased and applied. Thanks, @gsmet! I've added one more commit for making one method of the extension private which apparently accidentally was public before.

PR #847, opened 2017-09-07
HV-1481 Legacy cascading support broken when the collection/array type cannot be detected at runtime

gsmet:
https://hibernate.atlassian.net/browse/HV-1481
https://hibernate.atlassian.net/browse/HV-1482

I kept the original approach of modifying the cascading metadata as it's faster if we can use it.

Note that this patch adds some overhead even when we have something like:
```
@Valid
private MyBean bean;
```

but I don't think we can avoid it, unfortunately...

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -54,7 +54,6 @@ import org.hibernate.validator.internal.metadata.aggregated.BeanMetaData;
 import org.hibernate.validator.internal.metadata.aggregated.CascadingMetaData;
 import org.hibernate.validator.internal.metadata.aggregated.ContainerCascadingMetaData;
 import org.hibernate.validator.internal.metadata.aggregated.ExecutableMetaData;
-import org.hibernate.validator.internal.metadata.aggregated.NonContainerCascadingMetaData;
 import org.hibernate.validator.internal.metadata.aggregated.ParameterMetaData;
 import org.hibernate.validator.internal.metadata.aggregated.PropertyMetaData;
 import org.hibernate.validator.internal.metadata.aggregated.ReturnValueMetaData;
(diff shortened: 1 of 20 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/NonContainerCascadingMetaData.java
```diff
@@ -31,6 +51,22 @@
...
+	private static final Set<ValueExtractorDescriptor> ARRAY_VALUE_EXTRACTOR_DESCRIPTOR_SET = CollectionHelper.asSet(
+			BooleanArrayValueExtractor.DESCRIPTOR,
+			ByteArrayValueExtractor.DESCRIPTOR,
+			CharArrayValueExtractor.DESCRIPTOR,
+			DoubleArrayValueExtractor.DESCRIPTOR,
+			FloatArrayValueExtractor.DESCRIPTOR,
+			IntArrayValueExtractor.DESCRIPTOR,
+			LongArrayValueExtractor.DESCRIPTOR,
+			ObjectArrayValueExtractor.DESCRIPTOR,
+			ShortArrayValueExtractor.DESCRIPTOR
```
marko-bekhta: What if place these collections in `ValueExtractorManager` and either make them public or provide  "static getters" for them. Then there will be no need in exposing extractor implementations out of the package. Or is the manager just not the right place for them ?

gsmet: Yes, sure, we can discuss this.

I thought about putting them in the manager but I didn't find that compelling in the end.

marko-bekhta: I just saw those visibility changes from default to public so I thought I'll ask (to better understand the preferences with these visibility things :) ). If it's OK for those extractors to be exposed then I guess everything is fine :)

I think this is a bit related to that question about those cycles between packages and how tightly the packages are coupled with other packages and things like that...

gsmet: I'm thinking about introducing a LegacyCollectionValueExtractors class in the said package. But not sure it has much value.

The idea is to reduce the visibility as much as possible but we have a lot of public stuff that are not really public and are just public because we isolated code in a package for clarity.

I admit though it's a bit weird to have part of the extractors public and part of them package protected.

gsmet: @marko-bekhta OK, did that and made the visibility of the value extractors consistent. Some of them were already uselessly public.

marko-bekhta: nice, and thanks for the explanation about the visibility :)

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/NonContainerCascadingMetaData.java
```diff
@@ -95,6 +101,34 @@ public boolean isContainer() {
 		return false;
 	}
 
+	public ContainerCascadingMetaData getLegacyContainerCascadingMetaData(Class<?> valueClass) {
+		if ( List.class.isAssignableFrom( valueClass ) ) {
+			return new ContainerCascadingMetaData( List.class, List.class.getTypeParameters()[0], List.class, List.class.getTypeParameters()[0],
```
gunnarmorling: Could we use singleton instances of these meta-data types perhaps? At least for the case where there are no group conversions?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -583,6 +584,21 @@ private void validateCascadedConstraints(ValidationContext<?> validationContext,
 					// validate cascading on the annotated object
 					if ( cascadingMetaData.isCascading() ) {
 						validateCascadedAnnotatedObjectForCurrentGroup( value, validationContext, valueContext, cascadingMetaData );
+
+						if ( !cascadingMetaData.isContainer() ) {
+							// We might be in the case where the legacy @Valid support for collections and arrays has been used on a type for which
+							// it cannot be detected at bootstrap time: thus we need to check with the runtime type.
```
gunnarmorling: I'd clarify that comment a bit:

> Handle the case where the element is marked with `@Valid` but its static type isn't a container type (e.g. @Valid Object list = new ArrayList()), as the metadata won't be marked as a container then; if the static type is a container type the metadata will be marked as such and the regular container logic below kicks in.

gsmet:
@gunnarmorling I think I addressed your comments. It's a bit more clear now, I think as the thing is not directly included in the ValidatorImpl flow.

gunnarmorling:
Rebased and applied; thanks, @gsmet!

PR #848, opened 2017-09-15
HV-1478 Integration test for OSGi/CDI on Felix/Payara

arjantijms:
HV-1478 Test that specifically tests whether the hibernate-validator-cdi jar's OSGi headers are correct, Felix is able to process them and startup, and then an actual CDI and BV test is able to pass.

Note that to pass this depends on the work done in https://github.com/hibernate/hibernate-validator/pull/846

### DIFF
--- a/osgi/felixtest/cdi-validation/pom.xml
+++ b/osgi/felixtest/cdi-validation/pom.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Hibernate Validator, declare and validate application constraints
+  ~
+  ~ License: Apache License, Version 2.0
+  ~ See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+  -->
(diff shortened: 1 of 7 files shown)


gunnarmorling:
ok to test

gsmet:
@arjantijms looks like there is a test failure:

```
Failed tests: 
  CDITest.testAll9Invalid Expected test to throw exception with message a string containing "invalid number"

Tests run: 3, Failures: 1, Errors: 0, Skipped: 0
```

arjantijms:
@gsmet 
>looks like there is a test failure:

If the test is executed against Hibernate Validator without https://github.com/hibernate/hibernate-validator/pull/846 applied this is expected, since it will not load the CDI extension with that version and thus not throw the expected exception.

But if 846 was applied and tested with this I'll need to check again.

gsmet:
Ah sure, sorry about that :). I'll take another look tomorrow morning!

gsmet:
Closing in favor of https://github.com/hibernate/hibernate-validator/pull/851 .

PR #850, opened 2017-09-18
HV-1480

gunnarmorling:
Hey @gsmet, this is a small experiment based on your perf improvement work. You pointed out that the allocation of nodes and paths is a problem and I agree.

We went for the immutable approach back in the day as we often had issues with paths which were referenced e.g. in a constraint violation and then be altered later on in the course of subsequent validation of other constraints. I think the best way forward would be to separate mutable and immutable path and node types. The engine would in general work with the mutable representations and only "freeze out" an immutable reference when needed (constraint violation creation, storing processed path). 

As a quick attempt, this PR is doing this very hackishly for paths, could you check whether it already makes a noticable difference? I'd expect a bigger difference when doing the same for nodes, but an early check may be useful to see whether it's in the right direction or not.


gunnarmorling:
Hey @gsmet, what should we do about this one? Would you like to pursue the mutable/immutable split for paths and nodes at this time? If not, you could at least integrate my first commit and we should be good to go once you've addressed that one remark about a hash code.

On the split above, instead of using `Stack` for mutable paths, we also may consider to make this a linked list of dedicated node implementations which should help with reducing memory allocation.

PR #853, opened 2017-09-20
HV-1485 Make ConstraintViolationImpl serializable again

gsmet:
* https://hibernate.atlassian.net/browse/HV-1485


gunnarmorling:
Merged; thanks, @gsmet. Cool you found that existing serialization test, wasn't aware of that one really :)

PR #854, opened 2017-09-20
HV-1478 Only enable the new Felix/CDI test on JDK 8

gsmet:


gunnarmorling:
Rebased and merged; thanks, @gsmet!

PR #856, opened 2017-10-05
 HV-1487 Allow to disable the TraversableResolver result cache

gsmet:
* https://hibernate.atlassian.net/browse/HV-1487

### DIFF
--- a/documentation/src/test/java/org/hibernate/validator/referenceguide/chapter09/validation-DisableTraversableResolverResultCache.xml
+++ b/documentation/src/test/java/org/hibernate/validator/referenceguide/chapter09/validation-DisableTraversableResolverResultCache.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<validation-config
+        xmlns="http://xmlns.jcp.org/xml/ns/validation/configuration"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/validation/configuration validation-configuration-2.0.xsd"
+        version="2.0">
+    <default-provider>org.hibernate.validator.HibernateValidator</default-provider>
+
+    <property name="hibernate.validator.enable_traversable_resolver_result_cache">false</property>
+</validation-config>
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
@@ -122,9 +122,14 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 	private final boolean failFast;
 
 	/**
-	 * Hibernate validator specific flags to relax constraints on parameters.
+	 * Hibernate Validator specific flags to relax constraints on parameters.
 	 */
 	private final MethodValidationConfiguration methodValidationConfiguration;
+
+	/**
+	 * Hibernate Validator specific flag to disable the {@code TraversableResolver} result cache.
+	 */
+	private final boolean traversableResolverResultCacheEnabled;
 
 	/**
 	 * Metadata provider for XML configuration.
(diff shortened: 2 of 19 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -409,9 +428,9 @@ Validator createValidator(ConstraintValidatorFactory constraintValidatorFactory,
 		return metaDataProviders;
 	}
 
-	private boolean checkPropertiesForBoolean(Map<String, String> properties, String propertyKey, boolean programmaticValue) {
+	private boolean checkPropertiesForFailFast(Map<String, String> properties, boolean programmaticValue) {
```
gsmet: @gunnarmorling 

So this method was in fact very specific to the failFast case. See especially the log message of the error thrown.

I'm not sure what we really want to do here:
1/ report the error only for failFast
2/ report the error for the other properties too but in this case, we should generalize the log message. And we will probably have to change the way these booleans are initialized as the current logic only works if the default is true. In the case of the traversable resolver cache, the default is true so we can't use the same logic (we would probably need to have `Boolean` with null values in case the value is not specified).

I went with 1/ for now.

engine/src/main/java/org/hibernate/validator/internal/engine/resolver/TraverseAllTraversableResolver.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * {@code TraversableResolver} considering that all properties are reachable and cascadable.
+ * <p>
+ * This is the default behavior if JPA is not detected in the classpath.
+ *
+ * @author Guillaume Smet
+ */
+public class TraverseAllTraversableResolver implements TraversableResolver {
+
+	public static final TraverseAllTraversableResolver INSTANCE = new TraverseAllTraversableResolver();
```
gsmet: Having a static instance might be a bit counter productive if we use a custom `TraversableResolver` as we instantiate an object that won't be used. But I did it so we can use an equality check instead of an `instanceof` when wrapping the `TraversableResolver` with a caching one.

engine/src/main/java/org/hibernate/validator/internal/engine/resolver/CachingJPATraversableResolverForSingleValidation.java
```diff
@@ -0,0 +1,105 @@
...
+
+	@Override
+	public boolean isCascadable(Object traversableObject, Path.Node traversableProperty, Class<?> rootBeanType, Path pathToTraversableObject,
+			ElementType elementType) {
+		// JPATraversableResolver returns true for isCascadable() per spec so we can avoid the overhead of caching.
+
+		return true;
+	}
+
+	private static final class TraversableHolder {
```
gunnarmorling: Seems we could avoid some redundancy by extracting a base class for the two caching implementations?

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter09/validation-DisableTraversableResolverResultCache.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<validation-config
+        xmlns="http://xmlns.jcp.org/xml/ns/validation/configuration"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/validation/configuration validation-configuration-2.0.xsd"
+        version="2.0">
+    <default-provider>org.hibernate.validator.HibernateValidator</default-provider>
+
+    <property name="hibernate.validator.enable_traversable_resolver_result_cache">false</property>
+</validation-config>
```
gunnarmorling: New line is missing.

engine/src/main/java/org/hibernate/validator/internal/engine/resolver/CachingJPATraversableResolverForSingleValidation.java
gsmet: Yeah I thought about doing it but then I thought it might be introducing some complexity for no real advantage. But if you think so too, I suppose we should do it. Will push an update.

gsmet:
@gunnarmorling force pushed an update fixing your remarks.

About having the user use custom implementation, as mentioned on HipChat, the wrapping has to be made per validation call so we cannot do that.

gsmet:
Apparently, there's something wrong going on. Checking.

gsmet:
Fixed, it should be OK now.

gunnarmorling:
Cool; thanks, @gsmet!

PR #857, opened 2017-10-06
HV-1490 Don't fallback to the TCCL when detecting features in ConstraintHelper

gsmet:
We need to be able to access the class from the HV CL when we extract
the validated type from the validator generics.

 * https://hibernate.atlassian.net/browse/HV-1490

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
@@ -990,7 +990,7 @@ public class ConstraintHelper {
 
 	private static boolean isClassPresent(String className) {
 		try {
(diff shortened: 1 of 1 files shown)


gsmet:
Jenkins, retest this please

gunnarmorling:
Could we alternative also *load* the class using the TCCL? I.e. check and load the class using the same class loader? I think we had this discussion in the past but can't quite remember the outcome. Assuming the constraint validator class doesn't require any HV specifics (or we can restructure their code so that they don't) this also may let WF users use the Money constraints if they provide this as part of the deployment. WDYT?

gunnarmorling:
After discussing with @gsmet we agreed that loading via the TCCL isn't feasible as the constraint validator classes of HV won't be visible there.

Merging as proposed.

PR #858, opened 2017-10-10
Some documentation improvements

gsmet:
* https://hibernate.atlassian.net/browse/HV-1415
 * https://hibernate.atlassian.net/browse/HV-1417


gunnarmorling:
Applied; thanks, @gsmet! Note I've added one more commit to reference the BV spec URL via the property we've defined for it.

PR #859, opened 2017-10-14
HV-1492 Invoking log method with correct argument order

gunnarmorling:
https://hibernate.atlassian.net/projects/HV/issues/HV-1492


gsmet:
Merged, thanks!

PR #860, opened 2017-10-17
HV-1496 Add @CodePointLength that validate code point length

kazuki43zoo:
See https://hibernate.atlassian.net/browse/HV-1496 .


gunnarmorling:
Ok to test

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/CodePointLengthValidator.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.validation.ConstraintValidatorContext;
+
+/**
+ * Check that the code point length of character sequence is between min and max.
+ *
+ * @author Kazuki Shimizu
+ */
+public class CodePointLengthValidator implements ConstraintValidator<CodePointLength, CharSequence> {
+
+	private static final Log log = LoggerFactory.make();
```
gunnarmorling: Could you name this `LOG` instead of `log` as it's static (it's probably still wrong in some existing classes but it'd be nice to do it right in new code).

engine/src/main/java/org/hibernate/validator/constraints/CodePointLength.java
```diff
@@ -0,0 +1,58 @@
...
+import java.lang.annotation.Target;
+
+import javax.validation.Constraint;
+import javax.validation.Payload;
+
+import org.hibernate.validator.constraints.CodePointLength.List;
+
+
+/**
+ * Validate that the code point count of character sequence is between min and max included.
```
gunnarmorling: s/of/of a/

kazuki43zoo:
@gsmet thanks for your comments! I'll update this PR at later.

gsmet:
@kazuki43zoo Thanks!

FYI, we have a release planned for tomorrow morning (CEST time) so it would be nice if we could include this one in it.

documentation/src/main/asciidoc/ch02.asciidoc
```diff
@@ -687,6 +687,10 @@ With one exception also these constraints apply to the field/property level, onl
 	Supported data types::: `CharSequence`
 	Hibernate metadata impact::: Column length will be set to max
 
+`@CodePointLength(min=, max=)`:: Validates that code point length of the annotated character sequence is between `min` and `max` included
+	Supported data types::: `CharSequence`
+	Hibernate metadata impact::: Column length will be set to max
```
marko-bekhta: Impact should probably be None, @gunnarmorling @gsmet ?

kazuki43zoo:
@gunnarmorling @gsmet 

I want to support "Unicode normalization" option as follow:

```java
public @interface CodePointLength {
	int min() default 0;

	int max() default Integer.MAX_VALUE;

	NormalizationStrategy normalizationStrategy() default NormalizationStrategy.NONE;

	String message() default "{org.hibernate.validator.constraints.CodePointLength.message}";

	Class<?>[] groups() default { };

	Class<? extends Payload>[] payload() default { };

	/**
	 * Defines several {@code @CodePointLength} annotations on the same element.
	 */
	@Retention(RUNTIME)
	@Documented
	public @interface List {
		CodePointLength[] value();
	}

	public enum NormalizationStrategy {

		/**
		 * No normalization.
		 */
		NONE(null),

		/**
		 * Normalization by canonical decomposition.
		 */
		NFD(Normalizer.Form.NFD),

		/**
		 * Normalization by canonical decomposition, followed by canonical composition.
		 */
		NFC(Normalizer.Form.NFC),

		/**
		 * Normalization by compatibility decomposition.
		 */
		NFKD(Normalizer.Form.NFKD),

		/**
		 * Normalization by compatibility decomposition, followed by canonical composition.
		 */
		NFKC(Normalizer.Form.NFKC);

		private final Normalizer.Form form;

		NormalizationStrategy(Normalizer.Form form) {
			this.form = form;
		}

		/**
		 * Normalize specified character sequence.
		 * @param value target value
		 * @return normalized value
		 */
		public CharSequence normalize(CharSequence value) {
			if (this.form == null || value == null) {
				return value;
			}
			return Normalizer.normalize(value, this.form);
		}

	}

}
```

What do you think?

documentation/src/main/asciidoc/ch02.asciidoc
gsmet: Yes, good catch. I don't think we can safely set the length in the database.

gsmet:
@kazuki43zoo and you would use the normalized form to check the code point length?

Could you give us a use case for that?

Your implementation looks good so feel free to do it if you can give us a use case.

Btw, I will wrap the release soon so we would need this change sooner rather than later :).

Thanks!

kazuki43zoo:
For example, there is the case that normalizes "が"(one character) and "か゛"(two characters) of Hiragana and persists a normalized value to the database or external system. In such a case, we need to validate a normalized value.

gsmet:
@kazuki43zoo OK so it looks like it's something useful!

Please proceed, it looks like a nice addition to what you already did.

documentation/src/main/asciidoc/ch02.asciidoc
kazuki43zoo: ok! I will change to "None".

kazuki43zoo:
@gsmet @gunnarmorling 

I've added the normalization option and updated the document.
Should I squash to one commit?

gsmet:
@kazuki43zoo I'll review, squash it and merge it, thanks!

gsmet:
@kazuki43zoo I squashed everything, amended your commit to fix very minor formatting/wording issues here and there and merged.

Great contribution! Thanks a lot for that.

PR #861, opened 2017-10-17
HV-1493 Add time based constraints with clock skew tolerance

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1493

I took slightly different approach than @gunnarmorling  was suggesting in comments to https://hibernate.atlassian.net/browse/BVAL-700. It fits the current flows without changes, so that's why I'm thinking that it might work. 

@gunnarmorling @gsmet what do you think about it in general?

With some additional changes we would be able to reuse abstract validators for BV time constraints and for this new one. Though we would to have a list of implementations for all types for both BV and this new constraint...

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
+++ b/engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
@@ -6,10 +6,15 @@
  */
 package org.hibernate.validator;
 
+import java.time.Duration;
 import java.util.Set;
 
 import javax.validation.Configuration;
 import javax.validation.TraversableResolver;
+import javax.validation.constraints.Future;
+import javax.validation.constraints.FutureOrPresent;
+import javax.validation.constraints.Past;
+import javax.validation.constraints.PastOrPresent;
 import javax.validation.valueextraction.ValueExtractor;
 
 import org.hibernate.validator.cfg.ConstraintMapping;
@@ -85,6 +90,15 @@ public interface HibernateValidatorConfiguration extends Configuration<Hibernate
 	 */
 	@Incubating
 	String SCRIPT_EVALUATOR_FACTORY_CLASSNAME = "hibernate.validator.script_evaluator_factory";
+
+	/**
+	 * Property for configuring clock skew tolerance, allowing to set the acceptable margin of error when comparing
+	 * date/time in time related constraints. In milliseconds.
+	 *
+	 * @since 6.0.5
+	 */
+	@Incubating
+	String CLOCK_SKEW_TOLERANCE = "hibernate.validator.clock_skew_tolerance";
 
 	/**
 	 * <p>
@@ -264,4 +278,17 @@ public interface HibernateValidatorConfiguration extends Configuration<Hibernate
 	 */
 	@Incubating
 	HibernateValidatorConfiguration scriptEvaluatorFactory(ScriptEvaluatorFactory scriptEvaluatorFactory);
+
+	/**
+	 * Allows set the acceptable margin of error when comparing date/time in time related constraints
+	 * like {@link Past}/{@link PastOrPresent} and {@link Future}/{@link FutureOrPresent}.
+	 *
+	 * @param clockSkewTolerance the acceptable tolerance as {@link Duration}
+	 *
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 6.0.5
+	 */
+	@Incubating
+	HibernateValidatorConfiguration clockSkewTolerance(Duration clockSkewTolerance);
 }
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
@@ -1,0 +1,36 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.constraintvalidation;
+
+import org.hibernate.validator.Incubating;
+import org.hibernate.validator.spi.scripting.ScriptEvaluator;
+import org.hibernate.validator.spi.scripting.ScriptEvaluatorFactory;
+import org.hibernate.validator.spi.scripting.ScriptEvaluatorNotFoundException;
+
+/**
+ * Provides contextual data and operations when initializing constraint validator.
+ *
+ * @author Marko Bekhta
+ * @since 6.1
+ */
+@Incubating
+public interface HibernateConstraintValidatorInitializationContext {
+
+	/**
+	 * Returns a {@link ScriptEvaluator} created based on the {@link ScriptEvaluatorFactory}
+	 * passed at bootstrap.
+	 *
+	 * @param languageName the name of the scripting language
+	 *
+	 * @return a script executor for the given language. Never null.
+	 *
+	 * @throws ScriptEvaluatorNotFoundException in case no {@link ScriptEvaluator} was
+	 * found for a given {@code languageName}
+	 */
+
+	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName);
+}
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.validator.constraintvalidation;
 
+import java.time.Duration;
+
 import org.hibernate.validator.Incubating;
 import org.hibernate.validator.spi.scripting.ScriptEvaluator;
 import org.hibernate.validator.spi.scripting.ScriptEvaluatorFactory;
@@ -33,4 +35,13 @@ public interface HibernateConstraintValidatorInitializationContext {
 	 */
 
 	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName);
+
+	/**
+	 * Returns clock skew tolerance as {@link Duration} which is used to determine
+	 * acceptable margin of error in milliseconds, which is allowed
+	 * when comparing date/time in time related constraints.
+	 *
+	 * @return a tolerance as  {@link Duration}
+	 */
+	Duration getClockSkewTolerance();
 }
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractInstantBasedTimeValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractInstantBasedTimeValidator.java
@@ -9,12 +9,15 @@ package org.hibernate.validator.internal.constraintvalidators.bv.time;
 import java.lang.annotation.Annotation;
 import java.lang.invoke.MethodHandles;
 import java.time.Clock;
+import java.time.Duration;
 import java.time.Instant;
 
 import javax.validation.ClockProvider;
-import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
+import javax.validation.metadata.ConstraintDescriptor;
 
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidator;
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorInitializationContext;
 import org.hibernate.validator.internal.util.logging.Log;
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
 
@@ -24,9 +27,16 @@ import org.hibernate.validator.internal.util.logging.LoggerFactory;
  * @author Alaa Nassef
  * @author Guillaume Smet
  */
-public abstract class AbstractInstantBasedTimeValidator<C extends Annotation, T> implements ConstraintValidator<C, T> {
+public abstract class AbstractInstantBasedTimeValidator<C extends Annotation, T> implements HibernateConstraintValidator<C, T> {
 
 	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
+
+	protected Duration tolerance;
+
+	@Override
+	public void initialize(ConstraintDescriptor<C> constraintDescriptor, HibernateConstraintValidatorInitializationContext initializationContext) {
+		this.tolerance = initializationContext.getClockSkewTolerance();
+	}
 
 	@Override
 	public boolean isValid(T value, ConstraintValidatorContext context) {
@@ -45,7 +55,7 @@ public abstract class AbstractInstantBasedTimeValidator<C extends Annotation, T>
 			throw LOG.getUnableToGetCurrentTimeFromClockProvider( e );
 		}
 
-		int result = getInstant( value ).compareTo( reference.instant() );
+		int result = getInstant( value ).compareTo( adjustedReferenceValue( reference.instant() ) );
 
 		return isValid( result );
 	}
@@ -54,6 +64,12 @@ public abstract class AbstractInstantBasedTimeValidator<C extends Annotation, T>
 	 * Returns the {@link Instant} measured from Epoch.
 	 */
 	protected abstract Instant getInstant(T value);
+
+	/**
+	 * @param value a value to adjust
+	 * @return an adjusted value based on {@link #tolerance} and constraint type
+	 */
+	protected abstract Instant adjustedReferenceValue(Instant value);
 
 	/**
 	 * Returns whether the result of the comparison between the validated value and the time reference is considered
--- a/engine/src/main/java/org/hibernate/validator/HibernateValidatorFactory.java
+++ b/engine/src/main/java/org/hibernate/validator/HibernateValidatorFactory.java
@@ -7,6 +7,8 @@
 
 package org.hibernate.validator;
 
+import java.time.Duration;
+
 import javax.validation.ValidatorFactory;
 
 import org.hibernate.validator.constraints.ParameterScriptAssert;
@@ -33,6 +35,17 @@ public interface HibernateValidatorFactory extends ValidatorFactory {
 	 */
 	@Incubating
 	ScriptEvaluatorFactory getScriptEvaluatorFactory();
+
+	/**
+	 * Returns clock skew tolerance as {@link Duration} i.e. the acceptable margin of error
+	 * when comparing date/time in time related constraints.
+	 *
+	 * @return a tolerance as {@link Duration}
+	 *
+	 * @since 6.0.5
+	 */
+	@Incubating
+	Duration getClockSkewTolerance();
 
 	/**
 	 * Returns a context for validator configuration via options from the
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidator.java
@@ -1,0 +1,43 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.constraintvalidation;
+
+import java.lang.annotation.Annotation;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.ConstraintValidatorFactory;
+import javax.validation.metadata.ConstraintDescriptor;
+
+import org.hibernate.validator.Incubating;
+
+/**
+ * Extends the contract of {@link ConstraintValidator} by adding {@link #initialize(ConstraintDescriptor, HibernateConstraintValidatorInitializationContext)}
+ * which gives more flexibility during initialization of a constraint validator.
+ *
+ * @author Marko Bekhta
+ * @since 6.1
+ */
+@Incubating
+public interface HibernateConstraintValidator<A extends Annotation, T> extends ConstraintValidator<A, T> {
+
+	/**
+	 * Initializes the validator in preparation for {@link #isValid(Object, ConstraintValidatorContext)} calls.
+	 * It is an alternative to {@link #initialize(Annotation)} method. Should be used if any additional information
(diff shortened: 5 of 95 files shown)


marko-bekhta:
Hi @gsmet !

I agree - having some additional context during init would be nice :)

As for your second suggestion - I was actually thinking about it as well (having an annotation instead of a class) but as it wouldn't allow to generalize these 4 cases for time constraints I thought I'll give class attribute a try instead, as there are no parameters to time constraints except the regular ones (groups, payload, message) which could be set for this wrapper instead. 
But I guess we could have 4 different ones and then just put the `@Tolerance` as a separate thing so it can be reused in all of them.

Also maybe as a completely different approach to this problem of  tolerance we could add another configuration property which would allow user to set a tolerance during `Validator` configuration (similar to having possibility to configure `Clock`). But it would apply to all time related constraints for that validator.

gunnarmorling:
I was leaning more towards the approach described by @gsmet first, i.e. have a separate `@SkewTolerance` annotation to be given next to `@Past` et al.

I don't think it's a big problem that one cannot refer to a specific constraint in that approach: a) I don't expect multiple temporal constraints on one element anyways, b) if there indeed are multiple ones, chances are they are using the same tolerance and c) people can always build their custom annotation. So I'd consider this good enough until we can add a property to the constraints themselves in a future spec revision.

The point of initialization is ugly. We could hack it using a thread local set and removed by `ConstraintValidatorManager` or indeed look for an extended `initialize()` contract. I don't think it'd be that much work actually:

    if ( validator implements HibernateConstraintValidator ) {
        ((HCV)validator).initialize( descriptor );
    }
    else {
        validator.initialize( descriptor.getAnnotation() );
    }

gsmet:
@gunnarmorling yeah I don't think there is a point about targeting a specific constraint for the current case but my idea was to generalize this mechanism so than we could extend existing constraints with additional annotations more easily as it's a good way to experiment additions for the future.

Maybe it's not worth it though.

As for having 2 `initialize()` methods, I was thinking about calling both in a specific order as it would be a good way to inject it in the spec later. I don't think we would be able to remove the existing contract in favor of the new one and we won't have a way to make a distinction between the ones implementing the first method and the ones implementing the new method. We could create a new contract for `ConstraintValidator` in the spec but good luck to find a name :).

gunnarmorling:
For the spec I think it wouldn't be difficult:

    default initialize(Annotation annotation, Context ctx) {
        initialize(annotation);
    }

A 2.1 validator could override that new method and access the context.

gsmet:
Ah OK, so we would keep the method but not call it directly in the impl?

I'm not really worried about the technical side of it but more on the spec level. E.g. how we would explain the 2 methods, and typically what happens when both are overridden (IIRC we cannot deprecate an API in the spec).

But I suppose we could make it work with very specific wording.

gunnarmorling:
> Ah OK, so we would keep the method but not call it directly in the impl?

Yes, exactly.

> But I suppose we could make it work with very specific wording.

I think so, too.

gsmet:
So my vision for this one is that we could add a generic feature to allow getting additional information for an annotation in other annotations.

So I would add a field called `complementaryAnnotationDescriptors` in `ConstraintDescriptorImpl` (and probably introduce an `HibernateValidatorConstraintDescriptor` interface to host the getter methods).

And I would allow to declare the complementary annotations in `ConstraintHelper` with the constraint annotations supported . Thus we could very easily extend Bean Validation annotations with additional annotation.

For now, we could be naive in the approach and attach all the complementary annotations to the annotations that declares it (i.e. adds the `@ClockSkewTolerance` annotation to all the `@Past` annotations present on the field).

In the future (or if is is not too much work, we could do it right away), we could imagine supporting a `constraint` and `constraintIndex` parameters to target a specific constraint (in the same vein of what we do with `@OverridesAttribute`).

And then we would need to think about what we could do for the XML support.

gunnarmorling:
@gsmet so would such auxilliary annotation be marked with the type it applies to:

    @ConstraintAmendment(appliesTo={ @Past.class, @Future.class })
    @interface ClockSkewTolerance {
        int seconds();
    }

?

gunnarmorling:
Do you see other use cases for such generic mechanism? If it's just that one, it may not be worth it. Not sure.

gsmet:
Yes. But for now, to be honest, I'm not sure I would have let the users play with it and declare new ones.

An hardcoded map of complementary annotation -> supported constraints would have been OK to me as a first step.

For now, I see it more as a way for us to easily extend a BV constraint before normalizing a new feature in BV.

gsmet:
Well, I would like to have it handy so that we can move quickly if another use case comes up.

And tbh, I think it will be easier to have a generic mechanism than to support it solely for this use case.

gunnarmorling:
Yeah, makes sense. +1 for just hard-coding the link between constraint type and auxilliary annotation.

gsmet:
@marko-bekhta so I see this one as a follow-up to https://github.com/hibernate/hibernate-validator/pull/867 as you will need to have the `ConstraintDescriptor` in the `initialize()` method.

@gunnarmorling I like the name `amendment`, updating my proposal to use it.

marko-bekhta:
@gsmet yes, that's why I wanted to push forward the conversation on the other one :smiley:

gsmet:
@marko-bekhta cool. I think my proposal above should be easy to implement (I expect the building of the ``ConstraintDescriptor``s to be a bit tricky though).

gsmet:
@marko-bekhta we discussed it a bit further with Gunnar this afternoon, and, in the end, it's probably better to add a `ValidatorFactory` level method to be able to set the clock skew tolerance globally.

Otherwise you would have to set it locally on each temporal annotation, which does not really makes sense.

This clock skew tolerance could be passed via the context to the new `initialize()` method. Probably best to make it a `Duration`. This would allow to use `plus()` and `minus()` directly on the java.time objects.

I still think the extension point I envisioned has value but we will have to wait for another use case to make it useful :). Probably better to let it slide for now and keep the idea for later.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -87,6 +92,15 @@
 	String SCRIPT_EVALUATOR_FACTORY_CLASSNAME = "hibernate.validator.script_evaluator_factory";
 
 	/**
+	 * Property for configuring clock skew tolerance, allowing to set a margin of error in milliseconds, which is allowed
+	 * when comparing date/time in time related constraints.
+	 *
+	 * @since 6.1
+	 */
+	@Incubating
+	String CLOCK_SKEW_TOLERANCE_DURATION = "hibernate.validator.clock_skew_tolerance_duration";
```
gsmet: Let's remove the `_DURATION`/`_duration` suffix.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -87,6 +92,15 @@
 	String SCRIPT_EVALUATOR_FACTORY_CLASSNAME = "hibernate.validator.script_evaluator_factory";
 
 	/**
+	 * Property for configuring clock skew tolerance, allowing to set a margin of error in milliseconds, which is allowed
+	 * when comparing date/time in time related constraints.
+	 *
+	 * @since 6.1
```
gsmet: I think we will introduce it in the next 6.0, so let's target 6.0.5.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -264,4 +278,17 @@
...
+
+	/**
+	 * Allows to set a margin error in milliseconds, which is allowed when comparing date/time in time related constraints
+	 * like {@link Past}/{@link PastOrPresent} and {@link Future}/{@link FutureOrPresent}.
+	 *
+	 * @param clockSkewTolerance the acceptable tolerance as {@link Duration}
+	 *
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 6.1
```
gsmet: Same here about 6.0.5.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -87,6 +92,15 @@
 	String SCRIPT_EVALUATOR_FACTORY_CLASSNAME = "hibernate.validator.script_evaluator_factory";
 
 	/**
+	 * Property for configuring clock skew tolerance, allowing to set a margin of error in milliseconds, which is allowed
+	 * when comparing date/time in time related constraints.
```
gsmet: Property for configuring clock skew tolerance, allowing to set the acceptable margin of error when comparing date/time in time related constraints. In milliseconds.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractInstantBasedTimeValidator.java
```diff
@@ -1,64 +0,0 @@
...
-
-		try {
-			ClockProvider clockProvider = context.getClockProvider();
-			reference = clockProvider.getClock();
-		}
-		catch (Exception e) {
-			throw LOG.getUnableToGetCurrentTimeFromClockProvider( e );
-		}
-
-		int result = getInstant( value ).compareTo( reference.instant() );
```
gsmet: Hmmm, I'm a bit worried about replacing all the validator implementations in a micro release.

In an earlier comment, I pointed you to `minus()` and `plus()`. I think we should simply adjust the reference instant using the clock skew tolerance (in + or - depending if we validate a past or a future constraint). All the hierarchy is already ready for that. We just need to add an abstract method implemented in the abstract Past/Future implementations.

That would limit the changes to the minimum. And I don't think it's less clean.

WDYT?

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -0,0 +1,61 @@
...
+	 * passed at bootstrap.
+	 *
+	 * @param languageName the name of the scripting language
+	 *
+	 * @return a script executor for the given language. Never null.
+	 *
+	 * @throws ScriptEvaluatorNotFoundException in case no {@link ScriptEvaluator} was
+	 * found for a given {@code languageName}
+	 */
+
```
gsmet: Useless new line.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -0,0 +1,61 @@
...
+	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName);
+
+	/**
+	 * Returns clock skew tolerance as {@link Duration} which is used to determine
+	 * acceptable margin of error in milliseconds, which is allowed
+	 * when comparing date/time in time related constraints.
+	 *
+	 * @return a tolerance as  {@link Duration}
+	 */
+	Duration getClockSkewTolerance();
```
gsmet: Thinking more about it, I think it should rather be part of `HibernateConstraintValidatorContext` and that we might allow to overload it on a context basis.

So, in the end, we don't need the initialization context for this PR at all.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -0,0 +1,61 @@
...
+	/**
+	 * Returns the provider for obtaining the current time in the form of a {@link Clock},
+	 * e.g. when validating the {@code Future} and {@code Past} constraints.
+	 *
+	 * @return the provider for obtaining the current time, never {@code null}. If no
+	 * specific provider has been configured during bootstrap, a default implementation using
+	 * the current system time and the current default time zone as returned by
+	 * {@link Clock#systemDefaultZone()} will be returned.
+	 */
+	ClockProvider getClockProvider();
```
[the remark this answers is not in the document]
marko-bekhta: hmm, my understanding was that `ClockProvider` got into `ConstraintValidatorContext` only because there was no other way to pass it to validators at init time. I was even thinking about caching `Clock` in validator during init rather than a provider ...

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractInstantBasedTimeValidator.java
marko-bekhta: yeah.. I've used those methods when I added this method:
```java
protected Temporal adjustCurrentTime(Temporal nowValue) {
	return isFuture() ? nowValue.plus( tolerance ) : nowValue.minus( tolerance );
}
```
These changes that I've made are definitely not for the micro release :)
Let me try the "keeping current hierarchy" approach and I'll get back :)

marko-bekhta: in a mean time here's one more of those "crazy" ideas :smile: 
What if instead of simple duration we pass something like:
```java
interface ToleranceSupplier {
	<A extends Annotation> Duration toleranceFor(Class timeClazz, Class<A> constraint);
}
```
by default we would use some
```java
public static class ConstantToleranceSupplier implements ToleranceSupplier{
	private final Duration tolerance;

	public ConstantToleranceSupplier(Duration tolerance) {
		this.tolerance = tolerance;
	}

	@Override
	public <A extends Annotation> Duration toleranceFor(Class timeClazz, Class<A> constraint) {
		return tolerance;
	}
}
```
which would create when a property is passed in xml config. And otherwise user would be able to somehow customize the tolerance based on a constraint and property type (for example +/- 1 day for `LocalDate` and +/- 10 sec for a `LocalDateTime`)

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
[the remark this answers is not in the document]
gsmet: Hmmm. It sure is appealing to move the `ClockProvider` to the initialization context and caching the `Clock` would be a good idea in this case.

But I don't think we would be able to move it to the initialization context while following the spec.

And I'm not even sure we would be able to find a way to change the spec to support it in the future. As we would still have a `ClockProvider` in the `ConstraintValidatorContext`.

@gunnarmorling what's your take on this one?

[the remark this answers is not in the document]
mapstruct-website-builder: I'm not following on the caching idea. You want to query what "now" is each time when validating the constraint and not when initialising it.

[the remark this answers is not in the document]
gunnarmorling: Oops, sorry, was replying from wrong browser with wrong user session :)

gsmet:
@marko-bekhta btw, I don't know if you noticed it but the TCK tests related to the `ClockProvider` are broken.

See http://ci.hibernate.org/job/hibernate-validator-PR/907/ .

To run a specific TCK test, you need to use the `TCKRunner` class, remove the package filter, change the class filter to point to your class, then Run as a Java application.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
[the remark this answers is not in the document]
gsmet: A `Clock` gets you a new now every time you ask for it. At least that's my understanding (expect if you use a fixed one of course).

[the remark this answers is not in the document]
gunnarmorling: Interesting, I always perceived `Clock` as a representation of a single instant, but it seems I was wrong. Yeah, caching would work then. It seems not like something though we could do right now.

marko-bekhta:
@gsmet I've made some changes. 
as MonthDay is not Temporal there's no plus/minus on it so still needed to introduce one more abstract class for each constraint  for `Temporal`. Also as there no plus minus what I've done instead is:
```java
	@Override
	protected MonthDay getReferenceValue(Clock reference) {
		LocalDateTime now = LocalDateTime.now( reference ).minus( tolerance );
		return MonthDay.of( now.getMonth(), now.getDayOfMonth() );
	}
```
WDYT ?
As for the TCK tests failing - I've seen that and it actually was the same problem that there's now with the simple test that I've added `ClockToleranceTest`. It's throwing stuff like:
```
...
Caused by: java.time.temporal.UnsupportedTemporalTypeException: Unsupported unit: Seconds
	at java.time.chrono.ChronoLocalDateImpl.plus(ChronoLocalDateImpl.java:211)
	at java.time.chrono.JapaneseDate.plus(JapaneseDate.java:631)
```
So before going forward I wanted to confirm. It seems that for all those types without time part we would need to do something similar to what I've done for `MonthDay`. Is that fine ?

Also I didn't do any changes in initialization context part - as they are in a different PR and branch. I've started this one based on the HV-1497 one, so I'm not sure whether I should make changes here (or rather in that other PR and when/if it is merged, do a rebase if we would still want to use init context here). 
And as from the discussion above it doesn't seem that we'd not be able to cache clock, there seems no reason to use initContext in such case. I'll rework the first commit then to use `HibernateConstraintValidatorContext` instead, a bit later today.

gsmet:
Yeah, I just came across the test issue when I was thinking about your proposal of having a class to provide a different tolerance depending on the class.

The fact is that `LocalDate.minus(Duration)` won't work at all. You would have to use a `Period` instead of a `Duration` in this case...

The java.time API is really hard to manipulate as you have a lot of API calls that seem to work but are not supported at all. It's really a minefield.

I initially thought this would be easy and beneficial to HV as we would have contributed a nice extension mechanism.

Now, we see it requires a lot of work, it's a minefield and it doesn't contribute the extension mechanism I thought would be nice.

I see three outcomes for this one:
1. either we support this mechanism only for java.time types for which it makes sense. E.g. this mechanism wouldn't support `LocalDate`, `MonthDay` and so on. If you want a clock skew tolerance to be supported there, you would have to declare your own validators (and decide if it's OK to have a tolerance of full day/month). It wouldn't be very consistent though but maybe documenting the supported types would be good enough.
2. just drop the idea and recommend the OP to declare his own validators which introduce a clock skew tolerance. Maybe we could document an example.
3. provide the extension point you were drafting above but I think it will be very hard for people to get it right. E.g. the API would return a `TemporalAmount` and people would have to guess which `TemporalAmount` type will work with each java.time types.

For now I think I would lean towards 2 (I know it will be a disappointment for you, sorry about that). @gunnarmorling WDYT?

marko-bekhta:
yep... If it were up to me I'd be choosing between 2 and 3 in this case. For the extension we could think about different approaches - one can be based on allowing to provide different tolerances based on type, mentioned earlier.
Another interesting option could be to expose the way we compare objects. This extension can go beyond just java.time types and for example can be used when floats/doubles are compared (as they are the same kind of types that somebody would want to compare with some tolerance ...). Or maybe you have some other ides for extensions.
And it's not a disappointment. Some work was done, and if the result just shows that the idea didn't work out as expected, it's still ok - somebody should test out the ideas as well right ? 😃

gsmet:
OK, let's close it then. Maybe we will find a better idea in the future :). Thanks for working on this one!

danielwegener:
regarding the initial PR (and its complexity regarding different time and calendar types). have you thought about adjusting the reference clock using https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html#offset-java.time.Clock-java.time.Duration- instead of adjusting the reference value (which is, as you mentioned, not always that easy)?

gsmet:
@danielwegener 

Good catch this `offset()` method, I missed it. It solves the java.time API issue but it does not solve the biggest issue we have: you still need a different precision for different types. E.g. having a tolerance of 5 ms won't help you with your `YearMonth` which can be one month off if you're not lucky.

It's an extreme example but `LocalDate` for instance might be a more common case.

The API proposed by Marko allowing to return a different `Duration` depending on the type tries to solves this issue but I'm not sure it's worth it. What's your opinion on this?

Your proposal solves the tricky part for the user, though (e.g. getting a temporal amount compatible with each type).

danielwegener:
Okay maybe let me restate my original problem then :)
The original proposal was about adding a **skew tolerance** so a tolerance to compensate clock drift in distributed systems. The whole request was not meant as new generic way to parameterize existing constraints but rather as addition to the *already existing specially handled case `ClockProvider`* to make it works for distributed scenarios (which all the cool micro- and macro-service-kids are doing).

Consider the following example:
**Given HibernateValidator is configured with a skewTolerance of 5 minutes**
```java
@PastOrNow YearMonth goodTime;
```
`A` and `B` have a clock drift of 6 minutes (possibly not realistic, just for clarification)

1. **A** with LocalTime: `00:00:00 01. Feb. .2017` would pick `YearMonth.now()` with `{"goodTime": YearMonth.of(2017, Month.FEBRUARY`} and send it to **B**
2. **B with LocalTime 23:54:00 31.Jan. 2017 fails to validate because it is not Feb. 2017 yet on his local clock (not even by shifting its clock 5 minutes in the future)

3. **C** with LocalTime: 00:04:00 01. Feb. .2017 would pick YearMonth.now() with `{"goodTime": YearMonth.of(2017, Month.FEBRUARY`} and send it to **B**
2. **B with LocalTime 23:58:00 31.Jan. 2017 validates it successfully (by shifting its clock 5 minutes into the future and check if the received YearMonth ist still "PastOrNow")

This resolves to: For 5 minutes, it is okay to be one month ahead. And this is exactly what I'd expect, regardless the granularity of the field. The whole skey topic applies whenever the validator compares something with the wall-clocks `now` (which has nothing todo with the granularity in itself).

In my opinion, one global skewTolerance duration parameter is enough, although it would be nice if one could optionally overwrite it with a new annotationParameter (which still, would always be a `Duration` (mesurable offset in seconds) and not a `Period` (calendar units)).

Some rather special cases would be to compine multiple half-open constraints like `@PastOrPresent` with `@FutureOrPresent` which should work like a `@Present` with a tolerance into both directions.

gsmet:
Hmmm, thinking more about it, moving the clock reference might indeed make it work.

I'll take a shot at it this afternoon to see how it goes.

marko-bekhta:
@gsmet here's an attempt on that approach - https://github.com/marko-bekhta/hibernate-validator/commit/54abe175338578677b94e17c510b1749078d6803 
`ClockToleranceTest` pass with the changes

PR #862, opened 2017-10-18
HV-1494 hibernate validator error when @NotEmpty used on return type

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1494

Previous implementation were not giving a chance for composite constraints to execute if they were placed on a method with non void return type and parameters - as for those both `hasGenericValidators` and `hasCrossParameterValidator` where `false` and there's no way to set `validationAppliesTo` so `ConstraintDescriptorImpl#determineConstraintType` was just skipping to:

```java
// Now we are out of luck
if ( constraintType == null ) {
  throw LOG.getImplicitConstraintTargetInAmbiguousConfigurationException( annotationType );
}
```
hence there was no way to use them on methods with both return value and parameters....

Let me know what you think about these changes :)

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/descriptor/ConstraintDescriptorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/descriptor/ConstraintDescriptorImpl.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.validation.Constraint;
 import javax.validation.ConstraintTarget;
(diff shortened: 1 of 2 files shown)


marko-bekhta:
@gsmet sooo you were asking about this one in JIRA. Here's what I have atm.

marko-bekhta:
closing this one in favor of https://github.com/hibernate/hibernate-validator/pull/863

engine/src/main/java/org/hibernate/validator/internal/metadata/descriptor/ConstraintDescriptorImpl.java
```diff
@@ -474,9 +479,31 @@ else if ( hasParameters && !hasReturnValue ) {
 			}
 		}
 
-		// Now we are out of luck
 		if ( constraintType == null ) {
-			throw LOG.getImplicitConstraintTargetInAmbiguousConfigurationException( annotationType );
+			// check if it's a composite constraint and is built with constraints for which a type can be detected
+			Set<ConstraintType> constraintTypes = findComposingConstraints( constraintHelper )
```
gunnarmorling: We already call `parseComposingConstraints()`; could we somehow combine this, instead of running the annotation tree another time?

marko-bekhta: Ok, I'll see what can be done about it and open another PR if something good comes out.

PR #863, opened 2017-10-18
HV-1494 Hibernate Validator specific @NotEmpty used on return type throws an exception

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1494

I found that only `NotEmpty` `Range` and `TituloEleitoral` were the composite constraints without validators. So I've added `SupportedValidationTarget` for them.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraints/NotEmpty.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/NotEmpty.java
@@ -24,6 +24,8 @@ import javax.validation.Payload;
 import javax.validation.ReportAsSingleViolation;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Size;
+import javax.validation.constraintvalidation.SupportedValidationTarget;
+import javax.validation.constraintvalidation.ValidationTarget;
 
 import org.hibernate.validator.constraints.NotEmpty.List;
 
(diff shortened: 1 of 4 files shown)


gsmet:
@marko-bekhta you removed the rest of your changes? How does it work for custom composite constraints?

Note that we can require that users define `SupportedValidationTarget` on them.

What you did in this PR is what Gunnar originally had in mind but I'm curious on why you removed the rest of your changes?

marko-bekhta:
@gsmet well they are in that other branch (in the closed PR, I left them there if you'd like to check those, as otherwise they wouldn't be visible if I update the commit and force push the changes right?)
And this one is a clean one. With just what Gunnar was suggesting. 

I just thought that it was an over-complicated solution (because I didn't know about `SupportedValidationTarget`). For the custom ones - there's one in the test. If `SupportedValidationTarget` is placed on it - all works fine. 

So I thought if there's an existing solution to the problem - it'll be better to use it rather than make the `ConstraintDescriptorImpl#determineConstraintType` more complicated ...

gsmet:
@marko-bekhta sure. And this one will be easily backportable too. Thanks for that!

gunnarmorling:
That looks good +1. We can still go for the more automated route in a next release, only see my comment about combining the annotation tree traversals if possible in the closed PR.

gsmet:
Merged with a few minor changes I included in your commit.

gsmet:
Backported to 5.4 and 5.3.

PR #864, opened 2017-10-18
HV-1479 Describing requirement for java.xml.bind module in the docume…

gunnarmorling:
…ntation

https://hibernate.atlassian.net/browse/HV-1479


gsmet:
Merged, thanks!

PR #866, opened 2017-10-20
HV-1373 Validator calls objects hashCode() after failed @NotNull validation

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1373

It turned out that those 3 `*.hashcode()`calls were potentially dangerous as any of them could throw an exception - so I replaced all of them with `System#identityHashCode()`


gsmet:
Merged, thanks!

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -327,9 +327,9 @@ public String toString() {
 	private int createHashCode() {
 		int result = interpolatedMessage != null ? interpolatedMessage.hashCode() : 0;
```
gunnarmorling: Not strictly related, but it seems we could get rid of a few null checks here, for all the attributes we know will exist?

PR #867, opened 2017-10-21
HV-1497 Extend the ConstraintValidator#initialize() contract

marko-bekhta:
Hi @gunnarmorling , @gsmet so I wanted to finish that other task for time tolerance constraints before moving to the other ones. I liked the idea of having a possibility to pass more information to init method of constraint validator. After thinking about it - this idea of callbacks came to my mind. What if instead of adding another method for initialization, which we initially were thinking about, we add a "callback" methods like postInit and maybe preInit. At the moment I cannot think of a good use case for a preInit though :).

I think it can be a good alternative to our initial approach. This would definitely solve the problem of having multiple init methods which Gunnar mentioned and would be probably much easier to add to BV later. 

So I made the changes in this PR (just for the postInit) and updated the `ScriptAssert` constraints as an example of using such postInit callback.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ParameterScriptAssertValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ParameterScriptAssertValidator.java
@@ -15,11 +15,12 @@ import java.util.Map;
 import javax.validation.ConstraintValidatorContext;
 import javax.validation.constraintvalidation.SupportedValidationTarget;
 import javax.validation.constraintvalidation.ValidationTarget;
+import javax.validation.metadata.ConstraintDescriptor;
 
 import org.hibernate.validator.constraints.ParameterScriptAssert;
 import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorContext;
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorInitializationContext;
 import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorContextImpl;
-import org.hibernate.validator.internal.engine.messageinterpolation.util.InterpolationHelper;
 import org.hibernate.validator.internal.util.Contracts;
 
 /**
@@ -33,11 +34,10 @@ import org.hibernate.validator.internal.util.Contracts;
 public class ParameterScriptAssertValidator extends AbstractScriptAssertValidator<ParameterScriptAssert, Object[]> {
 
 	@Override
-	public void initialize(ParameterScriptAssert constraintAnnotation) {
+	public void initialize(ConstraintDescriptor<ParameterScriptAssert> constraintDescriptor, HibernateConstraintValidatorInitializationContext initializationContext) {
+		ParameterScriptAssert constraintAnnotation = constraintDescriptor.getAnnotation();
 		validateParameters( constraintAnnotation );
-		this.languageName = constraintAnnotation.lang();
-		this.script = constraintAnnotation.script();
-		this.escapedScript = InterpolationHelper.escapeMessageParameter( constraintAnnotation.script() );
+		initialize( constraintAnnotation.lang(), constraintAnnotation.script(), initializationContext );
 	}
 
 	@Override
@@ -51,7 +51,7 @@ public class ParameterScriptAssertValidator extends AbstractScriptAssertValidato
 
 		Map<String, Object> bindings = getBindings( arguments, parameterNames );
 
-		return getScriptAssertContext( constraintValidatorContext ).evaluateScriptAssertExpression( bindings );
+		return scriptAssertContext.evaluateScriptAssertExpression( bindings );
 	}
 
 	private Map<String, Object> getBindings(Object[] arguments, List<String> parameterNames) {
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
@@ -22,7 +22,6 @@ import java.util.stream.Collectors;
 
 import javax.validation.ConstraintDeclarationException;
 import javax.validation.ConstraintValidator;
-import javax.validation.ConstraintValidatorFactory;
 import javax.validation.ConstraintViolation;
 import javax.validation.ValidationException;
 
@@ -61,7 +60,7 @@ public class ConstraintTree<A extends Annotation> {
 
 	/**
 	 * Either the initialized constraint validator for the default constraint validator factory or
-	 * {@link #DUMMY_CONSTRAINT_VALIDATOR}.
+	 * {@link ConstraintValidatorManager#DUMMY_CONSTRAINT_VALIDATOR}.
 	 */
 	private volatile ConstraintValidator<A, ?> constraintValidatorForDefaultConstraintValidatorFactory;
 
@@ -119,7 +118,6 @@ public class ConstraintTree<A extends Annotation> {
 			ConstraintValidatorContextImpl constraintValidatorContext = new ConstraintValidatorContextImpl(
 					validationContext.getParameterNames(),
 					validationContext.getClockProvider(),
-					validationContext.getScriptEvaluatorFactory(),
 					valueContext.getPropertyPath(),
 					descriptor
 			);
@@ -184,8 +182,7 @@ public class ConstraintTree<A extends Annotation> {
 				synchronized ( this ) {
 					validator = constraintValidatorForDefaultConstraintValidatorFactory;
 					if ( validator == null ) {
-						validator = getInitializedConstraintValidator( validationContext.getConstraintValidatorManager(),
-								validationContext.getConstraintValidatorFactory() );
+						validator = getInitializedConstraintValidator( validationContext );
 						constraintValidatorForDefaultConstraintValidatorFactory = validator;
 					}
 				}
@@ -196,8 +193,7 @@ public class ConstraintTree<A extends Annotation> {
 			// factory. Creating a lot of CHM for that cache might not be a good idea and we prefer being conservative
 			// for now. Note that we have the ConstraintValidatorManager cache that mitigates the situation.
 			// If you come up with a use case where it makes sense, please reach out to us.
-			validator = getInitializedConstraintValidator( validationContext.getConstraintValidatorManager(),
-					validationContext.getConstraintValidatorFactory() );
+			validator = getInitializedConstraintValidator( validationContext );
 		}
 
 		if ( validator == DUMMY_CONSTRAINT_VALIDATOR ) {
@@ -208,12 +204,13 @@ public class ConstraintTree<A extends Annotation> {
 	}
 
 	@SuppressWarnings("unchecked")
-	private ConstraintValidator<A, ?> getInitializedConstraintValidator(ConstraintValidatorManager constraintValidatorManager,
-			ConstraintValidatorFactory constraintValidatorFactory) {
-		ConstraintValidator<A, ?> validator = constraintValidatorManager.getInitializedValidator(
+	private ConstraintValidator<A, ?> getInitializedConstraintValidator(ValidationContext<?> validationContext) {
+		ConstraintValidator<A, ?> validator = validationContext.getConstraintValidatorManager().getInitializedValidator(
 				validatedValueType,
 				descriptor,
-				constraintValidatorFactory );
+				validationContext.getConstraintValidatorFactory(),
+				validationContext.getConstraintValidatorInitializationContext()
+		);
 
 		if ( validator != null ) {
 			return validator;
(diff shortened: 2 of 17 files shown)


gunnarmorling:
Hi, @marko-bekhta, I don't think it's significantly different from what we had discussed before. In fact, I believe "call back" is not the correct term here, rather "lifecycle method". A call back would typically receive data from the method it was passed to, the receiver in `ValueExtractor` is a good example.

Apart from that the design seems reasonable. Alternatives I could think of would be injection into fields with some marker annotation (i.e. some very basic DI) or support for constructors with a defined signature instead of the no-arg one. The latter is quite elegant, but it wouldn't work with custom constraint validator factories, so not really viable unfortunately. A new `initialize()` method (whatever the name would be), seems like the best we can do at the level of the RI.

gsmet:
In fact, I was pondering passing the `ConstraintDescriptor` to the `initialize()` method instead of just the annotation.

If we do so, it wouldn't make sense to have a `postInitialize()` method and we'd rather have a new `initialize()` with a default implementation that calls the old `initialize()` version as Gunnar suggested it initially.

marko-bekhta:
Lifecycle method is definitely a more appropriate name for it :) and yes it's basically similar to what was previously discussed about the `initialize()` method. The difference is just in the naming of the method itself. I just think that having two `initialize()` methods might be more confusing - which one should be overridden? are both of them called? and so on. In case of `postInitialize()` the name itself would give the user the idea what that method is doing and we will know that both methods are called.

@gsmet I remember that you were suggesting passing `ConstraintDescriptor` and I also though about it. I'm just not sure if that would be enough. I might be wrong here, so please feel free to correct me :) I was thinking that having just `ConstraintDescriptor` would be too limiting. I mean it wouldn't allow to pass something like `ScriptEvaluatorFactory` or similar, but some `InitializationContext` on the other hand would be more flexible. We could for example add another method to that context like:

```java
public interface HibernateConstraintValidatorInitializationContext {
	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName);

	ConstraintDescriptor getCurrentConstraintDescriptor();
}
```
And maybe have the `HibernateConstraintValidatorInitializationContext#unwrap` method as there are in other interfaces in BV. This way I think it'll be easier to put it into BV later, and it would also be flexible enough for us to modify the context to add whatever we would need in the future.

gsmet:
@marko-bekhta I want the `ConstraintDescriptor` **and** the context :). So it's orthogonal to what you did but leans towards having another `initialize()` method instead of an additional one.

One `initialize()` method would be legacy. We won't be allowed to mark it as deprecated unfortunately but that's how it is.

You would be able to use either of them depending on what you want but not both. I think it's easy enough to understand.

gsmet:
Btw, it might not be the best moment to introduce this new method (especially if you're thinking about updating the tests to use this new method) as I have an ongoing PR touching a lot of things in this area.

marko-bekhta:
Hi @gsmet I guess that PR that you mentioned was the performance improvements, and is released now, right ? :) 
How should I proceed with this one? Leave this:
```java
public interface HibernateConstraintValidatorInitializationContext {
	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName);

	ConstraintDescriptor getCurrentConstraintDescriptor();
}

public interface HibernateConstraintValidator<A extends Annotation, T> extends ConstraintValidator<A, T> {
	default void postInitialize(HibernateConstraintValidatorInitializationContext initializationContext) {
	}
}
```
Or instead make it like this:
```java
public interface HibernateConstraintValidatorInitializationContext {
	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName);
}

public interface HibernateConstraintValidator<A extends Annotation, T> extends ConstraintValidator<A, T> {

	default void initialize(ConstraintDescriptor descriptor, HibernateConstraintValidatorInitializationContext initializationContext) {
	}
}
```
In second case should there be any additional checks to make sure that only one method is overridden, or we just leave it for the users to implement validators correctly and simply call both methods all the time ?

I still, kind of, like the postInit method more, as it looks to have less logic applied to it (no checks or restrictions, just call both methods). But I'm happy to do it the other way as well :smiley:  just want to be sure that we agreed on the approach first, so that I have something to work on a weekend :grin:

gsmet:
@marko-bekhta yes, the PR has been merged.

So my vote goes to the second one with this addition:
```java
public interface HibernateConstraintValidator<A extends Annotation, T> extends ConstraintValidator<A, T> {

	default void initialize(ConstraintDescriptor constraintDescriptor, HibernateConstraintValidatorInitializationContext initializationContext) {
                initialize( constraintDescriptor.getAnnotation() );
	}
}
```

So in the case of an `HibernateConstraintValidator`, we would only call the new `initialize()` method.

What I'm a bit worried about is that I think if we want to generalize this change, it will generate a lot of noise in the tests. At the moment, in quite a lot of tests, we initialize an annotation and we pass it to the validator `initialize()` method. With the new contract, we will have to initialize a full `ConstraintDescriptorImpl`. We will probably need a new test helper to reduce the noise as much as possible.

marko-bekhta:
rebased to latest master, let me know if anything is missing

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
```diff
@@ -208,12 +204,13 @@ private ValidationException getExceptionForNullValidator(Type validatedValueType
...
-	private ConstraintValidator<A, ?> getInitializedConstraintValidator(ConstraintValidatorManager constraintValidatorManager,
-			ConstraintValidatorFactory constraintValidatorFactory) {
-		ConstraintValidator<A, ?> validator = constraintValidatorManager.getInitializedValidator(
+	private ConstraintValidator<A, ?> getInitializedConstraintValidator(ValidationContext<?> validationContext) {
+		ConstraintValidator<A, ?> validator = validationContext.getConstraintValidatorManager().getInitializedValidator(
 				validatedValueType,
 				descriptor,
-				constraintValidatorFactory );
+				validationContext.getConstraintValidatorFactory(),
+				HibernateConstraintValidatorInitializationContextImpl.from( validationContext )
```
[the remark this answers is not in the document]
gsmet: I gave it some more thoughts and I think we should create the context at the `ValidatorImpl` level once and for all for now, and pass it to the `ValidationContext`. This would avoid creating a new object for each validation.

It might be necessary in the future but while it is not, I think it's worth avoiding the object creation.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ParameterScriptAssertValidator.java
```diff
@@ -33,11 +35,15 @@
...
 
 	@Override
-	public void initialize(ParameterScriptAssert constraintAnnotation) {
+	public void initialize(ConstraintDescriptor<ParameterScriptAssert> constraintDescriptor, HibernateConstraintValidatorInitializationContext initializationContext) {
+		ParameterScriptAssert constraintAnnotation = constraintDescriptor.getAnnotation();
 		validateParameters( constraintAnnotation );
+
 		this.languageName = constraintAnnotation.lang();
 		this.script = constraintAnnotation.script();
 		this.escapedScript = InterpolationHelper.escapeMessageParameter( constraintAnnotation.script() );
```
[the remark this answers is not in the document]
marko-bekhta: +1 was thinking about it as well, but then something distracted me :)

marko-bekhta:
@gsmet I think I've addressed all the comments. As for the initContext at `ValidatorImpl` level - I've added a filed for it there and then passed it to `ValidationContext` not to update all the methods down to a place where the context is actually needed. Also I've found that in `ValidationContext` there was unused reference to `ScriptEvaluatorFactory` so I just removed it (replaced with the init context really 😃)

marko-bekhta:
Also while working on this PR as well as the one about clock tolerance and previously the initial PR for `ScriptEvaluatorFactory` I've "discovered" that things like `ClockProvider`, `ScriptEvaluatorFactory`, failFast and probably some others appear all over the code and are passed to multiple different methods. I've noticed this as when these new parameters are added those method signatures are getting bigger and bigger. Maybe these things can be grouped together somehow and passed as one object ?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
```diff
@@ -223,11 +233,22 @@ public int numberOfCachedConstraintValidatorInstances() {
...
+				( (HibernateConstraintValidator) constraintValidator ).initialize( descriptor, initializationContext );
+			}
+			else {
+				constraintValidator.initialize( descriptor.getAnnotation() );
+			}
 		}
 		catch (RuntimeException e) {
+			if ( e instanceof ConstraintDeclarationException ) {
+				throw e;
+			}
```
gsmet: @marko-bekhta I just noticed this change. What does it have to do with this patch?

gsmet:
@marko-bekhta I added a commit to improve a few minor things. It looks good to me except for my question about the exception management code you added.

Thanks for your feedback!

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
marko-bekhta: it's from those script assert constraints. In `ScriptAssertValidatorTest#unknownLanguageNameRaisesException()`  `ConstraintDeclarationException` is expected when no language can be found. If run that test without this change we'd get:
```
javax.validation.ValidationException: HV000032: Unable to initialize org.hibernate.validator.internal.constraintvalidators.hv.ScriptAssertValidator.
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager.initializeValidator(ConstraintValidatorManager.java:258)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager.createAndInitializeValidator(ConstraintValidatorManager.java:170)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager.getInitializedValidator(ConstraintValidatorManager.java:126)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.getInitializedConstraintValidator(ConstraintTree.java:208)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.getInitializedConstraintValidator(ConstraintTree.java:185)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:115)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintTree.validateConstraints(ConstraintTree.java:86)
	at org.hibernate.validator.internal.metadata.core.MetaConstraint.doValidateConstraint(MetaConstraint.java:116)
	at org.hibernate.validator.internal.metadata.core.MetaConstraint.validateConstraint(MetaConstraint.java:109)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateMetaConstraint(ValidatorImpl.java:570)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForSingleDefaultGroupElement(ValidatorImpl.java:528)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForDefaultGroup(ValidatorImpl.java:497)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateConstraintsForCurrentGroup(ValidatorImpl.java:462)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validateInContext(ValidatorImpl.java:412)
	at org.hibernate.validator.internal.engine.ValidatorImpl.validate(ValidatorImpl.java:200)
	at org.hibernate.validator.test.internal.constraintvalidators.hv.ScriptAssertValidatorTest.unknownLanguageNameRaisesException(ScriptAssertValidatorTest.java:111)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:80)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:714)
	... 17 more
Caused by: javax.validation.ConstraintDeclarationException: HV000170: No JSR-223 scripting engine could be bootstrapped for language "foo".
	at org.hibernate.validator.internal.constraintvalidators.hv.AbstractScriptAssertValidator.initializeScriptContext(AbstractScriptAssertValidator.java:37)
	at org.hibernate.validator.internal.constraintvalidators.hv.ScriptAssertValidator.initialize(ScriptAssertValidator.java:48)
	at org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager.initializeValidator(ConstraintValidatorManager.java:248)
	... 38 more
Caused by: org.hibernate.validator.spi.scripting.ScriptEvaluatorNotFoundException: HV000232: No JSR 223 script engine found for language "foo".
	at org.hibernate.validator.internal.engine.scripting.DefaultScriptEvaluatorFactory.createNewScriptEvaluator(DefaultScriptEvaluatorFactory.java:66)
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1660)
	at org.hibernate.validator.spi.scripting.AbstractCachingScriptEvaluatorFactory.getScriptEvaluatorByLanguageName(AbstractCachingScriptEvaluatorFactory.java:41)
	at org.hibernate.validator.internal.engine.constraintvalidation.HibernateConstraintValidatorInitializationContextImpl.getScriptEvaluatorForLanguage(HibernateConstraintValidatorInitializationContextImpl.java:43)
	at org.hibernate.validator.internal.constraintvalidators.hv.AbstractScriptAssertValidator.initializeScriptContext(AbstractScriptAssertValidator.java:33)
	... 40 more
```

And personally I'd be OK with such error as the cause is clear, but I didn't want to change the behavior of which exception is thrown (maybe somebody is depending on it ?) WDYT?

gsmet: OK, makes sense.

gsmet:
Merged, thanks!

PR #868, opened 2017-10-24
Another set of performance related improvements

gsmet:
* https://hibernate.atlassian.net/browse/HV-1501
 * https://hibernate.atlassian.net/browse/HV-1502
 * https://hibernate.atlassian.net/browse/HV-1503

Bean Validation benchmark (lower is better, average of 2 runs):
 * 5.4.2.Final: 22938 ms
 * 6.0.3.Final: 10488.5 ms
 * 6.0.4-SNAPSHOT: 7418.5 ms


engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
```diff
@@ -56,21 +62,27 @@
 
 	private final Type validatedValueType;
 
-	public ConstraintTree(ConstraintDescriptorImpl<A> descriptor, Type validatedValueType) {
-		this( descriptor, validatedValueType, null );
+	private volatile Optional<ConstraintValidator<A, ?>> constraintValidatorForDefaultConstraintValidatorFactory;
```
gunnarmorling: What's the advantage of using `Optional`? It's not meant to be used as field really, using a nullable reference also saves the bytes of that wrapper.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
```diff
@@ -56,21 +62,27 @@
...
+
+	private volatile ConcurrentMap<ConstraintValidatorFactory, Optional<ConstraintValidator<A, ?>>> constraintValidatorPerConstraintValidatorFactory;
+
+	public ConstraintTree(ConstraintValidatorManager constraintValidatorManager, ConstraintDescriptorImpl<A> descriptor, Type validatedValueType) {
+		this( constraintValidatorManager, descriptor, validatedValueType, null );
 	}
 
-	private ConstraintTree(ConstraintDescriptorImpl<A> descriptor, Type validatedValueType, ConstraintTree<?> parent) {
+	private ConstraintTree(ConstraintValidatorManager constraintValidatorManager, ConstraintDescriptorImpl<A> descriptor, Type validatedValueType,
```
gunnarmorling: The `ConstraintValidatorManager` isn't really needed here, it's just passed forth and back.

gunnarmorling: I've just pushed a commit to remove it, it rippled through the entire change. Wondering whether it really isn't needed or you just "forgot" to make use of it here? Either way, feel free to include or drop that commit as needed.

gsmet: Nope, I needed it in the constructor before but you're right it's not needed anymore. Instead of undoing it and as we agreed on the volatile usage, I'll just remove the commit introducing it (it was a separate one and things should be OK).

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
gsmet: We need to make a difference between not yet defined and null. So `Optional` fits very well for this. And for the CHM, we have the same issue as we can't put null in the map.

Another solution is to use a dummy constraint validator, thus avoiding the wrapper but tbh I find it more readable with the `Optional`.

engine/src/main/java/org/hibernate/validator/cfg/AnnotationDef.java
```diff
@@ -110,14 +109,14 @@ protected C addAnnotationAsParameter(String key, AnnotationDef<?, ?> value) {
 		return getThis();
 	}
 
-	protected A createAnnotationProxy() {
-		AnnotationDescriptor<A> annotationDescriptor = new AnnotationDescriptor<>( annotationType );
+	protected AnnotationDescriptor<A> createAnnotationDescriptor() {
+		AnnotationDescriptor.Builder<A> annotationDescriptorBuilder = new AnnotationDescriptor.Builder<>( annotationType );
```
gunnarmorling: With that nice builder in place, did you think about using the builder as field directly and get rid of some/all of the intermediary maps?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
```diff
@@ -284,7 +285,7 @@ public boolean equals(Object o) {
 
 			CacheKey cacheKey = (CacheKey) o;
 
-			if ( annotation != null ? !annotation.equals( cacheKey.annotation ) : cacheKey.annotation != null ) {
+			if ( annotationDescriptor != null ? !annotationDescriptor.equals( cacheKey.annotationDescriptor ) : cacheKey.annotationDescriptor != null ) {
```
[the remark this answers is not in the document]
gsmet: `constraintValidatorFactory` cannot be null either so I simplified everything.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
```diff
@@ -148,9 +151,9 @@ private ConstraintTree(ConstraintDescriptorImpl<A> descriptor, Type validatedVal
 		}
 	}
 
-	private void throwExceptionForNullValidator(Type validatedValueType, String path) {
+	private ValidationException getExceptionForNullValidator(Type validatedValueType, String path) {
 		if ( descriptor.getConstraintType() == ConstraintDescriptorImpl.ConstraintType.CROSS_PARAMETER ) {
-			throw log.getValidatorForCrossParameterConstraintMustEitherValidateObjectOrObjectArrayException(
```
gunnarmorling: What's the motivation for throwing it separately instead doing it here? I think stack traces are a tiny bit confusing in this case due to line number mismatches, so I'd only do it if there's strong advantage.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
```diff
@@ -56,12 +60,11 @@
...
 	private final Type validatedValueType;
 
-	public ConstraintTree(ConstraintDescriptorImpl<A> descriptor, Type validatedValueType) {
-		this( descriptor, validatedValueType, null );
-	}
+	private volatile Optional<ConstraintValidator<A, ?>> constraintValidatorForDefaultConstraintValidatorFactory;
 
-	private ConstraintTree(ConstraintDescriptorImpl<A> descriptor, Type validatedValueType, ConstraintTree<?> parent) {
-		this.parent = parent;
+	private volatile ConcurrentMap<ConstraintValidatorFactory, Optional<ConstraintValidator<A, ?>>> constraintValidatorPerConstraintValidatorFactory;
```
gunnarmorling: Similar to the `Optional` above, what's the reason for using it as map value here?

gsmet:
@gunnarmorling I reorganized the commits to remove the noise.

The only thing that still requires discussion is the `Optional` usage. Just let me know if you feel strongly about it and I'll move to the dummy constraint validator approach.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
gunnarmorling: Sounds good, thanks for squashing it in!

engine/src/main/java/org/hibernate/validator/internal/util/annotation/AnnotationDescriptor.java
```diff
@@ -0,0 +1,286 @@
...
+	private int buildHashCode() {
+		int hashCode = 0;
+
+		for ( Entry<String, Object> member : attributes.entrySet() ) {
+			Object value = member.getValue();
+
+			int nameHashCode = member.getKey().hashCode();
+
+			int valueHashCode =
+					!value.getClass().isArray() ? value.hashCode() :
```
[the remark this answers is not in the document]
gsmet: It's coming from another class: `AnnotationProxy`, I just copied it (I don't think reindenting this is worth the time spent on it).

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
gsmet: At one point I needed it because I used `Optional.orElseThrow` but it's not strictly needed anymore.

That being said, I kept it because I like when the compiler is able to follow the flow and if we do the throw in another method it is not. So, I'd really prefer to keep it this way.

engine/src/main/java/org/hibernate/validator/cfg/AnnotationDef.java
gsmet: Not sure I understand what you're implying here?

gunnarmorling: What I mean is, can we e.g. get rid of the `parameters` map and work with a builder directly, instead of first adding everything to the maps and then creating + feeding a builder from the end? For sure not strictly needed, but it may further simplify this code here. I'll leave it to you :)

engine/src/main/java/org/hibernate/validator/internal/util/annotation/AnnotationAttributes.java
```diff
@@ -0,0 +1,88 @@
...
+import org.hibernate.validator.internal.util.logging.Log;
+import org.hibernate.validator.internal.util.logging.LoggerFactory;
+import org.hibernate.validator.internal.util.stereotypes.Immutable;
+
+public class AnnotationAttributes implements Serializable {
+
+	private static final Log LOG = LoggerFactory.make();
+
+	@Immutable
+	private final Map<String, Object> parameters;
```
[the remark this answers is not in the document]
gsmet: Taking a step back, yes, it's probably something we could remove. Working on it.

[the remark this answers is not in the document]
gsmet: Done.

engine/src/main/java/org/hibernate/validator/cfg/AnnotationDef.java
gsmet: I don't think it buys us much but done.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
gsmet: OK, so after discussing with myself in the shower, I decided to go for the dummy constraint validator approach. Having a wrapper in all cases for a corner case is not a good idea.

gsmet:
@gunnarmorling I think I fixed all your concerns. I also added a couple of cleanup commits that I kept for when we would agree on the approach.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/util/annotation/AnnotationDescriptor.java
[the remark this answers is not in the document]
gunnarmorling: Ok, makes sense.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
gunnarmorling: Ok.

gunnarmorling:
@gsmet thanks a lot for the quick fixes! It looks great overall. The one thing I'm still feeling a bit unsure about is the usage of `CHM` in `ConstraintTree` (also if it's just when not using the default `CVF`), which could lead to creation of thousands of instances.

Did you have the chance to compare memory usage before and after the change? I think `CHM` isn't using as much memory when empty or with a few elements as it used to, but perhaps we can instantiate it with more fine-tuned settings at least?

I'm terms of an alternative, I only can think of either a dedicated map implementation geared towards our use case of very few elements or "turning around" the entire model and have a model instance per `CVF`. It'd allow to have a single, final `CV` in `ConstraintTree`,  but it'd be a substantial change and it may end up using even more memory, though.

gsmet:
Yes, that's something that's been bothering me too. But I don't think having a `ContraintTree` per CVF is such a good idea.

I think our best bet for now would be to keep the current situation as it is if not the default CVF i.e. to rely entirely on the `ConstraintValidatorManager` cache if the CVF is not the default one. It's far more efficient now that the `CacheKey` is based on the `AnnotationDescriptor`.

gunnarmorling:
LGTM, merging. Thanks a lot, @gsmet!

PR #869, opened 2017-10-24
HV-1472 Update our WildFly integration to 11.0.0.Final

gsmet:
* https://hibernate.atlassian.net/browse/HV-1472

For now, we keep the 10.1 patch, waiting for the next development cycle
to start. Once it's started, we will switch to the next version.


gunnarmorling:
> waiting for the next development cycle to start.

Which one do you mean? HV 6.1 or WF 12? +1 for keeping the 10.1 file for the time being, it should be useful.

gsmet:
I mean the first alpha of WF 12.

PR #872, opened 2017-10-25
HV-1323 Avoid the warn message when ParameterMessageInterpolator is used

gsmet:
* https://hibernate.atlassian.net/browse/HV-1323


gunnarmorling:
And applied. Thanks, @gsmet!

PR #873, opened 2017-10-26
HV-1512 Improve initialization cost of all logger classes

Sanne:
https://hibernate.atlassian.net/browse/HV-1512

### DIFF
--- a/cdi/src/main/java/org/hibernate/validator/cdi/ValidationExtension.java
+++ b/cdi/src/main/java/org/hibernate/validator/cdi/ValidationExtension.java
@@ -8,6 +8,7 @@ package org.hibernate.validator.cdi;
 
 import java.lang.annotation.Annotation;
 import java.lang.annotation.ElementType;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.util.Collections;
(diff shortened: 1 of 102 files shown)


gunnarmorling:
Jenkins, retest this, please.

gunnarmorling:
Jenkins, retest this, please.

gunnarmorling:
LGTM, just waiting for the CI build.

gsmet:
@Sanne out of curiosity, where does it come from? Does it make a significant difference?

Sanne:
> @Sanne out of curiosity, where does it come from? Does it make a significant difference?

It's coming from you mentioning on the blog that we're slower to initialize. It reminded me that I had this idea in mind and never did it.

Yes it makes a very significant difference, but only on class initialization of the whole framework.

gunnarmorling:
Applied. Thanks, @Sanne!

Sanne:
Some more data; I run the testsuite using surefire's options reuseForks=false and forkCount=1 to ensure each test would run in a new fresh JVM.

after the patch: Total time: 05:06 min
before the patch: Total time: 06:39 min

That's quite good!

gunnarmorling:
Ok, but that's rather unrealistic, no one is going to reload their HV classes after using a validator factory just once ;) Still a very good improvement. Perhaps do the same for OGM? Don't expect such big difference there, but it's less hack-ish since the security manager subclass trick.

Sanne:
Sure I'm aware it's unrealistic, but knowing this would benefit exclusively the "cold boot" I think it's the only sensible way to measure if it makes sense at all.

I'm not shy to admit though that I'm interested in exploring such things more regularly to bring the "tool bootstrap time" down for e.g. the WildFly bootstrap time but also the "perception" people have of stuff being "not light". Totally agree that it makes not much sense for the _real_ value.

PR #875, opened 2017-10-31
HV-1505 Create ConstraintAnnotationDescriptor for typed access to groups() etc.

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1505


marko-bekhta:
That's something new from Jenkins :smiley:

gsmet:
@marko-bekhta yeah I don't see that as an improvement but we haven't figured out yet where it comes from :)

engine/src/main/java/org/hibernate/validator/cfg/AnnotationDef.java
```diff
@@ -64,12 +64,16 @@
 	 */
 	private final Map<String, Class<?>> annotationsAsParametersTypes;
 
-	protected AnnotationDef(Class<A> annotationType) {
-		this.annotationDescriptorBuilder = new AnnotationDescriptor.Builder<>( annotationType );
+	protected AnnotationDef(AnnotationDescriptor.Builder<A> annotationDescriptorBuilder) {
+		this.annotationDescriptorBuilder = annotationDescriptorBuilder;
```
gsmet: Unfortunately, we can't do that as we would expose internal classes in a public facing API.

We have the same issue with the `protected` trick we used to expose `createAnnotation()` without making it public. With `createAnnotationDescriptor()`, it does not work anymore as it exposes `AnnotationDescriptor` to the subclasses. We created https://hibernate.atlassian.net/browse/HV-1511 for this one (maybe you could fix it? Basically, it's about making `createAnnotationDescriptor()` private and calling it via reflection - with the usual `run()` method to check the security manager).

I think the only thing you can do is having the ability to build a `ConstaintAnnotationDescriptor.Builder` from an `AnnotationDescriptor.Builder`. The fact is that you don't use your `withMessage()` and al. methods in here so it shouldn't be a big change.

I think you can keep the `withMessage()` and al. methods even if they are only useful in the tests. I think I would call them `setMessage()` though as they are basically setters. What was your reasoning for using `with...()`?

marko-bekhta: yes... thought that might not be a good idea to change the visibility :) as for the methods they are used in the `ConstraintDescriptorImpl` as well. 
>What was your reasoning for using with...()?

I picked it up somewhere and start to call those builder methods `with*` :smiley: I'll change those to `set*` no problem. I'll see about that HV-1511 as well a bit later today.

gsmet: It's not an issue in `ConstraintDescriptorImpl` as it's an internal class and they are not exposed in the public API.

marko-bekhta:
rebased and applied some changes, let me know what you think :)

engine/src/main/java/org/hibernate/validator/internal/cfg/context/ConfiguredConstraint.java
```diff
@@ -94,31 +97,24 @@ public ConstraintLocation getLocation() {
 	}
 
 	public ConstraintAnnotationDescriptor<A> createAnnotationDescriptor() {
-		return constraint.createAnnotationDescriptor();
+		return run( CreateConstraintAnnotationDescriptor.action( constraint ) );
```
gsmet: After giving it some thoughts and considering we only manipulate a private HV method, I'm wondering if we really need the SM here (I know I recommended to use it but I'm wondering now). The method can for sure be called on an external class but it's still the same `AnnotationDef`private method.

@gunnarmorling thoughts on this?

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -816,4 +816,10 @@ ConstraintDeclarationException getNoValueExtractorFoundForTypeException(@FormatW
 	@LogMessage(level = DEBUG)
 	@Message(id = 234, value = "Using %1$s as ValidatorFactory-scoped %2$s.")
 	void logValidatorFactoryScopedConfiguration(@FormatWith(ClassObjectFormatter.class) Class<?> configuredClass, String configuredElement);
+
+	@Message(id = 235, value = "Unable to create annotation descriptor for %1$s.")
+	ValidationException getUnableToCreateAnnotationDescriptor(@FormatWith(ClassObjectFormatter.class) Class<?> configuredClass, @Cause Throwable e);
+
+	@Message(id = 236, value = "Unable to find declared methods of AnnotationDef")
```
gsmet: Maybe `Unable to find the method required to create the constraint annotation descriptor.`?

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -816,4 +816,10 @@ ConstraintDeclarationException getNoValueExtractorFoundForTypeException(@FormatW
 	@LogMessage(level = DEBUG)
 	@Message(id = 234, value = "Using %1$s as ValidatorFactory-scoped %2$s.")
 	void logValidatorFactoryScopedConfiguration(@FormatWith(ClassObjectFormatter.class) Class<?> configuredClass, String configuredElement);
+
+	@Message(id = 235, value = "Unable to create annotation descriptor for %1$s.")
```
gsmet: to create `an` annotation descriptor.

engine/src/main/java/org/hibernate/validator/internal/util/annotation/AnnotationDescriptor.java
```diff
@@ -191,20 +198,16 @@ private int buildHashCode() {
...
+			int valueHashCode = !value.getClass().isArray() ? value.hashCode()
+					: value.getClass() == boolean[].class ? Arrays.hashCode( (boolean[]) value )
+					: value.getClass() == byte[].class ? Arrays.hashCode( (byte[]) value )
+					: value.getClass() == char[].class ? Arrays.hashCode( (char[]) value )
+					: value.getClass() == double[].class ? Arrays.hashCode( (double[]) value )
+					: value.getClass() == float[].class ? Arrays.hashCode( (float[]) value )
+					: value.getClass() == int[].class ? Arrays.hashCode( (int[]) value )
+					: value.getClass() == long[].class ? Arrays.hashCode( (long[]) value )
+					: value.getClass() == short[].class ? Arrays.hashCode( (short[]) value )
+					: Arrays.hashCode( (Object[]) value );
```
gsmet: Thanks for cleaning this up!

engine/src/main/java/org/hibernate/validator/internal/util/annotation/ConstraintAnnotationDescriptor.java
```diff
@@ -0,0 +1,80 @@
...
+
+		public Builder(Class<S> type, Map<String, Object> attributes) {
+			super( type, attributes );
+		}
+
+		public Builder(S annotation) {
+			super( annotation );
+		}
+
+		public Builder withMessage(String message) {
```
gsmet: As discussed, let's call these methods `set*` to be consistent with `setAttribute()`.

gsmet:
@marko-bekhta Looks good. Added a couple of minor comments. I asked a question to Gunnar about the use of the SM, maybe it's not necessary.

Note that if we need it, we would need something like what we do in: https://github.com/hibernate/hibernate-validator/blob/51ca465d2ac88aaf76abc1243a957c71034eb3f1/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/FieldCascadable.java#L96 to check for the permission.

We also could probably make the privileged action more generic and used to get a `MethodHandle` in general, same as what we have with https://github.com/hibernate/hibernate-validator/blob/51ca465d2ac88aaf76abc1243a957c71034eb3f1/engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetDeclaredMethod.java instead of having a very specific privileged action.

But let's wait for Gunnar's feedback before making a move here.

HTH. (I will be mostly off tomorrow, public Holiday here in France)

engine/src/main/java/org/hibernate/validator/internal/util/annotation/AnnotationDescriptor.java
marko-bekhta: I just wasn't able to do other stuff in that file looking at that formatting :)

gsmet: Ah ah, yeah, I just close my eyes when I work on this file but I agree it's a bit harder to code this way :).

engine/src/main/java/org/hibernate/validator/internal/util/annotation/ConstraintAnnotationDescriptor.java
```diff
@@ -0,0 +1,80 @@
...
+
+	public ConstraintAnnotationDescriptor(A annotation) {
+		super( annotation );
+	}
+
+	public ConstraintAnnotationDescriptor(AnnotationDescriptor<A> descriptor) {
+		super( descriptor );
+	}
+
+	public String message() {
```
gsmet: Second thoughts on this one. As `ConstraintAnnotationDescriptor` is a JavaBean, let's stick to the bean convention and use `getMessage()`. Same for the others below.

marko-bekhta:
I've applied the changes and force-pushed the edited commits. I've added a SM check as it was a quick change. As for the action ... I thought about it first but then see how I initialized the method handle in static block so it would not be needed to do this each time the action is called (I thought that such approach should save some time during execution). Waiting for the answer now :)

gsmet:
@marko-bekhta rebased and added 2 more commits.

I discussed the security topic with Gunnar and what I did in my last commit should be safe.

gsmet:
@marko-bekhta waiting for CI and your feedback.

I think it's ready to be merged now.

marko-bekhta:
@gsmet looks good to me :) sorry took me a while to reply

gsmet:
Jenkins, retest this please

gsmet:
Looks like we have an issue with our CI. I'll merge it as soon as I can get a build to finish!

gsmet:
Jenkins, retest this please

gsmet:
And merged, thanks @marko-bekhta !

PR #876, opened 2017-10-31
HV-1493 Temporal validation tolerance

gsmet:
https://hibernate.atlassian.net/browse/HV-1493

@marko-bekhta @danielwegener

So here is my attempt at tackling the clock skew tolerance issue after my discussion with Daniel.

I think it now works as expected with very minimal changes to the validators.

@marko-bekhta it's based on what you did and it includes the fact that the `Clock` is now defined at initialization time. After giving it some thoughts, I don't see an issue with it, even spec-wise.

Feedback welcome.

### DIFF
--- a/documentation/src/main/asciidoc/ch09.asciidoc
+++ b/documentation/src/main/asciidoc/ch09.asciidoc
@@ -351,7 +351,8 @@ When implementing your own temporal constraints, you might need to have access t
 
 It can be obtained by calling the `HibernateConstraintValidatorInitializationContext#getTemporalValidationTolerance()` method.
 
-Note that to get access to this context at initialization, your constraint validator has to implement the `HibernateConstraintValidator` contract.
+Note that to get access to this context at initialization, your constraint validator has to implement
+the `HibernateConstraintValidator` contract (see <<constraint-validator-hibernateconstraintvalidator>>).
 This contract is currently marked as incubating: it might be subject to change in the future.
 ====
 
--- a/documentation/src/main/asciidoc/ch09.asciidoc
+++ b/documentation/src/main/asciidoc/ch09.asciidoc
@@ -287,7 +287,7 @@ to learn more about this specific implementation.
 ====
 
 [[section-clock-provider]]
-==== `ClockProvider`
+==== `ClockProvider` and temporal validation tolerance
 
 For time related validation (`@Past` and `@Future` constraints for instance), it might be useful to define what is
 considered `now`.
@@ -324,6 +324,35 @@ You can obtain the `ClockProvider` in your validators by calling the
 
 For instance, this might be useful if you want to replace the default message of the `@Future`
 constraint with a more explicit one.
+====
+
+When dealing with distributed architectures, you might need some tolerance when applying temporal constraints
+such as `@Past` or `@Future`.
+
+You can set a temporal validation tolerance by bootstrapping your `ValidatorFactory` as below:
+
+[[example-using-temporal-validation-tolerance]]
+.Using temporal validation tolerance
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/BootstrappingTest.java[tags=temporalValidationTolerance]
+----
+====
+
+Alternatively, you can define it in the XML configuration by setting the `hibernate.validator.temporal_validation_tolerance` property
+in your _META-INF/validation.xml_.
+
+The value of this property must be a `long` defining the tolerance in milliseconds.
+
+[NOTE]
+====
+When implementing your own temporal constraints, you might need to have access to the temporal validation tolerance.
+
+It can be obtained by calling the `HibernateConstraintValidatorInitializationContext#getTemporalValidationTolerance()` method.
+
+Note that to get access to this context at initialization, your constraint validator has to implement the `HibernateConstraintValidator` contract.
+This contract is currently marked as incubating: it might be subject to change in the future.
 ====
 
 [[section-bootstrapping-valueextractors]]
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
@@ -6,7 +6,10 @@
  */
 package org.hibernate.validator.constraintvalidation;
 
+import java.time.Clock;
 import java.time.Duration;
+
+import javax.validation.ClockProvider;
 
 import org.hibernate.validator.Incubating;
 import org.hibernate.validator.spi.scripting.ScriptEvaluator;
@@ -34,6 +37,16 @@ public interface HibernateConstraintValidatorInitializationContext {
 	 * found for a given {@code languageName}
 	 */
 	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName);
+
+	/**
+	 * Returns the provider for obtaining the current time in the form of a {@link Clock}, e.g. when validating the
+	 * {@code Future} and {@code Past} constraints.
+	 *
+	 * @return the provider for obtaining the current time, never {@code null}. If no specific provider has been
+	 * configured during bootstrap, a default implementation using the current system time and the current default time
+	 * zone as returned by {@link Clock#systemDefaultZone()} will be returned.
+	 */
+	ClockProvider getClockProvider();
 
 	/**
 	 * Returns the temporal validation tolerance i.e. the acceptable margin of error when comparing date/time in
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.validator.constraintvalidation;
 
+import java.time.Duration;
+
 import org.hibernate.validator.Incubating;
 import org.hibernate.validator.spi.scripting.ScriptEvaluator;
 import org.hibernate.validator.spi.scripting.ScriptEvaluatorFactory;
@@ -32,4 +34,12 @@ public interface HibernateConstraintValidatorInitializationContext {
 	 * found for a given {@code languageName}
 	 */
 	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName);
+
+	/**
+	 * Returns the temporal validation tolerance i.e. the acceptable margin of error when comparing date/time in
+	 * temporal constraints.
+	 *
+	 * @return the tolerance
+	 */
+	Duration getTemporalValidationTolerance();
 }
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
@@ -15,6 +15,7 @@ import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
+import java.time.Duration;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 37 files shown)


engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/bv/time/ClockSkewToleranceTest.java
```diff
@@ -0,0 +1,440 @@
...
+			minguoDate = MinguoDate.from( dateTime );
+			offsetDateTime = dateTime.toOffsetDateTime();
+			thaiBuddhistDate = ThaiBuddhistDate.from( dateTime );
+			year = Year.from( dateTime );
+			yearMonth = YearMonth.from( dateTime );
+			zonedDateTime = dateTime;
+			readableInstant = new org.joda.time.DateTime( dateTime.toEpochSecond() * 1_000 );
+		}
+	}
+}
```
danielwegener: missing empty line ;)

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -275,6 +277,14 @@ public HibernateValidatorConfiguration scriptEvaluatorFactory(ScriptEvaluatorFac
 		return this;
 	}
 
+	@Override
+	public HibernateValidatorConfiguration clockSkewTolerance(Duration clockSkewTolerance) {
+		Contracts.assertNotNull( clockSkewTolerance, MESSAGES.parameterMustNotBeNull( "clockSkewTolerance" ) );
```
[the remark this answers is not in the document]
gsmet: I already normalize it a bit later.

marko-bekhta:
Looks good. Nice to see that you decided to cache a clock in the init in the end :)

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -311,6 +320,7 @@ public Validator getValidator() {
 				clockProvider,
 				scriptEvaluatorFactory,
 				failFast,
+				clockSkewTolerance,
```
marko-bekhta: My concern here is that these methods start to grow in their parameter numbers each time we add some new config property ... But I guess https://hibernate.atlassian.net/browse/HV-1506 should help with this, right? If not directly then maybe we could add a "subcontainer" for these kind of properties...

gsmet: Yes, that's the idea of this issue. I think we would need to introduce some sort of `ValidatorContext`. That being said, it's a bit better than before because we included the new features in the initialization context and we don't need to pass them along everywhere.

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -326,6 +326,34 @@ For instance, this might be useful if you want to replace the default message of
...
+
+The value of this property must be a `long` defining the duration in milliseconds.
+
+[NOTE]
+====
+When implementing your own temporal constraints, you might need to have access to the clock skew tolerance.
+
+It can be obtained by calling the `HibernateConstraintValidatorInitializationContext#getClockSkewTolerance()` method.
+
+Note that to get access to this context at initialization, your validator has to implement the `HibernateConstraintValidator` contract.
```
marko-bekhta: Do we have/need anything for `HibernateConstraintValidator` in the documentation ?

gsmet: Glad you asked. I asked myself the same question when I wrote this part. I'm not totally sure as it's an incubating API so I'm not sure we want to advertise it.

I tried to think about how to present it but everything I can come with is a bit confusing.

So for now, I wouldn't advertise it apart for the clock skew tolerance usage. I added a note to state it's incubating.

What would be interesting would be to check if other validators could benefit from having some additional elements in the initialization context. If we put it in the spec, we should be careful to expose everything that might be needed.

marko-bekhta: That sounds like a good plan :)

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -326,6 +326,35 @@ For instance, this might be useful if you want to replace the default message of
 constraint with a more explicit one.
 ====
 
+When dealing with distributed architectures, you might need some tolerance for the clock skew in the temporal constraints
+e.g. `@Past` or `@Future`.
+
+You can set a clock skew tolerance by bootstrapping your `ValidatorFactory` as below:
```
gunnarmorling: "Clock skew" is an established thing with a different meaning: https://en.wikipedia.org/wiki/Clock_skew

Maybe we should find a different term, e.g. "temporal validation lenience/tolerance" or so?

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -326,6 +326,35 @@ For instance, this might be useful if you want to replace the default message of
...
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/BootstrappingTest.java[tags=clockSkewTolerance]
+----
+====
+
+Alternatively, you can define it in the XML configuration by setting the `hibernate.validator.clock_skew_tolerance` property
+in your _META-INF/validation.xml_.
+
+The value of this property must be a `long` defining the duration in milliseconds.
```
gunnarmorling: s/duration/tolerance/

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -32,4 +37,24 @@
...
+	 * {@link Clock#systemDefaultZone()} will be returned.
+	 */
+	ClockProvider getClockProvider();
+
+	/**
+	 * Returns clock skew tolerance as {@link Duration} which is used to determine
+	 * acceptable margin of error in milliseconds, which is allowed
+	 * when comparing date/time in time related constraints.
+	 *
+	 * @return a tolerance as {@link Duration}
```
gunnarmorling: No need to repeat the return type in the comment.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -32,4 +37,24 @@
...
+	 * @return the provider for obtaining the current time, never {@code null}. If no
+	 * specific provider has been configured during bootstrap, a default implementation using
+	 * the current system time and the current default time zone as returned by
+	 * {@link Clock#systemDefaultZone()} will be returned.
+	 */
+	ClockProvider getClockProvider();
+
+	/**
+	 * Returns clock skew tolerance as {@link Duration} which is used to determine
+	 * acceptable margin of error in milliseconds, which is allowed
```
gunnarmorling: Is the unit (milliseconds) actually relevant here?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -535,6 +552,30 @@ private static ScriptEvaluatorFactory getScriptEvaluatorFactory(ConfigurationSta
...
+				Duration tolerance = Duration.ofMillis( Long.parseLong( clockSkewToleranceProperty ) ).abs();
+				LOG.logClockSkewTolerance( tolerance );
+				return tolerance;
+			}
+			catch (Exception e) {
+				throw LOG.getUnableToParseClockSkewToleranceException( clockSkewToleranceProperty, e );
+			}
+		}
+
+		LOG.logClockSkewTolerance( Duration.ZERO  );
```
gunnarmorling: There are two blanks after `ZERO`.

engine/src/main/java/org/hibernate/validator/internal/util/logging/formatter/DurationFormatter.java
```diff
@@ -0,0 +1,28 @@
...
+ * Used with JBoss Logging to display durations in log messages.
+ *
+ * @author Marko Bekhta
+ */
+public class DurationFormatter {
+
+	private final String stringRepresentation;
+
+	public DurationFormatter(Duration duration) {
+		this.stringRepresentation = duration.toString();
```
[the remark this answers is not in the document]
marko-bekhta: That's on me :) I was thinking to use the same style of `Duration` formatting as the one used for `DurationMin/Max`:
```
${days == 0 ? '' : days == 1 ? ' 1 day' : ' ' += days += ' days'}${hours == 0 ? '' : hours == 1 ? ' 1 hour' : ' ' += hours += ' hours'}${minutes == 0 ? '' : minutes == 1 ? ' 1 minute' : ' ' += minutes += ' minutes'}${seconds == 0 ? '' : seconds == 1 ? ' 1 second' : ' ' += seconds += ' seconds'}${millis == 0 ? '' : millis == 1 ? ' 1 milli' : ' ' += millis += ' millis'}${nanos == 0 ? '' : nanos == 1 ? ' 1 nano' : ' ' += nanos += ' nanos'}
```
but then at some point we decided to close the ticket, and when it got reopened - I forgot about the formatter....

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
gsmet: Raah, it was in the original patch. I fixed it elsewhere but missed this occurrence. Will fix.

engine/src/main/java/org/hibernate/validator/internal/util/logging/formatter/DurationFormatter.java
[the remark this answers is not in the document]
gsmet: Yeah, I think it's worth it given how readable the `toString()` of `Duration` is.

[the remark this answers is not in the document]
gsmet: @marko-bekhta I let you provide a more sensible formatter? You will have to do some maths as `Duration` does not return the number of days/minutes...

documentation/src/main/asciidoc/ch09.asciidoc
gsmet: Maybe "temporal tolerance" would be enough?

documentation/src/main/asciidoc/ch09.asciidoc
gsmet: Fixed.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
gsmet: Fixed.

engine/src/main/java/org/hibernate/validator/internal/util/logging/formatter/DurationFormatter.java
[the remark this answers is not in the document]
gsmet: @marko-bekhta would you be willing to contribute a more sensible formatter?

documentation/src/main/asciidoc/ch09.asciidoc
danielwegener: Or maybe `Clock drift`? But both terms refer to the differences in frequency between clocks while the feature is actually about coping with imperfectly synchronized clocks (regardless the reason for the asynchrony itself).

engine/src/main/java/org/hibernate/validator/internal/util/logging/formatter/DurationFormatter.java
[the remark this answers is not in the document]
marko-bekhta: done :)

gsmet:
@marko-bekhta @gunnarmorling I amended the existing commits to change the name so that we have a clean history. Otherwise we would have had the last commit changing everything.

@marko-bekhta I also included your `DurationFormatter` commit in the first commit so that we have a clean formatter from the beginning (I modified it a bit though).

documentation/src/main/asciidoc/ch09.asciidoc
```diff
@@ -326,6 +326,35 @@ For instance, this might be useful if you want to replace the default message of
...
+
+The value of this property must be a `long` defining the tolerance in milliseconds.
+
+[NOTE]
+====
+When implementing your own temporal constraints, you might need to have access to the temporal validation tolerance.
+
+It can be obtained by calling the `HibernateConstraintValidatorInitializationContext#getTemporalValidationTolerance()` method.
+
+Note that to get access to this context at initialization, your constraint validator has to implement the `HibernateConstraintValidator` contract.
```
gunnarmorling: Is `HibernateConstraintValidator` described in the docs? If so, a link would be nice.

gsmet:
And merged! Thanks everyone!

PR #878, opened 2017-11-02
HV-1444 Cache value extractors resolved by value type

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1444

As promised :) @gsmet here's a simple test case for multiple containers at the same time. What bothers me is that 
```
.addValueExtractor( new AnotherCustomContainerValueExtractor() )
```
throws an error like:
```
javax.validation.valueextraction.ValueExtractorDefinitionException: HV000203: Value extractor type org.hibernate.validator.test.internal.engine.valueextraction.MultipleContainersAtTheSameTimeTest$AnotherCustomContainerValueExtractor fails to declare the extracted type parameter using @ExtractedValue.

	at org.hibernate.validator.internal.engine.valueextraction.ValueExtractorDescriptor.getExtractedTypeParameter(ValueExtractorDescriptor.java:96)
	at org.hibernate.validator.internal.engine.valueextraction.ValueExtractorDescriptor.<init>(ValueExtractorDescriptor.java:50)
	at org.hibernate.validator.internal.engine.ConfigurationImpl.addValueExtractor(ConfigurationImpl.java:216)
	at org.hibernate.validator.internal.engine.ConfigurationImpl.addValueExtractor(ConfigurationImpl.java:67)
	at org.hibernate.validator.test.internal.engine.valueextraction.MultipleContainersAtTheSameTimeTest.setupValidator(MultipleContainersAtTheSameTimeTest.java:41)
```

which is slightly unexpected for me :) The problem seems to be if the extractor is an inner class but if it is in its own file (`CustomContainerValueExtractor`) everything is fine. (and both extractors looks the same).

Otherwise - if that `AnotherCustomContainerValueExtractor` is removed the test pass. Which as I understand is not good, right ? I mean it should throw an exception as there are two possible value extractors to be applied to that class. Or I've done it wrong :smiley:


gsmet:
Well, yes, you encountered a recurring issue we have (see the JDK bugs we found here: https://hibernate.atlassian.net/projects/HV/versions/29900/tab/release-report-todo ).

I have opened an issue in the OpenJDK project a while ago but apparently, it's not something we can expect to be fixed soon...

And yes, you have the problem right. In the case of cascading, we should use the runtime type so we should have an error.

The problem is that, to support this very narrow use case, we will have to remove a very useful optimization. Thus, the fact that we should really work on this caching thing before even thinking fixing this issue.

marko-bekhta:
> Well, yes, you encountered a recurring issue we have

That's good to hear .. Because I've spent like 40 minutes trying to understand what am I doing wrong with the value extractor (initially I've added it as inner class...) :)

Now I can move to the other part (VE resolver cache) I'll push the changes to this PR when I'll have something.

gsmet:
Just send me an email or join us on HipChat when you have this sort of issues!

marko-bekhta:
I've added one more performance test to randomize collection creation and I tried to run them with this latest commit and without it. Here's what I've got from those runs:
with cache map:
```
Benchmark                                                                                                          Mode  Cnt    Score    Error   Units
o.h.v.p.cascaded.CascadedValidation.testCascadedValidation                                                        thrpt   20  232.499 ± 51.943  ops/ms
o.h.v.p.cascaded.CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                          thrpt   20  930.532 ± 20.195   ops/s
o.h.v.p.multilevel.MultiLevelContainerValidation.testMultiLevelPreGeneratedValidation                             thrpt   50    0.021 ±  0.003  ops/ms
o.h.v.p.multilevel.MultiLevelContainerValidation.testMultiLevelPreGeneratedWithRundomRuntimeContainersValidation  thrpt   50    0.067 ±  0.002  ops/ms
o.h.v.p.simple.SimpleValidation.testSimpleBeanValidation                                                          thrpt   20  673.511 ± 14.737  ops/ms
o.h.v.p.statistical.StatisticalValidation.testValidationWithStatisticalGraphDepthAndConstraintValidator           thrpt   10    0.181 ±  0.002  ops/ms
```
without the cache map:
```
Benchmark                                                                                                 Mode  Cnt    Score    Error   Units
o.h.v.p.cascaded.CascadedValidation.testCascadedValidation                                               thrpt   20  214.290 ± 46.570  ops/ms
o.h.v.p.cascaded.CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                 thrpt   20  862.916 ± 14.765   ops/s
o.h.v.p.multilevel.MultiLevelContainerValidation.testMultiLevelPreGeneratedValidation                    thrpt   50    0.050 ±  0.005  ops/ms
o.h.v.p.simple.SimpleValidation.testSimpleBeanValidation                                                 thrpt   20  564.996 ± 83.344  ops/ms
o.h.v.p.statistical.StatisticalValidation.testValidationWithStatisticalGraphDepthAndConstraintValidator  thrpt   10    0.171 ±  0.003  ops/ms
```

The one without the map is failing the new test, but I haven't looked at it yet (why exactly is it failing) with this kind of exception:
```
Caused by: java.lang.ClassCastException: org.hibernate.validator.internal.engine.ValidationContext$BeanPathMetaConstraintProcessedUnit cannot be cast to org.hibernate.validator.internal.engine.ValidationContext$BeanGroupProcessedUnit
	at org.hibernate.validator.internal.engine.ValidationContext$BeanGroupProcessedUnit.equals(ValidationContext.java:621)
	at java.util.HashMap.getNode(HashMap.java:572)
	at java.util.HashMap.containsKey(HashMap.java:596)
	at java.util.HashSet.contains(HashSet.java:204)
	at org.hibernate.validator.internal.engine.ValidationContext.isAlreadyValidatedForCurrentGroup(ValidationContext.java:441)
	at org.hibernate.validator.internal.engine.ValidationContext.isBeanAlreadyValidated(ValidationContext.java:270)
	at org.hibernate.validator.internal.engine.ValidatorImpl$CascadingValueReceiver.doValidate(ValidatorImpl.java:700)
	at org.hibernate.validator.internal.engine.ValidatorImpl$CascadingValueReceiver.indexedValue(ValidatorImpl.java:689)
	at org.hibernate.validator.internal.engine.valueextraction.ListValueExtractor.extractValues(ListValueExtractor.java:26)
	at org.hibernate.validator.internal.engine.valueextraction.ListValueExtractor.extractValues(ListValueExtractor.java:16)
	at org.hibernate.validator.internal.engine.valueextraction.ValueExtractorHelper.extractValues(ValueExtractorHelper.java:42)
	... 34 more
```

looking at these first numbers I'd say that the map is not introducing much of the problem, right?
Now I guess I need to figure out how to find an extractor based on a runtime type.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -155,19 +160,17 @@ else if ( maximallySpecificContainerElementCompliantValueExtractors.size() == 1
 	/**
 	 * Returns the maximally specific type-compliant and container-element-compliant value extractor
 	 * from a set of preselected value extractors or {@code null} if none was found.
-	 * <p>
```
gsmet: It's not a first time you remove a `<p>`, what's the rationale for this?

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -347,6 +351,124 @@ private static boolean isClassPresent(String className, boolean fallbackOnTCCL)
...
+		public ValueExtractorCache() {
+			this.valueExtractorCache = new ConcurrentHashMap<>();
+			this.nonContainerTypes = new HashSet<>();
+		}
+
+		public Set<ValueExtractorDescriptor> getValueExtractors( Set<ValueExtractorDescriptor> valueExtractorCandidates,
+							Class<?> valueType ) {
+			ValueExtractorCacheKey cacheKey = new ValueExtractorCacheKey( valueType, /*something here*/ null );
+
+			if ( nonContainerTypes.contains( cacheKey ) ) {
```
gsmet: You can avoid using the cache key here. If it's a non container type, the whole type is not a container. This means that in the simple case where it's not a container, we avoid instantiating the cache key.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -347,6 +351,124 @@ private static boolean isClassPresent(String className, boolean fallbackOnTCCL)
 		}
 	}
 
+	private final class ValueExtractorCache {
+
+		private final ConcurrentHashMap<ValueExtractorCacheKey, Set<ValueExtractorDescriptor>> valueExtractorCache;
```
gsmet: I don't think you should call it `valueExtractorCache`, it's rather disturbing as it's the name of the whole class.

Maybe `valueExtractorsByTypeArgument`? Not totally sold to it, maybe you have a better idea?

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -347,6 +351,124 @@ private static boolean isClassPresent(String className, boolean fallbackOnTCCL)
...
+		private final Set<ValueExtractorCacheKey> nonContainerTypes;
+
+		public ValueExtractorCache() {
+			this.valueExtractorCache = new ConcurrentHashMap<>();
+			this.nonContainerTypes = new HashSet<>();
+		}
+
+		public Set<ValueExtractorDescriptor> getValueExtractors( Set<ValueExtractorDescriptor> valueExtractorCandidates,
```
gsmet: So in my mind, I wouldn't have the candidates here. I would simply have the runtime type and the type parameter.

The whole notion of candidates would be removed. I thought that you could do it in 2 steps but maybe you can't.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -347,6 +351,124 @@ private static boolean isClassPresent(String className, boolean fallbackOnTCCL)
 		}
 	}
 
+	private final class ValueExtractorCache {
```
gsmet: Maybe `ValueExtractorResolutionCache`?

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -347,6 +351,124 @@ private static boolean isClassPresent(String className, boolean fallbackOnTCCL)
...
+					return false;
+				}
+				ValueExtractorCacheKey that = (ValueExtractorCacheKey) o;
+				return Objects.equals( this.runtimeType, that.runtimeType ) &&
+						Objects.equals( this.typeParameter, that.typeParameter );
+			}
+
+			@Override
+			public int hashCode() {
+				return Objects.hash( this.runtimeType, this.typeParameter );
```
gsmet: Better cache the hashCode.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -347,6 +351,124 @@ private static boolean isClassPresent(String className, boolean fallbackOnTCCL)
...
+//			}
+//			return getMaximallySpecificValueExtractors( key.runtimeType, containerElementCompliantExtractors );
+//		}
+
+		private class ValueExtractorCacheKey {
+
+			private final Class<?> runtimeType;
+			private final TypeVariable<?> typeParameter;
+
+			ValueExtractorCacheKey(Class<?> runtimeType, TypeVariable<?> typeParameter) {
```
gsmet: Wondering if just having the `TypeVariable<?>`as a key wouldn't be sufficient as it contains the Class too. In my experience `TypeVariable` hashCode and equals are quite slow so maybe we still need to encapsulate it in an object to cache the hashCode (one of my project is to get rid of `TypeVariable` and have a `TypeParameterDescriptor` similar to our `AnnotationDescriptor` but let's keep that for later).

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
marko-bekhta: It's the IDE reformatting things in such way. I finally found that setting and checked it so it shouldn't do it anymore. I'll send a PR for the codestyle repo with the update for that.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
marko-bekhta: yes, I just used the candidates here to try and see how the map would behave in general. As this value will be cached there shouldn't be a difference if we look just through the candidates or through all extractors, as in my understanding it should slow down only the first call and then the cache will be used, right ?

marko-bekhta: I also thought at first to store just a resulting value extractor 
```java
ConcurrentHashMap<ValueExtractorCacheKey, ValueExtractorDescriptor> valueExtractorCache
```
but then it wouldn't allow us to know if there are two or more extractors and we should throw an error. The alternative might be to have
```java
private final ConcurrentHashMap<ValueExtractorCacheKey, Supplier<ValueExtractorDescriptor>> valueExtractorCache;
```
and then if there are more than 1 extractor we could store something like
```java
() -> throw LOG.getUnableToGetMostSpecificValueExtractorDueToSeveralMaximallySpecificValueExtractorsDeclaredException( valueType,
					ValueExtractorHelper.toValueExtractorClasses( maximallySpecificContainerElementCompliantValueExtractors )
			);
```
but I wasn't sure that that would be a good idea...

gsmet: Yes, storing a Set looks like the right thing to do.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -614,7 +614,7 @@ public boolean equals(Object o) {
 			if ( this == o ) {
 				return true;
 			}
-			if ( o == null || getClass() != BeanGroupProcessedUnit.class ) {
+			if ( o == null || o.getClass() != BeanGroupProcessedUnit.class ) {
```
gsmet: Annoying one... Please make it a separate JIRA and commit.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -351,102 +348,82 @@ private static boolean isClassPresent(String className, boolean fallbackOnTCCL)
...
+				typeParameterBoundToExtractorType = typeParameterForBinding;
+			}
+			return Objects.equals( typeParameterToCompare, typeParameterBoundToExtractorType );
+		}
 
 		private class ValueExtractorCacheKey {
 
 			private final Class<?> runtimeType;
 			private final TypeVariable<?> typeParameter;
+			private final int hash;
```
gsmet: We call it `hashCode` everywhere else. Let's stick to the conventions.

performance/src/main/java-bv2/org/hibernate/validator/performance/multilevel/MultiLevelContainerValidation.java
```diff
@@ -93,7 +93,7 @@ public void testMultiLevelPreGeneratedValidation(MultiLevelContainerState state,
 	@Threads(50)
 	@Warmup(iterations = 10)
 	@Measurement(iterations = 50)
-	public void testMultiLevelPreGeneratedWithRundomRuntimeContainersValidation(MultiLevelContainerState state, Blackhole bh) {
+	public void testMultiLevelPareGeneratedWithRandomRuntimeContainersValidation(MultiLevelContainerState state, Blackhole bh) {
```
gsmet: You introduced a typo here (Pre became Pare)

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -347,6 +348,104 @@ private static boolean isClassPresent(String className, boolean fallbackOnTCCL)
...
 	}
 
+	private final class ValueExtractorResolutionCache {
+
+		private final ConcurrentHashMap<ValueExtractorCacheKey, Set<ValueExtractorDescriptor>> possibleValueExtractorsByRuntimeTypes;
+		private final Set<Class<?>> nonContainerTypes;
+
+		public ValueExtractorResolutionCache() {
+			this.possibleValueExtractorsByRuntimeTypes = new ConcurrentHashMap<>();
+			this.nonContainerTypes = new HashSet<>();
```
gsmet: Let's initialize them directly when declaring the variable. There's no value doing it in the constructor (and we can remove the constructor).

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ContainerCascadingMetaData.java
```diff
@@ -77,21 +76,13 @@
 	 */
 	private final boolean hasContainerElementsMarkedForCascading;
 
-	/**
-	 * The set of value extractors which are type compliant and container element compliant with the element where
-	 * the cascaded validation was declared. The final value extractor is chosen among these ones, based on the
-	 * runtime type.
-	 */
-	private final Set<ValueExtractorDescriptor> valueExtractorCandidates;
-
```
gsmet: You can also remove `getValueExtractorCandidatesForCascadedValidation()` from `ValueExtractorManager` and the 2 private methods it calls.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -351,102 +348,82 @@ private static boolean isClassPresent(String className, boolean fallbackOnTCCL)
...
+			ValueExtractorCacheKey cacheKey = new ValueExtractorCacheKey( runtimeType, typeParameter );
 
-			Set<ValueExtractorDescriptor> valueExtractorDescriptors = valueExtractorCache.get( cacheKey );
+			Set<ValueExtractorDescriptor> valueExtractorDescriptors = possibleValueExtractorsByRuntimeTypes.get( cacheKey );
 			if ( valueExtractorDescriptors == null ) {
-				valueExtractorDescriptors = getMaximallySpecificValueExtractors(
-						valueType,
-						valueExtractorCandidates
-				);
+				Set<ValueExtractorDescriptor> possibleValueExtractors = getMaximallySpecificValueExtractors( runtimeType );
```
gsmet: I'm not sure you got the logic right here.

We should select the type compliant VEs before filtering to get the maximally specific ones. If you filter the maximally specific VEs first, you might miss a VE that is not maximally specific in the general case BUT is in fact the maximally specific one that is container element compliant.

You cannot use `getMaximallySpecificValueExtractors(Class<?> valueType)` as it should only be used when we have an automatic unwrapping. We can also use it in the case the `@Valid` annotation is on the whole container as in this case we don't have to be container element compliant but definitely not when we are looking for a container element compliant one.

See the resolution algorithm in detail here: http://beanvalidation.org/2.0/spec/#constraintdeclarationvalidationprocess-validationroutine-valueextractorresolution-algorithm-cascaded .

I'm not sure it's easy to build a test case to make what you did fail but I think it would be worth it to at least try.

If I'm not clear, don't hesitate to ping me, it's not an easy subject...

gsmet:
Another additional thought: I'm wondering if we could reuse an identical logic for container element constraints and cascaded validation. In the case of container element constraints, the runtime type and the declared type would be identical (which can also happen in the cascaded case so maybe we need a shortcut for that).

As for the case where we have:
```
@Valid
private Object eventuallyContainer;
```
we would probably have to reuse the logic we use to find a VE with the unwrap options. As it's basically the same thing: either we find zero or one VE for the given runtime container type, or we fail because we don't define a type argument.

marko-bekhta:
@gsmet I've cleaned up the commits and rebased to latest master. Comments about VE filtering logic are not yet addressed.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java
```diff
@@ -37,6 +37,7 @@
 /**
  * @author Gunnar Morling
  * @author Guillaume Smet
+ * @author Marko Bekhta
  */
 public class ValueExtractorManager {
```
gsmet: So... about this one... I'm wondering if we should have the `ValueExtractorManager` and have a separate class named `ValueExtractorResolver` that would contain all the resolving logic (and the cache directly in it i.e. `ValueExtractorResolver` would replace `ValueExtractorResolutionCache`). We would keep the `ValueExtractorManager` as the entry point that delegates the calls to the resolver (I don't want us to carry 2 objects everywhere).

I'm wondering if we could make the methods used for container element constraints also use the cache, using the same methods, just having declaredType == runtimeType? Not sure it would work but I prefer to ask. It would be nicer I think (I would keep separate shortcut public methods though).

I think it would also be nice for each public methods involved in the resolution to have comments explaining when it is used (container element constraints/cascaded validation on container element/cascaded validation on the whole container).

marko-bekhta:
@gsmet I've made a few tiny changes :) 

- first commit should contain all cleanups that you've mentioned (except some of the javadoc ones)
- next I've tried your idea to use the cache for the declared fields as well - it worked so I guess we can keep it ?
- then I've created a VE resolver and simply moved all the resolving and caching logic there
- next commit is additional cleanups in the resolver. Didn't want to do them with the previous commit so that the changes could be more visible.
- and with the last one - you can see an attempt to store found VE in metadata.

didn't have the time to run benchmarks yet. Maybe sometime later today.

gsmet:
@marko-bekhta there is a checkstyle error in your last commit:
[ERROR] /mnt/jenkins-workdir/workspace/hibernate-validator-PR/engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorManager.java:153:94: '(' is not followed by whitespace. [ParenPad]

marko-bekhta:
hmm, and I thought I fixed that.... Force pushed an updated commit.

gsmet:
Thanks!

marko-bekhta:
No problem :)
I still owe a few JavaDoc description updates, I left TODOs so I don't forget :)

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorResolver.java
```diff
@@ -0,0 +1,331 @@
...
+	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
+
+	private final ConcurrentHashMap<ValueExtractorCacheKey, Set<ValueExtractorDescriptor>> possibleValueExtractorsByRuntimeTypeAndTypeParameter = new ConcurrentHashMap<>();
+	private final ConcurrentHashMap<Class<?>, Set<ValueExtractorDescriptor>> possibleValueExtractorsByRuntimeType = new ConcurrentHashMap<>();
+	private final Set<Class<?>> nonContainerTypes = new HashSet<>();
+
+	private final List<ValueExtractorDescriptor> valueExtractors;
+
+	ValueExtractorResolver(ValueExtractorManager valueExtractorManager) {
+		this.valueExtractors = valueExtractorManager.getValueExtractors().values().stream().collect(
```
gsmet: I would say let's pass the list here rather than having a dependency back to `ValueExtractorManager`.

gsmet:
Jenkins, retest this please

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -659,22 +657,17 @@ private BeanPathMetaConstraintProcessedUnit(Object bean, Path path, MetaConstrai
...
 		@Override
 		public boolean equals(Object o) {
-			if ( this == o ) {
-				return true;
-			}
-			if ( o == null || o.getClass() != BeanPathMetaConstraintProcessedUnit.class ) {
-				return false;
-			}
-
+			// no need to check for instance equality or if the calss match because of
```
gsmet: s/calss/class/

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -659,22 +657,17 @@ private BeanPathMetaConstraintProcessedUnit(Object bean, Path path, MetaConstrai
...
+			// how this class is used in the set.
 			BeanPathMetaConstraintProcessedUnit that = (BeanPathMetaConstraintProcessedUnit) o;
 
 			if ( bean != that.bean ) {  // instance equality
```
gsmet: Can't you keep that one at the top? Let's avoid the cast if we can.

marko-bekhta: there's `that` hiding in the if statement so cannot avoid the cast

gsmet:
Jenkins, retest this please

gsmet:
Closing this one. We keep it for archive. Will open a rebased/squashed new one.

PR #880, opened 2017-11-03
Making some code cleanups

marko-bekhta:
Hi @gsmet here are a few cleanups here and there. Mostly around creating empty arrays or lists with one element. This probably wouldn't be a visible performance improvement though :smiley:

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -559,14 +559,9 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		boolean success = true;
 
 		if ( isValidationRequired( validationContext, valueContext, metaConstraint ) ) {
-			Object valueToValidate = null;
 
 			if ( parent != null ) {
-				valueToValidate = valueContext.getValue( parent, metaConstraint.getLocation() );
-				valueContext.setCurrentValidatedValue( valueToValidate );
-			}
-			else {
-				valueToValidate = valueContext.getCurrentValidatedValue();
+				valueContext.setCurrentValidatedValue( valueContext.getValue( parent, metaConstraint.getLocation() ) );
 			}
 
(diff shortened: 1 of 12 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -565,9 +565,6 @@ private boolean validateMetaConstraint(ValidationContext<?> validationContext, V
 				valueToValidate = valueContext.getValue( parent, metaConstraint.getLocation() );
 				valueContext.setCurrentValidatedValue( valueToValidate );
 			}
-			else {
-				valueToValidate = valueContext.getCurrentValidatedValue();
-			}
```
gsmet: We could probably get rid of the variable altogether.

engine/src/main/java/org/hibernate/validator/internal/xml/MetaConstraintBuilder.java
```diff
@@ -51,6 +51,7 @@
 	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
 	private static final Pattern IS_ONLY_WHITESPACE = Pattern.compile( "\\s*" );
+	private static final Class[] EMPTY_CLASSES_ARRAY = { };
```
gsmet: Maybe keep the same notation you used before? new Class[0]?

test-utils/src/main/java/org/hibernate/validator/testutil/ValidationInvocationHandler.java
```diff
@@ -21,6 +21,8 @@
  */
 public class ValidationInvocationHandler implements InvocationHandler {
 
+	private static final Object[] EMPTY_ARGS = { };
```
gsmet: Same here.

marko-bekhta:
Yes, I'll create a JIRA and address the suggestions.

gsmet:
Merged, thanks!

PR #881, opened 2017-11-03
HV-1516 Support punycode-encoded domain names

gsmet:
https://hibernate.atlassian.net/browse/HV-1516

@marko-bekhta can you check that it makes sense to you? Thanks!

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/DomainNameUtil.java
@@ -14,6 +14,7 @@ import java.util.regex.Pattern;
 
 /**
  * @author Marko Bekhta
+ * @author Guillaume Smet
  */
 public final class DomainNameUtil {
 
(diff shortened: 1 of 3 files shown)


marko-bekhta:
Ha :) domain names once again. It seems we'll never stop updating the rules for them :)
I was just writing that we could simplify this by adding the dash to the `DOMAIN_CHARS_WITHOUT_DASH` and it all should work then without additional regex but then it occurred to me that there might be even easier solution to this. 
What if instead of matching a plain string we do this:
```java
		Matcher matcher = pattern.matcher( IDN.toUnicode( domain ) );
		if ( !matcher.matches() ) {
			return false;
		}
```

It will cover the cases from the issue as xn--* stuff will be converted and we would be matching the string without xn-- ?

gsmet:
Yeah, I'm not sure it's a good idea to bet that `IDN.toUnicode()` will work on already Unicode encoded strings.

Moreover, I'm far from sure the operation is fast.

I don't think what I added is that bad.

gsmet:
@marko-bekhta is the new version of the patch OK with you? If so, I'll merge it!

marko-bekhta:
@gsmet looks good to me, thanks!

gsmet:
Merged! @marko-bekhta thanks for the review.

PR #885, opened 2017-11-17
HV-1506 Regroup the Validator scoped properties in ValidationContext

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1506

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
@@ -10,6 +10,7 @@ import static org.hibernate.validator.internal.util.CollectionHelper.newHashSet;
 
 import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Executable;
+import java.time.Duration;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.Iterator;
@@ -26,6 +27,7 @@ import javax.validation.ParameterNameProvider;
 import javax.validation.Path;
 import javax.validation.TraversableResolver;
 import javax.validation.ValidationException;
+import javax.validation.Validator;
 import javax.validation.metadata.ConstraintDescriptor;
 
 import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorInitializationContext;
@@ -39,6 +41,7 @@ import org.hibernate.validator.internal.metadata.core.MetaConstraint;
 import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
 import org.hibernate.validator.internal.util.logging.Log;
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
+import org.hibernate.validator.spi.scripting.ScriptEvaluatorFactory;
 
 /**
  * Context object keeping track of all required data for a validation call.
@@ -111,11 +114,6 @@ public class ValidationContext<T> {
 	private final Set<ConstraintViolation<T>> failingConstraintViolations;
 
 	/**
-	 * The message resolver which should be used in this context.
-	 */
-	private final MessageInterpolator messageInterpolator;
-
-	/**
 	 * The constraint factory which should be used in this context.
 	 */
 	private final ConstraintValidatorFactory constraintValidatorFactory;
@@ -126,19 +124,9 @@ public class ValidationContext<T> {
 	private final TraversableResolver traversableResolver;
 
 	/**
-	 * Parameter name provider which should be used in this context.
+	 * Context containing all validator level configurations.
 	 */
-	private final ExecutableParameterNameProvider parameterNameProvider;
-
-	/**
-	 * Clock provider which should be used in this context.
-	 */
-	private final ClockProvider clockProvider;
-
-	/**
-	 * Whether or not validation should fail on the first constraint violation.
-	 */
-	private final boolean failFast;
+	private final ValidatorScopedContext validatorScopedContext;
 
 	/**
 	 * The constraint validator initialization context.
@@ -151,13 +139,10 @@ public class ValidationContext<T> {
 	private String validatedProperty;
 
 	private ValidationContext(ConstraintValidatorManager constraintValidatorManager,
-			MessageInterpolator messageInterpolator,
 			ConstraintValidatorFactory constraintValidatorFactory,
+			ValidatorScopedContext validatorScopedContext,
 			TraversableResolver traversableResolver,
-			ExecutableParameterNameProvider parameterNameProvider,
-			ClockProvider clockProvider,
 			HibernateConstraintValidatorInitializationContext constraintValidatorInitializationContext,
-			boolean failFast,
 			T rootBean,
 			Class<T> rootBeanClass,
 			BeanMetaData<T> rootBeanMetaData,
@@ -165,13 +150,10 @@ public class ValidationContext<T> {
 			Object[] executableParameters,
 			Object executableReturnValue) {
 		this.constraintValidatorManager = constraintValidatorManager;
-		this.messageInterpolator = messageInterpolator;
+		this.validatorScopedContext = validatorScopedContext;
 		this.constraintValidatorFactory = constraintValidatorFactory;
 		this.traversableResolver = traversableResolver;
-		this.parameterNameProvider = parameterNameProvider;
-		this.clockProvider = clockProvider;
 		this.constraintValidatorInitializationContext = constraintValidatorInitializationContext;
-		this.failFast = failFast;
 
 		this.rootBean = rootBean;
 		this.rootBeanClass = rootBeanClass;
@@ -189,22 +171,18 @@ public class ValidationContext<T> {
 	public static ValidationContextBuilder getValidationContextBuilder(
 			BeanMetaDataManager beanMetaDataManager,
 			ConstraintValidatorManager constraintValidatorManager,
-			MessageInterpolator messageInterpolator,
 			ConstraintValidatorFactory constraintValidatorFactory,
+			ValidatorScopedContext validatorScopedContext,
 			TraversableResolver traversableResolver,
-			ClockProvider clockProvider,
-			HibernateConstraintValidatorInitializationContext constraintValidatorInitializationContext,
-			boolean failFast) {
+			HibernateConstraintValidatorInitializationContext constraintValidatorInitializationContext) {
 
 		return new ValidationContextBuilder(
 				beanMetaDataManager,
 				constraintValidatorManager,
-				messageInterpolator,
 				constraintValidatorFactory,
+				validatorScopedContext,
 				traversableResolver,
-				clockProvider,
-				constraintValidatorInitializationContext,
-				failFast
+				constraintValidatorInitializationContext
 		);
 	}
 
@@ -229,7 +207,7 @@ public class ValidationContext<T> {
 	}
 
 	public boolean isFailFastModeEnabled() {
-		return failFast;
+		return validatorScopedContext.isFailFast();
 	}
 
 	public ConstraintValidatorManager getConstraintValidatorManager() {
@@ -244,15 +222,15 @@ public class ValidationContext<T> {
 	 * created for parameter validation, {@code null} otherwise.
 	 */
 	public List<String> getParameterNames() {
-		if ( parameterNameProvider == null ) {
+		if ( executableParameters == null ) {
 			return null;
 		}
 
-		return parameterNameProvider.getParameterNames( executable );
+		return validatorScopedContext.getParameterNameProvider().getParameterNames( executable );
 	}
 
 	public ClockProvider getClockProvider() {
-		return clockProvider;
+		return validatorScopedContext.getClockProvider();
 	}
 
 	public HibernateConstraintValidatorInitializationContext getConstraintValidatorInitializationContext() {
@@ -399,7 +377,7 @@ public class ValidationContext<T> {
 		);
 
 		try {
-			return messageInterpolator.interpolate(
+			return validatorScopedContext.getMessageInterpolator().interpolate(
 					messageTemplate,
 					context
 			);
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -91,7 +90,7 @@
 	private final Object executableReturnValue;
 
 	/**
-	 * The set of already processed unit of works. See {@link ProcessedUnit}.
+	 * The set of already processed unit of works. See {@code ProcessedUnit}.
```
gsmet: Didn't you change that in another commit by pointing to the real classes?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -121,19 +115,9 @@
 	private final TraversableResolver traversableResolver;
 
 	/**
-	 * Parameter name provider which should be used in this context.
-	 */
-	private final ExecutableParameterNameProvider parameterNameProvider;
-
-	/**
-	 * Clock provider which should be used in this context.
+	 * Context containing all validator level configuraions .
```
gsmet: Typo in `configuraions`

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -459,44 +431,34 @@ private void markCurrentBeanAsProcessedForCurrentGroup(Object bean, Class<?> gro
...
 		private ValidationContextBuilder(
 				BeanMetaDataManager beanMetaDataManager,
 				ConstraintValidatorManager constraintValidatorManager,
-				MessageInterpolator messageInterpolator,
 				ConstraintValidatorFactory constraintValidatorFactory,
-				TraversableResolver traversableResolver,
-				ClockProvider clockProvider,
 				HibernateConstraintValidatorInitializationContext constraintValidatorInitializationContext,
-				boolean failFast) {
+				ValidatorContextProperties validatorContextProperties) {
```
gsmet: Let's put it before the initialization context as it's more general.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -459,44 +431,34 @@ private void markCurrentBeanAsProcessedForCurrentGroup(Object bean, Class<?> gro
...
 			this.beanMetaDataManager = beanMetaDataManager;
 			this.constraintValidatorManager = constraintValidatorManager;
-			this.messageInterpolator = messageInterpolator;
 			this.constraintValidatorFactory = constraintValidatorFactory;
-			this.traversableResolver = traversableResolver;
-			this.clockProvider = clockProvider;
+			this.traversableResolver = validatorContextProperties.getWrapedWithCachingForSingleValidation();
```
gsmet: I don't think it's a good idea to incorporate the `TraversableResolver` in it. It might not be `Validator` scoped so I don't think it's such a good idea to make it look like it.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextProperties.java
```diff
@@ -0,0 +1,173 @@
...
+
+	public boolean isTraversableResolverResultCacheEnabled() {
+		return this.traversableResolverResultCacheEnabled;
+	}
+
+	public TraversableResolver getWrapedWithCachingForSingleValidation() {
+		return TraversableResolvers.wrapWithCachingForSingleValidation( traversableResolver, traversableResolverResultCacheEnabled );
+	}
+
+	public static class Builder {
```
gsmet: Wondering if this builder pattern is bringing something to the plate. Looks like it's just noise. Moreover it might be error prone as we might forget to set up one of the necessary elements.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextProperties.java
```diff
@@ -0,0 +1,173 @@
...
+import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
+import org.hibernate.validator.spi.scripting.ScriptEvaluatorFactory;
+
+/**
+ * Context object keeping storing all possible validator configurable properties.
+ * There should be one per {@link Validator} instance.
+ *
+ * @author Marko Bekhta
+ */
+public class ValidatorContextProperties {
```
gsmet: I don't like the name but I can see how it is difficult to find a good name given we already have a `ValidatorContext` in `javax.validation`. I can't think of anything better than `ValidatorContext` though.

That being said, if we take into account my remark about limiting the class to `ValidatorImpl` and `ValidationContext`, I think that might be acceptable to name it `ValidatorContext` as the class will be very localized.

@gunnarmorling thoughts?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextProperties.java
```diff
@@ -0,0 +1,173 @@
...
+
+	/**
+	 * The default message interpolator for this factory.
+	 */
+	private final MessageInterpolator messageInterpolator;
+
+	/**
+	 * The default traversable resolver for this factory.
+	 */
+	private final TraversableResolver traversableResolver;
```
gsmet: As mentioned earlier, I think it would be a mistake to include this one. Let's keep it out of it to make it clear it can be validation scoped.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextProperties.java
```diff
@@ -0,0 +1,173 @@
...
+
+	public boolean isFailFast() {
+		return this.failFast;
+	}
+
+	public boolean isTraversableResolverResultCacheEnabled() {
+		return this.traversableResolverResultCacheEnabled;
+	}
+
+	public TraversableResolver getWrapedWithCachingForSingleValidation() {
```
gsmet: There are 2 p at `Wraped` but I think this method will go away.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -69,42 +69,16 @@
  * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
  * @author Chris Beckey &lt;cbeckey@paypal.com&gt;
  * @author Guillaume Smet
+ * @author Marko Bekhta
  */
 public class ValidatorFactoryImpl implements HibernateValidatorFactory {
```
gsmet: I think I would have kept the changes to `ValidatorImpl` and I wouldn't have included `ValidatorFactoryImpl` and `ValidatorContextImpl` in the change. I would have built the context in the `ValidatorImpl` constructor.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -238,15 +210,15 @@ public ConstraintValidatorManager getConstraintValidatorManager() {
 	 * created for parameter validation, {@code null} otherwise.
 	 */
 	public List<String> getParameterNames() {
-		if ( parameterNameProvider == null ) {
+		if ( executableParameters == null ) {
```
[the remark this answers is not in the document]
marko-bekhta: well ... honestly it bugs me, I mean not this particular change but Validation context. In particular two things:
- this null check was previously for the name provider as we were not passing it if the context wasn't for the parameters
- and `createConstraintViolation` method, this part in particular https://github.com/marko-bekhta/hibernate-validator/blob/d976b480573871c37c844ba8f44ffa17b268d9b6/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java#L294 (If we have a `@NotNull` on return value and the value is actually null it will slip this if and fall to the last one. It is not visible in a constraint but still it kind of bugs me...)

So as I'm not sure about it I wanted to ask you - if it would be a good idea to split this context into 3 different ones ?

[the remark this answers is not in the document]
gsmet: Yes, you're right, the test in `createConstraintViolation()` is buggy.

About separating `ValidationContext` in three contexts, I think it's too big of a change for a minor. We could maybe consider it for 6.1 but I'm not sure it's worth it.

I think what we are missing here is a `ValidationType` enum to define the type of validation currently at stake. Then we could base the conditions on the type and that would be totally safe. It would be a private member of `ValidationContext` so the change would be very local.

WDYT?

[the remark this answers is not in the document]
marko-bekhta: sounds good. I'll add an enum for now and create a ticket for splitting the context, so we can decide later if it is worth it or not ...

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextProperties.java
marko-bekhta: Just needed to use something different than `ValidatorContext` :). How about `InstanceScopeValidatorContext` ?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextProperties.java
marko-bekhta: well, I just don't like constructors with a huge amount of parameters :), but other than that, I think  there's not much benefit from it. Let me know if I should just remove it.

gsmet: I tend to agree but I think it would be an error in this case. We instantiate the context in several places and with the builder, we could forget one of the arguments.

I think in this case, a multi-parameters constructor is best.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextProperties.java
gsmet: Yuk :).

Maybe `ValidatorScopedContext`? And maybe we make it a package protected inner class of `ValidationContext`and reference it as `ValidationContext.ValidatorScopedContext`?

That's the best I can think of right now.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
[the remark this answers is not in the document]
gsmet: Maybe, it's better to talk about "operation" instead of "type".

marko-bekhta:
I've rebased to current master and reorganized the commits a bit:
- First one contains only changes for `ValidatorScopedContext` with the fixes to the things you've mentioned.
- In the next one I've cleaned up a constructor of `ValidationFactory` so now all configuration property extraction should follow the same pattern.
- And in the last one - I've added `ValidatioOperation` enum as were suggested.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextImpl.java
```diff
@@ -188,4 +188,87 @@ public Validator getValidator() {
 				traversableResolverResultCacheEnabled
 		);
 	}
+
+	/**
+	 * Context object storing all possible validator configurable properties.
+	 * There should be just one per {@link Validator} instance.
+	 */
+	static class ValidatorScopedContext {
```
gunnarmorling: Maybe just `ValidatorContext`? And how about adding all the things scoped to the VF, too (or having another, dedicated context for that)? That way we'd get back to much leaner signatures of the methods passing through these objects.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -238,15 +220,15 @@ public ConstraintValidatorManager getConstraintValidatorManager() {
 	 * created for parameter validation, {@code null} otherwise.
 	 */
 	public List<String> getParameterNames() {
-		if ( parameterNameProvider == null ) {
+		if ( !ValidationOperation.PARAMETER_VALIDATION.equals( validationOperation ) ) {
```
gunnarmorling: Or even raise an exception? It seems calling this method in this case would indicate a programming error?

gsmet: +1

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorContextImpl.java
gsmet: We already have `ValidatorContext` in the spec. That's why we tried to have another name.

@marko-bekhta my idea was to put this one in `ValidationContext` not `ValidatorContextImpl`. I don't think it has anything to do with `ValidatorContextImpl`.

gsmet: > And how about adding all the things scoped to the VF, too (or having another, dedicated context for that)? That way we'd get back to much leaner signatures of the methods passing through these objects.

I haven't thought about this yet. But, anyway, let's keep it for another PR.

marko-bekhta: @gsmet that was my mistake - classes are together and I dropped the new one into a wrong one. 
@gunnarmorling +1 for updating the factory as well. Let's do it in another PR.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
marko-bekhta: This is, kind of, related to suggestion to split the `ValidationContext` into multiple ones (https://hibernate.atlassian.net/browse/HV-1526). The way how this method is used:
```java
// create a constraint validator context
ConstraintValidatorContextImpl constraintValidatorContext = new ConstraintValidatorContextImpl(
		validationContext.getParameterNames(),
		validationContext.getClockProvider(),
		valueContext.getPropertyPath(),
		descriptor
);
```
wouldn't allow to just throw an exception here. So I was thinking that if we create different kinds of `ValidationContext`s we could then also have different implementations of `HibernateConstraintValidatorContext` and then we wouldn't need such kind of checks here, as well as in `ValidationContext#createConstraintViolation()`.

gsmet:
Added a commit with cosmetic improvements and comment fixes and merged. Thanks!

PR #886, opened 2017-11-18
HV-1522 Improve performance of CV hashCode method

marko-bekhta:
So the problematic method turned out to be `System#identityHashCode()`. I gave it some thought and I think that we can skip evaluating those hashcodes. It will not break the hash/equals contract only might cause additional calls to equals. As for the equals method, I thought about two things 

- the order of those "if checks" is important so the less expensive checks should go first 
- using equals methods on value/rootBean/leafBeanInstance is probably not good. We don't know the implementation of those methods, hence their performance, as they are user defined. I think that for the CV it is enough to do == check instead. 

after such changes the profiling picture looks like this:
hash_after_change
`ConstraintViolationImpl#hashCode` completely disappeared. And here are some results running the `SimpleValidation#testSimpleBeanValidation` before the change:
```
# Run complete. Total time: 00:00:37

Benchmark                                   Mode  Cnt    Score    Error   Units
SimpleValidation.testSimpleBeanValidation  thrpt   20  585.949 ± 42.153  ops/ms
```

and after:
```
# Run complete. Total time: 00:00:36

Benchmark                                   Mode  Cnt    Score    Error   Units
SimpleValidation.testSimpleBeanValidation  thrpt   20  657.163 ± 53.447  ops/ms
```

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
@@ -278,28 +278,25 @@ public class ConstraintViolationImpl<T> implements HibernateConstraintViolation<
 		if ( interpolatedMessage != null ? !interpolatedMessage.equals( that.interpolatedMessage ) : that.interpolatedMessage != null ) {
 			return false;
 		}
+		if ( messageTemplate != null ? !messageTemplate.equals( that.messageTemplate ) : that.messageTemplate != null ) {
+			return false;
+		}
 		if ( propertyPath != null ? !propertyPath.equals( that.propertyPath ) : that.propertyPath != null ) {
 			return false;
 		}
-		if ( rootBean != null ? !rootBean.equals( that.rootBean ) : that.rootBean != null ) {
+		if ( rootBean != null ? ( rootBean != that.rootBean ) : that.rootBean != null ) {
 			return false;
 		}
-		if ( leafBeanInstance != null ? !leafBeanInstance.equals( that.leafBeanInstance ) : that.leafBeanInstance != null ) {
+		if ( leafBeanInstance != null ? ( leafBeanInstance != that.leafBeanInstance ) : that.leafBeanInstance != null ) {
+			return false;
+		}
+		if ( value != null ? ( value != that.value ) : that.value != null ) {
 			return false;
 		}
 		if ( constraintDescriptor != null ? !constraintDescriptor.equals( that.constraintDescriptor ) : that.constraintDescriptor != null ) {
 			return false;
 		}
 		if ( elementType != null ? !elementType.equals( that.elementType ) : that.elementType != null ) {
-			return false;
-		}
-		if ( messageTemplate != null ? !messageTemplate.equals( that.messageTemplate ) : that.messageTemplate != null ) {
-			return false;
-		}
-		if ( rootBeanClass != null ? !rootBeanClass.equals( that.rootBeanClass ) : that.rootBeanClass != null ) {
-			return false;
-		}
-		if ( value != null ? !value.equals( that.value ) : that.value != null ) {
 			return false;
 		}
 
@@ -334,7 +331,6 @@ public class ConstraintViolationImpl<T> implements HibernateConstraintViolation<
 		result = 31 * result + System.identityHashCode( value );
 		result = 31 * result + ( constraintDescriptor != null ? constraintDescriptor.hashCode() : 0 );
 		result = 31 * result + ( messageTemplate != null ? messageTemplate.hashCode() : 0 );
-		result = 31 * result + ( rootBeanClass != null ? rootBeanClass.hashCode() : 0 );
 		result = 31 * result + ( elementType != null ? elementType.hashCode() : 0 );
 		return result;
 	}
--- a/engine/src/main/java/org/hibernate/validator/messageinterpolation/AbstractMessageInterpolator.java
+++ b/engine/src/main/java/org/hibernate/validator/messageinterpolation/AbstractMessageInterpolator.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.validator.messageinterpolation;
 
+import static org.hibernate.validator.internal.util.ConcurrentReferenceHashMap.ReferenceType.SOFT;
+
 import java.lang.invoke.MethodHandles;
 import java.util.EnumSet;
 import java.util.List;
@@ -30,8 +32,6 @@ import org.hibernate.validator.internal.util.logging.LoggerFactory;
 import org.hibernate.validator.resourceloading.PlatformResourceBundleLocator;
 import org.hibernate.validator.spi.resourceloading.ResourceBundleLocator;
 
-import static org.hibernate.validator.internal.util.ConcurrentReferenceHashMap.ReferenceType.SOFT;
-
 /**
  * Resource bundle backed message interpolator.
  *
@@ -39,6 +39,7 @@ import static org.hibernate.validator.internal.util.ConcurrentReferenceHashMap.R
  * @author Hardy Ferentschik
  * @author Gunnar Morling
  * @author Adam Stawicki
+ * @author Marko Bekhta
  *
  * @since 5.2
  */
@@ -238,9 +239,9 @@ public abstract class AbstractMessageInterpolator implements MessageInterpolator
 
 	/**
 	 * Runs the message interpolation according to algorithm specified in the Bean Validation specification.
-	 * <br/>
+	 * <p>
 	 * Note:
-	 * <br/>
+	 * <p>
 	 * Look-ups in user bundles is recursive whereas look-ups in default bundle are not!
 	 *
 	 * @param message the message to interpolate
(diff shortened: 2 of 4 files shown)


gsmet:
Hum. This is an interesting one. Out of curiosity, which JDK and which version of VisualVM do you use to get these results?

I made a lot of profiling and never got this profile (I would have thought it might be a profiling artifact but your benchmarking seems to prove at least partially your point).

Another thing we could improve is `AbstractMessageInterpolator#replaceEscapedLiterals()`. First I think we should try to replace the four regexps with only one capturing the replacement character and see if it improves things. I don't think the order has an importance? At least in the spec, it's not said that the replacement should be made in any particular order (http://beanvalidation.org/2.0/spec/#validationapi-message-defaultmessageinterpolation). WDYT?

Btw, it would be interesting to see if https://github.com/greenlaw110/java-str-benchmark is right and if we should import the string replacement method of https://github.com/osglworks/java-tool/tree/develop . The license is Apache 2 so it wouldn't be an issue (the Fast method in this benchmark is LGPL 3 and I'm not sure we could import the code in HV - pretty sure we can't).

I would do that in steps:
 * see if only one regexp with capturing is faster than 4
 * replace the fastest way with OSGL replace method and see what's going on
 * keep the fastest
(if we go with the OSGL method, it solves the order question altogether!)

We could also test if the string contains the char '\' before replacing anything here: https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/messageinterpolation/AbstractMessageInterpolator.java#L352 . I don't think it will add a lot of overhead and that would avoid the replacement altogether.

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -329,9 +329,6 @@ public String toString() {
 	private int createHashCode() {
 		int result = interpolatedMessage != null ? interpolatedMessage.hashCode() : 0;
 		result = 31 * result + ( propertyPath != null ? propertyPath.hashCode() : 0 );
-		result = 31 * result + System.identityHashCode( rootBean );
-		result = 31 * result + System.identityHashCode( leafBeanInstance );
-		result = 31 * result + System.identityHashCode( value );
```
gsmet: Could you try to benchmark with and without it?

From my previous experiments, I often had `System.identityHashCode()` in my profiling but it was more of an artifact.

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -278,28 +278,28 @@ public boolean equals(Object o) {
 		if ( interpolatedMessage != null ? !interpolatedMessage.equals( that.interpolatedMessage ) : that.interpolatedMessage != null ) {
 			return false;
 		}
-		if ( propertyPath != null ? !propertyPath.equals( that.propertyPath ) : that.propertyPath != null ) {
+		if ( messageTemplate != null ? !messageTemplate.equals( that.messageTemplate ) : that.messageTemplate != null ) {
```
gsmet: Interesting change.

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -278,28 +278,28 @@ public boolean equals(Object o) {
...
 			return false;
 		}
-		if ( rootBean != null ? !rootBean.equals( that.rootBean ) : that.rootBean != null ) {
+		if ( propertyPath != null ? !propertyPath.equals( that.propertyPath ) : that.propertyPath != null ) {
 			return false;
 		}
-		if ( leafBeanInstance != null ? !leafBeanInstance.equals( that.leafBeanInstance ) : that.leafBeanInstance != null ) {
+		if ( rootBean != null ? ( rootBean != that.rootBean ) : that.rootBean != null ) {
```
gsmet: Hmmm, you're right about that. I think it's even a bug to use `equals()`. Everywhere we used `System.identityHashCode()` for the hash code computation, we should use `==`.

There might be other occurrences of this pattern in the code base.

gsmet:
And that lets us with `BeanPathMetaConstraintProcessedUnit#createhashCode()`. I'm a bit surprised it takes so much time in the profiling. Wondering if it's not a profiling artifact. This `Path` thing is really a pain...

Another thing that surprises me is that you have `CachingTraversableResolverForSingleValidation` in the profile. Do we have the JPA API in the classpath when running the benchmarks? it shouldn't be there if we don't have the JPA APi present.

gsmet:
Crazy idea but I'm wondering if we should try to base the `hashCode` of `BeanPathMetaConstraintProcessedUnit` on the `hasCode` of `toString()` of the `Path` instead of its `hashCode`.

There shouldn't be that much collision and it might be faster. It's worth a test IMHO. We would have to keep the `toString()` of the `Path` sensible but I think we would want that anyway.

Another idea would be to try to understand what is so slow in this `hashCode` (if it's not a profiling artifact, of course) and see if the `hashCode` still makes sense without these elements.

marko-bekhta:
Hi @gsmet , finally got to this...
So the VisualVm is `1.3.9` and the JDK:
```
java version "1.8.0_151"
Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.151-b12, mixed mode)
```

As for the `CachingTraversableResolverForSingleValidation` it's probably because I was running a test from engine, as I was also doing some debugging and trying to find why there's a difference, so I used the same test to run some profiling. 

I didn't have much time to run the tests for changes to message interpolation that you've suggested. From what I was able to do it looked that there's no much difference ... But nevertheless I'll try to do it more properly, maybe later today.

Also not entirely related to the current discussion... I was thinking that all these hash computations comes from the fact that we are using `Set` and `Map` in the code quite a lot. I understand that we wouldn't be able to change the public facing API but what about internals? Do you think that using for example `List` in internals might help? And then only if a `Set` is needed to be returned to a  user we would do a conversion?

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
marko-bekhta: Do you mean just with this change ? or a simple before the changes from this PR ? If it's the later one, than here's a result:
old_hash_impl
I've attached it to the JIRA but not here ...

gsmet:
OK, not so surprised about the message interpolator thing. It's the issue when doing profiling, sometimes you have things that pop up but are not really in the way. It's really funny your profile is so much different from the one I have here. That being said, I really think at least adding a test to check if the string contains `\` would be beneficial.

About using `List` instead of `Set`, you have 2 cases:
- case where we could have `List` and make the transformation at the end: but in this case, it won't change anything as we only do the computation once (we add every element to the `Set` and there's very little chance there is an already existing element - CV for instance);
- case where we use the `Set`/`Map` as a map (with `contains()`/`containsKey()`) and there's a good chance a `List` would be slower in this case (case of the *ProcessedUnit).

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
gsmet: I would like to see the results with all your changes except this one. From what I've seen in my benchmarking campaings (and Sanne confirmed it), `System.identityHashCode()` is often high in the profile for now good reasons. So I'm wondering what would be the result if we keep them here but apply all your other changes.

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -278,28 +278,28 @@ public boolean equals(Object o) {
...
-		if ( messageTemplate != null ? !messageTemplate.equals( that.messageTemplate ) : that.messageTemplate != null ) {
+		if ( constraintDescriptor != null ? !constraintDescriptor.equals( that.constraintDescriptor ) : that.constraintDescriptor != null ) {
 			return false;
 		}
-		if ( rootBeanClass != null ? !rootBeanClass.equals( that.rootBeanClass ) : that.rootBeanClass != null ) {
+		if ( elementType != null ? !elementType.equals( that.elementType ) : that.elementType != null ) {
 			return false;
 		}
-		if ( value != null ? !value.equals( that.value ) : that.value != null ) {
+		if ( rootBeanClass != null ? !rootBeanClass.equals( that.rootBeanClass ) : that.rootBeanClass != null ) {
```
marko-bekhta: One more thing .... Do we need to check this at all ? if rootBean != that.rootBean  CV are not equals and if rootBean == that.rootBean classes should already be equal so no need to check, right ?

gsmet: Hmmm, yes, I think you're right.

marko-bekhta:
So here are some before and afters:
```
Current master:

  678.539 ±(99.9%) 8.915 ops/ms [Average]
  (min, avg, max) = (649.331, 678.539, 689.641), stdev = 10.267
  CI (99.9%): [669.624, 687.454] (assumes normal distribution)


# Run complete. Total time: 00:00:35

Benchmark                                   Mode  Cnt    Score   Error   Units
SimpleValidation.testSimpleBeanValidation  thrpt   20  678.539 ± 8.915  ops/ms

-----------------------------------------------------------------
with changes from HV-1522 Updating CV hash/equals method 

  835.344 ±(99.9%) 9.710 ops/ms [Average]
  (min, avg, max) = (816.321, 835.344, 854.220), stdev = 11.182
  CI (99.9%): [825.634, 845.054] (assumes normal distribution)


# Run complete. Total time: 00:00:34

Benchmark                                   Mode  Cnt    Score   Error   Units
SimpleValidation.testSimpleBeanValidation  thrpt   20  835.344 ± 9.710  ops/ms
```
and also screens from visualvm: 
Before
before
and after:
after

These are probably more similar to the ones that you've had :smiley: 

I also found two more places where equals probably shouldn't be used so I changed them in the second commit. Also it's probably nothing but message key construction can be omitted if cache is disabled (so done that in the second commit as well)

engine/src/main/java/org/hibernate/validator/messageinterpolation/AbstractMessageInterpolator.java
```diff
@@ -251,10 +251,11 @@ public String interpolate(String message, Context context, Locale locale) {
 	 */
 	private String interpolateMessage(String message, Context context, Locale locale)
 			throws MessageDescriptorFormatException {
-		LocalizedMessage localisedMessage = new LocalizedMessage( message, locale );
+		LocalizedMessage localisedMessage = null;
 		String resolvedMessage = null;
 
 		if ( cachingEnabled ) {
+			localisedMessage = new LocalizedMessage( message, locale );
```
gsmet: Mmmh, this is a bold move... I don't like this pattern at all as the compiler is all lost.

I think the issue comes from the fact that we didn't have Java 8 at this time.

What about:
```java
		String resolvedMessage = null;

		if ( cachingEnabled ) {
			resolvedMessage = resolvedMessages.computeIfAbsent( new LocalizedMessage( message, locale ), lm -> resolveMessage( message, locale ) );
		} else {
			resolvedMessage = resolveMessage( message, locale );
		}
```

with `resolveMessage()` being:

```java
private String resolveMessage(String message, Locale locale) {
		String resolvedMessage = message;

		ResourceBundle userResourceBundle = userResourceBundleLocator
				.getResourceBundle( locale );

		ResourceBundle constraintContributorResourceBundle = contributorResourceBundleLocator
				.getResourceBundle( locale );

		ResourceBundle defaultResourceBundle = defaultResourceBundleLocator
				.getResourceBundle( locale );

		String userBundleResolvedMessage;
		boolean evaluatedDefaultBundleOnce = false;
		do {
			// search the user bundle recursive (step1)
			userBundleResolvedMessage = interpolateBundleMessage(
					resolvedMessage, userResourceBundle, locale, true
			);

			// search the constraint contributor bundle recursive (only if the user did not define a message)
			if ( !hasReplacementTakenPlace( userBundleResolvedMessage, resolvedMessage ) ) {
				userBundleResolvedMessage = interpolateBundleMessage(
						resolvedMessage, constraintContributorResourceBundle, locale, true
				);
			}

			// exit condition - we have at least tried to validate against the default bundle and there was no
			// further replacements
			if ( evaluatedDefaultBundleOnce
					&& !hasReplacementTakenPlace( userBundleResolvedMessage, resolvedMessage ) ) {
				break;
			}

			// search the default bundle non recursive (step2)
			resolvedMessage = interpolateBundleMessage(
					userBundleResolvedMessage,
					defaultResourceBundle,
					locale,
					false
			);
			evaluatedDefaultBundleOnce = true;
		} while ( true );
		
		return resolvedMessage;
	}
```

(we probably need to rework the comments with the steps)

marko-bekhta: +1 that would be better

engine/src/main/java/org/hibernate/validator/messageinterpolation/AbstractMessageInterpolator.java
```diff
@@ -356,10 +357,12 @@ private String interpolateMessage(String message, Context context, Locale locale
 	}
 
 	private String replaceEscapedLiterals(String resolvedMessage) {
-		resolvedMessage = LEFT_BRACE.matcher( resolvedMessage ).replaceAll( "{" );
-		resolvedMessage = RIGHT_BRACE.matcher( resolvedMessage ).replaceAll( "}" );
-		resolvedMessage = SLASH.matcher( resolvedMessage ).replaceAll( Matcher.quoteReplacement( "\\" ) );
-		resolvedMessage = DOLLAR.matcher( resolvedMessage ).replaceAll( Matcher.quoteReplacement( "$" ) );
+		if ( resolvedMessage.indexOf( '\\' ) > -1 ) {
```
gsmet: BTW, I think we could use the same optimization before triggering the interpolation mechanisms: we don't need to interpolate if the string does not contain '{' - in fact, we could even skip the EL interpolation if the string does not contain "${" but I'm not sure it's worth the additional call to `contains()`.

It's probably worth a quick benchmark to be sure the additional `indexOf()` does not slow us down.

engine/src/main/java/org/hibernate/validator/messageinterpolation/AbstractMessageInterpolator.java
gsmet: Mmmh, and we could probably also use `computeIfAbsent()` for `tokenizedParameterMessages` and `tokenizedELMessages`. It will make the code more readable.

gsmet:
@marko-bekhta btw, for your profiling, I recommend you to wait the end of the warmup phase before attaching your profiler.

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -329,10 +326,10 @@ public String toString() {
 	private int createHashCode() {
 		int result = interpolatedMessage != null ? interpolatedMessage.hashCode() : 0;
 		result = 31 * result + ( propertyPath != null ? propertyPath.hashCode() : 0 );
+		result = 31 * result + ( constraintDescriptor != null ? constraintDescriptor.hashCode() : 0 );
 		result = 31 * result + System.identityHashCode( rootBean );
 		result = 31 * result + System.identityHashCode( leafBeanInstance );
 		result = 31 * result + System.identityHashCode( value );
-		result = 31 * result + ( constraintDescriptor != null ? constraintDescriptor.hashCode() : 0 );
 		result = 31 * result + ( messageTemplate != null ? messageTemplate.hashCode() : 0 );
 		result = 31 * result + ( rootBeanClass != null ? rootBeanClass.hashCode() : 0 );
```
gsmet: Shouldn't we remove it from the hashCode too?

marko-bekhta: removed this one as well.

marko-bekhta:
And here are some before/after with the latest changes:
```
before

  1470.870 ±(99.9%) 41.745 ops/ms [Average]
  (min, avg, max) = (1328.400, 1470.870, 1513.702), stdev = 48.074
  CI (99.9%): [1429.124, 1512.615] (assumes normal distribution)


# Run complete. Total time: 00:00:32

Benchmark                                   Mode  Cnt     Score    Error   Units
SimpleValidation.testSimpleBeanValidation  thrpt   20  1470.870 ± 41.745  ops/ms
---------------------------------------------------------------------------------------------
after:

  1986.396 ±(99.9%) 30.573 ops/ms [Average]
  (min, avg, max) = (1907.107, 1986.396, 2040.153), stdev = 35.208
  CI (99.9%): [1955.823, 2016.970] (assumes normal distribution)


Benchmark                                   Mode  Cnt     Score    Error   Units
SimpleValidation.testSimpleBeanValidation  thrpt   20  1986.396 ± 30.573  ops/ms
```

gsmet:
Amended the commit messages and merged. Nice team work!

PR #887, opened 2017-11-20
HV-1524 Log temporal validation tolerance only if it has been explicitly set

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1524

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
@@ -572,7 +572,6 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 			}
 		}
 
-		LOG.logTemporalValidationTolerance( Duration.ZERO );
 		return Duration.ZERO;
 	}
 


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -556,23 +556,26 @@ private Duration getTemporalValidationTolerance(ConfigurationState configuration
 		if ( configurationState instanceof ConfigurationImpl ) {
 			ConfigurationImpl hibernateSpecificConfig = (ConfigurationImpl) configurationState;
 			if ( hibernateSpecificConfig.getTemporalValidationTolerance() != null ) {
-				LOG.logTemporalValidationTolerance( hibernateSpecificConfig.getTemporalValidationTolerance() );
+				if ( !Duration.ZERO.equals( hibernateSpecificConfig.getTemporalValidationTolerance() ) ) {
+					LOG.logTemporalValidationTolerance( hibernateSpecificConfig.getTemporalValidationTolerance() );
+				}
```
gsmet: Is it null if it hasn't been defined at all? If so, maybe we should consider logging it anyway as it has been explicitly set and only remove the `LOG.logTemporalValidationTolerance( Duration.ZERO );` line below?

marko-bekhta: Yes, I actually also thought about this while walking home - we should log the value even if it's 0 but it was explicitly defined. Cleaned up the commit and forced pushed.

gsmet:
Amended the commit message and merged. Thanks!

PR #888, opened 2017-11-22
HV-1527 Separate Group and Path processed units in ValidationContext

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1527

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
@@ -91,9 +91,14 @@ public class ValidationContext<T> {
 	private final Object executableReturnValue;
 
 	/**
-	 * The set of already processed unit of works. See {@link ProcessedUnit}.
+	 * The set of already processed path meta constraint units ({@link BeanPathMetaConstraintProcessedUnit}).
 	 */
-	private final Set<Object> processedUnits;
+	private final Set<BeanPathMetaConstraintProcessedUnit> processedPathUnits;
+
+	/**
+	 * The set of already processed group units ({@link BeanGroupProcessedUnit}).
+	 */
+	private final Set<BeanGroupProcessedUnit> processedGroupUnits;
 
 	/**
 	 * Maps an object to a list of paths in which it has been validated. The objects are the bean instances.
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -91,9 +91,15 @@
 	private final Object executableReturnValue;
 
 	/**
-	 * The set of already processed unit of works. See {@link ProcessedUnit}.
+	 * The sets of already processed path meta constraint unit of works({@link BeanPathMetaConstraintProcessedUnit})
+	 * {@link }.
```
gsmet: There's something wrong with the javadoc here.

+ I think you can remove "of works" and make unit a plural.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -91,9 +91,15 @@
...
 	 */
-	private final Set<Object> processedUnits;
+	private final Set<BeanPathMetaConstraintProcessedUnit> processedPathUnits;
+
+	/**
+	 * The sets of already processed group unit of works({@link BeanGroupProcessedUnit})
```
gsmet: Same here about the second remark.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -611,13 +618,8 @@ private BeanGroupProcessedUnit(Object bean, Class<?> group) {
 
 		@Override
 		public boolean equals(Object o) {
-			if ( this == o ) {
-				return true;
-			}
```
gsmet: Do we really gain something by removing this? If not I would vote for keeping it (and update the comment below).

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -659,22 +661,17 @@ private BeanPathMetaConstraintProcessedUnit(Object bean, Path path, MetaConstrai
...
```
gsmet: Same here.

marko-bekhta: I've checked. It don't really make a difference, results are almost identical. So I've added these back.

gsmet:
Jenkins, retest this please

gsmet:
OK, looks good now. I'll rework some of the comments and merge tomorrow morning.

Thanks!

gsmet:
Merged, thanks!

PR #889, opened 2017-11-24
HV-1444 Rebased/Squashed version of Marko's PR

gsmet:
* https://github.com/hibernate/hibernate-validator/pull/878
 * https://hibernate.atlassian.net/browse/HV-1444


marko-bekhta:
and some numbers:
```
with changes applied:
# Run complete. Total time: 00:01:51

Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  241.063 ±  8.683  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  613.279 ± 10.496   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  821.396 ±  9.824   ops/s


  Raw validation time:        1096ms {12342}
  Raw parsing time:              0ms {0}
  First parse and validate:      0ms {0}
  Multithreaded validation:      0ms 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.192 s - in org.apache.bval.bench.BenchTest

======================================================
current (without last commit on potentially CMD):
# Run complete. Total time: 00:01:52

Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  212.603 ± 26.591  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  592.819 ± 20.631   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  785.133 ± 13.543   ops/s


  Raw validation time:         884ms {12342}
  Raw parsing time:              0ms {0}
  First parse and validate:      0ms {0}
  Multithreaded validation:      0ms 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.143 s - in org.apache.bval.bench.BenchT

======================================================
master:

# Run complete. Total time: 00:01:53

Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  226.021 ± 37.921  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  607.903 ±  8.669   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  860.026 ± 11.842   ops/s



  Raw validation time:         903ms {12342}
  Raw parsing time:              0ms {0}
  First parse and validate:      0ms {0}
  Multithreaded validation:      0ms 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.113 s - in org.apache.bval.bench.BenchTest
```

gsmet:
Mmmh, did you mess up with the BV benchmark number or is your last version slower than everything and the current version faster than master?

it looks weird to me.

marko-bekhta:
it's possible that I did ... So I've run the tests again :)
```
current master:
  Raw validation time:         946ms {12342}
  Raw parsing time:              0ms {0}
  First parse and validate:      0ms {0}
  Multithreaded validation:      0ms 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.263 s - in org.apache.bval.bench.BenchTest

with all changes:
  Raw validation time:        1056ms {12342}
  Raw parsing time:              0ms {0}
  First parse and validate:      0ms {0}
  Multithreaded validation:      0ms 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.126 s - in org.apache.bval.bench.BenchTest
```
they are pretty close, and I think that it's faster with these latest changes because in case we have `NonContainerCascadingMetaData` we don't call `NonContainerCascadingMetaData#addRuntimeLegacyCollectionSupport` anymore, which is still called in the code form master (well there is a similar call but in case of `NonContainerCascadingMetaData` it will just return this).

marko-bekhta:
Also about that BV benchmark.... I think it's possible to convert it to JMH test. We would need to have two modules though. One to generate beans and in the other one we can write a JMH benchmark using those beans.

gsmet:
@marko-bekhta it's slower in your "with all changes" case. Note that the unit here is ms. So the lower the better.

I don't have anything against moving the existing code to a submodule and adding a new module with some JMH testing. It would be nice if we had the same possibilities than the existing benchmark: the ability to test the raw parsing speed and the raw validation speed (having both mixed is of no use IMHO).

marko-bekhta:
Hah, yes ..... It's in ms !!! :smiley: but I did try to build a project with `git` today instead of `mvn` so I'm not surprised that I missed the units of a benchmark...
And I just took the "best" numbers from a couple of runs. 

Here I've run it once more and copied 5 results for each:
```
master:
Raw validation time:         900ms {12342}
Raw validation time:         791ms {12342}
Raw validation time:         825ms {12342}
Raw validation time:         886ms {12342}
Raw validation time:         901ms {12342}

changes:
Raw validation time:         834ms {12342}
Raw validation time:         834ms {12342}
Raw validation time:         866ms {12342}
Raw validation time:        1005ms {12342}
Raw validation time:        1123ms {12342}
```

gsmet:
Mmmmf. I think the only conclusion we can draw is that we really need to move to JMH :).

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaDataBuilder.java
```diff
@@ -317,40 +321,37 @@ public boolean equals(Object obj) {
...
-		}
-		else if ( cascadableClass.isArray() ) {
+		if ( cascadableClass.isArray() ) {
 			// for arrays, we need to add an ArrayElement cascading metadata: it's the only way arrays support cascading at the moment.
 			return addArrayElementCascadingMetaData( cascadableClass, containerElementTypesCascadingMetaData, groupConversions );
 		}
 		else {
-			return containerElementTypesCascadingMetaData;
+			Map<TypeVariable<?>, CascadingMetaDataBuilder> cascadingMetaData = containerElementTypesCascadingMetaData;
+			for ( Entry<TypeVariable<?>, ValueExtractorDescriptor> extractorDescriptorEntry : possibleValueExtractorCandidatesForCascadedValidation.entrySet() ) {
```
gsmet: Does it really work? I'm especially surprised that we might have several VED on different type arguments. I don't see how this could work.

If we have at least 2 VED for different type arguments, we will throw an error at one point, don't we?

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PotentiallyContainerCascadingMetaData.java
```diff
@@ -0,0 +1,129 @@
...
+
+	public static PotentiallyContainerCascadingMetaData of(CascadingMetaDataBuilder cascadingMetaDataBuilder, Set<ValueExtractorDescriptor> potentialValueExtractorDescriptors, Object context) {
+		return new PotentiallyContainerCascadingMetaData( cascadingMetaDataBuilder, potentialValueExtractorDescriptors );
+	}
+
+	private PotentiallyContainerCascadingMetaData(CascadingMetaDataBuilder cascadingMetaDataBuilder, Set<ValueExtractorDescriptor> potentialValueExtractorDescriptors) {
+		this(
+				potentialValueExtractorDescriptors,
+				GroupConversionHelper.of( cascadingMetaDataBuilder.getGroupConversions() )
+				);
```
gsmet: Indentation issue.

marko-bekhta:
Finally had a chance to spent some time on this. I've edited the last commit and force-pushed it. 
I tried to add explanations in the comments with some examples. Hope those would make it easier to read and understand the changes.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaDataBuilder.java
```diff
@@ -211,12 +195,57 @@ public CascadingMetaDataBuilder merge(CascadingMetaDataBuilder otherCascadingTyp
...
-			return ContainerCascadingMetaData.of( valueExtractorManager, this, context );
+
+		// now we are at the case when we might have something like @Valid SomeType something;
+		// First we should check if there are any available VE that can be applied to the current declared type (SomeType)
+		// If such are found we promote the cascading to the type variables that we have VEs for.
+		// In case of the Map call to valueExtractorManager.getPossibleValueExtractorCandidatesForCascadedValidation()
+		// will not return MapKeyExtractor, hence by default it will only return MapValueExtractor which matches BV Spec.
+		//
+		// If we found more than one type variable and an extractor for it we convert the metadata to something that we would
+		// have if the declaration was SomeContainer<@Valid VarWithVE1, @Valid VarWithVE2, VarWithoutVE ..., >
```
gsmet: OK, this is wrong. We need to throw an exception in this case as I mentioned in an earlier comment.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaDataBuilder.java
```diff
@@ -211,12 +195,57 @@ public CascadingMetaDataBuilder merge(CascadingMetaDataBuilder otherCascadingTyp
 	public CascadingMetaData build(ValueExtractorManager valueExtractorManager, Object context) {
 		validateGroupConversions( context );
 
-		if ( containerElementTypesCascadingMetaData.isEmpty() ) {
+		// we are 100% sure that we have a container and its elements are marked for cascading
+		if ( !containerElementTypesCascadingMetaData.isEmpty() && hasContainerElementsMarkedForCascading ) {
+			return ContainerCascadingMetaData.of( valueExtractorManager, this, context );
```
gsmet: Wondering how we would react to:

```java
@Valid
private Map<@Valid KeyBean, ValueBean> map;
```

The spec advises you to not use it but because there is a risk for the values to be validated twice.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaDataBuilder.java
gsmet: I was referring to this comment.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaDataBuilder.java
gsmet: Maybe we should have all this logic like this:

```java
if ( !cascading ) {
    if ( !containerElementTypesCascadingMetaData.isEmpty() && hasContainerElementsMarkedForCascading ) {
         // [...]
    }
    else {
       return NonContainerCascadingMetaData.of( this, context );
    }
}
```

gsmet: And we obviously need a new test case.

marko-bekhta:
It's not related to VE directly, just encountered one more case where annotations are not read correctly, here's an example:
```java
@Test
public void testTypeVariableAnnotationsIssue() throws Exception {
	class Barrr {
	}
	class MapTest {
		private final Map<@Valid List<@NotBlank String>, @NotNull @Valid Barrr> container;

		MapTest(Map<List<String>, Barrr> container) {
			this.container = container;
		}
	}

	AnnotatedType[] annotatedActualTypeArguments = ( (AnnotatedParameterizedType) MapTest.class.getDeclaredFields()[0].getAnnotatedType() ).getAnnotatedActualTypeArguments();
	assertEquals( annotatedActualTypeArguments.length, 2 );
	assertEquals( annotatedActualTypeArguments[0].getAnnotations().length, 1 );
	assertEquals( annotatedActualTypeArguments[1].getAnnotations().length, 2 ); // this part fails
}
```
Last assert fails if the type is a local one (`Barrr`). If it is replaced with for example `Integer` - everything is ok.

gsmet:
@marko-bekhta might be nice to add an issue with a test case to our JDK issues milestone.

gunnarmorling:
So where are we with that one? Should I do a review pass, too?

gsmet:
Here is the situation:
 * I want to make a last pass to change a few method names and get the overall thing easier to understand
 * Marko is working in parallel on the conversion of the BV benchmark to JMH so that we can have stable results and determine the real consequences of this patch

I'll ping you if I think an additional review is required but we should really wait for the final changes and the benchmark results.

marko-bekhta:
+ in addition I wanted to give you some "free-time-from-Marko" during your F2F meeting :smiley: hence not much activity on this front atm.

gsmet:
@marko-bekhta I'm finally done with this PR. We can check that the performances are not too bad, now :).

Will you do it or should I?

marko-bekhta:
@gsmet :tada: I have some time for it. I'll give it a shot and post results here.

marko-bekhta:
```bash
## First run:
INFO: HV000001: Hibernate Validator 6.0.7.Final:

# Run complete. Total time: 00:02:04
Benchmark                                            Mode  Cnt    Score    Error   Units
ParsingBeansSpeedBenchmark.testCascadedValidation   thrpt   20   71.851 ±  6.902  ops/ms
RawValidationSpeedBenchmark.testCascadedValidation  thrpt   30  174.241 ± 11.016   ops/s

# Run complete. Total time: 00:01:49

Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  171.268 ± 17.361  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  443.933 ±  5.771   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  719.041 ±  8.809   ops/s

########################################################################################
INFO: HV000001: Hibernate Validator - 6.0.8-SNAPSHOT

# Run complete. Total time: 00:02:00

Benchmark                                            Mode  Cnt    Score   Error   Units
ParsingBeansSpeedBenchmark.testCascadedValidation   thrpt   20   77.417 ± 6.520  ops/ms
RawValidationSpeedBenchmark.testCascadedValidation  thrpt   30  214.495 ± 5.391   ops/s

# Run complete. Total time: 00:01:50

Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  176.687 ± 14.551  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  479.938 ±  7.135   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  648.564 ± 14.705   ops/s
```
and once more:
```bash
## Second run:
INFO: HV000001: Hibernate Validator 6.0.7.Final:

# Run complete. Total time: 00:02:06

Benchmark                                            Mode  Cnt    Score    Error   Units
ParsingBeansSpeedBenchmark.testCascadedValidation   thrpt   20   72.326 ± 10.849  ops/ms
RawValidationSpeedBenchmark.testCascadedValidation  thrpt   30  183.109 ±  4.638   ops/s

# Run complete. Total time: 00:01:50

Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  163.579 ±  5.531  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  455.314 ±  7.435   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  740.634 ± 17.353   ops/s

########################################################################################
INFO: HV000001: Hibernate Validator - 6.0.8-SNAPSHOT

# Run complete. Total time: 00:01:58

Benchmark                                            Mode  Cnt    Score   Error   Units
ParsingBeansSpeedBenchmark.testCascadedValidation   thrpt   20   70.965 ± 5.177  ops/ms
RawValidationSpeedBenchmark.testCascadedValidation  thrpt   30  210.119 ± 4.036   ops/s

# Run complete. Total time: 00:01:50

Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  168.120 ±  8.574  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  503.153 ±  9.834   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  716.397 ± 13.306   ops/s
```

Later today I'll try to run these on a different machine. I also tried some profiling on `CascadedWithLotsOfItemsAndMoreConstraintsValidation`, nothing unusual there:
```
org.hibernate.validator.internal.engine.path.PathImpl.buildHashCode() | 10.282879 | 81,893 ms (10.3%) | 81,893 ms | 81,893 ms | 81,893 ms
org.hibernate.validator.internal.engine.ValidationContext.markConstraintProcessed() | 7.529805 | 59,967 ms (7.5%) | 59,967 ms | 59,967 ms | 59,967 ms
org.hibernate.validator.internal.engine.ValidationContext.markCurrentBeanAsProcessedForCurrentPath() | 5.771458 | 45,964 ms (5.8%) | 45,964 ms | 59,804 ms | 59,804 ms
org.hibernate.validator.internal.util.CollectionHelper.newHashSet() | 5.6387296 | 44,906 ms (5.6%) | 44,906 ms | 44,906 ms | 44,906 ms
```

gsmet:
Hi @marko-bekhta ,

Thanks for the numbers.

It would probably be nice to try to understand what we do differently in CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems with the new code to make it noticeably slower.

gsmet:
I also think it would be beneficial to run the BV benchmark.

gsmet:
Hi @marko-bekhta,

As the `ReflectionHelper#getValue()` PR ends up being more complicated than expected, I think we should focus on trying to understand why we have a perf regression on `CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems` and try to get this one in (finally!).

marko-bekhta:
Also some fresh numbers. (note that last commit doesn't really make any difference it's just that it can never happen and I just think that it would be easier to read without it). So the benchmarks:

single thread 1000 items:
```bash
# *Hibernate Validator 6.0.7.Final*
Benchmark                                                                                   Mode  Cnt     Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   176.892 ±  3.046  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20   877.112 ± 13.760   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1198.482 ± 17.046   ops/s

# *Hibernate Validator 6.0.8-SNAPSHOT*
Benchmark                                                                                   Mode  Cnt     Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   194.530 ±  2.667  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20   857.824 ± 11.153   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1177.166 ± 16.300   ops/s
```
single thread 2000 items (regular amount):
```bash
# *Hibernate Validator 6.0.7.Final*
Benchmark                                                                                   Mode  Cnt    Score   Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  180.171 ± 2.331  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  431.718 ± 4.602   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  593.008 ± 6.853   ops/s

# *Hibernate Validator 6.0.8-SNAPSHOT*
Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  190.509 ±  2.453  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  422.300 ±  4.079   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  581.905 ± 10.974   ops/s

```
10 threads 2000 items (regular amount):
```bash
# *Hibernate Validator 6.0.7.Final*
Benchmark                                                                                   Mode  Cnt     Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   179.879 ±  2.063  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1369.362 ± 31.438   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1845.036 ± 25.291   ops/s

# *Hibernate Validator 6.0.8-SNAPSHOT*
Benchmark                                                                                   Mode  Cnt     Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   194.068 ±  2.552  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1641.228 ± 22.088   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1951.470 ± 24.349   ops/s

```
20 threads 2000 items (regular amount):
```bash
# *Hibernate Validator 6.0.7.Final*
Benchmark                                                                                   Mode  Cnt     Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   179.347 ±  2.173  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1344.456 ± 26.267   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1721.242 ± 30.364   ops/s

# *Hibernate Validator 6.0.8-SNAPSHOT*
Benchmark                                                                                   Mode  Cnt     Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   192.206 ±  2.391  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1468.377 ± 43.692   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1753.757 ± 26.439   ops/s

```
40 threads 2000 items (regular amount):
```bash
# *Hibernate Validator 6.0.7.Final*
Benchmark                                                                                   Mode  Cnt     Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   179.848 ±  1.553  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1368.312 ± 26.805   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1638.132 ± 29.546   ops/s

# *Hibernate Validator 6.0.8-SNAPSHOT*
Benchmark                                                                                   Mode  Cnt     Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   192.664 ±  1.830  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1255.655 ± 26.588   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1723.268 ± 34.982   ops/s

```
100 threads 2000 items (regular amount):
```bash
# *Hibernate Validator 6.0.7.Final*
Benchmark                                                                                   Mode  Cnt     Score     Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   176.352 ±   1.623  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20   957.121 ±  98.533   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1551.714 ± 107.216   ops/s

# *Hibernate Validator 6.0.8-SNAPSHOT*
Benchmark                                                                                   Mode  Cnt     Score     Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20   191.282 ±   2.465  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20   972.603 ±  74.376   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1520.292 ± 107.742   ops/s
```

for all results above  I've kept the `CascadedValidation.testCascadedValidation` with a single thread.

marko-bekhta:
Looking at these latest benchmarks `CascadedWithLotsOfItems*` are both just a bit slower with these VE changes up to some number of threads. I've run each of the benchmarks a couple times and the results were consistent all the time.
I did some more debugging, staring at the code changes :) and comparing method call logs yesterday and there's nothing that showed up interesting. That's why I've started to think that maybe it's something to do with multiple threads ? Hence those benchmarks for different number of threads. Based on this could it be that this "strange" behavior of the benchmarks, that ought to behave the same, has something to do with the CHM access?  After the first run when the metadata is available we would only access the CHM just once when getting a runtime VE from the set of candidates.

engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorResolver.java
```diff
@@ -0,0 +1,431 @@
...
+	private static final Object NON_CONTAINER_VALUE = new Object();
+
+	@Immutable
+	private final Set<ValueExtractorDescriptor> registeredValueExtractors;
+
+	private final ConcurrentHashMap<ValueExtractorCacheKey, Set<ValueExtractorDescriptor>> possibleValueExtractorsByRuntimeTypeAndTypeParameter = new ConcurrentHashMap<>();
+
+	private final ConcurrentHashMap<Class<?>, Set<ValueExtractorDescriptor>> possibleValueExtractorsByRuntimeType = new ConcurrentHashMap<>();
+
+	private final ConcurrentHashMap<Class<?>, Object> nonContainerTypes = new ConcurrentHashMap<>();
```
gunnarmorling: Should this be using `ConcurrentReferenceHashMap` instead? The key of `Class` makes me a bit nervous in terms of deployment classes leaking.

Sanne: Since it's not static, what are you concerned about exactly?

But also: since the key is a Class, there's a hyper officient specialized map implementation for such cases in the JDK. I forgot the name :( but it was originally introduced to optimize "invoke dynamic".

gunnarmorling: My concern is about a `ValidatorFactory` managed by the container outliving the deployment. Probably you're right, though, and this should not happen in practice. FWIW, we use CRHM in other places where we keep metadata by class.

On that hyper-efficient impl, above we're instantiating `CHM` directly, so we wouldn't benefit from that? Or are you saying that CHM is internally using an optimized path if keys are class objects?

Sanne: It's a different collection container, not sure it implements the Map or the ConcurrentMap API but it's essentially an equivalent of Map<Class,X>. I wish I could remember its name, as I'm needing to try it out in another context too :-/

gsmet:
@marko-bekhta I added the commit b73fe21 with a couple of micro optimizations and here are my results:

changes with the micro optimizations
```
Result "org.hibernate.validator.performance.cascaded.CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems":
  1373.747 ±(99.9%) 8.356 ops/s [Average]
  (min, avg, max) = (1356.278, 1373.747, 1389.036), stdev = 9.623
  CI (99.9%): [1365.391, 1382.104] (assumes normal distribution)
```

changes without
```
Result "org.hibernate.validator.performance.cascaded.CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems":
  1281.256 ±(99.9%) 9.410 ops/s [Average]
  (min, avg, max) = (1265.115, 1281.256, 1301.452), stdev = 10.837
  CI (99.9%): [1271.846, 1290.666] (assumes normal distribution)
```

So I think we are probably back on track compared to the previous behavior.

Could you take a look at the last 3 commits? I don't expect any issue there but let's be on the safe side. Once I have your go, I'll merge it.

gsmet:
@marko-bekhta yeah, it's not my idea. @Sanne suggested it before in a similar case, that's why I did it.

gsmet:
And finally merged!

@marko-bekhta thanks for your hard work on this one, glad to see it finally in.

gunnarmorling:
Thanks, Marko and Guillaume!

PR #890, opened 2017-11-24
HV-1529 Add dynamic payload to HibernateConstraintValidator

mkurz:
@marko-bekhta and @gsmet probably are the persons who should take a look at this :wink: 
All the information can be found here: https://hibernate.atlassian.net/browse/HV-1529

Basically we can now pass a **_dynamic_** payload to a `(Hibernate)ConstraintValidator` via:

```java
someValidator
    .unwrap(ValidatorImpl.class)
    .withDynamicConstraintValidatorInitializationPayload(payloadObject)
    .validate(objectToValidate)
```

And retrieve it via `getDynamicPayload`:

```java
public class MyFutureValidator implements HibernateConstraintValidator<MyFuture, Instant> {

	private SomeType dynamicPayload;

	@Override
	public void initialize(ConstraintDescriptor<MyFuture> constraintDescriptor,
			HibernateConstraintValidatorInitializationContext initializationContext) {
		this.dynamicPayload = initializationContext.getDynamicPayload(SomeType.class);
	}

	@Override
	public boolean isValid(Instant instant, ConstraintValidatorContext constraintContext) {
		// Here I can use the dynamic payload now
		dynamicPayload.getSomeData();
		//...
	}
}
```


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -156,7 +158,17 @@ public ValidatorImpl(ConstraintValidatorFactory constraintValidatorFactory,
...
 		this.traversableResolver = traversableResolver;
 		this.validatorScopedContext = new ValidatorScopedContext( messageInterpolator, parameterNameProvider, clockProvider, temporalValidationTolerance, scriptEvaluatorFactory, failFast, traversableResolverResultCacheEnabled );
-		this.constraintValidatorInitializationContext = new HibernateConstraintValidatorInitializationContextImpl( validatorScopedContext.getScriptEvaluatorFactory(), validatorScopedContext.getClockProvider(), validatorScopedContext.getTemporalValidationTolerance() );
+		this.scriptEvaluatorFactory = validatorScopedContext.getScriptEvaluatorFactory();
+		this.constraintValidatorInitializationContext = new HibernateConstraintValidatorInitializationContextImpl( validatorScopedContext.getScriptEvaluatorFactory(), validatorScopedContext.getClockProvider(), validatorScopedContext.getTemporalValidationTolerance(), null );
+	}
+
+	public ValidatorImpl withDynamicConstraintValidatorInitializationPayload(Object dynamicPayload) {
+		this.constraintValidatorInitializationContext = new HibernateConstraintValidatorInitializationContextImpl(
+			this.scriptEvaluatorFactory,
```
mkurz: Instead of saving a reference to the `ScriptEvaluatorFactory` here inside the `ValidatorImpl` class we could also add a `getScriptEvaluatorFactory()` to the `(Hibernate)ConstraintValidatorInitializationContext(Impl)` - if it would be ok to expose that `ScriptEvaluatorFactory` from that context.
We could then replace this line here with `this.constraintValidatorInitializationContext.getScriptEvaluatorFactory()` instead. I leave it up to you which approach to use.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -55,4 +56,12 @@
 	 * @return the tolerance
 	 */
 	Duration getTemporalValidationTolerance();
+
+	/**
+	 * @param type The type of payload to retrieve
+	 * @return an instance of the specified type set by the user via
```
gsmet: Please make the long version the description of the method and keep a short one for the `@return`.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -55,4 +56,12 @@
...
+	 * {@link ValidatorImpl#withDynamicConstraintValidatorInitializationPayload(Object)} or {@code null} if no constraint payload
+	 * if the given type has been set.
```
gsmet: s/if the given type/of the given type/

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -55,4 +56,12 @@
 	 * @return the tolerance
 	 */
 	Duration getTemporalValidationTolerance();
+
+	/**
+	 * @param type The type of payload to retrieve
+	 * @return an instance of the specified type set by the user via
+	 * {@link ValidatorImpl#withDynamicConstraintValidatorInitializationPayload(Object)} or {@code null} if no constraint payload
```
gsmet: s/constraint payload/constraint validator payload/

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gsmet: We sure don't want to expose it (see the `clear()` method for instance).

What you could do instead is creating another class that wraps the original `HibernateConstraintValidatorInitializationContextImpl` and adds the payload. The original `HibernateConstraintValidatorInitializationContextImpl` would simply return null for the payload.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -156,7 +158,17 @@ public ValidatorImpl(ConstraintValidatorFactory constraintValidatorFactory,
 		this.validationOrderGenerator = validationOrderGenerator;
 		this.traversableResolver = traversableResolver;
 		this.validatorScopedContext = new ValidatorScopedContext( messageInterpolator, parameterNameProvider, clockProvider, temporalValidationTolerance, scriptEvaluatorFactory, failFast, traversableResolverResultCacheEnabled );
-		this.constraintValidatorInitializationContext = new HibernateConstraintValidatorInitializationContextImpl( validatorScopedContext.getScriptEvaluatorFactory(), validatorScopedContext.getClockProvider(), validatorScopedContext.getTemporalValidationTolerance() );
+		this.scriptEvaluatorFactory = validatorScopedContext.getScriptEvaluatorFactory();
+		this.constraintValidatorInitializationContext = new HibernateConstraintValidatorInitializationContextImpl( validatorScopedContext.getScriptEvaluatorFactory(), validatorScopedContext.getClockProvider(), validatorScopedContext.getTemporalValidationTolerance(), null );
+	}
+
+	public ValidatorImpl withDynamicConstraintValidatorInitializationPayload(Object dynamicPayload) {
```
gsmet: I think I would stick to as simple name. Something like `withConstraintValidatorDynamicPayload()`. The fact that it is related to the initialization can be removed I think. We don't really care (mostly because it can still be used at runtime by adding an attribute to the constraint validator).

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -130,11 +130,13 @@
...
 	private final ValidatorScopedContext validatorScopedContext;
 
+	private final ScriptEvaluatorFactory scriptEvaluatorFactory;
+
 	/**
 	 * The constraint initialization context is stored at this level to prevent creating a new instance each time we
 	 * initialize a new constraint validator as, for now, it only contains Validator scoped objects.
 	 */
-	private final HibernateConstraintValidatorInitializationContext constraintValidatorInitializationContext;
+	private HibernateConstraintValidatorInitializationContext constraintValidatorInitializationContext;
```
gsmet: A wise man said to me one day: "Every time you remove a `final`, God kills a kitten." (yes, he's so wise he uses backticks when he speaks).

You're paving the way to concurrency issues. Keep in mind that the `ValidatorImpl` is shared across thread.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -156,7 +158,17 @@ public ValidatorImpl(ConstraintValidatorFactory constraintValidatorFactory,
 		this.validationOrderGenerator = validationOrderGenerator;
 		this.traversableResolver = traversableResolver;
 		this.validatorScopedContext = new ValidatorScopedContext( messageInterpolator, parameterNameProvider, clockProvider, temporalValidationTolerance, scriptEvaluatorFactory, failFast, traversableResolverResultCacheEnabled );
-		this.constraintValidatorInitializationContext = new HibernateConstraintValidatorInitializationContextImpl( validatorScopedContext.getScriptEvaluatorFactory(), validatorScopedContext.getClockProvider(), validatorScopedContext.getTemporalValidationTolerance() );
+		this.scriptEvaluatorFactory = validatorScopedContext.getScriptEvaluatorFactory();
+		this.constraintValidatorInitializationContext = new HibernateConstraintValidatorInitializationContextImpl( validatorScopedContext.getScriptEvaluatorFactory(), validatorScopedContext.getClockProvider(), validatorScopedContext.getTemporalValidationTolerance(), null );
+	}
+
+	public ValidatorImpl withDynamicConstraintValidatorInitializationPayload(Object dynamicPayload) {
+		this.constraintValidatorInitializationContext = new HibernateConstraintValidatorInitializationContextImpl(
```
gsmet: Concurrency issue. You're not overriding the context for your `validate()` call here but also for all the other random threads running. And the other threads will override it too.

I think we won't be able to make your approach work.

My opiniong is that you'd be better served by having the payload at the `ValidatorFactory` level AND the `HibernateValidatorContext` level.

The drawback is that doing you calls to `validate()` you would have to instantiate the `Validator` first by unwrapping the context and setting the payload. The `Validator` creation is quite lightweight so I don't think it will be an issue in practice.

Could you try this approach and see how it goes with your use case?

gunnarmorling:
Hi @mkurz, a couple of comments/questions:

* What's the use case this is trying to address? In the issue you mention a country/language, but that's where the message interpolator and its locale handling comes in. The code example shows a temporal constraint, but no actual usage of the proposed feature.
* `j.v.Validator` is not meant to be recreated for every call to `validate()`, which I think your proposal implies. Instead it seems a context parameter would have to be added to the `validate()` method.
* `ValidatorImpl` is an internal class, i.e. it shouldn't be exposed to clients. So if we wanted to provide some additional functionality, we'd have to do so via a new interface which one then can unwrap to.

Note there are two existing ways for passing in contextual information:

* Use a class-level constraint, allowing you to obtain e.g. a language from the bean hosting the constrained property; admittedly that may be complex, but then again it'd help to better understand the usecase this is about.
* Use dependency injection into constraint validators via a custom `ConstraintValidatorFactory`, which e.g. returns CDI or Spring enabled validator instances. They could have request-scoped contextual information injected, e.g. the locale of the current request's users.

gsmet:
@gunnarmorling 

> What's the use case this is trying to address?

Matthias gave an example in the JIRA issue: "think of zip codes which differ for each country".

> j.v.Validator is not meant to be recreated for every call to validate(), which I think your proposal implies. Instead it seems a context parameter would have to be added to the validate() method.

That's not what the PR does (and it indeed introduces concurrency issues I mentioned in my review). Frankly, taking a look at the `ValidatorImpl` constructor, I don't think it would be that bad to instantiate a `Validator` for each call. And we could imagine a CHM of ``Validator`` per country code if required.

Do you think it would be better to add a parameter to the various `validate...()` calls? I don't like that very much.

> ValidatorImpl is an internal class, i.e. it shouldn't be exposed to clients. So if we wanted to provide some additional functionality, we'd have to do so via a new interface which one then can unwrap to.

Yup, I didn't mention it because we can't introduce the feature here anyway (concurrency issues).

gsmet:
Btw, I think it makes sense to consider it `Validator` scoped.

If the same user makes several `validate()` calls at once, I would expect them to be run with the same payload.

gunnarmorling:
> "think of zip codes which differ for each country".

Ok, so the requirement is then:

```java
public class Address {
    @ValidZipCode // acts differently based on the country
    String zipCode;

    Country country;
```

?

As said, I think the easiest would be to use a class-level constraint or context injection into the validator then.

> Do you think it would be better to add a parameter to the various validate...() calls? I don't like that very much.

Actually, the problem with either approach is that the user typically isn't in the business of obtaining a validator (in the context of one request/transaction etc.) nor of calling `validate()`.

Instead this will be done by some framework for them. I.e. usually passing in such context will not be available to them, whereas the other two approaches allow to get hold of contextual information also in such managed environment.

In a way it's related to [BVAL-237](https://hibernate.atlassian.net/browse/BVAL-237), too, i.e. the idea of making the root or leaf bean (or at least their properties in a detyped manner) available to constraint validators.

gunnarmorling:
Btw. there already is a way for passing `Validator`-scoped metadata: `ValidatorFactory#usingContext()`. `HibernateValidatorContext` could be expanded to allow for the propagation of arbitrary data, but as said I'm not convinced that this helps the discussed use case much.

gsmet:
> Btw. there already is a way for passing Validator-scoped metadata: ValidatorFactory#usingContext(). HibernateValidatorContext could be expanded to allow for the propagation of arbitrary data, but as said I'm not convinced that this helps the discussed use case much.

Yeah, that's exactly what I proposed in my review :).

gsmet:
@gunnarmorling I thought about it a bit more and I don't think having the ability to pass some context to constraint validators is useless.

It might be handy to manage differences between 2 environments or 2 different apps using the same jars.

Having them at VF and context level would be enough.

@mkurz could you give us more details about your use case? As you might have noticed, @gunnarmorling is a bit skeptic about the usefulness of such addition so having more details on why it's not practical for you to use class level constraints and some data in the model would help.

Thanks!

mkurz:
@gsmet and @gunnarmorling thank you both very much for your comprehensive replies. I really appreciate that.

**So what is my use case?**
In my free time I work as an open source contributor for the [Play Framework](https://www.playframework.com). Right now Play stores various information about a request in a [`play.mvc.Http.Context`](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play/src/main/java/play/mvc/Http.java#L56) object. That would be e.g. [the locale](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play/src/main/java/play/mvc/Http.java#L83) of the current request (or the one the user has set explicit via a cookie) and various other things, e.g. there is an [`args` map](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play/src/main/java/play/mvc/Http.java#L346-L349) in which a framework user can put various other stuff inside that context.
Right now that http-context is stored in [a `ThreadLocal`](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play/src/main/java/play/mvc/Http.java#L58) and therefore we can access the [current http-context](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play/src/main/java/play/mvc/Http.java#L65) from basically everywhere, including the `isValid` method of a `ConstraintValidator`.

However in near future that `ThreadLocal` will be removed an that http-context has to be passed to every method were it is needed. That means we somehow have to pass that http-context (or just parts of it, e.g. the locale) inside the `isValid` method because right now we use e.g. the locale of the current request for validation (like the zip code example @gunnarmorling gave). And therefore that has to be done **for each request** - therefore it has to be "dynamic".

**That is how hibernate-validator is set up in Play righ now**
We [bind](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-java-forms/src/main/java/play/data/validation/ValidatorsModule.java#L18-L19) the components we need via dependency injection. We [set up a Validator Factory](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-java-forms/src/main/java/play/data/validation/ValidatorProvider.java#L27-L30) that [provides us a new validator](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-java-forms/src/main/java/play/data/validation/ValidatorProvider.java#L39) each time we need one. (I know now these validators could be cached, but that is not the point here. I will probably add caching soon. Because I didn't know caching them is allowed is why I stumpled into that concurrency issues in my pull request....) The `ConstraintValidatorFactory` simply [returns us a new instance](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-java-forms/src/main/java/play/data/validation/DefaultConstraintValidatorFactory.java#L26) of the desired constraint.
So now a `Validator` gets injected into the `FormFactory` [here](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-java-forms/src/main/java/play/data/FormFactory.java#L24-L25) and eventually we call the `validate(...)` method that `javax.validation.Validator` (we don't unwrap it right now) provides [here](https://github.com/playframework/playframework/blob/2.6.7/framework/src/play-java-forms/src/main/java/play/data/Form.java#L408-L410).

Actually at first I was thinking the same like what @gunnarmorling proposed - to add a parameter to all the various (unwrapped) `validate...()` calls. However I also don't like that approach, that is why I came up with the `with...Dynamic...Payload()` method.

**About the suggestions @gunnarmorling mentioned**
- *Using a class-level constraint*
I am aware of that feature, however I don't think this is what we are looking for here. The only way I can think of to get that working is that a user would have to pollute his/her form classes with "special" properties into which the framework sets e.g. the locale. (A form class in Play is simply a POJO that represents a HTML form on the server side and that uses bean validation/hibernate constraints annotations for validation). However that would mean the framework would have to provide the user some kind of contract of how he/she has to set up that "special" properties (which values will later set by the framework) e.g. the user woud have to sub-class his/her forms of a specific super-class that the framework provides and where the locale can be set on... I really don't like that. For users, nothing should change and everything should just work out of the box as simple as possible (without a framework user having taking care of this)
So I think with the design Play is following this approach isn't really working. 

- *Request-scoped DI*
Play wont work with request-scoped dependencies (right now). See https://github.com/playframework/playframework/pull/6798 - it was closed this year and it looks like that wont happen. So that is also not an option for us.

**Status quo of this pull request**
(First of all I had to rebase against the latest `master` because of the changes introduces for [HV-1531](https://hibernate.atlassian.net/browse/HV-1531)).
I updated the pull request as suggested so the dynamic payload is now at `ValidatorFactory` and `HibernateValidatorContext` level. As far as I can tell in Play we now would have to inject the `ValidatorFactory` instead of a `Validator` itself into the `FormFactory` and then just before calling `validate(...)` unwrap the context and set the http-context (or parts of it): `validatorFactory.usingContext().dynamicPayload(ctx.getLocale()).getValidator()` like [shown in the docs](https://docs.jboss.org/hibernate/validator/6.0/reference/en-US/html_single/#section-configuring-validator).
I think that should work for us.

However I just set up [another branch](https://github.com/mkurz/hibernate-validator/tree/withConstraintValidatorDynamicPayload) which - like my original idea - adds a `withConstraintValidatorDynamicPayload` (on top of that pull request here) which now returns a **new** validator instance - which is basically a copy of the origin one - but with the given payload set. For that I just added a `private ValidatorImpl()` constructor and a copy constructor for `ValidatorScopedContext`. I think that way concurrency isn't a problem anymore.
This way setting a dynamic payload is **really** dynamic. You can just "copy" your validators and set the payload you need. That would also work when caching the validators. I think this is what we are really looking for here for the Play Framework.

What do you think about all of this? Am I missing something?
I really would like to find a solution for this and get it into one of the next HV releases. Frankly I think that stuff makes sense, not just for Play.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -183,7 +183,8 @@ public ValidatorFactoryImpl(ConfigurationState configurationState) {
 				getTemporalValidationTolerance( configurationState, properties ),
 				getScriptEvaluatorFactory( configurationState, properties, externalClassLoader ),
 				getFailFast( hibernateSpecificConfig, properties ),
-				getTraversableResolverResultCacheEnabled( hibernateSpecificConfig, properties )
+				getTraversableResolverResultCacheEnabled( hibernateSpecificConfig, properties ),
+				null
```
mkurz: I am not sure about just passing `null` here. Suggestions?

gsmet:
> This way setting a dynamic payload is really dynamic. You can just "copy" your validators and set the payload you need. That would also work when caching the validators. I think this is what we are really looking for here for the Play Framework.

I don't understand what is the value of that. I think it's better to pass the `ValidatorFactory` and create new validators from it using a context. Which benefits do you see in your approach?

You could either have a `Validator` cache per locale, or at least try to reuse the same `Validator` for the whole request.

mkurz:
> I don't understand what is the value of that. I think it's better to pass the ValidatorFactory and create new validators from it using a context. Which benefits do you see in your approach?

In case we don't pass a locale as dynamic payload but e.g. the whole play request context which could contain different data on each request we can't cache that validator like we could when just passing the locale as payload. So when no caching is possible, "copying" a validator and passing the payload is cheaper than using `ValidatorFactory.getValidator()` each time? Only of course if we could cache a "base" `Validator` which would be copied each time. What do you think? Or am I wrong?

gsmet:
@mkurz just to let you know that I'm thinking about this. Will get back to you shortly.

mkurz:
@gsmet thanks for your feedback! I will update the pull request asap - I am very busy these days.

gsmet:
@mkurz ah ah, funny thing: i had on my TODO list for today to check if you were still interested in pursuing this. Let me know if you need any help or if you have questions!

mkurz:
@gsmet I finally found time to update this pull request. Please check the commits I added (the last one probably is the important one).
And about commit f0418e0 - is that what you meant by "*...I think we can reduce the overhead of creating a brand new `Validator` a bit by making the `ValidationOrderGenerator` global to the `ValidatorFactory`...*"?

Also CI does not pass. Do you know what's going on? Can you help out and checkout my branch and see if you get it working? Thanks!

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -512,6 +512,17 @@ private Duration getTemporalValidationTolerance(ConfigurationState configuration
...
 
+	private Object getConstraintValidatorPayload(ConfigurationState configurationState, Map<String, String> properties) {
+		if ( configurationState instanceof ConfigurationImpl ) {
+			ConfigurationImpl hibernateSpecificConfig = (ConfigurationImpl) configurationState;
+			if ( hibernateSpecificConfig.getConstraintValidatorPayload() != null ) {
+				LOG.logConstraintValidatorPayload( hibernateSpecificConfig.getConstraintValidatorPayload() );
+				return hibernateSpecificConfig.getConstraintValidatorPayload();
+			}
+		}
+		return null;
```
mkurz: @gsmet question: Here I am simply returning `null` and not trying to get a result out of the `properties` (like in `getTemporalValidationTolerance(...)` above. Is that OK/enough here? As I understand looking into the `properties` is the way to handle the xml config, however does it make sense to define a payload in the xml at all? If so it would probably just be of type string (even tough the payload could be "object") Do you know what I mean? (Its late... :wink:)

marko-bekhta:
@mkurz Hi, about that build failure I think it's because of this error in javadoc:
```
[ERROR] /var/lib/jenkins/workspace/hibernate-validator-PR/distribution/../engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java:305: error: @param name not found
```
there's a mismatch between the param name in javadoc `dynamicPayload` and the name of the actual param `constraintValidatorPayload`

mkurz:
@marko-bekhta Thanks! I fixed the problem (squashed the fix into the last commit were it belongs)

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
gsmet: Yeah, no need to implement it in the properties. It does not make much sense.

I simplified the parameters of the method as we don't need the `properties` at all.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -511,6 +512,17 @@ private Duration getTemporalValidationTolerance(ConfigurationState configuration
 		return Duration.ZERO;
 	}
 
+	private Object getConstraintValidatorPayload(ConfigurationState configurationState, Map<String, String> properties) {
+		if ( configurationState instanceof ConfigurationImpl ) {
+			ConfigurationImpl hibernateSpecificConfig = (ConfigurationImpl) configurationState;
+			if ( hibernateSpecificConfig.getConstraintValidatorPayload() != null ) {
+				LOG.logConstraintValidatorPayload( hibernateSpecificConfig.getConstraintValidatorPayload() );
```
gsmet: I'm reducing the log level of this one to `DEBUG` as it might be a bit annoying to have the payload object logged (typically in your use case, if the context object has a valid `toString()` that is a bit long, it's going to be a pain). As we don't have the properties in this case, the value being used is pretty clear: it's the one set.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -804,6 +825,16 @@ public Builder setTraversableResolverResultCacheEnabled(boolean traversableResol
 				return this;
 			}
 
+			public Builder setDynamicPayload(Object dynamicPayload) {
+				if ( scriptEvaluatorFactory == null ) {
```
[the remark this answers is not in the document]
mkurz: Haha, yeah copy/paste ain't so good :wink:

mkurz:
Thanks for cleaning up my code! Great work!

> There is one BIG thing missing: we need tests. I let you add them?

I can have a look, but I don't know when I find time. Maybe this weekend? But I can't promise.

gsmet:
> I can have a look, but I don't know when I find time. Maybe this weekend? But I can't promise.

Well, you're the one who wants this in ;).

I think it's the only piece missing. It should be added in the `org.hibernate.validator.test.internal.constraintvalidators` package (at the root). You can take a look at `TemporalValidationToleranceTest` to see how you can initialize the validators in the test (yours will be a lot simpler, it's just a matter of checking that the payload is passed to the constraint validator when we use the different ways to pass a payload).

I think the easiest way to do this is to:
 * create a specific constraint that takes a class as a parameter (this way you will be able to inject different payloads and check that the payload has the right class - might be useful if you want to test that the one of the VC overrides the one of the VF)
 * declare a custom validator testing the payload in src/test/resources/META-INF/services/javax.validation.ConstraintValidator
(see the classes in the same package as ConstraintDefinitionContributorTest)

I don't think it will take a lot of time. Feel free to ask for help if you have any issue, I'll help!

gsmet:
Found a better way to declare your validator, just use this when initializing your Validator (it will avoid declaring another global validator):
```java
HibernateValidatorConfiguration configuration = Validation
				.byProvider( HibernateValidator.class )
				.configure();

ConstraintMapping constraintMapping = configuration.createConstraintMapping();
constraintMapping
		.constraintDefinition( YourConstraint.class )
		.validatedBy( YourConstraintValidatorTestingThePayload.class );

configuration.addMapping( constraintMapping );
```

mkurz:
@gsmet Please have a look at my last commit and let me know what you think :wink:

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -298,4 +298,16 @@
...
+	 * Allows to set a payload which will be passed to the constraint validators.
+	 *
+	 * @param constraintValidatorPayload the payload passed to constraint validators
+	 *
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 6.0.8
+	 */
+	@Incubating
+	HibernateValidatorConfiguration constraintValidatorPayload(Object constraintValidatorPayload);
```
gunnarmorling: How comes this is exposed at the `Configuration` level? IIRC the originally discussed use case, this feature was more indented for a specific context (e.g. the current user's locale) which typically will differ between multiple requests. While I probably can get behind supporting this on a per-validator basis, I don't think exposing this here is very useful. A VF should be constructed once and kept around, you definitely don't want to create a new one for each request.

mkurz:
I had to rebase, build passes now.

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
mkurz: With my PR you don't have to create a new VF for each request, you can just use `vf.unwrap( HibernateValidatorFactory.class ).usingContext().constraintValidatorPayload(...)`. Exposing the payload at `Configuration` level was never my original intention, however during the work on this pull request @gsmet thought we should do that as well, probably to define a "default" payload.

gsmet: This was added at my specific request. I don't think it's that absurd to imagine someone needing to pass information at the VF level to the validators (might be info about the environment for instance, staging/production, with validation rules differing depending on the environment).

And it was also because I wanted the symmetry with the other features.

I think it's better with it than without.

mkurz: I have to agree with @gsmet - it's better with it than without.

engine/src/test/java/org/hibernate/validator/test/constraintvalidator/ConstraintValidatorPayloadTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+	@Test
+	public void testHibernateValidatorContextPayload() {
+		HibernateValidatorConfiguration configuration = getConfiguration();
+
+		final Validator validator = configuration.buildValidatorFactory().unwrap( HibernateValidatorFactory.class )
+				.usingContext().constraintValidatorPayload( Integer.class )
+				.getValidator();
+
+		assertNoViolations(	validator.validate( new PayloadDummyEntity( Integer.class ) ) );
```
gsmet: Minor formatting issue here.

engine/src/test/java/org/hibernate/validator/test/constraintvalidator/ConstraintValidatorPayloadTest.java
```diff
@@ -0,0 +1,126 @@
...
+ * @author Matthias Kurz
+ */
+@TestForIssue(jiraKey = "HV-1529")
+public class ConstraintValidatorPayloadTest {
+
+	@Test
+	public void testHibernateValidatorContextPayload() {
+		HibernateValidatorConfiguration configuration = getConfiguration();
+
+		final Validator validator = configuration.buildValidatorFactory().unwrap( HibernateValidatorFactory.class )
```
gsmet: We don't need it to be final, let's remove the noise.

engine/src/test/java/org/hibernate/validator/test/constraintvalidator/ConstraintValidatorPayloadTest.java
mkurz: Ah, wondering this didn't trigger a checkstyle error...

engine/src/test/java/org/hibernate/validator/test/constraintvalidator/ConstraintValidatorPayloadTest.java
```diff
@@ -0,0 +1,126 @@
...
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorInitializationContext;
+import org.hibernate.validator.testutil.TestForIssue;
+import org.hibernate.validator.testutils.ValidatorUtil;
+import org.testng.annotations.Test;
+
+/**
+ * @author Matthias Kurz
+ */
+@TestForIssue(jiraKey = "HV-1529")
+public class ConstraintValidatorPayloadTest {
```
gsmet: Generally, it looks good. The only issue is that we don't check at all that the validator is validating anything: I mean if the constraint was not exercised at all, you would have the exact same result.

It would be nice if you could add failure cases too so that we are sure the validator is working.

Testing with another type in the entity and checking that the violation returned is a `PayloadConstraint` constraint would be nice. You  can use this:
```
import static org.hibernate.validator.testutil.ConstraintViolationAssert.assertThat;
import static org.hibernate.validator.testutil.ConstraintViolationAssert.violationOf;

[...]

assertThat( violations ).containsOnlyViolations(
		violationOf( PayloadConstraint.class )
);
```

engine/src/test/java/org/hibernate/validator/test/constraintvalidator/ConstraintValidatorPayloadTest.java
gsmet: Yeah, it's unfortunate but IIRC Checkstyle is only picky about having at least one whitespace after the parenthesis, it doesn't trigger an error if there are more. Probably because you could have had a new line then several tabs if you wanted to indent it differently and it would still respect the rule.

mkurz:
@gsmet Updated! Removed `final` and the tab. Please check the `testPayloadViolation` test.
BTW: Seems link travis doesn't start...

mkurz:
Alright, it's green now....

gsmet:
@mkurz thanks, I'll take it from here (I would like to add something to the documentation about this feature but I still need to figure out where).

I'll try to get this done this week so that we can merge it before the end of the week.

mkurz:
@gsmet Great! Thank you very much for your help :+1: Looking forward to the next release :stuck_out_tongue:

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
gunnarmorling: My concern is that this this may be people lead to do the wrong thing and provide volatile contextual information to the VF instead of getting a (rather cheap) validator instance with such information.

Really that's the purpose of `VF#usingContext()`, I don't think adding another means of passing in "context" is a very good idea. About that environment use case, you'd just pass that to the `Validator` and keep that around. I don't see what's gained by exposing this at the VF level?

gsmet: Why are you supposing that the information is volatile? My environment example is not about volatile information.

I would say it makes as much sense as any other configuration knobs we have at the VF level: we want it to be the default value for all Validator creations.

`usingContext()` is exclusively used to override VF configuration for a given Validator creation. That's how it's documented and coded.

What you're asking is special casing this one, not the other way around.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -55,4 +55,14 @@
...
+
+	/**
+	 * Returns an instance of the specified type or {@code null} if no constraint validator payload of the given type
+	 * has been set.
+	 *
+	 * @param type the type of payload to retrieve
+	 * @return an instance of the specified type or null if no payload of this type has been set
+	 * @since 6.0.8
+	 */
+	<C> C getPayload(Class<C> type);
```
gunnarmorling: I was about to say that it should be named `getConstraintValidatorPayload()` for the sake of consistency with the other new API methods. But actually I'm wondering whether this all should be named "payload" at all. That term already is used in the Bean Validation API (`payload()` on constraints, and also `HibernateConstraintViolation#getDynamicPayload()`.

These existing payloads are exposed on the resulting constraint violation. So my question is should we a) use another name here (if it's about some contextual information meant for the constraint validator information) or b) expose that new payload information on resulting constraint violations as we do for the other payloads (which are scoped to a single constraint and its violations, though).

I feel finding another name might be the best thing to avoid confusion with the existing payload concepts (of course a given validator implementation can still decide to forward this new contextual information as a dynamic constraint payload in emitted violations).

Any thoughts?

engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
mkurz: Hmmm... specific to my use case maybe we could run different instances of an application - each instance for a different country. Each country should use it's own locale - no matter what the user locale in the request is. E.g. app instances that server `.de` domains would configure the payload to be german, `.com` instances would be configured to be english, etc.
Just an idea of how that **could** be used (even if that may be not a real good example, just came into my mind right now)

gsmet: @gunnarmorling @mkurz Yes, that's basically what I had in mind.

Or a different behavior in staging for instance because staging doesn't have access to some component required for validation (external service only available for production for instance). I know you shouldn't do that but I also know from experience it happens.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
mkurz: /cc @gsmet

gsmet: @gunnarmorling well I asked on HipChat and you told me to name it whatever I feel it to be the most appropriate so that's what I did :). Considering we are in the validator context, I think the context is clear enough, that's why I went for the short form. I couldn't think of another payload we would want it here. I don't have a strong opinion though as I hesitated between the consistency and the impression of redundancy. So if you prefer the full name, let's name it that way, I'm +0 on this.

As for the payload term, I think it's the appropriate generic term for this sort of things so I don't think it's an issue to use it there and I don't think it collides with the other payloads as it's a very generic term.

gunnarmorling: > I asked on HipChat and you told me to name it whatever I feel it to be the most appropriate so that's what I did :)

Well, of course I was expecting you to do a wise choice ;)

My concern is that using the same name may lead people to expect that it will automatically be exposed via `getDynamicPayload()`, which won't be the case unless a specific validator implementation decides to do so. The problem is though I can't think of another good name either, apart from yet another "context" flavor.

gsmet: I think payload is a more appropriate term than context.

Let's rename `getPayload()` to `getConstraintValidatorPayload()` and be extra clear in the documentation I still have to write :).

gsmet: One other thing to note is that it's only exposed in the *initialization* context passed to the validator. It's not exposed in the `isValid()` context. That might help with the confusion as only people willing to use the advanced `initialize()` method will notice it.

gunnarmorling: Ok, `getConstraintValidatorPayload()` it is. I see it's all marked as incubating, so we got some liberty to adjust down the road, if needed. Thanks!

mkurz: Alright, @gsmet will you do the renaming when you write the docs?

gsmet: Yes, will do.

mkurz: Alright!

gsmet:
@gunnarmorling @mkurz I added a commit with some documentation. Feedback welcome.

Once the dust is settled on the documentation, we can (finally) merge this feature.

Thanks.

mkurz:
@gsmet Had a quick look at the documentation you added, LGTM. :shipit:

documentation/src/main/asciidoc/ch06.asciidoc
```diff
@@ -210,6 +210,66 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter06/MyFutureVa
...
+[[example-constraint-validator-payload-definition-validatorcontext]]
+.Defining a constraint validator payload using a `Validator` context
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter06/constraintvalidatorpayload/ConstraintValidatorPayloadTest.java[tags=setConstraintValidatorPayloadInValidatorContext]
+----
+====
+
+Once you have set the constraint validator payload, it is time to use it in your constrait validators as shown in the example below:
```
gunnarmorling: s/constrait/constraint/

documentation/src/main/asciidoc/ch06.asciidoc
```diff
@@ -210,6 +210,66 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter06/MyFutureVa
...
+[[example-constraint-validator-payload-usage]]
+.Using the constraint validator payload in a constraint validator
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter06/constraintvalidatorpayload/ZipCodeValidator.java[tags=include]
+----
+====
+
+`HibernateConstraintValidatorInitializationContext#getConstraintValidatorPayload()` has a type parameter
```
gunnarmorling: Question on that one, what happens if I have registered two payloads of type `A` and `B`, `B` extends `A` and I request `A`? Is an order of precedence defined?

documentation/src/main/asciidoc/ch06.asciidoc
```diff
@@ -210,6 +210,66 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter06/MyFutureVa
...
+and returns the payload only if the payload is of the given type.
+
+[NOTE]
+====
+It is important to note that the constraint validator payload is different from the dynamic payload you can include in
+the constraint violation raised.
+
+The whole purpose of this constraint validator payload is to be used to condition the behavior of your constraint validators.
+It is not included in the constraint violations.
```
gunnarmorling: , unless a specific `ConstraintValidator` implementation passes on the payload to emitted constraint violations by calling xyz.

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter06/constraintvalidatorpayload/ConstraintValidatorPayloadTest.java
```diff
@@ -0,0 +1,48 @@
...
+	@Test
+	public void setConstraintValidatorPayloadInValidatorContext() {
+		//tag::setConstraintValidatorPayloadInValidatorContext[]
+		HibernateValidatorFactory hibernateValidatorFactory = Validation.byDefaultProvider()
+				.configure()
+				.buildValidatorFactory()
+				.unwrap( HibernateValidatorFactory.class );
+
+		Validator validator = hibernateValidatorFactory.usingContext()
+				.constraintValidatorPayload( "US" )
```
gunnarmorling: Now that I'm seeing the API used, I'm wondering whether it should be `constraintValidatorPayload( String.class, "US" )`. I can see two advantages:

* It makes it clearer that there can be multiple payloads (for different types)
* One can register a given implementation type under the key of a (public) interface

Btw. what happens if I register multiple payloads of the same type? Last one wins, or is an exception raised?

gunnarmorling: Hum, no, it seems the implementation actually doesn't allow what I had in mind. Wondering whether that's not a bit too limited?

gsmet: We had the exact same discussion about the other payloads recently and you answered me the user could use a Map or a bean of some sort.

I don't think it's limited and it's consistent with the other payloads in the API.

documentation/src/main/asciidoc/ch06.asciidoc
mkurz: If I understand your question right you can't have two payloads, because the last one wins.

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter06/constraintvalidatorpayload/ConstraintValidatorPayloadTest.java
mkurz: > ...and you answered me the user could use a Map or a bean of some sort.
>
> I don't think it's limited and it's consistent with the other payloads in the API.

I fully agree.

documentation/src/main/asciidoc/ch06.asciidoc
mkurz: I leave that up to @gsmet if he wants to mention that.

documentation/src/main/asciidoc/ch06.asciidoc
gsmet: Fixed.

gunnarmorling:
Rebased and applied. Also added a commit for some JavaDoc adjustments. Thanks, everyone!

mkurz:
:tada: Woohoo! Thanks everyone for your feedback and the time you invested! You do a real great job :+1: :smile:

gsmet:
@mkurz thanks for your contribution.

gsmet:
@mkurz just a heads up that we released 6.0.8.Final today with the constraint validator payload feature. Time to use in the Play framework :).

mkurz:
@gsmet Thanks for letting me know! Hopefully next week I will find time to set up a pull request for Play :+1:

mkurz:
@gsmet I tried to make use the hibernate constraint validator payload feature within the Play Framework this week, however I ran into an issue with the caching behaviour of `HibernateConstraintValidator` instances. I am pretty sure it's broken.
See https://hibernate.atlassian.net/browse/HV-1589 and for a fix #939.

PR #896, opened 2017-12-05
HV-1536

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1536


engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetDeclaredMethodHandle.java
```diff
@@ -35,10 +35,6 @@ public static GetDeclaredMethodHandle action(Lookup lookup, Class<?> clazz, Stri
 		return new GetDeclaredMethodHandle( lookup, clazz, methodName, false, parameterTypes );
 	}
 
-	/**
-	 * Before using this method on arbitrary classes, you need to check the {@code HibernateValidatorPermission.ACCESS_PRIVATE_MEMBERS}
-	 * permission against the security manager.
-	 */
```
gsmet: Why did you remove this comment?

gunnarmorling: Because it doesn't apply for this action. We only use it for accessing our own private field in the programmatic API (recent change). We don't need the calling code base to have a permission for that (and indeed the code location using this method isn't applying that check).

gsmet: Yeah, for now. But if we use this class for other things, we need to be careful about this. So maybe change it to a warning that you need to be aware of this issue and check the permission if need be but I wouldn't remove the comment altogether.

gunnarmorling:
@gsmet Reworded the comment as per your request.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetDeclaredMethodHandle.java
```diff
@@ -37,7 +37,7 @@ public static GetDeclaredMethodHandle action(Lookup lookup, Class<?> clazz, Stri
 
 	/**
 	 * Before using this method on arbitrary classes, you need to check the {@code HibernateValidatorPermission.ACCESS_PRIVATE_MEMBERS}
-	 * permission against the security manager.
+-	 * permission against the security manager, if the calling class exposes the handle to clients.
```
gsmet: You put a dash at the beginning of the line to check I would really review it? ;)

Will fix.

gsmet:
Amended and merged! Thanks.

PR #898, opened 2017-12-08
HV-1537 Reintroduce ReflectionParameterNameProvider to avoid API breakage

gsmet:
* https://hibernate.atlassian.net/browse/HV-1537
 * https://hibernate.atlassian.net/browse/HV-1538

I decided to mark it as deprecated so that we can imagine removing it at some point.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/parameternameprovider/ReflectionParameterNameProvider.java
+++ b/engine/src/main/java/org/hibernate/validator/parameternameprovider/ReflectionParameterNameProvider.java
@@ -1,0 +1,51 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.parameternameprovider;
+
+import static org.hibernate.validator.internal.util.CollectionHelper.newArrayList;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.List;
+
+import javax.validation.ParameterNameProvider;
+
+/**
+ * Uses Java 8 reflection to get the parameter names.
+ *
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/validator/parameternameprovider/ReflectionParameterNameProvider.java
```diff
@@ -0,0 +1,53 @@
...
+import static org.hibernate.validator.internal.util.CollectionHelper.newArrayList;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.List;
+
+import javax.validation.ParameterNameProvider;
+
+/**
```
gunnarmorling: I'd suggest to remove the documentation apart from the deprecation note, making it even less attractive to use the type. It's a pattern sucessfully employed by other frameworks, too.

engine/src/main/java/org/hibernate/validator/parameternameprovider/ReflectionParameterNameProvider.java
```diff
@@ -0,0 +1,53 @@
...
+ * Uses Java 8 reflection to get the parameter names.
+ *
+ * <p>For this provider to return the actual parameter names, classes must be compiled with the '-parameters' compiler
+ * argument. Otherwise, the JDK will return synthetic names in the form {@code arg0}, {@code arg1}, etc.</p>
+ *
+ * <p>See also <a href="http://openjdk.java.net/jeps/118">JEP 118</a></p>
+ * @author Khalid Alqinyah
+ *
+ * @since 5.2
+ * @deprecated since 6.0 - getting the parameter names via reflection is now enabled by default
```
gunnarmorling: Let's also add a remark that it will be removed in a future version.

engine/src/main/java/org/hibernate/validator/parameternameprovider/ReflectionParameterNameProvider.java
gsmet: Done.

gsmet:
@gunnarmorling addressed your concerns. Should be ready to go.

PR #899, opened 2017-12-12
HV-1533 Some small performance improvements

gsmet:
* https://hibernate.atlassian.net/browse/HV-1533

Found out while working with Marko on the benchmarking of this PR: https://github.com/hibernate/hibernate-validator/pull/889 .

+ a small change requested by John.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -6,7 +6,6 @@
  */
 package org.hibernate.validator.internal.engine;
 
-import static org.hibernate.validator.internal.util.CollectionHelper.newHashMap;
 import static org.hibernate.validator.internal.util.logging.Messages.MESSAGES;
 
 import java.lang.annotation.ElementType;
@@ -17,6 +16,7 @@ import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -434,7 +434,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 
 	private <U> void validateConstraintsForDefaultGroup(ValidationContext<?> validationContext, ValueContext<U, Object> valueContext) {
 		final BeanMetaData<U> beanMetaData = valueContext.getCurrentBeanMetaData();
-		final Map<Class<?>, Class<?>> validatedInterfaces = newHashMap();
+		final Map<Class<?>, Class<?>> validatedInterfaces = new HashMap<>();
 
 		// evaluating the constraints of a bean per class in hierarchy, this is necessary to detect potential default group re-definitions
 		for ( Class<? super U> clazz : beanMetaData.getClassHierarchy() ) {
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -434,7 +434,7 @@ private void validateConstraintsForCurrentGroup(ValidationContext<?> validationC
 
 	private <U> void validateConstraintsForDefaultGroup(ValidationContext<?> validationContext, ValueContext<U, Object> valueContext) {
 		final BeanMetaData<U> beanMetaData = valueContext.getCurrentBeanMetaData();
-		final Map<Class<?>, Class<?>> validatedInterfaces = newHashMap();
+		final Map<Class<?>, Class<?>> validatedInterfaces = new HashMap<>();
```
gsmet: @johnaohara Can you confirm it's the change you had in mind? Thanks!

(I plan to release 6.0.6 today to push it to WF 12)

johnaohara: @gsmet yes that was the change for HashMap

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/LocalizedMessage.java
```diff
@@ -31,10 +34,10 @@ public boolean equals(Object o) {
 
 		LocalizedMessage that = (LocalizedMessage) o;
 
-		if ( locale != null ? !locale.equals( that.locale ) : that.locale != null ) {
+		if ( message != null ? !message.equals( that.message ) : that.message != null ) {
```
marko-bekhta: It might not be much of the value, but the way `LocalizedMessage` is used we shouldn't have `locale` or `message` as `null` and also that if statement above:
```java
if ( o == null || getClass() != o.getClass() ) {
    return false;
}
``` 
I don't remember what we've done in the previous case, did we removed them, or not?

gsmet: Ah sure. I did remove the test for the `hashCode()` but forgot to do the same for `equals()`. Will do.

gsmet: @marko-bekhta fixed thanks

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gunnarmorling: Interesting, why is it better? But either way it's fine with me :)

gunnarmorling:
LGTM, merging. Thanks!

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
gsmet: From what John said, it prevents the compiler from doing optimizations. But apparently, that wasn't expected by John. So we fixed it there for now but it probably requires more investigation on the compiler side.

johnaohara: The change allowed ValidatorImpl.validateConstraintsForDefaultGroup() and ValidatorImpl.validateMetaConstraint() to be inlined

Also, there was contention on a futex call from the JVM, changing from a static method call reduced overall the cpu time spent spinning on futexes. This I have not seen before and need to understand better.

PR #900, opened 2017-12-12
HV-1545 Reintroduce hibernate.validator.constraint_mapping_contributor

gsmet:
Mark it as planned for removal so that it's clear we are going to remove
it in a future version.

This reverts commit 1c03bace19e2bfb59de9e81b049ab0dac84481de and adds a
test to be sure the feature works as expected.


engine/src/main/java/org/hibernate/validator/HibernateValidatorConfiguration.java
```diff
@@ -62,6 +62,17 @@
 	 */
 	String ALLOW_PARALLEL_METHODS_DEFINE_PARAMETER_CONSTRAINTS = "hibernate.validator.allow_parallel_method_parameter_constraint";
 
+	/**
```
gunnarmorling: Did you keep the docs on purpose? I'd say let's remove it.

gsmet: Doc removed.

gunnarmorling:
Applied. Thx, @gsmet!

PR #902, opened 2017-12-13
HV-1547 Refactor ValidatorImpl to allow C2 compiler to inline methods

johnaohara:
@gsmet This PR contains refactoring that reduces the bytecode size of ValidatorImpl.determineGroupValidationOrder()


gsmet:
Hi @johnaohara ,

Is the compiler able to inline the 2 extracted methods too? Otherwise if I understand these things coreectly, we would end up with 2 method calls instead of one?

johnaohara:
After further profiling with the changes, the extracted methods are not being compiled. Even though ValidatorImpl.determineGroupValidationOrder() is now being inlined we still need to call the extracted methods which are now interpreted and won't be inlined

PR #903, opened 2017-12-21
Travis build improvements

yrodiere:
https://hibernate.atlassian.net/browse/HV-1549
https://hibernate.atlassian.net/browse/HV-1550

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,9 @@ install:
   - mvn -N io.takari:maven:wrapper
   - ./mvnw -v
   # first run to download all the Maven dependencies without logging
-  - travis_wait ./mvnw $BUILD_OPTIONS -U -B -q -DskipTests=true install
+  - travis_wait ./mvnw $BUILD_OPTIONS -U -B -q -Dcheckstyle.skip -DskipTests=true install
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,12 +16,19 @@ addons:
...
+  - travis_wait ./mvnw $BUILD_OPTIONS -U -B -q -Dcheckstyle.skip -DskipTests=true install
 before_script:
 script:
-  - ./mvnw -s settings-example.xml clean verify
+  - ./mvnw $BUILD_OPTIONS -DskipDistro=true checkstyle:check
+    && BUILD_OPTIONS+=' -Dcheckstyle.skip'
+    && ./mvnw $BUILD_OPTIONS clean install
+before_cache:
+  # Do not put Hibernate Validator artifacts into the cache
+  - rm -r $HOME/.m2/repository/org/hibernate/validator/*
```
gsmet: I think we also need to remove the `$HOME/.m2/repository/org/hibernate/hibernate-validator*` directories. We still have the relocation artifacts.

yrodiere: Right; done!

PR #904, opened 2018-01-07
HV-1551 @Size annotation fails on not parametrized Map property

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1551

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorDescriptor.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorDescriptor.java
@@ -12,7 +12,6 @@ import java.lang.reflect.AnnotatedParameterizedType;
 import java.lang.reflect.AnnotatedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
-import java.lang.reflect.WildcardType;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
@@ -75,7 +74,7 @@ public class ValueExtractorDescriptor {
 			AnnotatedParameterizedType parameterizedExtractedType = (AnnotatedParameterizedType) containerType;
 			int i = 0;
 			for ( AnnotatedType typeArgument : parameterizedExtractedType.getAnnotatedActualTypeArguments() ) {
-				if ( !isUnboundWildcard( typeArgument.getType() ) ) {
+				if ( !TypeHelper.isUnboundWildcard( typeArgument.getType() ) ) {
 					throw LOG.getOnlyUnboundWildcardTypeArgumentsSupportedForContainerTypeOfValueExtractorException( extractorImplementationType );
 				}
 				if ( typeArgument.isAnnotationPresent( ExtractedValue.class ) ) {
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/size/SizeValidatorForCollection.java
```diff
@@ -20,7 +20,7 @@
  *
  * @author Hardy Ferentschik
  */
-public class SizeValidatorForCollection implements ConstraintValidator<Size, Collection<?>> {
+public class SizeValidatorForCollection implements ConstraintValidator<Size, Collection> {
```
gsmet: Having a `SuppressWarnings` and a comment explaining why we did it like that would be nice.

It's not possible to keep the wildcard and fix it in another place?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/size/SizeValidatorForMap.java
```diff
@@ -20,7 +20,7 @@
  *
  * @author Hardy Ferentschik
  */
-public class SizeValidatorForMap implements ConstraintValidator<Size, Map<?, ?>> {
+public class SizeValidatorForMap implements ConstraintValidator<Size, Map> {
```
gsmet: Same here about the `SuppressWarnings` and comment if we can't fix it another way.

engine/src/test/java/org/hibernate/validator/test/constraints/annotations/SizeOnParametrizedAndNonParametrizedContainersTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+import org.hibernate.validator.testutil.TestForIssue;
+
+import org.testng.annotations.Test;
+
+/**
+ * @author Marko Bekhta
+ */
+@TestForIssue(jiraKey = "HV-1551")
+public class SizeOnParametrizedAndNonParametrizedContainersTest extends AbstractConstrainedTest {
```
gsmet: Paramet*e*rized * 2 here.

And the same in the method names.

gsmet:
I think it would also be interesting to check if it's a regression. And if so, what caused it exactly.

marko-bekhta:
@gsmet I checked this against 5.4 and there's same issue. 
The problem is that when we are filtering through available validators `TypeHelper#isAssignable()` receives as a `supertype` parametrized type (`Map<?, ?>`) and as a `type` - simple `Map`. So in the end it calls `TypeHelper#isSuperAssignable()` where `getResolvedSuperclass( type )` is null and initial method returns `false`.

At first I thought about updating the `TypeHelper#isAssignable()` with something similar to:

```java
if ( supertype instanceof ParameterizedType ) {
// the first statement in this ^ `if`
    if ( type.equals( getErasedType( supertype ) ) ) {
        return true;
    }
}
```

But I wasn't 100% sure if it'll be a good idea to change this helper method as it's used in a couple more places... And also applying such change would make the next test fail
```java
public class TypeHelperTest {
    //.....    
    @Test
    public void isAssignableWithClassFromParameterizedType() {
        assertAsymmetricallyAssignable( List.class, TypeHelper.parameterizedType( List.class, Integer.class ) );
    }
    //.....
}
```

gsmet:
@marko-bekhta I tried another approach, which I think is more correct (note that there are probably a lot of other ways to break `TypeHelper` but I think this will do for now).

All tests are passing.

gsmet:
After a discussion on HipChat with Marko, I squashed the commits and merged. Thanks Marko for the initial work!

PR #905, opened 2018-01-09
 HV-1555  HV-1556 Fixes related to OSGi testing

gsmet:
We have some issues with one of our OSGi tests randomly failing. I hope this will help to fix the issue.

* https://hibernate.atlassian.net/browse/HV-1555
* https://hibernate.atlassian.net/browse/HV-1556

### DIFF
--- a/osgi/felixtest/pom.xml
+++ b/osgi/felixtest/pom.xml
@@ -20,7 +20,6 @@
     <description>Hibernate Validator OSGi integration test for Felix using Payara Server</description>
 
     <properties>
-        <hibernate.validator.version>6.0.3-SNAPSHOT</hibernate.validator.version>
         <hibernate-validator-parent.path>../..</hibernate-validator-parent.path>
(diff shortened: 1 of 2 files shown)


gsmet:
@marko-bekhta can you take a quick look? (it's a very simple one)

osgi/felixtest/pom.xml
```diff
@@ -20,7 +20,6 @@
     <description>Hibernate Validator OSGi integration test for Felix using Payara Server</description>
 
     <properties>
-        <hibernate.validator.version>6.0.3-SNAPSHOT</hibernate.validator.version>
```
marko-bekhta: Ahh so maybe that's what caused those random test failures ?

gsmet: That's what I hope :). But I'm not sure. The server is complaining about a missing dependency in a microprofile so that might be it.

PR #906, opened 2018-01-09
HV-1551 Proper fix

gsmet:
* https://hibernate.atlassian.net/browse/HV-1551

Better reviewed commit per commit as some commits are just here to fix typos.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/UniqueElementsValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/UniqueElementsValidator.java
@@ -29,7 +29,10 @@ import org.hibernate.validator.internal.util.CollectionHelper;
  * @author Tadhg Pearson
  * @author Guillaume Smet
  */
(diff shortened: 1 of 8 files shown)


engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/UniqueElementsValidator.java
```diff
@@ -38,7 +41,7 @@
 	 * @return true if the input collection is null or does not contain duplicate elements
 	 */
 	@Override
-	public boolean isValid(Collection<?> collection, ConstraintValidatorContext constraintValidatorContext) {
+	public boolean isValid(Collection collection, ConstraintValidatorContext constraintValidatorContext) {
```
gsmet: I had to keep the `Collection<?>` in the last method of this class as, otherwise, the stream generics just go berserk.

gsmet:
OK, thanks, let's merge it and get CI back to green. Funnily enough, the JDK 8 jobs are working and the JDK 9 jobs are failing consistently.

gunnarmorling:
LGTM, too. Thanks, @gsmet!

Happy 2018, everyone :)

PR #909, opened 2018-01-10
HV-1552 Adding a new MinAge Constraint

Hilmerc:
This is a new constraint, it is working only with LocalDate, perhaps we can extend it to util.Date and sql.Date. Please let me know your thoughts  about it.


engine/src/main/java/org/hibernate/validator/constraints/MinAge.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * The annotated element must be a date where the number of years go by to today must be
+ * greater or equal to the specified value
+ * <p>
+ * <p>
+ * The supported type is {@code LocalDate}. {@code null} is considered valid.
+ * <p>
+ *
+ * @author Hillmer Chona
+ * @since 6.0.x
```
gsmet: Let's make it 6.0.8 for now, we can adjust later if need be.

engine/src/main/java/org/hibernate/validator/constraints/MinAge.java
```diff
@@ -0,0 +1,66 @@
...
+public @interface MinAge {
+
+	String message() default "{org.hibernate.validator.constraints.MinAge.message}";
+
+	Class<?>[] groups() default {};
+
+	Class<? extends Payload>[] payload() default {};
+
+	/**
+	 * @return value the age must be greater or equal to
```
gsmet: It should be clear that it is the number of years.

engine/src/main/java/org/hibernate/validator/constraints/MinAge.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * The annotated element must be a date where the number of years go by to today must be
+ * greater or equal to the specified value
+ * <p>
+ * <p>
+ * The supported type is {@code LocalDate}. {@code null} is considered valid.
+ * <p>
+ *
+ * @author Hillmer Chona
```
gsmet: Isn't it Hilmer with one L? (same in the other classes)

That's how we wrote it in the copyright.txt file so we need to fix it if it's with 2 L.

engine/src/main/java/org/hibernate/validator/constraints/MinAge.java
```diff
@@ -0,0 +1,66 @@
...
+ * @since 6.0.x
+ */
+
+@Target({METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE})
+@Retention(RUNTIME)
+@Repeatable(MinAge.List.class)
+@Documented
+@Constraint(validatedBy = {})
+public @interface MinAge {
```
gsmet: As mentioned in my email, let's make it `AgeMin` to be similar to `DecimalMin`.

engine/src/main/java/org/hibernate/validator/constraints/MinAge.java
```diff
@@ -0,0 +1,66 @@
...
+
+	Class<?>[] groups() default {};
+
+	Class<? extends Payload>[] payload() default {};
+
+	/**
+	 * @return value the age must be greater or equal to
+	 */
+	int value();
+
```
gsmet: As mentioned in my email, we would need an `inclusive()` option, as in DecimalMin.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/MinAgeValidator.java
```diff
@@ -0,0 +1,45 @@
...
+import org.hibernate.validator.constraints.MinAge;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+import java.time.LocalDate;
+import java.time.temporal.ChronoUnit;
+
+/**
+ *
```
gsmet: Useless new line.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/MinAgeValidator.java
```diff
@@ -0,0 +1,45 @@
...
+ * Checks that the number of years from a given date to today is greater or equal to
+ * a specified value
+ *
+ * @author Hillmer Chona
+ * @since 6.0.x
+ */
+
+
```
gsmet: Same here. Be careful about the formatting.

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
```diff
@@ -50,4 +51,4 @@ org.hibernate.validator.constraints.pl.NIP.message                  = Invalid VA
...
-org.hibernate.validator.constraints.time.DurationMin.message        = must be longer than${inclusive == true ? ' or equal to' : ''}${days == 0 ? '' : days == 1 ? ' 1 day' : ' ' += days += ' days'}${hours == 0 ? '' : hours == 1 ? ' 1 hour' : ' ' += hours += ' hours'}${minutes == 0 ? '' : minutes == 1 ? ' 1 minute' : ' ' += minutes += ' minutes'}${seconds == 0 ? '' : seconds == 1 ? ' 1 second' : ' ' += seconds += ' seconds'}${millis == 0 ? '' : millis == 1 ? ' 1 milli' : ' ' += millis += ' millis'}${nanos == 0 ? '' : nanos == 1 ? ' 1 nano' : ' ' += nanos += ' nanos'}
+org.hibernate.validator.constraints.time.DurationMin.message        = must be longer than${inclusive == true ? ' or equal to' : ''}${days == 0 ? '' : days == 1 ? ' 1 day' : ' ' += days += ' days'}${hours == 0 ? '' : hours == 1 ? ' 1 hour' : ' ' += hours += ' hours'}${minutes == 0 ? '' : minutes == 1 ? ' 1 minute' : ' ' += minutes += ' minutes'}${seconds == 0 ? '' : seconds == 1 ? ' 1 second' : ' ' += seconds += ' seconds'}${millis == 0 ? '' : millis == 1 ? ' 1 milli' : ' ' += millis += ' millis'}${nanos == 0 ? '' : nanos == 1 ? ' 1 nano' : ' ' += nanos += ' nanos'}
```
gsmet: Don't remove the empty line at the end, please. It's a pain otherwise.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/AgeValidatorTest.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * A set of tests for {@link MinAge} constraint validator ({@link MinAgeValidator}), which
+ * make sure that validation is performed correctly.
+ *
+ * @author Hillmer Chona
+ * @since 6.0.x
+ */
+@TestForIssue(jiraKey = "HV-1552" )
+public class AgeValidatorTest {
+	private MinAgeValidator ageValidator;
```
gsmet: Let's make it clear it's for AgeMin in the name of the property.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/AgeValidatorTest.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
+
+/**
+ * A set of tests for {@link MinAge} constraint validator ({@link MinAgeValidator}), which
+ * make sure that validation is performed correctly.
+ *
+ * @author Hillmer Chona
+ * @since 6.0.x
```
gsmet: Just add the `@since` tag on public facing API e.g. the `AgeMin` annotation and the (future) `AgeMinDef` class.

engine/src/main/java/org/hibernate/validator/constraints/MinAge.java
gunnarmorling: I'm wondering whether we should provide an attribute for the time unit, too? It'd default to `ChronoUnit#YEARS`, but it'd allow for more fine-grained configuration if needed. But I'm not sure what use cases there are? Any thoughts?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/MinAgeValidator.java
```diff
@@ -0,0 +1,45 @@
...
+public class MinAgeValidator implements ConstraintValidator<MinAge, LocalDate> {
```
gunnarmorling: `LocalDateTime` seems worth supporting, too?

gsmet: Yeah, what I discussed with Hilmer is that we should have one type supported well. Then we would add the other types on the same model once we have converged.

gsmet:
> I'm wondering whether we should provide an attribute for the time unit, too? It'd default to ChronoUnit#YEARS, but it'd allow for more fine-grained configuration if needed. But I'm not sure what use cases there are? Any thoughts?

Yup, I was also wondering about that (either that or have years, months, days attributes but your proposal will be easier and probably sufficient). That's part of the second step I have in mind once we have settled everything else. I would like to be sure that we have a way to extract the values for all the types we want to support.

marko-bekhta:
Hi @Hilmerc thanks for the PR, it looks like an interesting addition to a set of standard constraints. 

Here are some general thoughts:
- We should make use of clock provider in validators for this new constraint. (for example on how it can be accessed, see next validator for the time constraints that are already present - https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/time/AbstractJavaTimeValidator.java#L41)
- I'm also wondering if we should make use of HV specific time tolerance property for such constraint. This question is probably more to @gsmet and @gunnarmorling. My thoughts are next - if we keep this constraint just for years then it probably doesn't make sense to apply tolerance, but if we allow other time units it might be useful.

Hilmerc:
I will be working on all your comments.

engine/src/main/java/org/hibernate/validator/constraints/MinAge.java
Hilmerc: My name is Hillmer with double L

engine/src/main/java/org/hibernate/validator/constraints/MinAge.java
Hilmerc: Done

gsmet:
> I'm also wondering if we should make use of HV specific time tolerance property for such constraint. This question is probably more to @gsmet and @gunnarmorling. My thoughts are next - if we keep this constraint just for years then it probably doesn't make sense to apply tolerance, but if we allow other time units it might be useful.

Yes, we need to take into account the tolerance because, even in years, you can be one more year old for a few milliseconds.

But let's do it step by step.

engine/src/main/java/org/hibernate/validator/constraints/MinAge.java
gsmet: I fixed it in copyright.txt, thanks. BTW, your email configuration says "Hilmer" when you're sending emails, that's why I got confused.

Hilmerc:
Hi @marko-bekhta I am working with all suggestions, I implemented  the clock provider as the example you told me does, but I have some problems with Test,  testProgrammaticDefinition works perfect,  but other tests don't. 

I am trying to validate a simple value instance of LocalDate: 
_assertTrue( ageMinValidator.isValid( LocalDate.now().minusYears( 18 ), null ), birthDate + "  valid " );_

I get a NullPointer because the initialize() method called is from ConstraintValidator and this method doesn't create clock provider.

What do you think can be the problem?

marko-bekhta:
Hi @Hilmerc, right ... In case you followed `AbstractJavaTimeValidator` example which implements `HibernateConstraintValidator` you should also use  `ConstraintValidatorInitializationHelper#initialize()` method in your tests. It will pass a dummy initialization context to your validators init method (an example using this method is - https://github.com/hibernate/hibernate-validator/blob/master/engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/ScriptAssertValidatorTest.java#L210). 
Looking at the code that is already in this PR, your test probably should look something like:
```java
@Test
public void testSimpleAgeMin() throws Exception {
    ConstraintValidator<AgeMin, LocalDate> constraintValidator = getInitializedValidator( 18 );
    constraintValidator.isValid( LocalDate.now().minusYears( 18 ), null );
}

/**
 * @return an initialized {@link ConstraintValidator} using {@code DUMMY_CONSTRAINT_VALIDATOR_INITIALIZATION_CONTEXT}
 */
private ConstraintValidator<AgeMin, LocalDate> getInitializedValidator(int value) {
    HibernateConstraintValidator<AgeMin, LocalDate> validator = new AgeMinValidator();

    ConstraintAnnotationDescriptor.Builder<AgeMin> descriptorBuilder = new ConstraintAnnotationDescriptor.Builder<>( AgeMin.class );

    descriptorBuilder.setAttribute( "value", value );

    ConstraintAnnotationDescriptor<AgeMin> descriptor = descriptorBuilder.build();

    ConstraintValidatorInitializationHelper.initialize( validator, descriptor );
    return validator;
}
```

Let me know if this is helpful.

Hilmerc:
Thanks @marko-bekhta , now Test is working very well.

Hilmerc:
I closed this PR and submitted a new one.

PR #910, opened 2018-01-11
HV-1557 Remove old release scripts from the repository

gsmet:
* https://hibernate.atlassian.net/browse/HV-1557

Noticed them as GitHub was complaining about the security issue with nokogiri.


gsmet:
OK, I don't think there is much to discuss here so I rebased and merged.

PR #911, opened 2018-01-11
HV-1558 Put the distribution files to upload into distribution/target/dir so that release scripts find them

yrodiere:
https://hibernate.atlassian.net//browse/HV-1558

See also https://github.com/hibernate/hibernate-noorm-release-scripts/pull/4 for more information about how this `dist` directory will be uploaded.

Note this also fixes an issue: the WildFly patch files were not being uploaded to Sourceforge. Now they will.


gsmet:
Merged, thanks!

PR #912, opened 2018-01-12
HV-1560 Removing Google Guava as testing dependency from engine

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1560


gsmet:
Fixed a minor formatting glitch and merged, thanks!

PR #916, opened 2018-01-31
HV-1542 Reduce the number of calls to ReflectionHelper#getValue()

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1542

@gsmet I found that patch that I mentioned for this change... I run a few benchmarks for it and here are some of the results:
```bash
INFO: HV000001: Hibernate Validator 6.0.7.Final

# Run complete. Total time: 00:01:31

Benchmark                                            Mode  Cnt    Score   Error  Units
RawValidationSpeedBenchmark.testCascadedValidation  thrpt   30  202.036 ± 3.562  ops/s

# Run complete. Total time: 00:01:06

Benchmark                                                                   Mode  Cnt     Score     Error  Units
RawValidationSpeedWithMultipleConstraintsBenchmark.testCascadedValidation  thrpt   30  4373.006 ± 161.747  ops/s


INFO: HV000001: Hibernate Validator 6.0.8-SNAPSHOT

# Run complete. Total time: 00:01:32

Benchmark                                            Mode  Cnt    Score   Error  Units
RawValidationSpeedBenchmark.testCascadedValidation  thrpt   30  200.153 ± 6.807  ops/s

# Run complete. Total time: 00:01:06

Benchmark                                                                   Mode  Cnt     Score    Error  Units
RawValidationSpeedWithMultipleConstraintsBenchmark.testCascadedValidation  thrpt   30  4751.330 ± 69.204  ops/s
```

Some difference is more or less visible on the second benchmark. It's a very simple test where a bean has multiple constraints defined for the same property. Here's the relevant code for it:
```java
@Benchmark
@BenchmarkMode(Mode.Throughput)
@OutputTimeUnit(TimeUnit.SECONDS)
@Fork(value = 1)
@Threads(50)
@Warmup(iterations = 20) // it seems that as there are a lot of beans it takes some time to warmup
@Measurement(iterations = 30)
public void testCascadedValidation(RawValidationSpeedState state, Blackhole bh) {
    for ( Foo o : state.foos ) {
        Set<ConstraintViolation<Foo>> constraintViolations = state.validator.validate( o );
        bh.consume( constraintViolations );
    }
}

private static class Foo {

    private static final Random RANDOM = new Random();
    private static final String[] VALS = { null, "", "\t\t", "text", "email@mail.com", "1234567" };

    @NotNull
    @NotBlank
    @NotEmpty
    @Email
    @Size(max = 1000)
    @Pattern(regexp = ".*")
    private final String string;

    private Foo(String string) {
        this.string = string;
    }

    public static Foo random() {
        return new Foo( VALS[RANDOM.nextInt( VALS.length )] );
    }
}
```

where a list of foos - `state.foos` is initialized with 100 random `Foo`s

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -510,17 +510,6 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 	}
 
 	private void validateMetaConstraints(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, Object parent,
-			Iterable<MetaConstraint<?>> constraints) {
-
-		for ( MetaConstraint<?> metaConstraint : constraints ) {
-			validateMetaConstraint( validationContext, valueContext, parent, metaConstraint );
-			if ( shouldFailFast( validationContext ) ) {
-				break;
-			}
-		}
-	}
-
-	private void validateMetaConstraints(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, Object parent,
 			Map<ConstraintLocation, Set<MetaConstraint<?>>> constraints) {
 
 		for ( Map.Entry<ConstraintLocation, Set<MetaConstraint<?>>> entry : constraints.entrySet() ) {
@@ -529,28 +518,6 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 				break;
 			}
 		}
-	}
-
-	private boolean validateMetaConstraint(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, Object parent, MetaConstraint<?> metaConstraint) {
-		ValueContext.ValueState<Object> originalValueState = valueContext.getCurrentValueState();
-		valueContext.appendNode( metaConstraint.getLocation() );
-		boolean success = true;
-
-		if ( isValidationRequired( validationContext, valueContext, metaConstraint ) ) {
-
-			if ( parent != null ) {
-				valueContext.setCurrentValidatedValue( valueContext.getValue( parent, metaConstraint.getLocation() ) );
-			}
-
-			success = metaConstraint.validateConstraint( validationContext, valueContext );
-
-			validationContext.markConstraintProcessed( valueContext.getCurrentBean(), valueContext.getPropertyPath(), metaConstraint );
-		}
-
-		// reset the value context to the state before this call
-		valueContext.resetValueState( originalValueState );
-
-		return success;
 	}
 
 	private boolean validateMetaConstraintsOnLocation(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, Object parent, Set<MetaConstraint<?>> metaConstraints, ConstraintLocation location) {
@@ -971,13 +938,14 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 	}
 
 	private <T> void validateParametersForSingleGroup(ValidationContext<T> validationContext, Object[] parameterValues, ExecutableMetaData executableMetaData, Class<?> currentValidatedGroup) {
-		if ( !executableMetaData.getCrossParameterConstraints().isEmpty() ) {
+		Map<ConstraintLocation, Set<MetaConstraint<?>>> crossParameterConstraints = executableMetaData.getCrossParameterConstraints();
+		if ( !crossParameterConstraints.isEmpty() ) {
 			ValueContext<T, Object> valueContext = getExecutableValueContext(
 					validationContext.getRootBean(), executableMetaData, executableMetaData.getValidatableParametersMetaData(), currentValidatedGroup
 			);
 
 			// 1. validate cross-parameter constraints
-			validateMetaConstraints( validationContext, valueContext, parameterValues, executableMetaData.getCrossParameterConstraints() );
+			validateMetaConstraints( validationContext, valueContext, parameterValues, crossParameterConstraints );
 			if ( shouldFailFast( validationContext ) ) {
 				return;
 			}
@@ -1010,7 +978,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 				}
 			}
 
-			validateMetaConstraints( validationContext, valueContext, parameterValues, parameterMetaData );
+			validateMetaConstraints( validationContext, valueContext, parameterValues, parameterMetaData.getAllConstraints() );
 			if ( shouldFailFast( validationContext ) ) {
 				return;
 			}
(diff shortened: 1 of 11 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -542,6 +553,32 @@ private boolean validateMetaConstraint(ValidationContext<?> validationContext, V
 		return success;
 	}
 
+	private boolean validateMetaConstraintsOnLocation(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, Object parent, Set<MetaConstraint<?>> metaConstraints, ConstraintLocation location) {
+		ValueContext.ValueState<Object> originalValueState = valueContext.getCurrentValueState();
+		valueContext.appendNode( location );
```
gsmet: Can't we do this only once somehow?

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -477,15 +477,15 @@ private void validateConstraintsForCurrentGroup(ValidationContext<?> validationC
 	}
 
 	private <U> boolean validateConstraintsForSingleDefaultGroupElement(ValidationContext<?> validationContext, ValueContext<U, Object> valueContext, final Map<Class<?>, Class<?>> validatedInterfaces,
-			Class<? super U> clazz, Set<MetaConstraint<?>> metaConstraints, Group defaultSequenceMember) {
+			Class<? super U> clazz, Map<ConstraintLocation, Set<MetaConstraint<?>>> metaConstraintsByLocation, Group defaultSequenceMember) {
```
gsmet: I think you did the change the risky way (but i agree it's cute :)). And indeed a TCK test is broken.

I think you should still call this `validateConstraintsForSingleDefaultGroupElement()` method once per location. I think it's going to be more code changes to do it the way I think it should be but it will be safer.

IIRC I posted the pseudo code I think we should use in the JIRA issue. And it's in line with my other comment about making the `appendNode()` only once if we can.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaData.java
```diff
@@ -87,13 +89,22 @@
 	 */
 	Set<MetaConstraint<?>> getMetaConstraints();
```
gsmet: Couldn't we get rid of the 2 existing methods? I think they could go away and we should remove the 2 fields in the impl (no need to store the elements twice).

(I just did a quick check so I might be wrong)

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/BeanMetaDataImpl.java
```diff
@@ -482,22 +460,19 @@ else if ( defaultGroupSequence != null && !defaultGroupSequence.isEmpty() ) {
 		return context;
 	}
 
-	private Set<MetaConstraint<?>> getDirectConstraints() {
-		Set<MetaConstraint<?>> constraints = newHashSet();
-
+	private Map<ConstraintLocation, Set<MetaConstraint<?>>> getDirectConstraints() {
```
gsmet: Let's call this method buildDirectMetaConstraintsByLocationMap()

gsmet:
I ran a couple of benchmarks on my laptop and HV is definitely slower with these commits.

Could you also try to run the simple HV JMH benchmarks with and without these commits to check I'm not the only one with this behavior?

BTW you were mentioning some weird behaviors comparing 6.0.7.Final and snapshots before these commits. It would be nice to check we haven't introduced a perf regression somewhere. We upgraded a few libraries but that's about it and I don't see how the commits in master could make master slower than 6.0.7.Final.

If we can't get it to be faster, I think we should withdraw this patch. It's a bit unfortunate as my guts tell me it should be faster but...

gunnarmorling:
The map seems a big price to pay for the cases where there's just a single constraint on one element. Did you see my comment on the issue, suggesting to keep it as a set, but properly ordered by location -- so all constraints on one element are adjecent in that set --, which would mean that the value just needs to be fetched if the location of the next element is different from the current one. This might be cheaper than doing the map-of-sets approach.

marko-bekhta:
hi @gunnarmorling thanks for checking this :) I've actually spent some more time on this, doing a bit of profiling to see if there was anything interesting. Nothing obvious came up from that though... And I've also tried to change the map to a set of pairs (`ConstraintLocation`,`Set<MetaConstraint>` ) but that obviously didn't help much :). 
I've left the sorted set for the end as it will need to change the logic a bit and I wanted to try if this map approach can be somehow improved. 
Right now it looks that this change slows down all regular cases but speeds up the case where there are multiple constraints defined on the same location (see the benchmark and results from the first comment in the PR). 
So I'll try the sorted set approach and will get back to you with the results of that :)

gsmet:
@gunnarmorling yeah I saw your comment. I wasn't sold on the SortedSet approach as I think it's going to make the logic less readable. As `ValidatorImpl`is already hard to grasp, I'm not sure I'll go with it. But if Marko wants to try it, we can decide once it's done.

gunnarmorling:
Good point about readability. It wouldn't literally have to be a SortedSet,
though. Some specifically tailed type might mitigate this concern:

    interface ConstraintsByLocation {
        boolean hasNext(); // is there another location?
        void next(); // advance to next location
        ConstraintLocation getLocation(); // get current location
        Set<MetaConstraint<?>> getConstraints(); // get constraints of
current location
    }

This could be implemented using a single (sorted) set internally but
encapsulate the complexity of transition from one location to the next.
Just some food for thought :)

marko-bekhta:
I've put together a sorted set approach just to see how it will work out. Here's the results I've got:
```bash
HV000001: Hibernate Validator 6.0.8-SNAPSHOT

Benchmark                                                                    Mode  Cnt    Score    Error   Units
MultipleSimpleConstraintsOnSameLocationValidation.testSimpleBeanValidation  thrpt   20  579.386 ± 13.231  ops/ms

# Run complete. Total time: 00:00:32

Benchmark                                                                 Mode  Cnt     Score    Error  Units
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1843.622 ± 21.652  ops/s

```
```bash
HV000001: Hibernate Validator 6.0.7.Final

Benchmark                                                                    Mode  Cnt    Score    Error   Units
MultipleSimpleConstraintsOnSameLocationValidation.testSimpleBeanValidation  thrpt   20  527.489 ± 16.798  ops/ms
# Run complete. Total time: 00:00:33

Benchmark                                                                 Mode  Cnt     Score    Error  Units
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1923.361 ± 44.022  ops/s
```
So it's still slower in regular cases and the boost in case when multiple constraints are present on the same location isn't that high.... Based on this, I'd say it'll be better to improve the performance of `ReflectionHelper#getValue()` with MH and leave the logic in `ValidatorImpl` as is, until we decide to do some refactoring in there. WDYT ?

gsmet:
@marko-bekhta thanks for experimenting. Let's close this PR then.

gsmet:
@marko-bekhta btw, could you create a PR with your Set experiment so that we have it archived?

Thanks!

marko-bekhta:
@gsmet sure I'll create one later today. (Have those changes on a different machine).

PR #918, opened 2018-02-05
HV-1566 Cache type variable index for performance reasons

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1566

and same results as in the JIRA ticket:
```bash
current master:

# Run complete. Total time: 00:01:52

Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  160.152 ± 11.923  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  471.770 ±  8.797   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  633.382 ± 11.182   ops/s

with changes

# Run complete. Total time: 00:01:50

Benchmark                                                                                   Mode  Cnt    Score    Error   Units
CascadedValidation.testCascadedValidation                                                  thrpt   20  170.439 ± 15.699  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  471.262 ± 20.874   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  703.661 ± 39.131   ops/s
```

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValueContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValueContext.java
@@ -171,20 +171,19 @@ public class ValueContext<T, V> {
 	 * Sets the container element information.
 	 *
 	 * @param containerClass the class of the container
-	 * @param typeParameter the actual type parameter (i.e. not one of our internal type parameters)
+	 * @param typeParameterIndex the index of the actual type parameter
 	 *
 	 * @see TypeVariables#getContainerClass(TypeVariable)
 	 * @see TypeVariables#getActualTypeParameter(TypeVariable)
 	 * @see AnnotatedObject
 	 * @see ArrayElement
 	 */
-	public final void setTypeParameter(Class<?> containerClass, TypeVariable<?> typeParameter) {
+	public final void setTypeParameter(Class<?> containerClass, Integer typeParameterIndex) {
 		if ( containerClass == null ) {
 			return;
 		}
 
-		propertyPath.setLeafNodeTypeParameter( containerClass,
-				typeParameter != null ? TypeVariables.getTypeParameterIndex( typeParameter ) : null );
+		propertyPath.setLeafNodeTypeParameter( containerClass, typeParameterIndex );
 	}
 
 	public final void setCurrentGroup(Class<?> currentGroup) {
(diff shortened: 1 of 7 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValueContext.java
```diff
@@ -171,20 +171,19 @@ public final void markCurrentPropertyAsIterableAndSetIndex(Integer index) {
 	 * Sets the container element information.
 	 *
 	 * @param containerClass the class of the container
-	 * @param typeParameter the actual type parameter (i.e. not one of our internal type parameters)
+	 * @param typeParameterIndex the index of actual type parameter (i.e. not one of our internal type parameters)
```
gsmet: the index of **the** actual type parameter

gsmet:
OK, this looks good and it's a good thing to improve the container element case. I don't think we have benchmarked it a lot. Will merge.

engine/src/main/java/org/hibernate/validator/internal/engine/ValueContext.java
```diff
@@ -171,20 +171,19 @@ public final void markCurrentPropertyAsIterableAndSetIndex(Integer index) {
 	 * Sets the container element information.
 	 *
 	 * @param containerClass the class of the container
-	 * @param typeParameter the actual type parameter (i.e. not one of our internal type parameters)
+	 * @param typeParameterIndex the index of the actual type parameter (i.e. not one of our internal type parameters)
```
gunnarmorling: Seems the (...) part is superfluous now?

marko-bekhta: yes it seems so, as we just pass an already calculated index to this method. Removed the (...) part.

gsmet:
Merged with a few further minor tweaks: in our getter, we usually don't use `this.` (at least not in the modified classes) so I removed them.

Thanks!

marko-bekhta:
> we usually don't use this.

Ahh, ok. That's just how the IDE generates them. Will fix that for HV settings :)

> wondering if we shouldn't introduce our own structure (TypeParameter?) wrapping the values

That's actually very close to what I have in mind for a different PR I wanted to discuss. I was just waiting while we finish with the opened ones :)

PR #919, opened 2018-02-12
HV-1568 Prevent possible NullPointerException in ValidatorFactoryScopedContext.Builder

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1568

As default context cannot be null at this point we should have an assertion
instead of initialization on null. I've also looked at the logic before this change - we had a reference to `ValidatorFactoryImpl` in `ValidatorContextImpl` which was used to get default values if a null was passed. Now that we have all those things grouped in the context object the context cannot be null as it is initialized in the constructor of `ValidatorFactoryImpl` hence that check in the builder constructor wasn't really needed.


gsmet:
Looks good to me. Will merge tomorrow morning.

gunnarmorling:
Thx, @marko-bekhta!

gsmet:
Merged, thanks!

PR #921, opened 2018-02-26
HV-1572 Make the custom container type value extractor example extra clear

gsmet:
https://hibernate.atlassian.net/browse/HV-1572


gunnarmorling:
Thanks, @gsmet!

PR #925, opened 2018-03-05
HV-1576 Upgrade the integration tests and the base WildFly version to WildFly 12

gsmet:
* https://hibernate.atlassian.net/browse/HV-1576

and then:
 * https://hibernate.atlassian.net/browse/HV-1581
 * https://hibernate.atlassian.net/browse/HV-1582

Better reviewed commit per commit, especially because the first one rejiggers things around.

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -215,7 +215,7 @@
                             javax.el.*;version="[2.0.0,4.0.0)";resolution:=optional,
                             org.xml.sax.*;version="0",
                             org.jboss.logging.*;version="[3.1.0,4.0.0)",
-                            com.fasterxml.classmate.*;version="[${classmate.version},2.0.0)",
+                            com.fasterxml.classmate.*;version="[1.3,2.0.0)",
                             org.joda.time.*;version="[2.0.0,3.0.0)";resolution:=optional,
                             org.jsoup.*;version="[1.5.2,2.0.0)";resolution:=optional,
                             javax.money;version="[1.0.0,2.0.0)";resolution:=optional,
--- a/osgi/felixtest/pom.xml
+++ b/osgi/felixtest/pom.xml
@@ -111,6 +111,7 @@
                         <id>felix-integration-test</id>
                         <goals>
                             <goal>integration-test</goal>
+                            <goal>verify</goal>
                         </goals>
                     </execution>
                 </executions>
--- a/cdi/pom.xml
+++ b/cdi/pom.xml
@@ -88,7 +88,7 @@
         </dependency>
         <dependency>
             <groupId>org.jboss.weld</groupId>
-            <artifactId>weld-core</artifactId>
+            <artifactId>weld-core-impl</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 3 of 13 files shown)


gsmet:
Mmmh, there is a failure in the Felix CDI test run on Payara on CI. A bit surprised as I ran the tests locally. They also run successfully on Travis so let's try to trigger the tests again.

gsmet:
Jenkins retest this please

gsmet:
See https://hibernate.atlassian.net/browse/HV-1581 for why the build doesn't fail on Travis or locally. The test fails though, working on fixing it.

gsmet:
Jenkins, retest this please

osgi/felixtest/pom.xml
```diff
@@ -111,6 +111,7 @@
                         <id>felix-integration-test</id>
                         <goals>
                             <goal>integration-test</goal>
+                            <goal>verify</goal>
```
gunnarmorling: Uff.

engine/pom.xml
```diff
@@ -215,7 +215,7 @@
                             javax.el.*;version="[2.0.0,4.0.0)";resolution:=optional,
                             org.xml.sax.*;version="0",
                             org.jboss.logging.*;version="[3.1.0,4.0.0)",
-                            com.fasterxml.classmate.*;version="[${classmate.version},2.0.0)",
+                            com.fasterxml.classmate.*;version="[1.3,2.0.0)",
```
gunnarmorling: What's the rationale behind that? Fixing it seems like a step back on first sight?

gsmet: Well, it's fixed for all the others so I would say it's a step forward for consistency :).

The rationale behing that is that Payara/GlassFish is still using 1.3.3 so upgrading to 1.3.4 makes the Payara/GlassFish CDI/OSGi tests fail.

I could use the build-helper-plugin to extract the major.minor version but I'm not sure it's worth it. And I'm not even sure we would want to require 1.4 if we upgrade to 1.4. That's why I did it like that.

osgi/felixtest/pom.xml
gsmet: Yeah :). To be fair, it's not the most natural setup.

engine/pom.xml
gsmet: FWIW, when I will work on making the pom files consistent with JBoss parent/WF parent, the whole range will be parameterized and put in the `<properties>` part of the parent pom; they won't be hidden in engine.

That's how it's done in Search now and I think it's better: we define the range we want but it's still very visible so we don't forget to update it if required.

This is for 6.1 though.

gunnarmorling: Ok, that sounds good. On that consistency effort, I hope this won't mean we'll inherit from any JBoss or WF POM, right?

gunnarmorling:
Rebased and applied; thx, @gsmet!

engine/pom.xml
gsmet: > On that consistency effort, I hope this won't mean we'll inherit from any JBoss or WF POM, right?

Nope. That's the limit we set with Sanne.

PR #931, opened 2018-03-07
HV-1450 BeanMetaDataImpl.BeanMetaDataBuilder#build() can choose ConstraintMetaData w/o constraints

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1450

In the end it turned out that this could be fixed with not so much changes:
- changed the logic on how executable builder determines if it should accept a constrained element. Used a most specific sub type from the hierarchy under inspection to resolve executables and check if they override each other somewhere in the hierarchy.
- changed the order of metadata providers. Made annotation based as a first provider. This should build all possible builder delegates after the first iteration on metadata providers and data from other providers will be just merged in.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/BeanMetaDataManager.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/BeanMetaDataManager.java
@@ -127,9 +127,6 @@ public class BeanMetaDataManager {
 		this.parameterNameProvider = parameterNameProvider;
 		this.validationOrderGenerator = validationOrderGenerator;
 
-		this.metaDataProviders = newArrayList();
-		this.metaDataProviders.addAll( optionalMetaDataProviders );
-
 		this.methodValidationConfiguration = methodValidationConfiguration;
 
 		this.beanMetaDataCache = new ConcurrentReferenceHashMap<>(
@@ -141,15 +138,19 @@ public class BeanMetaDataManager {
 				EnumSet.of( IDENTITY_COMPARISONS )
 		);
 
-		AnnotationProcessingOptions annotationProcessingOptions = getAnnotationProcessingOptionsFromNonDefaultProviders();
+		AnnotationProcessingOptions annotationProcessingOptions = getAnnotationProcessingOptionsFromNonDefaultProviders( optionalMetaDataProviders );
 		AnnotationMetaDataProvider defaultProvider = new AnnotationMetaDataProvider(
 					constraintHelper,
 					typeResolutionHelper,
 					valueExtractorManager,
 					annotationProcessingOptions
 			);
-
+		this.metaDataProviders = newArrayList();
+		// default annotation based metadata provider should go first in the list of providers to prevent possible issues
+		// similar to HV-1450. This way after the first run we will have all possible constrained elements in the list and
+		// should only merge in additional constraint information from other providers.
 		this.metaDataProviders.add( defaultProvider );
+		this.metaDataProviders.addAll( optionalMetaDataProviders );
 	}
 
 	@SuppressWarnings("unchecked")
@@ -195,9 +196,9 @@ public class BeanMetaDataManager {
 	/**
 	 * @return returns the annotation ignores from the non annotation based meta data providers
 	 */
-	private AnnotationProcessingOptions getAnnotationProcessingOptionsFromNonDefaultProviders() {
+	private AnnotationProcessingOptions getAnnotationProcessingOptionsFromNonDefaultProviders(List<MetaDataProvider> optionalMetaDataProviders) {
 		AnnotationProcessingOptions options = new AnnotationProcessingOptionsImpl();
-		for ( MetaDataProvider metaDataProvider : metaDataProviders ) {
+		for ( MetaDataProvider metaDataProvider : optionalMetaDataProviders ) {
 			options.merge( metaDataProvider.getAnnotationProcessingOptions() );
 		}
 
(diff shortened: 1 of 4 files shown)


gsmet:
After these changes, `org.hibernate.validator.test.cdi.internal.methodvalidation.BasicMethodValidationTest.testReturnValueValidation` fails.

marko-bekhta:
I dig a bit more into that test failure. If I understand it correctly the problem is with CDI's (Weld?) proxies and what they return for `clazz.getGenericInterfaces()`. 
In case of simple class, which in our case is `org.hibernate.validator.test.cdi.internal.methodvalidation.DefaultRepeater` for `getGenericInterfaces()` we would get:
```
result = {Type[1]@4948} 
 0 = {ParameterizedTypeImpl@4863} "org.hibernate.validator.test.cdi.internal.methodvalidation.Repeater<java.lang.String>"
```
But in case of a proxy `org.hibernate.validator.test.cdi.internal.methodvalidation.DefaultRepeater$Proxy$_$$_WeldSubclass` for `getGenericInterfaces()` we would get next:
```
 result = {Class[6]@4880} 
 0 = {Class@3049} "interface org.hibernate.validator.test.cdi.internal.methodvalidation.Repeater"
 1 = {Class@327} "interface java.io.Serializable"
 2 = {Class@4636} "interface org.jboss.weld.proxy.WeldConstruct"
 3 = {Class@4638} "interface org.jboss.weld.interceptor.util.proxy.TargetInstanceProxy"
 4 = {Class@4637} "interface org.jboss.weld.interceptor.proxy.LifecycleMixin"
 5 = {Class@4639} "interface org.jboss.weld.bean.proxy.ProxyObject"
```

We could ignore all the other interfaces except the `Repeater`. You could see that it doesn't have the `<String>` part and is not a `ParameterizedTypeImpl`, hence when it comes to resolving methods in the `ExecutableHelper` we get next situation in`instanceMethodParametersResolveToSameTypes`: 
```
resolvedMethods = {ResolvedMethod[2]@5130} 
 0 = {ResolvedMethod@5132} "reverse"
  _declaringType = {ResolvedInterfaceType@5157} "org.hibernate.validator.test.cdi.internal.methodvalidation.Repeater"
  _member = {Method@5159} "public abstract java.lang.Object org.hibernate.validator.test.cdi.internal.methodvalidation.Repeater.reverse(java.lang.Object)"
  _type = {ResolvedObjectType@4294} "java.lang.Object"
 1 = {ResolvedMethod@5133} "reverse"
  _declaringType = {ResolvedObjectType@5147} "org.hibernate.validator.test.cdi.internal.methodvalidation.DefaultRepeater"
  _member = {Method@5149} "public java.lang.String org.hibernate.validator.test.cdi.internal.methodvalidation.DefaultRepeater.reverse(java.lang.String)"
  _type = {ResolvedObjectType@5150} "java.lang.String"
```

One of the resolved methods has `String` type while the other - `Object` and as a result the methods do not override each other, and we get into the similar situation as in the original problem, there are more builder delegates than needed and metadata is not correctly aggregated. 
But if a regular class is passed (for example DefaultRepeater) we get two `reverse` methods with `String` type and all is fine.
So unless we can make the proxied class return the "correct" interface it will give us such errors. As a workaround we could make the next change:

```java
// instead of simply returning :
return instanceMethodParametersResolveToSameTypes( mainSubType, left, right );

//we can return:
return ClassHierarchyHelper.getHierarchy( mainSubType ).stream()
        .filter( type -> instanceMethodParametersResolveToSameTypes( type, left, right ) )
        .findAny()
        .isPresent();
```
this allows the tests to pass as we'd get the class behind the proxy in that hierarchy and the methods would resolve correctly, but I'd prefer to get the `getGenericInterfaces()` working.

gsmet:
@marko-bekhta interesting detective work.

So, first, I think we should ping the Weld folks to understand if it's a limitation or a bug.

Second, even if they fix it, I don't think we would want to enforce a specific version of Weld on the users, so, if we have a workaround, we should use it.

marko-bekhta:
I've sent a message to weld-dev about this problem earlier today. I'll let you know if there's any answer on it. I also rebased this branch and applied a fix to the first commit and force-pushed the changes.

I used a slightly different approach then the one mentioned in a previous comment as that one actually failed a test in the engine.

gsmet:
@marko-bekhta so you consider it good to merge?

marko-bekhta:
hmmm.... @gsmet to be honest I'm 50/50 on this one. I mean the tests are fine, that last fix excludes the proxy so it looks like it can go as it is. But on the other hand I'd like to hear back from the Weld devs what would they say. And also even though the changes looks small I'm leaning more towards leaving this one for 6.1, we did break CDI couldn't we break something else as well ?

Also on kinda related topic - should we maybe add a couple simple integration tests with Spring ?

marko-bekhta:
@gsmet I've got an answer from the Weld dev and I think we are on the right track with these changes. There, it was also suggested to detecting proxies and retrieving the superclass from them. I've checked the `org.hibernate.validator.internal.util.classhierarchy.Filters.WeldProxyFilter` that we have in HV and it has less checks than was suggested in the answer. So I wonder if we should make any changes to it ? From Martin (on Weld dev) : 
> Since 2.3.0.Beta1 and 3.0.0.Alpha11 classes generated by Weld are marked with the SYNTHETIC modifier. And the name of the proxy class should always contain "$Proxy$". In Weld 3.0.1+ you can even use org.jboss.weld.proxy.WeldConstruct and org.jboss.weld.proxy.WeldClientProxy interfaces to distinguish such classes.

gsmet:
@marko-bekhta about your last message why not but let's do it in another issue.

We have to be very careful to stay compatible with a large portion of the Weld installed base.

gsmet:
Merged, thanks!

PR #932, opened 2018-03-07
HV-1586 Various improvements to performance tests

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1586

Changes:
- changed the validated model in `SimpleValidation` to constantly have same amount of violations
- moved model preparation in `CascadedValidation` from benchmark itself to state preparation
- updated the version for the hv-6.0 profile to the latest final

### DIFF
--- a/performance/pom.xml
+++ b/performance/pom.xml
@@ -202,7 +202,7 @@
             </activation>
             <properties>
                 <beanvalidation-impl.name>Hibernate Validator</beanvalidation-impl.name>
-                <beanvalidation-impl.version>6.0.7.Final</beanvalidation-impl.version>
+                <beanvalidation-impl.version>6.0.8.Final</beanvalidation-impl.version>
             </properties>
             <dependencies>
                 <dependency>
--- a/performance/src/main/java/org/hibernate/validator/performance/simple/SimpleValidation.java
+++ b/performance/src/main/java/org/hibernate/validator/performance/simple/SimpleValidation.java
@@ -8,7 +8,6 @@ package org.hibernate.validator.performance.simple;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
@@ -37,30 +36,16 @@ import org.openjdk.jmh.infra.Blackhole;
  */
 public class SimpleValidation {
 
-	private static final String[] names = {
-			null,
-			"Jacob",
-			"Isabella",
-			"Ethan",
-			"Sophia",
-			"Michael",
-			"Emma",
-			"Jayden",
-			"Olivia",
-			"William"
-	};
-
 	@State(Scope.Benchmark)
 	public static class ValidationState {
 		public volatile Validator validator;
-		public volatile Random random;
+		public volatile Driver driver;
 
 		{
 			ValidatorFactory factory = Validation.buildDefaultValidatorFactory();
 			validator = factory.getValidator();
-			random = new Random();
+			driver = new Driver( null, 17, false );
 		}
-
 	}
 
 	@Benchmark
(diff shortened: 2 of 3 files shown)


performance/pom.xml
```diff
@@ -202,7 +202,7 @@
             </activation>
             <properties>
                 <beanvalidation-impl.name>Hibernate Validator</beanvalidation-impl.name>
-                <beanvalidation-impl.version>6.0.7.Final</beanvalidation-impl.version>
+                <beanvalidation-impl.version>6.0.8.Final</beanvalidation-impl.version>
```
marko-bekhta: wanted to ask if anybody knows of a nice way to do these kind of changes "automatically". I thought of something like:
```xml
<beanvalidation-impl.version>[6.0.0,${project.version})</beanvalidation-impl.version>
```
I tried it and it gives the result:
```
[INFO] --- maven-shade-plugin:3.1.0:shade (default) @ hibernate-validator-performance ---
.....
[INFO] Including javax.validation:validation-api:jar:2.0.1.Final in the shaded jar.
[INFO] Including org.hibernate.validator:hibernate-validator:jar:6.0.8.Final in the shaded jar.
....
```
but I guess it could be unclear. 
Also I've noticed the `previous.stable` property in main pom.xml :
```xml
<previous.stable>6.0.4.Final</previous.stable>
```
should we update this one? and maybe we should use it in performance tests pom as well ?

gsmet: No good idea here. The fact is that they might be different as `previous.stable` could refer to the last version of the previous major.minor release whereas we might want to test the performance against the latest micro.

performance/src/main/java/org/hibernate/validator/performance/simple/SimpleValidation.java
```diff
@@ -71,13 +56,12 @@
 	@Warmup(iterations = 10)
 	@Measurement(iterations = 20)
 	public void testSimpleBeanValidation(ValidationState state, Blackhole bh) {
-		DriverSetup driverSetup = new DriverSetup( state );
-		Set<ConstraintViolation<Driver>> violations = state.validator.validate( driverSetup.getDriver() );
-		assertThat( violations ).hasSize( driverSetup.getExpectedViolationCount() );
+		Set<ConstraintViolation<Driver>> violations = state.validator.validate( state.driver );
+		assertThat( violations ).hasSize( 3 );
 		bh.consume( violations );
```
gsmet: I thought about that and I wonder if we should have 2 different benchmarks: one without violation and one with.

Why? Because most of the time, we expect HV to not return any violation so getting this path as fast as possible is definitely the priority. Whereas having the violation path a bit slower is not that big a deal.

Does it make sense?

marko-bekhta: yes it does. You are right - users expect their models to be valid. 
+1 for two benchmarks.

gsmet:
By the way, another improvement I had in mind: allow to override the version used for the hv-6.0 profile.

gsmet:
@marko-bekhta I gave it some more thoughts and, frankly, the benchmark is quite stable as it is, even with the randomization (probably because we have a lot of iterations).

Maybe we should just keep it that way as it's not that far from a real usage (having valid beans and invalid ones).

gsmet:
So, I merged the changes to `CascadedValidation`. The update of the base 6.0 version was done in another PR and for now let's keep SimpleValidation as is.

marko-bekhta:
@gsmet OK good! I also thought about:

>allow to override the version used for the hv-6.0 profile

And it looks that if we do next:
```bash
mvn package -Dvalidator=hv-6.0 -Dbeanvalidation-impl.version=6.0.4.Final 
```
another option - we could introduce another property only for minor version and pass just it.
Or did you have something else on your mind ?

PR #933, opened 2018-03-07
HV-1585 Omit the cache lookup in AbstractMessageInterpolator for the simple text messages

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1585

Added a check to prevent map lookups in case of simple user defined message that does not require any message resolution and should not be stored in cache (if it is enabled).


gsmet:
Refined the comments and merged. Thanks!

PR #934, opened 2018-03-08
HV-1587 Reduce significantly the tracking of already validated beans and constraints

gsmet:
* https://hibernate.atlassian.net/browse/HV-1587

@gunnarmorling @marko-bekhta I think these changes are safe but I'm really interested in your feedback on this one as it's a rather bold change.

Before commit:
```
  1624.415 ±(99.9%) 21.333 ops/ms [Average]
  (min, avg, max) = (1555.542, 1624.415, 1668.294), stdev = 24.567
  CI (99.9%): [1603.082, 1645.748] (assumes normal distribution)
```

After commit:
```
  2023.440 ±(99.9%) 26.463 ops/ms [Average]
  (min, avg, max) = (1972.431, 2023.440, 2080.609), stdev = 30.475
  CI (99.9%): [1996.977, 2049.903] (assumes normal distribution)
```

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
@@ -16,6 +16,7 @@ import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -38,6 +39,7 @@ import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintVi
 import org.hibernate.validator.internal.engine.path.PathImpl;
 import org.hibernate.validator.internal.metadata.BeanMetaDataManager;
 import org.hibernate.validator.internal.metadata.aggregated.BeanMetaData;
+import org.hibernate.validator.internal.metadata.aggregated.ExecutableMetaData;
 import org.hibernate.validator.internal.metadata.core.MetaConstraint;
 import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
 import org.hibernate.validator.internal.util.logging.Log;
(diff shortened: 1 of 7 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -348,10 +363,22 @@ public void addConstraintFailures(Set<ConstraintViolation<T>> failingConstraintV
 	}
 
 	public boolean hasMetaConstraintBeenProcessed(Object bean, Path path, MetaConstraint<?> metaConstraint) {
+		// this is only useful if the constraint is defined for more than 1 group as in the case it's only
+		// defined for one group, there is no chance it's going to be called twice.
+		if ( metaConstraint.getGroupList().size() <= 1 ) {
```
marko-bekhta: maybe we can replace this with `MetaConstraint#isInMultipleGroups()` we could cache that value as well?

gsmet: Good idea. Will do and force push.

gsmet: Done. It doesn't change anything though but it's cleaner anyway.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidationContext.java
```diff
@@ -372,6 +413,33 @@ public String toString() {
...
+	private static boolean buildDisableAlreadyValidatedBeanTracking(ValidationOperation validationOperation, BeanMetaData<?> rootBeanMetaData,
+			Optional<ExecutableMetaData> executableMetaData) {
+		switch ( validationOperation ) {
+			case BEAN_VALIDATION:
+			case PROPERTY_VALIDATION:
+			case VALUE_VALIDATION:
+				// note that in the case of property and value validation, we are considering the root bean, whereas we
+				// could consider the bean of the property or the value. We don't really have the info here though so it
+				// will do for now.
+				return !rootBeanMetaData.hasCascadingProperties();
```
gunnarmorling: Ok, this should work. I still believe it'd have been simpler if `hasCascadables()` would be exposed on `Validatable` (all three objects you check here implement that interface, so there could be a single method for it). But it's fine this way, too.

PR #935, opened 2018-03-13
HV-1588 Bail out faster during cascading validation when in fail fast mode

gsmet:
* https://hibernate.atlassian.net/browse/HV-1588

If we don't do that, we do quite a lot of useless work.

Note that the tests was added to ensure this case also worked correctly.


gsmet:
Closing this one. Thinking more about it, I'm not sure it's such a good idea to add more overhead when we don't have any violation.

PR #938, opened 2018-03-15
HV-1363 support for non-standard Java beans

marko-bekhta:

I think with an opened PR it'll be easier to discuss this. What I've done so far is next:
- First I wanted to remove the dependency on reflection in `ConstraintLocation` and that resulted in the bigger part of the changes in this PR.
- After that change the fail fast tests started to behave randomly in case there are two constraints on the same parameter. That's why the last commit was needed.
- And probably the most relevant part - I've moved out all the logic related to detecting getters to a separate class - `PropertyFilter`.  And that actually uncovered a couple of new challenges. 

Having simple filtering for property methods seems to be not enough. I've extracted all the things that we do with the getters to this interface:

```java
public interface PropertyFilter {
    boolean isProperty(Executable executable);
    String getPropertyName(Method method);
    Set<String> getPossibleMethodNamesForProperty(String propertyName);
}
```
Why such methods?
- we need to determine if something is a property or not
- if it is a property we need to extract the name part from it
- and in case of XML or programmatic configuration we receive the name of the property and we need to construct back the full name to find a corresponding executable.

Next step that I was planning to do was to continue with the approach suggested by @gunnarmorling in the previous PR for this improvement, i.e. having a `PropertySelector` and refactor `AnnotationMetaDataProvider` to use it. 

```java
public interface PropertySelector<T extends ConstrainableType> {
    Stream<Property> getProperties(T type);
}
```

But actually extracting the getter logic into `PropertyFilter` uncovered the other question. In case of XML or programmatic configuration we need to have such filter right away, when we read xml or build programmatic constraints. And it would mean that we cannot expose the `PropertyFilter` interface as a configuration option on `HibernateValidatorContext`. As exposing it there would allow users to change the filter after the XML config was read. (the similar problem to what we had when we tried to cache the parameter name in the constraint location).
So the question is - do we want to expose this configuration through the context or not ? Seems like exposing it would require to delay the metadata creation from XML and programmatic configuration to the creation of Validator ?


gsmet:
> So the question is - do we want to expose this configuration through the context or not ?

Quick answer for this one: I don't think it's a good idea to do so. It makes sense for it to be a `ValidatorFactory` only setting.

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -304,7 +315,17 @@ public MethodValidationConfiguration getMethodValidationConfiguration() {
...
+		if ( propertyFilter == null ) {
+			propertyFilterToUse = new DefaultPropertyFilter();
+			LOG.creatingConstraintMappingPriorToBuildingFactoryWithoutDefinedFilter();
+		}
+		else {
+			propertyFilterToUse = propertyFilter;
+		}
+		LOG.creatingConstraintMappingPriorToBuildingFactory( propertyFilterToUse.getClass() );
+
+		return new DefaultConstraintMapping( propertyFilterToUse );
```
marko-bekhta: @gsmet even though we expose this "property selection functionality" only through the `HibernateValidatorConfiguration` we still would have a "problem", as it's at the same level where we add constraint mappings so if user adds mappings and then change the `propertyFilter` it most likely would not end good. I've added a few warnings here, as I don't see a nice solution for now. Unless we want to just collect the info here and delay all the work until later time, when building the factory. 
Other than that I think you can give it a first look, when you have time for it :smiley:.

gsmet: Hmmm, it's the second time we hit this wall (we already had this issue when we discussed resolving the parameter names once and for all - it's a bit different here though as we don't have the `HIbernateValidatorContext` issue). Taking a step back, I think it was a pretty bad decision to allow to define the mappings during the configuration phase. It should have been a separate step.

Wondering if we should revisit this idea of postponing the mappings creation to the factory building phase.

But it's something we can do after getting this patch in, if we think it makes sense.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetDeclaredFieldValueWithMethodHandle.java
```diff
@@ -0,0 +1,61 @@
...
+import org.hibernate.validator.internal.util.logging.Log;
+import org.hibernate.validator.internal.util.logging.LoggerFactory;
+
+/**
+ * Returns the declared field value with the specified name, using {@link MethodHandle} or
+ * throws an exception if it does not exist or cannot be accessed.
+ *
+ * @author Marko Bekhta
+ */
+public final class GetDeclaredFieldValueWithMethodHandle implements PrivilegedAction<MethodHandle> {
```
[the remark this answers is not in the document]
gsmet: @marko-bekhta looks like you forgot this one.

cdi/src/main/java/org/hibernate/validator/cdi/ValidationExtension.java
```diff
@@ -259,7 +262,8 @@ else if ( bean.getTypes().contains( Validator.class ) || bean instanceof Validat
 		for ( AnnotatedMethod<? super T> annotatedMethod : type.getMethods() ) {
 			Method method = annotatedMethod.getJavaMember();
 
-			boolean isGetter = ReflectionHelper.isGetterMethod( method );
+			//TODO: need to update the getter logic here:
+			boolean isGetter = getterPropertyMatcherHelper.isProperty( method );
```
gsmet: Is it normal that there's still a TODO here?

gsmet:
@marko-bekhta looks like we have some issues with Karaf. Do you reproduce the failures on your laptop? If so, taking a look at the karaf log file might help.

marko-bekhta:
@gsmet I have the same strange behavior as in the stax PR. I did `mvn clean install` before pushing the code and the build passed for all modules. But then there are these failures on CI. I made a couple more small changes after that but haven't pushed them yet as wanted to figure out what's happening with the osgi tests...

marko-bekhta:
As for the first question:
> We had this issue reported in the forum recently: https://discourse.hibernate.org/t/primefaces-chips-web-component-sizevalidatorforcharsequence/374 . I would like us to fix it at some point. I.e. be able to say that the getter type is different from the property type.  ....

With the current implementation the next test pass:
```java
@Test
public void testGetterPropertyMatcher() {
    Validator validator = Validation.byProvider( HibernateValidator.class )
            .configure()
            .getterPropertyMatcher( new BarGetterPropertyMatcher() )
            .buildValidatorFactory()
            .getValidator();
    Set<ConstraintViolation<Bar>> violations = validator.validate( new Bar() );

    ConstraintViolationAssert.assertThat( violations ).containsOnlyViolations(
            violationOf( AssertTrue.class ).withProperty( "bar" ),
            violationOf( NotBlank.class ).withProperty( "bar" ),
            violationOf( NotNull.class ).withProperty( "bar" ),
            violationOf( Min.class ).withProperty( "bar" ),

            violationOf( NotEmpty.class ).withProperty( "tags" ),
            violationOf( Size.class ).withProperty( "tags" )
    );
}

private static class Bar {

    @Min( 10 )
    private final int bar = 1;

    @Size(min = 10, max = 2147483647)
    private final String tags = "";

    @NotEmpty
    public List<String> getTags() {
        return Arrays.stream( tags.split( "," ) ).filter( tag -> !tag.trim().isEmpty() )
                .collect( Collectors.toList() );
    }

    @AssertTrue
    boolean isBar() {
        return false;
    }

    @NotBlank
    String getBar(){
        return "";
    }

    @NotNull
    Set<String> bar(){
        return null;
    }
}

public static class BarGetterPropertyMatcher implements GetterPropertyMatcher {

    @Override
    public boolean isProperty(ConstrainableExecutable executable) {
        return executable.getParameterTypes().length == 0;
    }

    @Override
    public String getPropertyName(ConstrainableExecutable method) {
        String name = method.getName();
        char[] chars = name.substring( name.startsWith( "is" ) ? 2 : name.startsWith( "get" ) ? 3 : 0 ).toCharArray();
        chars[0] = Character.toLowerCase( chars[0] );
        return new String( chars );
    }
}
```

we have a `GetterPropertyMatcher` that would take any method without parameters as _getter_ and extract the name without a prefix. 
So if I understood the problem correctly - it is already solved, right?

gsmet:
@marko-bekhta yes looks good. It would be nice if we could add a test (not as convoluted as your example, just with the String/List and the default `GetterPropertyMatcher`).

marko-bekhta:
@gsmet done :) - https://github.com/marko-bekhta/hibernate-validator/blob/9a8c7a4b68392c3132edd18f4b34f6ca4ca42a78/engine/src/test/java/org/hibernate/validator/test/properties/GetterPropertyMatcherTest.java#L100

and about the second question ... Should it be something like `javax.json.JsonObject` or just a string, or something like `Map<String, Object>` ?

gsmet:
I think we could try to experiment with the new `javax.json` and see how it goes. But yeah, it's the same with a Map. Our ultimate goal is to be able to validate free form objects.

The idea would be to see if we could validate something without adding new stuff to the HV engine proper.

marko-bekhta:
Hi @gsmet I've tried a simple example with `javax.json` and here's a commit with my experiment - https://github.com/marko-bekhta/hibernate-validator/commit/020e49c0a8b9bbb48cd27303c352da666227d13a
I didn't want to push it here to keep this PR cleaner. What I've done there is:
- Added a "mocked" version of `JsonMetadataProvider` which returns a couple of constraints for properties like name and email. 
- Added `JsonProperty` and `JsonConstrainableType` which extend from the property API that we have. Note that I've hardcoded the list of possible properties into `JsonConstrainableType` but I real case I think we wouldn't need that at all and we would just programmatically define some constraints (probably) without any validation if the property is present or not ...
- And there's this simple test to check how it would work:

```java
@Test
public void testValidatePropertyWithRedefinedDefaultGroupOnMainEntity() {
    Validator validator = getValidator();
    JsonObject jsonObject = Json.createObjectBuilder()
            .add( "firstName", "name" )
            .add( "lastName", "" )
            .add( "age", BigDecimal.valueOf( 20 ) )
            .add( "email", "not-an-email-string" )
            .build();
    Set<ConstraintViolation<JsonObject>> constraintViolations = validator.validate( jsonObject );
    assertThat( constraintViolations ).containsOnlyViolations(
            violationOf( Email.class ).withProperty( "email" ),
            violationOf( NotBlank.class ).withProperty( "lastName" )
    );
}
```
But here are some questions I didn't find the answers for right away and I'm not sure how we would like to do it:
- How a user would define constraints for Json objects ? I thought of a possibility to do it with something similar to our programmatic API. But there's the next question:
- Would it mean that we could define only one set of global constraints for json properties ?

gsmet:
Agreed to not push this here. We should probably even move this discussion to another PR.

Yes, I was thinking about having some sort of programmatic API. TBH, what would have been nice would be to be able to use the existing API but I don't think it will work.

The type could be some sort of key to define a JSON schema (e.g. that would answer your question about having only a global set of constraints, you would define different types and it would be a parameter of your `validate()` call).

What I'm not very clear about is how you would define the metadata for nested properties. You either need a way to define the path of the property or you would use some sort of cascading mechanism and you would need an API to set the type of a nested object so that it properly cascades.

Another thing is how we would define the metadata, I suppose we would have to build them for every validation call as the JSON information might be different (types for instance) and I'm not sure we would like to force the user to define the whole schema of his JSON content.

@yrodiere have you already thought about it as part of Search 6?

marko-bekhta:
@gsmet I've created another PR for json discussion. https://github.com/hibernate/hibernate-validator/pull/949

yrodiere:
> @yrodiere have you already thought about it as part of Search 6?

@gsmet If you're talking about non-standard Java beans, I did. Essentially I added an abstraction layer so that the core of the mapper implementation does not care about how properties are accessed.

* [PojoTypeModel](https://github.com/hibernate/hibernate-search-6-poc/blob/master/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/PojoTypeModel.java) allows to retrieve properties [by name](https://github.com/hibernate/hibernate-search-6-poc/blob/master/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/PojoTypeModel.java#L16) or [as a Stream](https://github.com/hibernate/hibernate-search-6-poc/blob/b0166e9fe7905629a3871cc125704cd10bcc1966/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/PojoRawTypeModel.java)
* [PojoPropertyModel](https://github.com/hibernate/hibernate-search-6-poc/blob/b0166e9fe7905629a3871cc125704cd10bcc1966/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/PojoPropertyModel.java) gives access to the property metadata and to the [property handle](https://github.com/hibernate/hibernate-search-6-poc/blob/b0166e9fe7905629a3871cc125704cd10bcc1966/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/PojoPropertyModel.java#L26)
* [PropertyHandle](https://github.com/hibernate/hibernate-search-6-poc/blob/b0166e9fe7905629a3871cc125704cd10bcc1966/mapper-pojo/src/main/java/org/hibernate/search/v6poc/entity/pojo/model/spi/PropertyHandle.java) allows to get the property value at runtime

One thing that is missing is the ability to explicitly select the access type: getter or field. For now in the Hibernate ORM mapper, I comply with the JPA Access Type, and in the JavaBeans mapper, I use getters (because that's the rule with standard JavaBeans). But we should probably offer some way of selecting the access type explicitly.

Chains of properties are not modeled as part of this SPI, though, because they can be implemented independently from how properties are accessed. I added such model as [BoundPojoModelPath](https://github.com/hibernate/hibernate-search-6-poc/pull/38/files#diff-0d2771cb8350ae9779a7b57eca39671b) and its various subtypes in my most recent PR.

If you were talking about JSON... Well, I took great care to allow the introduction of mappers that are not targeting POJOs, but target something else such as JSON or Maps.  but I suspect we won't be able to expose the same APIs/SPIs, because the requirements are different:

* Java annotations make no sense
* the access type (getter/field) makes no sense
* even random access to properties might not make sense: we might want to handle JSON as a stream of data, keeping the memory footprint minimal.

I think the best way to solve all this is to try to implement such a mapper, and see what we can re-use from the POJO mapper. We might be able to move a subset of it to a common project that both the POJO mapper and the JSON mapper would depend on.

gsmet:
@marko-bekhta could you rebase this one? I plan to make a last pass on it and discuss with Gunnar about the API names when he's back.

Thanks!

marko-bekhta:
rebased and ready.

engine/src/main/java/org/hibernate/validator/internal/properties/DefaultGetterPropertyMatcher.java
```diff
@@ -47,52 +41,38 @@
...
 
 	@Override
-	public String getPropertyName(Method method) {
+	public String getPropertyName(ConstrainableExecutable executable) {
 		String name = null;
-		String methodName = method.getName();
+		String methodName = executable.getName();
 		for ( String prefix : PROPERTY_ACCESSOR_PREFIXES ) {
 			if ( methodName.startsWith( prefix ) ) {
 				name = StringHelper.decapitalize( methodName.substring( prefix.length() ) );
```
gsmet: You should `break;` from here. Otherwise if you have an `isgethas()`, you will have some issues (agreed it's not recommended but let's be thorough).

engine/src/main/java/org/hibernate/validator/internal/cfg/context/PropertyConstraintMappingContextImpl.java
```diff
@@ -23,6 +23,7 @@
 import org.hibernate.validator.internal.metadata.raw.ConstrainedProperty;
 import org.hibernate.validator.internal.properties.Callable;
 import org.hibernate.validator.internal.properties.Property;
+import org.hibernate.validator.internal.properties.javabean.JavaBeanField;
```
gsmet: It would be better if we could squash this commit with the first one. They are definitely related.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/TypeConstraintMappingContextImpl.java
```diff
@@ -74,9 +73,12 @@
 	private List<Class<?>> defaultGroupSequence;
 	private Class<? extends DefaultGroupSequenceProvider<? super C>> defaultGroupSequenceProviderClass;
 
-	TypeConstraintMappingContextImpl(DefaultConstraintMapping mapping, Class<C> beanClass) {
+	private final GetterPropertyMatcher getterPropertyMatcher;
```
gsmet: I know I already made you change the name but I'm wondering if `GetterPropertySelectionStrategy` would be better?

gsmet:
@marko-bekhta let's close this one as it's barely readable now with all those comments, most of them related to the abstraction.

We can rebase and open a new (smaller) one.

Two comments are still valid though:
- https://github.com/hibernate/hibernate-validator/pull/938#discussion_r186500125
- https://github.com/hibernate/hibernate-validator/pull/938#discussion_r186503273

PR #939, opened 2018-03-16
 HV-1589 Correctly cache HibernateConstraintValidators

mkurz:
https://hibernate.atlassian.net/browse/HV-1589

I think it's the easiest if you review this pull request per commit. (I want to make your life as easy as possible that's why I created small commits you hopefully understand better than one big commit ;)

/cc @gsmet @marko-bekhta

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
@@ -34,9 +34,11 @@ import org.hibernate.validator.HibernateValidatorConfiguration;
 import org.hibernate.validator.HibernateValidatorContext;
 import org.hibernate.validator.HibernateValidatorFactory;
 import org.hibernate.validator.cfg.ConstraintMapping;
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorInitializationContext;
 import org.hibernate.validator.internal.cfg.context.DefaultConstraintMapping;
 import org.hibernate.validator.internal.engine.constraintdefinition.ConstraintDefinitionContribution;
 import org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorManager;
+import org.hibernate.validator.internal.engine.constraintvalidation.HibernateConstraintValidatorInitializationContextImpl;
 import org.hibernate.validator.internal.engine.groups.ValidationOrderGenerator;
 import org.hibernate.validator.internal.engine.scripting.DefaultScriptEvaluatorFactory;
 import org.hibernate.validator.internal.engine.valueextraction.ValueExtractorManager;
@@ -177,8 +179,6 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 						getAllowParallelMethodsDefineParameterConstraints( hibernateSpecificConfig, properties )
 				).build();
 
-		this.constraintValidatorManager = new ConstraintValidatorManager( configurationState.getConstraintValidatorFactory() );
-
 		this.validatorFactoryScopedContext = new ValidatorFactoryScopedContext(
 				configurationState.getMessageInterpolator(),
 				configurationState.getTraversableResolver(),
@@ -189,6 +189,11 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 				getFailFast( hibernateSpecificConfig, properties ),
 				getTraversableResolverResultCacheEnabled( hibernateSpecificConfig, properties ),
 				getConstraintValidatorPayload( hibernateSpecificConfig )
+		);
+
+		this.constraintValidatorManager = new ConstraintValidatorManager(
+				configurationState.getConstraintValidatorFactory(),
+				this.validatorFactoryScopedContext.getConstraintValidatorInitializationContext()
 		);
 
 		this.validationOrderGenerator = new ValidationOrderGenerator();
@@ -702,7 +707,15 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 		 */
 		private final boolean traversableResolverResultCacheEnabled;
 
+		/**
+		 * The constraint validator payload.
+		 */
 		private final Object constraintValidatorPayload;
+
+		/**
+		 * The constraint validator initialization context.
+		 */
+		private final HibernateConstraintValidatorInitializationContextImpl constraintValidatorInitializationContext;
 
 		private ValidatorFactoryScopedContext(MessageInterpolator messageInterpolator,
 				TraversableResolver traversableResolver,
@@ -713,6 +726,22 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 				boolean failFast,
 				boolean traversableResolverResultCacheEnabled,
 				Object constraintValidatorPayload) {
+			this( messageInterpolator, traversableResolver, parameterNameProvider, clockProvider, temporalValidationTolerance, scriptEvaluatorFactory, failFast,
+					traversableResolverResultCacheEnabled, constraintValidatorPayload,
+					new HibernateConstraintValidatorInitializationContextImpl( scriptEvaluatorFactory, clockProvider,
+							temporalValidationTolerance, constraintValidatorPayload ) );
+		}
+
+		private ValidatorFactoryScopedContext(MessageInterpolator messageInterpolator,
+				TraversableResolver traversableResolver,
+				ExecutableParameterNameProvider parameterNameProvider,
+				ClockProvider clockProvider,
+				Duration temporalValidationTolerance,
+				ScriptEvaluatorFactory scriptEvaluatorFactory,
+				boolean failFast,
+				boolean traversableResolverResultCacheEnabled,
+				Object constraintValidatorPayload,
+				HibernateConstraintValidatorInitializationContextImpl constraintValidatorInitializationContext) {
 			this.messageInterpolator = messageInterpolator;
 			this.traversableResolver = traversableResolver;
 			this.parameterNameProvider = parameterNameProvider;
@@ -722,6 +751,7 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 			this.failFast = failFast;
 			this.traversableResolverResultCacheEnabled = traversableResolverResultCacheEnabled;
 			this.constraintValidatorPayload = constraintValidatorPayload;
+			this.constraintValidatorInitializationContext = constraintValidatorInitializationContext;
 		}
 
 		public MessageInterpolator getMessageInterpolator() {
@@ -760,6 +790,10 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 			return this.constraintValidatorPayload;
 		}
 
+		public HibernateConstraintValidatorInitializationContext getConstraintValidatorInitializationContext() {
+			return this.constraintValidatorInitializationContext;
+		}
+
 		static class Builder {
 			private final ValidatorFactoryScopedContext defaultContext;
 
@@ -772,6 +806,7 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 			private boolean failFast;
 			private boolean traversableResolverResultCacheEnabled;
 			private Object constraintValidatorPayload;
+			private HibernateConstraintValidatorInitializationContextImpl constraintValidatorInitializationContext;
 
 			Builder(ValidatorFactoryScopedContext defaultContext) {
 				Contracts.assertNotNull( defaultContext, "Default context cannot be null." );
@@ -786,6 +821,7 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 				this.failFast = defaultContext.failFast;
 				this.traversableResolverResultCacheEnabled = defaultContext.traversableResolverResultCacheEnabled;
 				this.constraintValidatorPayload = defaultContext.constraintValidatorPayload;
+				this.constraintValidatorInitializationContext = defaultContext.constraintValidatorInitializationContext;
 			}
 
 			public Builder setMessageInterpolator(MessageInterpolator messageInterpolator) {
@@ -874,7 +910,14 @@ public class ValidatorFactoryImpl implements HibernateValidatorFactory {
 						scriptEvaluatorFactory,
 						failFast,
 						traversableResolverResultCacheEnabled,
-						constraintValidatorPayload
+						constraintValidatorPayload,
+						HibernateConstraintValidatorInitializationContextImpl.of(
+								constraintValidatorInitializationContext,
+								scriptEvaluatorFactory,
+								clockProvider,
+								temporalValidationTolerance,
+								constraintValidatorPayload
+						)
 				);
 			}
 		}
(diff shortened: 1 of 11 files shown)


marko-bekhta:
Hi @mkurz thanks for reporting the problem and sending this PR!
I know that I'm a bit late to this payload party :smiley:  but looking at this particular caching issue it got me thinking if it should really be in the initialization context, i.e. if the payload can change from request to request maybe a better place for it would be `ConstraintValidatorContext` ? That way there will be no caching problems caused by payload..

mkurz:
@marko-bekhta thanks for you feedback!
You mean something like this:
```java
@Override
public boolean isValid(String object, ConstraintValidatorContext constraintContext) {
    String payload = constraintContext.getConstraintValidatorPayload( String.class );

    // Be aware: there is already similiar method, but with a different purpose:
    constraintContext.withDynamicPayload( somePayloadObject );
}
```
For me that would be perfectly OK. The `ConstraintValidatorContext` probably isn't cached.

However `HibernateConstraintValidatorInitializationContext` passed to the `initialize(...)` method should still be part of the caching key for the constraint validator cache to make sure we cache per `scriptEvaluatorFactory`, `cockProvider`, etc.

gsmet:
Hum good catch @mkurz .

Yeah, so even if it's not per validation call, but per Validator, having the payload specific to a Validator makes it a bad candidate for the initialization context. I initially thought it would be a good idea to have it there as you might have some heavy objects to initialize. But considering it will be hard to cache it, maybe it doesn't make sense and you would have to initialize all the heavy objects for all the payload values.

So, I suppose I'm +1 to move the payload to the `ConstraintValidatorContext` and yes, we should fix the cache issue for the `ScriptEvaluatorFactory` and the `ClockProvider`.

Let's make it 2 different JIRA issues.

mkurz:
Great, so let's change it. I hope I will find time to work on this later today or tomorrow. Thanks!

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
```diff
@@ -327,6 +368,9 @@ private int createHashCode() {
 			int result = annotationDescriptor.hashCode();
 			result = 31 * result + validatedType.hashCode();
 			result = 31 * result + constraintValidatorFactory.hashCode();
+			result = 31 * result + scriptEvaluatorFactory.hashCode();
+			result = 31 * result + clockProvider.hashCode();
+			result = 31 * result + temporalValidationTolerance.hashCode();
```
gsmet: So the idea here would be to have `equals()` and `hashCode()` defined for the initialization context. Thus you wouldn't have to expose anything and it would be better isolated.

The hashCode in the context should be created in the constructor and cached (we have a `createHashCode()` pattern you can see in several places of the code).

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -39,6 +39,15 @@
...
 	ScriptEvaluator getScriptEvaluatorForLanguage(String languageName);
 
+	/**
+	 * Returns a {@link ScriptEvaluatorFactory} passed at bootstrap.
+	 *
+	 * @return a script evaluator factory. Never null.
+	 *
+	 * @since 6.0.9
+	 */
+	ScriptEvaluatorFactory getScriptEvaluatorFactory();
```
gsmet: This wouldn't need to be exposed if you implement `hashCode()` and `equals()` in the context.

mkurz: Yeah I understand.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
marko-bekhta: I guess we'd probably better also use `System#identityHashCode()` and `==` equality as well for `clockProvider` and `scriptEvaluatorFactory` as they can be provided by the users and we have no idea on their impl.

mkurz:
@marko-bekhta and @gsmet this PR is now updated and should be ready.
I still kept the payload for now because I will work on that in my next pull request (I think it's cleaner to not touch it here but have it in a separate PR).
Also please also check if the equals/hashCode method are ok for you.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/HibernateConstraintValidatorInitializationContextImpl.java
```diff
@@ -59,4 +62,43 @@ public Duration getTemporalValidationTolerance() {
...
+		if ( scriptEvaluatorFactory != null ? scriptEvaluatorFactory != hibernateConstraintValidatorInitializationContextImpl.scriptEvaluatorFactory : hibernateConstraintValidatorInitializationContextImpl.scriptEvaluatorFactory != null ) {
+			return false;
+		}
+		if ( clockProvider != null ? clockProvider != hibernateConstraintValidatorInitializationContextImpl.clockProvider : hibernateConstraintValidatorInitializationContextImpl.clockProvider != null ) {
+			return false;
+		}
+		if ( temporalValidationTolerance != null ? !temporalValidationTolerance.equals( hibernateConstraintValidatorInitializationContextImpl.temporalValidationTolerance ) : hibernateConstraintValidatorInitializationContextImpl.temporalValidationTolerance != null ) {
+			return false;
+		}
+		if ( constraintValidatorPayload != null ? constraintValidatorPayload != hibernateConstraintValidatorInitializationContextImpl.constraintValidatorPayload : hibernateConstraintValidatorInitializationContextImpl.constraintValidatorPayload != null ) {
```
marko-bekhta: As `scriptEvaluatorFactory` `clockProvider` and `temporalValidationTolerance` cannot be `null` we can skip the null checks for them, and also we do a `!=` check anyway.

mkurz: @marko-bekhta Updated, please check the `equals` method again (I squashed my fix in the original commit to keep a nice commit history)

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
```diff
@@ -281,12 +281,14 @@ else if ( TypeHelper.isAssignable( assignableTypes.get( i ), type ) ) {
 		private final ConstraintAnnotationDescriptor<?> annotationDescriptor;
 		private final Type validatedType;
 		private final ConstraintValidatorFactory constraintValidatorFactory;
```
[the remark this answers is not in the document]
mkurz: Waiting for @gsmet on this one.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
```diff
@@ -120,7 +121,13 @@ private ValidationException getExceptionForNullValidator(Type validatedValueType
...
 					if ( validator == null ) {
 						validator = getInitializedConstraintValidator( validationContext );
-						constraintValidatorForDefaultConstraintValidatorFactory = validator;
+
+						// HibernateConstraintValidator instances should not be cached within this constraint tree instance, only ConstraintValidatorManager
+						// should handle their caching. That's because the HibernateConstraintValidatorInitializationContext influences their caching behaviour.
+						// (HibernateConstraintValidatorInitializationContext could have changed during a ValidatorFactory lifetime - via usingContext())
+						if ( !( validator instanceof HibernateConstraintValidator ) ) {
+							constraintValidatorForDefaultConstraintValidatorFactory = validator;
+						}
```
gsmet: This I don't like. The fact is that in the end I would like to have the `HibernateConstraintValidator` `initialize()` contract moved to `ConstraintValidator` and that would defeat the purpose of this optimization.

Thus I think we should include the initialization context right now in all the caching mechanisms.

Will give it a try.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/HibernateConstraintValidatorInitializationContextImpl.java
```diff
@@ -59,4 +62,43 @@ public Duration getTemporalValidationTolerance() {
...
+
+	@Override
+	public int hashCode() {
+		return hashCode;
+	}
+
+	private int createHashCode() {
+		int result = System.identityHashCode( scriptEvaluatorFactory );
+		result = 31 * result + System.identityHashCode( clockProvider );
+		result = 31 * result + ( temporalValidationTolerance != null ? temporalValidationTolerance.hashCode() : 0 );
```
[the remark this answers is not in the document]
mkurz: Thanks for fixing!

engine/src/test/java/org/hibernate/validator/test/internal/engine/constraintvalidation/SimpleHibernateConstraintValidator.java
```diff
@@ -0,0 +1,32 @@
...
+
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.metadata.ConstraintDescriptor;
+
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidator;
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorInitializationContext;
+
+public class SimpleHibernateConstraintValidator implements HibernateConstraintValidator<SimpleHibernateConstraintValidatorConstraint, String> {
+
+	Duration duration;
```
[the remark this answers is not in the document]
mkurz: Good catch!

engine/src/test/java/org/hibernate/validator/test/internal/engine/constraintvalidation/ConstraintValidatorManagerTest.java
```diff
@@ -258,6 +268,196 @@ public void testValidatorsAreCachedPerConstraintAndAnnotationMembers() {
...
 	}
 
+	@Test
+	@TestForIssue(jiraKey = "HV-1589")
+	public void testValidatorsAreCachedPerConstraintAndAnnotationMembersAndScriptEvaluatorFactory() {
+		Validator validator = getConfiguration()
+				.addMapping(
+						ConstraintValidatorManagerTest.class.getResourceAsStream(
+								"hv-1589-mapping.xml"
+						)
```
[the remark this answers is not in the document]
mkurz: I more or less copied the test directly above (`testValidatorsAreCachedPerConstraintAndAnnotationMembers`) and simply modified it for my needs. This test also had a XML mapping, so I kept it. (The intention probably was to test mixed constraints coming from annotation and from XML).

engine/src/test/java/org/hibernate/validator/test/internal/engine/constraintvalidation/ConstraintValidatorManagerTest.java
```diff
@@ -258,6 +268,196 @@ public void testValidatorsAreCachedPerConstraintAndAnnotationMembers() {
...
+					.temporalValidationTolerance( Duration.ofHours( 7 ) )
+					.getValidator().validate( new Company() ).iterator().next().getMessage() ).isEqualTo( "PT7H" );
+
+			assertThat( factory
+					.unwrap( HibernateValidatorFactory.class ).usingContext()
+					.temporalValidationTolerance( Duration.ofDays( 5 ) )
+					.getValidator().validate( new Company() ).iterator().next().getMessage() ).isEqualTo( "PT120H" );
+		}
+	}
+
```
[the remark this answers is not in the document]
mkurz: You are welcome :wink:

engine/src/test/java/org/hibernate/validator/test/internal/engine/constraintvalidation/ConstraintValidatorManagerTest.java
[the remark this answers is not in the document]
gsmet: OK. I'll wait for Marko to review my changes before merging. Thanks again for your nice work.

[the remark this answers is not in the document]
mkurz: Great, after this got merged I already have the next pull request waiting in the line for [HV-1595](https://hibernate.atlassian.net/browse/HV-1595).

engine/src/test/java/org/hibernate/validator/testutils/ConstraintValidatorInitializationHelper.java
```diff
@@ -87,4 +87,32 @@ private ConstraintValidatorInitializationHelper() {
 	public static HibernateConstraintValidatorInitializationContext getDummyConstraintValidatorInitializationContext() {
 		return DUMMY_CONSTRAINT_VALIDATOR_INITIALIZATION_CONTEXT;
 	}
+
+	public static HibernateConstraintValidatorInitializationContext getConstraintValidatorInitializationContext(
```
marko-bekhta: With this method added we could initialize the `DUMMY_CONSTRAINT_VALIDATOR_INITIALIZATION_CONTEXT` above with it. Might look cleaner with just:
```java
private static final HibernateConstraintValidatorInitializationContext DUMMY_CONSTRAINT_VALIDATOR_INITIALIZATION_CONTEXT
			= getConstraintValidatorInitializationContext( new DefaultScriptEvaluatorFactory( null ), DefaultClockProvider.INSTANCE, Duration.ZERO, null );
```

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -702,8 +707,16 @@ public String toString() {
...
 
+		/**
+		 * The constraint validator payload.
+		 */
 		private final Object constraintValidatorPayload;
 
+		/**
+		 * The constraint validator initialization context.
+		 */
+		private final HibernateConstraintValidatorInitializationContextImpl constraintValidatorInitializationContext;
```
marko-bekhta: So with this `HibernateConstraintValidatorInitializationContext` added here I suppose we could use it in `ValidatorImpl` [here](https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java#L148). instead of creating one more context we could just
```java
this.constraintValidatorInitializationContext = validatorFactoryScopedContext.getConstraintValidatorInitializationContext();
```

engine/src/test/java/org/hibernate/validator/testutils/ConstraintValidatorInitializationHelper.java
gsmet: Yup, good catch, done.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
gsmet: Yeah, that was the whole point of my commit in fact :). I missed a step.

gsmet:
Merged! Thanks @mkurz  @marko-bekhta .

@mkurz looking forward to your next PR :)

mkurz:
@gsmet Thanks for merging! Stay tuned, should be up within the next one to two hours right after lunch :smile:

PR #941, opened 2018-03-20
HV-1592 Validate that the validator is defined with the right constraint

gsmet:
* https://hibernate.atlassian.net/browse/HV-1592

**This one can be reviewed but we need to wait for the 2.0.3.Final TCK to be released before merging it.**

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ClassBasedValidatorDescriptor.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ClassBasedValidatorDescriptor.java
@@ -25,28 +25,38 @@ import org.hibernate.validator.internal.util.logging.LoggerFactory;
  * Represents an implementation of {@link ConstraintValidator}.
  *
  * @author Gunnar Morling
+ * @author Guillaume Smet
  */
 class ClassBasedValidatorDescriptor<A extends Annotation> implements ConstraintValidatorDescriptor<A> {
 
-	private static final long serialVersionUID = -8207687559460098548L;
 	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
 	private final Class<? extends ConstraintValidator<A, ?>> validatorClass;
 	private final Type validatedType;
 	private final EnumSet<ValidationTarget> validationTargets;
 
-	public ClassBasedValidatorDescriptor(Class<? extends ConstraintValidator<A, ?>> validatorClass) {
+	private ClassBasedValidatorDescriptor(Class<? extends ConstraintValidator<A, ?>> validatorClass) {
 		this.validatorClass = validatorClass;
-		this.validatedType = TypeHelper.extractType( validatorClass );
+		this.validatedType = TypeHelper.extractValidatedType( validatorClass );
 		this.validationTargets = determineValidationTargets( validatorClass );
+	}
+
+	public static <T extends Annotation> ClassBasedValidatorDescriptor<T> of(Class<? extends ConstraintValidator<T, ?>> validatorClass,
+			Class<? extends Annotation> registeredConstraintAnnotationType) {
+		Type definedConstraintAnnotationType = TypeHelper.extractConstraintType( validatorClass );
+		if ( !registeredConstraintAnnotationType.equals( definedConstraintAnnotationType ) ) {
+			throw LOG.getConstraintValidatorDefinitionConstraintMismatchException( validatorClass, registeredConstraintAnnotationType,
+					definedConstraintAnnotationType );
+		}
+
+		return new ClassBasedValidatorDescriptor<T>( validatorClass );
 	}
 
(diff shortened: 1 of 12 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ClassBasedValidatorDescriptor.java
```diff
@@ -25,28 +25,38 @@
  * Represents an implementation of {@link ConstraintValidator}.
  *
  * @author Gunnar Morling
+ * @author Guillaume Smet
  */
 class ClassBasedValidatorDescriptor<A extends Annotation> implements ConstraintValidatorDescriptor<A> {
 
-	private static final long serialVersionUID = -8207687559460098548L;
-	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
+	private static final Log log = LoggerFactory.make( MethodHandles.lookup() );
```
gunnarmorling: Why the change in casing, it's static after all?

engine/src/main/java/org/hibernate/validator/internal/util/TypeHelper.java
```diff
@@ -324,24 +326,32 @@ private static Type getResolvedSuperclass(Type type) {
 		return validatorsTypes;
 	}
 
-	public static Type extractType(Class<? extends ConstraintValidator<?, ?>> validator) {
-		Map<Type, Type> resolvedTypes = newHashMap();
+	public static Type extractConstraintValidatorValidatedType(Class<? extends ConstraintValidator<?, ?>> validator) {
```
gunnarmorling: How about `extractValidatedType()` and `extractConstraintType()`? The context is clear from the parameter type and I had to read twice to see where the two method names are different.

engine/src/main/java/org/hibernate/validator/internal/util/TypeHelper.java
```diff
@@ -324,24 +326,32 @@ private static Type getResolvedSuperclass(Type type) {
...
+	}
+
+	public static Type extractConstraintValidatorTypeArgumentType(Class<? extends ConstraintValidator<?, ?>> validator, int typeArgumentIndex) {
+		Map<Type, Type> resolvedTypes = new HashMap<>();
 		Type constraintValidatorType = resolveTypes( resolvedTypes, validator );
 
 		//we now have all bind from a type to its resolution at one level
-		Type validatorType = ( (ParameterizedType) constraintValidatorType ).getActualTypeArguments()[VALIDATOR_TYPE_INDEX];
-		if ( validatorType == null ) {
+		Type constraintType = ( (ParameterizedType) constraintValidatorType ).getActualTypeArguments()[typeArgumentIndex];
```
gunnarmorling: Shouldn't it be just `type` here? Or perhaps `extractedType`. After all, it can be the constraint type _or_ the validator type.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ClassBasedValidatorDescriptor.java
gsmet: Raaah, I'm in Search mode when they do it the opposite way and are fixing it the opposite way. Will fix.

engine/src/main/java/org/hibernate/validator/internal/util/TypeHelper.java
gsmet: Yup, it's coming from a late minute refactoring, will fix.

engine/src/main/java/org/hibernate/validator/internal/util/TypeHelper.java
gsmet: Frankly, I don't like having such specific methods in the very generic `TypeHelper` (I opened a new issue for that), that's why I was over cautious. Can shorten the name though.

gsmet:
@gunnarmorling force-pushed fixes for your comments.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -841,4 +841,10 @@ ValidationException getUnableToAccessMethodException(Lookup lookup, @FormatWith(
 	@LogMessage(level = DEBUG)
 	@Message(id = 240, value = "Constraint validator payload set to %1$s.")
 	void logConstraintValidatorPayload(Object payload);
+
+	@Message(id = 241, value = "The constraint validator %1$s is registered for the constraint %2$s but is defined with the constraint %3$s.")
```
gunnarmorling: "registered" and "defined" may not be intuitive for everyone. Perhaps:

> Constraint @Foo references constraint validator type Bar, but this validator is defined for constraint type Qux

?

gsmet: Right, done.

gsmet:
And merged now that we have fixed the CDI issue.

PR #942, opened 2018-03-22
HV-1595 Move ConstraintValidatorPayload to ConstraintValidatorContext

mkurz:
https://hibernate.atlassian.net/browse/HV-1595

@gsmet @marko-bekhta Like promised. That one should be easy.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
@@ -8,6 +8,9 @@ package org.hibernate.validator.constraintvalidation;
 
 import javax.validation.ConstraintValidatorContext;
 
+import org.hibernate.validator.HibernateValidatorConfiguration;
+import org.hibernate.validator.Incubating;
+
 /**
  * A custom {@link ConstraintValidatorContext} which allows to set additional message parameters for
  * interpolation.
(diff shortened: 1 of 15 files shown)


engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorContext.java
```diff
@@ -107,4 +109,17 @@
...
+	 * the given type.
+	 *
+	 * @param type the type of payload to retrieve
+	 * @return an instance of the specified type or {@code null} if the current constraint validator payload isn't of
+	 * the given type
+	 *
+	 * @since 6.0.9
+	 * @see HibernateValidatorConfiguration#constraintValidatorPayload(Object)
+	 */
+	<C> C getConstraintValidatorPayload(Class<C> type);
```
gsmet: We need the `@Incubating` annotation here. In case we want to move it elsewhere in the next version ;).

mkurz: @gsmet Added!

gsmet:
Merged, thanks @mkurz , nice work.

mkurz:
Great, thanks! So... when will you release `6.0.9`? :stuck_out_tongue_winking_eye:

gsmet:
There's this issue that is bugging me: https://hibernate.atlassian.net/browse/HV-1596 . But for now I was unable to reproduce it with the information in the JIRA so waiting for the OP to provide more info.

Could you check with the current snapshot that you can get your Play integration to work? It would be nice to be sure.

Other than these 2 things, I think we already have sufficient improvements for a 6.0.9.Final.

mkurz:
I **already have** a working Play integration sitting on my machine waiting for `6.0.9` :smile: And it's working like a charm! (There are even tests on Play side as well to make sure that always a new payload gets passed for each request - that's how I ran into the original caching bug). I will reference you in the Play pull request as soon `6.0.9` is out.

gsmet:
OK, nice, so this just lets this annoying bug report. Hope we will know more soon.

PR #943, opened 2018-03-22
HV-1597 Allow overriding the constraint validator payload with null

gsmet:
* https://hibernate.atlassian.net/browse/HV-1597


gsmet:
@mkurz @marko-bekhta thoughts about this one? It feels natural to me.

marko-bekhta:
it does make sense to "reset" the payload if the previous is not needed anymore. And also maybe we could have:
```java
<C> Optional<C> getConstraintValidatorPayload(Class<C> type);
```
the payload getter returning the optional ? it would look more "modern"  :) but it's probably more for the https://github.com/hibernate/hibernate-validator/pull/942

gsmet:
Yeah, so it's not about giving the ability to reset it but about having the ability to override it with null for a given Validator. I think it makes sense.

About the `Optional`, good point. The thing is that it would not be consistent with the other payload-type things we already have so I'm a bit torn...

PR #944, opened 2018-03-22
HV-1598 Fix the behavior of XML default-validated-executable-types

gsmet:
* https://hibernate.atlassian.net/browse/HV-1598

**This one can be reviewed but we need to wait for the 2.0.3.Final TCK to be released before merging it.**


gsmet:
And merged now that we have fixed the CDI issue.

PR #945, opened 2018-03-22
HV-1596 Fix TypeVariableBindings when dealing with parent classes

gsmet:
* https://hibernate.atlassian.net/browse/HV-1596

So this one is a pretty bad one. It has been there since the very first implementation of value extractors and has gone unnoticed since.

I plan to release 6.0.9.Final next week, probably on Tuesday.


gunnarmorling:
LGTM, merging. Thanks, @gsmet!

PR #947, opened 2018-03-30
HV-1577 Use Stax instead of JAXB to parse the XML descriptors

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1577

All parsing logic is now based on `AbstractStaxBuilder`; each builder later adds reading and building logic depending on what elements it reads. In the last commits I've removed the JAXB related code and schema normalization as with Stax that seems to be redundant. I'll probably go through all those builders once more and maybe clean them up a bit (make some fields private, add additional comments and javadocs).


engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -468,14 +469,14 @@ ConstraintDeclarationException getMultipleGroupConversionsForSameSourceException
 
 	@Message(id = 133, value = "%1$s does not contain a constructor with the parameter types %2$s.")
 	ValidationException getBeanDoesNotContainConstructorException(@FormatWith(ClassObjectFormatter.class) Class<?> beanClass,
-			@FormatWith(CollectionOfClassesObjectFormatter.class) List<Class<?>> parameterTypes);
+			@FormatWith(ArrayOfClassesObjectFormatter.class) Class<?>[] parameterTypes);
```
gsmet: I don't see any value in this change. I saw at least once occurrence where you now convert the list to an array.

A list is fine, let's get rid of the diff noise.

engine/src/main/java/org/hibernate/validator/internal/xml/ConstrainedExecutableBuilder.java
```diff
@@ -102,7 +102,7 @@
 				throw LOG.getBeanDoesNotContainMethodException(
 						beanClass,
 						methodName,
-						parameterTypes
+						parameterTypes.toArray( new Class[parameterTypes.size()] )
```
gsmet: See here and just below :).

engine/src/main/java/org/hibernate/validator/internal/xml/ClassLoadingHelper.java
```diff
@@ -21,7 +21,7 @@
  *
  * @author Gunnar Morling
  */
-/*package*/ class ClassLoadingHelper {
+public class ClassLoadingHelper {
```
gsmet: I haven't my IDE handy but can't you move it to the package now using it? Is it now used by multiple packages?

engine/src/main/java/org/hibernate/validator/internal/xml/stax/mapping/ContainerElementTypeStaxBuilder.java
```diff
@@ -36,12 +36,11 @@
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
 import org.hibernate.validator.internal.xml.ClassLoadingHelper;
 import org.hibernate.validator.internal.xml.ContainerElementTypePath;
-import org.hibernate.validator.internal.xml.binding.ContainerElementTypeType;
 import org.hibernate.validator.internal.xml.stax.AbstractStaxBuilder;
 import org.hibernate.validator.internal.xml.stax.mapping.ContainerElementTypeConfigurationStaxBuilder.ContainerElementTypeConfiguration;
 
 /**
- * Builds the cascading and type argument constraints configuration from the {@link ContainerElementTypeType} elements.
+ * Builds the cascading and type argument constraints configuration from the {@code <container-element-type>} elements.
```
gsmet: Minor nitpick but just above in this commit, you replaced it with a link to `ContainerElementTypeStaxBuilder`, let's be consistent.

engine/pom.xml
```diff
@@ -273,39 +273,6 @@
         </plugins>
     </build>
     <profiles>
-        <profile>
```
gsmet: We can also clean up the parent pom as we used to require the `java.xml.bind` module (just search for it in the parent pom and you'll see the start of the clean up to do).

I don't think it's worth keeping the existing add-modules machinery as I don't think we will need it anymore.

We can also clean up the `-J--add-modules=java.xml.ws.annotation` compilation parameter while at it. We won't have any `@Generated` annotation anymore.

engine/src/main/java/org/hibernate/validator/internal/xml/MappingXmlParser.java
```diff
@@ -109,105 +90,44 @@ public MappingXmlParser(ConstraintHelper constraintHelper, TypeResolutionHelper
...
-		catch (JAXBException | SAXException | IOException | XMLStreamException e) {
-			throw LOG.getErrorParsingMappingFileException( e );
-		}
-	}
 
-	private ConstraintMappingsType unmarshal(JAXBContext jc, InputStream in) throws JAXBException, XMLStreamException, IOException, SAXException {
-		ClassLoader previousTccl = run( GetClassLoader.fromContext() );
+				// The validation is done first as we manipulate the XML document before pushing it to the unmarshaller
+				// and it might not be valid anymore as we might have switched the namespace to the latest namespace
+				// supported.
```
gsmet: This comment should be updated (or removed?) in the commit removing the namespace normalization as it's not accurate anymore.

engine/src/main/java/org/hibernate/validator/internal/xml/stax/AbstractStaxBuilder.java
```diff
@@ -64,7 +64,13 @@ public boolean process(XMLEventReader xmlEventReader, XMLEvent xmlEvent) {
...
 	protected String readSingleElement(XMLEventReader xmlEventReader) throws XMLStreamException {
 		// trimming the string value as it might contain leading/trailing spaces or \n
-		return xmlEventReader.nextEvent().asCharacters().getData().trim();
+		XMLEvent xmlEvent = xmlEventReader.nextEvent();
+		StringBuilder stringBuilder = new StringBuilder( xmlEvent.asCharacters().getData() );
+		while ( xmlEventReader.peek().isCharacters() ) {
+			xmlEvent = xmlEventReader.nextEvent();
+			stringBuilder.append( xmlEvent.asCharacters().getData() );
+		}
+		return stringBuilder.toString().trim();
```
gsmet: Could we maybe move this change to the first commit? It looks like it should have been there from the beginning.

engine/src/main/java/org/hibernate/validator/internal/xml/ValidationXmlParser.java
```diff
@@ -103,6 +103,7 @@ public final BootstrapConfiguration parseValidationXml() {
 			ValidationConfigStaxBuilder validationConfigStaxBuilder = new ValidationConfigStaxBuilder( xmlEventReader );
 
 			xmlEventReader.close();
+			in.reset();
```
gsmet: Let's move this change to the first commit too. Usually, I just apply the change there when editing the commit in the rebase and usually git ignores the change afterwards when continuing the rebase.

engine/src/main/java/org/hibernate/validator/internal/xml/stax/mapping/ContainerElementTypeConfigurationStaxBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+import org.hibernate.validator.internal.xml.ContainerElementTypePath;
+import org.hibernate.validator.internal.xml.binding.ContainerElementTypeType;
+
+/**
+ * Builds the cascading and type argument constraints configuration from the {@link ContainerElementTypeType} elements.
+ *
+ * @author Guillaume Smet
+ * @author Marko Bekhta
+ */
+class ContainerElementTypeConfigurationStaxBuilder {
```
gsmet: I was surprised it did not extend `AbstractStaxBuilder`? I suppose it's not needed?

engine/src/main/java/org/hibernate/validator/internal/xml/stax/mapping/MultiElementStaxBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+
+import javax.xml.stream.XMLEventReader;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.events.XMLEvent;
+
+import org.hibernate.validator.internal.xml.ClassLoadingHelper;
+import org.hibernate.validator.internal.xml.stax.AbstractStaxBuilder;
+
+/**
+ * An abstract builder for an elements that could have multiple {@code <value> ... </value>} entries.
```
gsmet: s/elements/element/

engine/src/main/java/org/hibernate/validator/internal/xml/stax/mapping/MultiElementStaxBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @author Marko Bekhta
+ */
+abstract class MultiElementStaxBuilder extends AbstractStaxBuilder {
```
gsmet: Let's name it `AbstractMultiValuedElementStaxBuilder`.

engine/src/main/java/org/hibernate/validator/internal/xml/stax/mapping/ValidStaxBuilder.java
```diff
@@ -4,26 +4,32 @@
...
-import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.events.XMLEvent;
 
+import org.hibernate.validator.internal.xml.stax.AbstractStaxBuilder;
+
 /**
  * @author Marko Bekhta
  */
-public class ConstraintMappingsStaxBuilder extends AbstractStaxBuilder {
+class ValidStaxBuilder extends AbstractStaxBuilder {
```
gsmet: Yeah, see, better remove it from the first commit as git goes berserk here :).

gsmet:
Some WildFly tests are failing. I wonder if it's due to the removal of this line: https://github.com/hibernate/hibernate-validator/pull/947/commits/3ab2b655fe7874fcb3ceb84b27388437afa1bccf#diff-67a3634393a73903fe765a7ee0d14e3aL181 .

engine/src/main/java/org/hibernate/validator/internal/xml/MappingXmlParser.java
```diff
@@ -109,105 +90,44 @@ public MappingXmlParser(ConstraintHelper constraintHelper, TypeResolutionHelper
...
+				String schemaResourceName = getSchemaResourceName( schemaVersion );
+				Schema schema = xmlParserHelper.getSchema( schemaResourceName );
+				Validator validator = schema.newValidator();
+				validator.validate( new StreamSource( new CloseIgnoringInputStream( in ) ) );
```
gsmet: I'm wondering if we shouldn't validate the namespace somehow. It was done before by JAXB and now we don't check anything at all meaning that the user could have a random XML file there and we wouldn't throw any error, it would just be ignored.

Am I right?

engine/src/main/java/org/hibernate/validator/internal/xml/stax/mapping/ContainerElementTypeConfigurationStaxBuilder.java
marko-bekhta: in this case the builder is just a container for the list `List<ContainerElementTypeStaxBuilder> containerElementTypeStaxBuilders` so maybe remove the `Stax` part from the name ?

engine/src/main/java/org/hibernate/validator/internal/xml/MappingXmlParser.java
gunnarmorling: Wouldn't this be caught as part of the schema validation?

marko-bekhta:
@gsmet thanks for the review :) I'll start to work on the changes. As for the failing tests it's strange I tried to run full builds locally and all the tests pass. If I understand it right, the part with changing classloaders was only needed for unmarshaling and it was reseted back for building the metadata. I actually had that code at first but with it was causing a lot of `ClassNotFoundException`s so I removed it completely.

gunnarmorling:
Wow, that's fantastic, @marko-bekhta. Thank you so much!

engine/src/main/java/org/hibernate/validator/internal/xml/MappingXmlParser.java
marko-bekhta: yes, I would think that `Validator` should handle such case. I'll try that and add a test with some random xml.

engine/src/main/java/org/hibernate/validator/internal/xml/ClassLoadingHelper.java
marko-bekhta: yeah ... wasn't really sure what to do about it. We create this helper in `org.hibernate.validator.internal.xml.MappingXmlParser` and this single instance is passed to almost all builders from `org.hibernate.validator.internal.xml.stax.mapping` where it is actually used.

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
gsmet: OK, sounds good then.

engine/src/main/java/org/hibernate/validator/internal/xml/ClassLoadingHelper.java
gsmet: I think I would have:
 * `org.hibernate.validator.internal.xml.mapping` without `stax` for all the mapping classes, including the `MappingXmlParser`
 * the same for `config`

That would solve the problem, wouldn't it?

engine/src/main/java/org/hibernate/validator/internal/xml/stax/mapping/ContainerElementTypeConfigurationStaxBuilder.java
gsmet: Yes, let's remove the `Stax` part.

engine/src/main/java/org/hibernate/validator/internal/xml/MappingXmlParser.java
gsmet: Yeah, sorry, don't know what I was thinking :).

gsmet: @marko-bekhta no need for an additional test here by the way.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/LoadClass.java
```diff
@@ -80,7 +91,9 @@ private LoadClass(String className, ClassLoader classLoader, boolean fallbackOnT
 			exception = e;
 		}
 		if ( fallbackOnTCCL ) {
-			ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+			ClassLoader contextClassLoader = initialThreadContextClassLoader != null
```
marko-bekhta: this seems to repeat once more, maybe we can init the `contextClassLoader` in the constructor ?

marko-bekhta:
@gsmet I found that `LocalNamespace` was also not used anymore (previously used for schema normalization) so I've removed it as well in the corresponding commit and force-pushed the changes.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/LoadClass.java
```diff
@@ -37,22 +38,32 @@
...
 	private final boolean fallbackOnTCCL;
 
 	public static LoadClass action(String className, ClassLoader classLoader) {
-		return new LoadClass( className, classLoader, true );
+		return action( className, classLoader, true );
 	}
 
 	public static LoadClass action(String className, ClassLoader classLoader, boolean fallbackOnTCCL) {
-		return new LoadClass( className, classLoader, fallbackOnTCCL );
+		return new LoadClass( className, classLoader, null, fallbackOnTCCL );
```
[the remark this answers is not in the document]
gsmet: Which `action()`? I don't think I have a matching one :). The other one does not have the `fallbackOnTCCL` parameter.

gsmet:
@marko-bekhta I replied to your comments. Feel free to shout at me if you disagree :).

I think this is good to go. Do you agree or do you still have something else in mind?

If not, I'll do a last review, get master ready for 6.1 and merge it.

marko-bekhta:
@gsmet heh 😄no shouting required 😄all is good
I think the last thing left was to reorganize the packages (drop the `stax` one and place `ValidationXmlParser` and `MappingXmlParser` into `config` and `mapping` ones. If we do that we would need to make `CloseIgnoringInputStream` public. But this seems better than exposing the `ClassLoadingHelper`. So that's what I've done in one of the commits that I've just added. Another one was to clean up the TODO that I've left. I think that these two commits could be squashed into the ones that introduced the corresponding classes but I though that it would be easier to see the changes (what I've moved and where) if you'll do another review.

gsmet:
Fixed a few more details, reorganized the commits, changed the commit comments, rebased and merged.

Nice work @marko-bekhta . Really nice to have this in.

PR #950, opened 2018-04-26
HV-1526 Create separate validation context for different validation kinds

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1526

Hi @gsmet, this is part of a preparation work for property holder validation. I tried to do smaller commits so it's easier to see what's happening. Mainly there are two bigger things and a couple small ones done so far.
- added multiple validation contexts for different validation operations. 
- extracted the interface from previous `ValidationContext` and split responsibilities into two interfaces. The idea behind this part is that context used in `ConstraintTree` is pretty generic to validation operations and we can and would want to use the same interface for future property holder contexts. 

I haven't started exploration of possible replacement/improvement of parameter validation context yet as it seems it would need more work and I want to get property holder validation in better shape first.


engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/ValidationContext.java
```diff
@@ -35,43 +30,34 @@
  * @author Gunnar Morling
  * @author Guillaume Smet
  */
-public interface ValidationContext<T> {
+public interface ValidationContext<T> extends ConstraintTreeValidationContext {
```
gsmet: So about the names...

I think `ConstraintTreeValidationContext` should be `ValidationContext` and that this one should have a different name as it's a specialized one. It's related to bean validation but the issue is that we already have a `BeanValidationContext` and I can't think of a better name for the other one.

Maybe we could have:
```
ValidationContext
   BaseBeanValidationContext
       BeanValidationContext
       ...
```

gsmet:
BTW, good call to push these commits separately.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/AbstractConstraintValidatorContext.java
```diff
@@ -198,19 +258,39 @@ public ConstraintValidatorContext addConstraintViolation() {
...
+		private CrossParameterConstraintViolationBuilderImpl(List<String> methodParameterNames,  String template, PathImpl path) {
 			super( template, path );
 			this.methodParameterNames = methodParameterNames;
 		}
 
+		@Override
+		public NodeBuilderDefinedContext addParameterNode(int index) {
+			if ( propertyPath.getLeafNode().getKind() != ElementKind.CROSS_PARAMETER ) {
+				throw LOG.getParameterNodeAddedForNonCrossParameterConstraintException( propertyPath );
+			}
```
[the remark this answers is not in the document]
marko-bekhta: I wasn't sure about all the logic in these builders, hence I just left it as is. I just pushed the commit after this one with the changes that I initially wanted to apply but wasn't sure about. If they are ok I'll just squash these two.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateCrossParameterConstraintValidatorContext.java
```diff
@@ -0,0 +1,28 @@
...
+import org.hibernate.validator.Incubating;
+
+/**
+ * A custom {@link ConstraintValidatorContext} which provides aditional functionality for parameter validation contexts.
+ *
+ * @author Marko Bekhta
+ * @since 6.1.0
+ */
+@Incubating
+public interface HibernateCrossParameterConstraintValidatorContext extends HibernateConstraintValidatorContext {
```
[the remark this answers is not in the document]
marko-bekhta: I agree that this change generated a lot of changes. Why I wanted to apply it is because I didn't quite like the part where we pass the parameter names all the time and sometimes they are `null` (in all cases except the parameter validation). and I just wanted to get rid of that. We could add another constructor without the parameter names argument and use it in other places, but still we will be doing the checks in this builder. 
As a side effect of this - I wasn't happy how we handled things in `ParameterScriptAssertValidator` where we first check if we can unwrap to `HibernateConstraintValidatorContext` but then just do the cast to `ConstraintValidatorContextImpl`, and that's why I've added this additional interface. 
With all that said :) I definitely would like to clean things up in `ParameterScriptAssertValidator` and as for the rest I'll trust your decision.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/AbstractConstraintValidatorContext.java
```diff
@@ -198,19 +258,39 @@ public ConstraintValidatorContext addConstraintViolation() {
...
 		}
 	}
 
-	private class ConstraintViolationBuilderImpl extends NodeBuilderBase implements ConstraintViolationBuilder {
+	private class CrossParameterConstraintViolationBuilderImpl extends ConstraintViolationBuilderImpl {
 
 		private final List<String> methodParameterNames;
 
-		private ConstraintViolationBuilderImpl(List<String> methodParameterNames, String template, PathImpl path) {
+		private CrossParameterConstraintViolationBuilderImpl(List<String> methodParameterNames,  String template, PathImpl path) {
```
gsmet: Looks like there are 2 spaces before `String template` here.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateCrossParameterConstraintValidatorContext.java
```diff
@@ -0,0 +1,28 @@
...
+package org.hibernate.validator.constraintvalidation;
+
+import java.util.List;
+
+import javax.validation.ConstraintValidatorContext;
+
+import org.hibernate.validator.Incubating;
+
+/**
+ * A custom {@link ConstraintValidatorContext} which provides aditional functionality for parameter validation contexts.
```
gsmet: s/aditional/additional/
s/parameter/cross parameter/

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/AbstractConstraintValidatorContext.java
```diff
@@ -171,8 +169,70 @@ private ConstraintViolationCreationContext getDefaultConstraintViolationCreation
 		);
 	}
 
-	public List<String> getMethodParameterNames() {
-		return methodParameterNames;
+	private static class SimpleConstraintValidatorContext extends AbstractConstraintValidatorContext {
```
gsmet: Let's move this class in a separate file.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/AbstractConstraintValidatorContext.java
```diff
@@ -171,8 +169,70 @@ private ConstraintViolationCreationContext getDefaultConstraintViolationCreation
...
+		public final ConstraintViolationBuilder buildConstraintViolationWithTemplate(String messageTemplate) {
+			return new ConstraintViolationBuilderImpl(
+					messageTemplate,
+					getCopyOfBasePath()
+			);
+		}
+	}
+
+
+	private static class CrossParameterConstraintValidatorContext extends AbstractConstraintValidatorContext implements
```
gsmet: Same here.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/AbstractConstraintValidatorContext.java
```diff
@@ -273,16 +273,16 @@ private CrossParameterConstraintViolationBuilderImpl(List<String> methodParamete
 
 		@Override
 		public NodeBuilderDefinedContext addParameterNode(int index) {
-			if ( propertyPath.getLeafNode().getKind() != ElementKind.CROSS_PARAMETER ) {
-				throw LOG.getParameterNodeAddedForNonCrossParameterConstraintException( propertyPath );
-			}
```
gsmet: As mentioned in my earlier comment, I think I would add a `Contracts.assertTrue()` condition to the constructor.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -357,8 +357,8 @@ private ValidationOrder determineGroupValidationOrder(Class<?>[] groups) {
 	 *
 	 * @return Set of constraint violations or the empty set if there were no violations.
 	 */
-	private <T, U> Set<ConstraintViolation<T>> validateInContext(ValidationContext<T> validationContext, ValueContext<U, Object> valueContext,
-			ValidationOrder validationOrder) {
+	private <T, U> Set<ConstraintViolation<T>> validateInContext(BaseBeanValidationContext<T> validationContext, ValueContext<U, Object> valueContext,
+																 ValidationOrder validationOrder) {
```
gsmet: Let's avoid this formatting, it's really painful to read in diffs.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -474,8 +474,8 @@ private void validateConstraintsForCurrentGroup(ValidationContext<?> validationC
 		}
 	}
 
-	private <U> boolean validateConstraintsForSingleDefaultGroupElement(ValidationContext<?> validationContext, ValueContext<U, Object> valueContext, final Map<Class<?>, Class<?>> validatedInterfaces,
-			Class<? super U> clazz, Set<MetaConstraint<?>> metaConstraints, Group defaultSequenceMember) {
+	private <U> boolean validateConstraintsForSingleDefaultGroupElement(BaseBeanValidationContext<?> validationContext, ValueContext<U, Object> valueContext, final Map<Class<?>, Class<?>> validatedInterfaces,
+																		Class<? super U> clazz, Set<MetaConstraint<?>> metaConstraints, Group defaultSequenceMember) {
```
gsmet: Same remark about the formatting.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -502,13 +502,13 @@ private void validateConstraintsForCurrentGroup(ValidationContext<?> validationC
...
-	private void validateConstraintsForNonDefaultGroup(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext) {
+	private void validateConstraintsForNonDefaultGroup(BaseBeanValidationContext<?> validationContext, ValueContext<?, Object> valueContext) {
 		validateMetaConstraints( validationContext, valueContext, valueContext.getCurrentBean(), valueContext.getCurrentBeanMetaData().getMetaConstraints() );
 		validationContext.markCurrentBeanAsProcessed( valueContext );
 	}
 
-	private void validateMetaConstraints(ValidationContext<?> validationContext, ValueContext<?, Object> valueContext, Object parent,
-			Iterable<MetaConstraint<?>> constraints) {
+	private void validateMetaConstraints(BaseBeanValidationContext<?> validationContext, ValueContext<?, Object> valueContext, Object parent,
+										 Iterable<MetaConstraint<?>> constraints) {
```
gsmet: Ditto.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -585,8 +585,8 @@ private void validateCascadedConstraints(ValidationContext<?> validationContext,
 		}
 	}
 
-	private void validateCascadedAnnotatedObjectForCurrentGroup(Object value, ValidationContext<?> validationContext, ValueContext<?, Object> valueContext,
-			CascadingMetaData cascadingMetaData) {
+	private void validateCascadedAnnotatedObjectForCurrentGroup(Object value, BaseBeanValidationContext<?> validationContext, ValueContext<?, Object> valueContext,
+																CascadingMetaData cascadingMetaData) {
```
gsmet: Same here. Looks like there is something to fix in your IDE config.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintTree.java
```diff
@@ -115,12 +115,12 @@ private ValidationException getExceptionForNullValidator(Type validatedValueType
...
 	}
 
-	protected final <T> ConstraintValidator<A, ?> getInitializedConstraintValidator(ValidationContext<T> validationContext, ValueContext<?, ?> valueContext) {
+	protected final ConstraintValidator<A, ?> getInitializedConstraintValidator(ValidationContext<?> validationContext, ValueContext<?, ?> valueContext) {
 		ConstraintValidator<A, ?> validator;
 
 		if ( validationContext.getConstraintValidatorFactory() == validationContext.getConstraintValidatorManager().getDefaultConstraintValidatorFactory()
 				&& validationContext.getConstraintValidatorInitializationContext() == validationContext.getConstraintValidatorManager()
-						.getDefaultConstraintValidatorInitializationContext() ) {
+				.getDefaultConstraintValidatorInitializationContext() ) {
```
gsmet: I think the formatting were right before.

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/ValidationContext.java
```diff
@@ -8,70 +8,74 @@
...
 
-	boolean isBeanAlreadyValidated(Object value, Class<?> group, PathImpl path);
-
-	void markCurrentBeanAsProcessed(ValueContext<?, ?> valueContext);
-
-	Set<ConstraintViolation<T>> getFailingConstraints();
-
+	/**
+	 * Creates a constraint violation {@link ConstraintViolation} based on passed parameters and adds it
+	 * to a set of failed constrains.
```
gsmet: We don't need to repeat `constraint violation`.

s/a set of failed constrains/**the** set of failed constrain**t**s/

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/ValidationContext.java
```diff
@@ -8,70 +8,74 @@
...
 
-	boolean hasMetaConstraintBeenProcessed(Object bean, Path path, MetaConstraint<?> metaConstraint);
-
-	void markConstraintProcessed(Object bean, Path path, MetaConstraint<?> metaConstraint);
-
-	default boolean canProcess(MetaConstraint<?> metaConstraint) {
-		return true;
-	}
+	/**
+	 * @return a set of constraint vilations created with {@link ValidationContext#addConstraintFailure(ValueContext, ConstraintViolationCreationContext, ConstraintDescriptor)}.
```
gsmet: s/vilations/vi**o**lations/

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/ValidationContext.java
```diff
@@ -8,70 +8,74 @@
...
+	 */
+	Set<ConstraintViolation<T>> getFailingConstraints();
 
+	/**
+	 * @return created constraint validator context based on passed constraint descriptor and path.
```
gsmet: It would be more readable as `a newly created constraint validator context based on the passed constraint descriptor and path.`

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/AbstractValidationContext.java
```diff
@@ -288,7 +288,7 @@ public AbstractConstraintValidatorContext createConstraintValidatorContextFor(Co
 	@Override
 	public String toString() {
 		final StringBuilder sb = new StringBuilder();
-		sb.append( "ValidationContext" );
+		sb.append( "BaseBeanValidationContext" );
```
gsmet: Not very excited about this toString() here. Let's at least use the concrete class name. And tbh, I would have implemented the `toString()` in each child so that I can add the proper context information.

gsmet:
@marko-bekhta I added a few more comments, mostly typos and formatting issues.

I suggested something about the change I found too invasive for a very small benefit, let's see if you agree with that. I think it would be a nice compromise for now and we can later refine it if it happens to be needed.

marko-bekhta:
@gsmet sorry for all the typos and formatting. I think I fixed the problem in the IDE and it should not mess up the existing code. 
As for extracting `CrossParameterConstraintValidatorContext` I've done that. I've thought about such approach initially but wasn't sure that making some of those inner builder classes visible would be good (e.g. `ConstraintViolationBuilderImpl` or `NodeBuilder`)

gsmet:
Finally got to the bottom of it. I made quite a lot of editorialization.

One thing for the future: for internal interfaces/classes, you don't need to write dumb javadoc. We expect javadoc for the API/SPI and if the purpose of the method should be explained. It avoids having to maintain it in the future.

In the same vein, I updated the `toString()` methods to use `getClass().getSimpleName()` to avoid having stalled class names in the future.

Apart from that, reworded a couple of comments and fixed a few warnings (missing `@Override` and type parameters, not sure your IDE warns you about them).

Anyway, nice work and glad to see this tackled. I remember the first time we played with this idea a few months ago with a bit of emotion ;).

Thanks!

PR #953, opened 2018-05-02
HV-1609 Change the way of filtering which beans to process in CDI extension

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1609

I've got the answer on our question about `@WithAnnotations` and it turns out that such behavior is dictated by the [CDI spec](http://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#process_annotated_type):

> The annotation @WithAnnotations may be applied to the event parameter. If the annotation is applied, the container must only deliver ProcessAnnotatedType events for types which contain at least one of the annotations specified.

Hence there's not much can be done in this direction. So I've removed the annotation and added a custom filter to determine if we should try to wrap a bean with validation proxy.

I tried to run the same tests for tck 2.0.3 that were failing and with these changes all seems good.

If such approach is OK I'll create a JIRA ticket and update the commit message with it.

Link to mailing list - http://lists.jboss.org/pipermail/weld-dev/2018-May/003670.html

### DIFF
--- a/cdi/src/main/java/org/hibernate/validator/cdi/ValidationExtension.java
+++ b/cdi/src/main/java/org/hibernate/validator/cdi/ValidationExtension.java
@@ -18,6 +18,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
+import java.util.function.Predicate;
 
 import javax.enterprise.event.Observes;
 import javax.enterprise.inject.Default;
@@ -32,12 +33,9 @@ import javax.enterprise.inject.spi.BeforeBeanDiscovery;
 import javax.enterprise.inject.spi.Extension;
 import javax.enterprise.inject.spi.ProcessAnnotatedType;
 import javax.enterprise.inject.spi.ProcessBean;
-import javax.enterprise.inject.spi.WithAnnotations;
 import javax.enterprise.util.AnnotationLiteral;
 import javax.validation.BootstrapConfiguration;
 import javax.validation.Configuration;
-import javax.validation.Constraint;
-import javax.validation.Valid;
 import javax.validation.Validation;
 import javax.validation.Validator;
 import javax.validation.ValidatorFactory;
@@ -47,6 +45,7 @@ import javax.validation.metadata.BeanDescriptor;
 import javax.validation.metadata.PropertyDescriptor;
 
 import org.hibernate.validator.cdi.internal.InheritedMethodsHelper;
+import org.hibernate.validator.cdi.internal.ValidateableBeanFilter;
 import org.hibernate.validator.cdi.internal.ValidationProviderHelper;
 import org.hibernate.validator.cdi.internal.ValidatorBean;
 import org.hibernate.validator.cdi.internal.ValidatorFactoryBean;
@@ -83,6 +82,7 @@ public class ValidationExtension implements Extension {
 			EnumSet.of( ExecutableType.CONSTRUCTORS, ExecutableType.NON_GETTER_METHODS, ExecutableType.GETTER_METHODS );
 	private static final EnumSet<ExecutableType> DEFAULT_EXECUTABLE_TYPES =
 			EnumSet.of( ExecutableType.CONSTRUCTORS, ExecutableType.NON_GETTER_METHODS );
+	private static final Predicate<Class<?>> VALIDATEABLE_BEAN_FILTER = new ValidateableBeanFilter();
 
 	@SuppressWarnings("serial")
 	private final Annotation defaultQualifier = new AnnotationLiteral<Default>() {
@@ -219,19 +219,23 @@ public class ValidationExtension implements Extension {
 	 * @param processAnnotatedTypeEvent event fired for each annotated type
 	 * @param <T> the annotated type
 	 */
-	public <T> void processAnnotatedType(@Observes @WithAnnotations({
-			Constraint.class,
-			Valid.class,
-			ValidateOnExecution.class
-	}) ProcessAnnotatedType<T> processAnnotatedTypeEvent) {
+	public <T> void processAnnotatedType(@Observes ProcessAnnotatedType<T> processAnnotatedTypeEvent) {
+		// NOTE: it is important to said here that we cannot use @WithAnnotations for filtering purposes here.
+		// This annotation does not consider checking implemented interfaces and super classes for annotations.
+		// Hence clean beans, without any annotation on them, that implement interfaces that have any annotation
+		// that is of iterest to us (@Valid, @Constraint, @ValidateOnExecution) will not be pushed to this method
+		// if @WithAnnotations is used.
+		// To prevent filling memory with unneeded metadata and redundant work in HV we have a custom VALIDATEABLE_BEAN_FILTER
+		// that checks for annotation presense in the way that we need.
 		Contracts.assertNotNull( processAnnotatedTypeEvent, "The ProcessAnnotatedType event cannot be null" );
 
-		// validation globally disabled
-		if ( !isExecutableValidationEnabled ) {
+		AnnotatedType<T> type = processAnnotatedTypeEvent.getAnnotatedType();
+
+		// validation globally disabled or annotated type has none of needed annotations:
+		if ( !isExecutableValidationEnabled || !VALIDATEABLE_BEAN_FILTER.test( type.getJavaClass() ) ) {
 			return;
 		}
 
-		AnnotatedType<T> type = processAnnotatedTypeEvent.getAnnotatedType();
 		Set<AnnotatedCallable<? super T>> constrainedCallables = determineConstrainedCallables( type );
 
 		if ( !constrainedCallables.isEmpty() ) {
--- a/cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
+++ b/cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
@@ -7,6 +7,8 @@
 package org.hibernate.validator.cdi.internal;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedParameterizedType;
+import java.lang.reflect.AnnotatedType;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Executable;
 import java.lang.reflect.Field;
@@ -68,6 +70,9 @@ public class ValidateableBeanFilter implements Predicate<Class<?>> {
 
 		// 2. Or on a field level
 		for ( Field field : clazz.getDeclaredFields() ) {
+			if ( isAnnotationPresent( field.getAnnotatedType(), processedAnnotations ) ) {
+				return true;
+			}
 			if ( isAnnotationPresentIn( field.getDeclaredAnnotations(), processedAnnotations ) ) {
 				return true;
 			}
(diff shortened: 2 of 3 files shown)


cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.stream.Stream;
+
+import javax.validation.Constraint;
+import javax.validation.Valid;
+import javax.validation.executable.ValidateOnExecution;
+
+import org.hibernate.validator.internal.util.classhierarchy.ClassHierarchyHelper;
+
+/**
+ * A filter that checks if the passed class has any of annotaitons:
```
gsmet: s/any of annotaitons/any of **these** annota**ti**ons/

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
```diff
@@ -0,0 +1,89 @@
...
+ * <ul>
+ * <li>{@link Valid}</li>
+ * <li>{@link Constraint}</li>
+ * <li>{@link ValidateOnExecution}</li>
+ * </ul>
+ * <p>
+ * on a type level, or on any member, or any parameter of any member.
+ * The annotation may be applied as a meta-annotation on any annotation considered
```
gsmet: Missing dot at the end of line.

We should also mention the fact that it takes into account the parent classes and the implemented interfaces.

cdi/src/main/java/org/hibernate/validator/cdi/ValidationExtension.java
```diff
@@ -219,19 +219,16 @@ else if ( bean.getTypes().contains( Validator.class ) || bean instanceof Validat
 	 * @param processAnnotatedTypeEvent event fired for each annotated type
 	 * @param <T> the annotated type
 	 */
-	public <T> void processAnnotatedType(@Observes @WithAnnotations({
-			Constraint.class,
-			Valid.class,
-			ValidateOnExecution.class
-	}) ProcessAnnotatedType<T> processAnnotatedTypeEvent) {
+	public <T> void processAnnotatedType(@Observes ProcessAnnotatedType<T> processAnnotatedTypeEvent) {
```
gsmet: Maybe we should add a comment about why we don't use `@WithAnnotations`.

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
```diff
@@ -0,0 +1,89 @@
...
+				.anyMatch( annotation -> isAnnotationAllowed( annotation, processedAnnotations ) )
+				// 3. or on any executable
+				// 3.1 constructors:
+				|| Arrays.stream( clazz.getDeclaredConstructors() )
+				.flatMap( this::fromExecutable )
+				.anyMatch( annotation -> isAnnotationAllowed( annotation, processedAnnotations ) )
+				// 3.2 methods:
+				|| Arrays.stream( clazz.getDeclaredMethods() )
+				.flatMap( this::fromExecutable )
+				.anyMatch( annotation -> isAnnotationAllowed( annotation, processedAnnotations ) );
```
gsmet: I know you love streams and lambdas very much :) but... we are trying to avoid them when there is no clear benefit as there is a cost to it.

In this case, I think a bunch of for loops would be more readable, especially since the `return` will serve as a `break`.

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
```diff
@@ -0,0 +1,89 @@
...
+
+	private boolean isAnnotationAllowed(Annotation annotationToCheck, Set<Annotation> processedAnnotations) {
+		// need to have this to prevent infinite loops for example on annotations like @Target
+		if ( !processedAnnotations.add( annotationToCheck ) ) {
+			return false;
+		}
+		Class<? extends Annotation> annotationType = annotationToCheck.annotationType();
+		return ALLOWED_ANNOTATIONS.contains( annotationType )
+				|| Arrays.stream( annotationType.getDeclaredAnnotations() )
+				.anyMatch( annotation -> isAnnotationAllowed( annotation, processedAnnotations ) );
```
gsmet: Same here, not sure this construct buys us much.

cdi/src/test/java/org/hibernate/validator/test/cdi/internal/ValidateableBeanFilterTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+	private static class Foo {
+
+		public void doNothing() {
+		}
+	}
+
+	private static class FooBar extends Foo implements Bar {
+
+		@Override public int bar() {
```
gsmet: We usually have the `@Override` on its own line.

cdi/src/test/java/org/hibernate/validator/test/cdi/internal/ValidateableBeanFilterTest.java
```diff
@@ -0,0 +1,86 @@
...
+			return 0;
+		}
+
+		public void doSomething(Foo foo) {
+		}
+	}
+
+	interface Bar {
+
+		@Min(10)
```
gsmet: It would be nice to also have a test for `@Valid` and `@ValidateOnExecution` in this very case (we don't need to test all the combinations of all cases but 2 more interfaces with both annotations would be nice.

cdi/src/test/java/org/hibernate/validator/test/cdi/internal/ValidateableBeanFilterTest.java
```diff
@@ -0,0 +1,86 @@
...
+		@Override public int compareTo(Baz o) {
+			return 0;
+		}
+
+		@Override public boolean test(Integer integer) {
+			return false;
+		}
+	}
+
+	private static class BarBaz extends Baz implements Bar {
```
gsmet: I think, in this case, it would make sense to name the classes with the test case they're supposed to check. It's a bit hard to follow otherwise.

gsmet:
Jenkins retest this please

gsmet:
BTW, if it wasn't clear, it's a big +1 for creating a JIRA and pursue this path :).

marko-bekhta:
Updated the description and the code. 
So it was my bad :) I understood that the exact behavior for interface/superclass was not specified  and that we might get it fixed.

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
```diff
@@ -0,0 +1,131 @@
...
+	private boolean isAnnotationPresentAnywhere(Class<?> clazz) {
+		Set<Annotation> processedAnnotations = new HashSet<>();
+		// 1. Is present on a type level:
+		if ( isAnnotationPresentIn( clazz.getDeclaredAnnotations(), processedAnnotations ) ) {
+			return true;
+		}
+
+		// 2. Or on a field level
+		for ( Field field : clazz.getDeclaredFields() ) {
+			if ( isAnnotationPresentIn( field.getDeclaredAnnotations(), processedAnnotations ) ) {
```
gsmet: @marko-bekhta Ah ah, when I took my shower this morning, I thought about the type argument constraints/`@Valid`. We should take them into account for fields, parameters and return values.

Always take a shower before merging a PR :).

Other than that, it looks good.

marko-bekhta: :) that's a good point. Wonder if `@WithAnnotations` would pick up such case.

gsmet: Yeah, not so sure, but better get it entirely right this time.

marko-bekhta:
@gsmet I've deliberately added a new commit so it'll be more visible what I've changed. But I'd say we should squash them together. 
I've added a couple more test cases as our felix osgi test started to fail. And it turned out that somehow this https://github.com/hibernate/hibernate-validator/pull/953/commits/06c769d7f29c3965378b18da2d0acc3639deb2ae#diff-66e9ba3cc4d09f83eb2ae799daeacab5R116 would pick an annotation from something like:
```java
private Min(10) int num;
```
but would fail to do so in case of a custom constraint:
```java
private @Custom SomeBean bean;
```
and the same applies for executable parameters...

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
```diff
@@ -42,6 +45,7 @@
  *
  * @author Marko Bekhta
  */
+@IgnoreForbiddenApisErrors(reason = "")
```
gsmet: Hum, why this?

cdi/src/test/java/org/hibernate/validator/test/cdi/internal/ValidateableBeanFilterTest.java
```diff
@@ -69,6 +84,43 @@ public void test() {
...
+		}
+	}
+
+	private static class TypeArgumentMethodParameter {
+		void strings(List<@NotNull String> strings) {
+		}
+	}
+
+	private static class TypeArgumentConstructorParameter {
+		TypeArgumentConstructorParameter(List<@NotNull String> strings) {
```
gsmet: Could you add a test for a nested one?

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
```diff
@@ -68,6 +72,9 @@ private boolean isAnnotationPresentAnywhere(Class<?> clazz) {
 
 		// 2. Or on a field level
 		for ( Field field : clazz.getDeclaredFields() ) {
+			if ( isAnnotationPresent( field.getAnnotatedType(), processedAnnotations ) ) {
+				return true;
+			}
 			if ( isAnnotationPresentIn( field.getDeclaredAnnotations(), processedAnnotations ) ) {
```
gsmet: Isn't this one already dealt with the call above?

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidateableBeanFilter.java
marko-bekhta: let's pretend it was never there :)
was doing some `System.out` and `printStackTrace()` when the Felix test start to fail. But apparently forgot about it when was pushing the changes. Should be ok now.

cdi/src/test/java/org/hibernate/validator/test/cdi/internal/ValidateableBeanFilterTest.java
marko-bekhta: done. Added:
```java
void doSomethinf(List<Map<String, List<Optional<Map<String, @NotNull String>>>>> strings){
}
...
private List<Map<String, @NotNull String>> strings;
```
in different classes.

gsmet:
@marko-bekhta so as I understand it, what you did does not work correctly with custom constraint and Felix?

What error do you have? I don't see anything risky that would fail with OSGi.

marko-bekhta:
I've explained it badly... So everything works fine with the first commit but we ignore the type argument annotations with it. When I've added support for type arguments I was assuming that having for example for fields only this check would be enough:
```java
// 2. Or on a field level
for ( Field field : clazz.getDeclaredFields() ) {
    if ( isAnnotationPresent( field.getAnnotatedType(), processedAnnotations ) ) {
        return true;
    }
}
```
i.e. getting the annotated type and checking both declared annotations and type arguments if any are available. 
And such check passed the initial `ValidateableBeanFilterTest` where we had a case:
```java
private static class AnnotatedField {
    @Min(10)
    private int num;
}
```
and I had same kind of checks for executable parameters. Just checking the `executable.getAnnotatedParameterTypes()`. Which also worked fine in initial `ValidateableBeanFilterTest`. So I run a full `mvn clean install` to discover that felix tests start to fail. It turned out that the reason for that was that the bean from those tests was not wrapped with validation proxy. That's when I've tried to add similar case to `ValidateableBeanFilterTest` and it turned out that if there's a custom annotation like `@ValidNumber` on field, parameter or method an annotated type doesn't contain that annotation, while it contains the `@Min`. So that's why I've added those additional checks for `something.getDeclaredAnnotations()` and all went back to normal after that.

I've left the inline comment about such behavior near parameter checks code, didn't want to copy/paste the same in the other places. If you'd like to see this behavior just try to comment out:
```java
if ( isAnnotationPresentIn( field.getDeclaredAnnotations(), processedAnnotations ) ) {
    return true;
}
```
and run the test for the filter. With this peace of code commented out, it's failing for me on:
```java
assertTrue( predicate.test( BeanWithCustomConstraintOnField.class ) );
```
while works perfectly fine on:
```java
assertTrue( predicate.test( AnnotatedField.class ) );
```
Hope this explains things better :)

gsmet:
OK, cool, I'll do a final review to tidy up some comment and merge.

gsmet:
Fixed a few typos here and there and changed a few method names, hopefully for the better :), and merged.

Thanks for taking care of this.

aguibert:
@gsmet @marko-bekhta this PR seems to cause a significant performance regression. Removing `@WithAnnotations` from the CDI extension is going to cause the method to be called far more often, additionally the `VALIDATEABLE_BEAN_FILTER.test()` call now done on every invocation of `processAnnotatedType()` is doing costly reflection on each call.

My performance team took a sampling of CDI application startup time going from Hibernate Validator 6.0.9.Final to 6.0.10.Final and measured a 6% regression with IBM JDK 8 sr6 and a 19% regression with Oracle JDK 8u171.

It seems the BeanValidation spec overstepped its bounds a bit by promising this behavior to work when CDI had no reasonable capability to do so.  I propose that we either:
 - revise the Bean Validation spec to remove this behavior claim or
 - open a spec enhancement request on CDI to provide this functionality in a future version, and at that time provide this capability from Bean Validation

gsmet:
@aguibert interesting. I'm a bit torn about it as it's a really surprising behavior and might lead to constraints being totally ignored.

One way to improve the situation might be to ignore classes from the `java.` package (that would exclude analyzing `Object` methods which might help). Don't know if it would make a real difference in practice, and, for sure, it wouldn't solve the entire scanning issue.

I would have added an option (or a parallel implementation of the extension) if it was possible but unfortunately I don't think it is...

gsmet:
@aguibert could your team test with https://github.com/gsmet/hibernate-validator/tree/cdi-performance-issue-6.0 ?

It won't entirely mitigate the issue but maybe it will help.

gsmet:
@aguibert any chance you could take a look at this before next week? I might cut a release soon to fix a JDK 11 issue so I could include this commit in the release.

Thanks!

PR #954, opened 2018-05-03
HV-1607 Make the build work with JDK 10

gsmet:
We need prevent the OSGi integration tests from running on JDK 10 for
now as pax-exam does not support JDK 10.

It will be fixed in the not yet released pax-exam 4.12.

 * https://hibernate.atlassian.net/browse/HV-1607

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -92,7 +92,6 @@
         <module>annotation-processor</module>
         <module>integration</module>
         <module>performance</module>
-        <module>osgi</module>
     </modules>
 
     <properties>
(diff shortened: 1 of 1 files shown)


marko-bekhta:
Looks good. More out of curiosity - was the forbiddenAPI failing on 10?

gsmet:
I don't think so, it's just that I experimented with the things deprecated for JDK 9 and 10 and kept the plugin upgrade.

For the record, `AccessibleObject#isAccessible()` is deprecated and should be replaced by `canAccess()` which is only present in JDK 9+. I suppose at some point, we will need MR jars...

marko-bekhta:
that probably will be fun ...
And yes, it's good to merge this one :)

PR #955, opened 2018-05-04
Added Dutch validation messages

Maurice-Betzel:
* https://hibernate.atlassian.net/browse/HV-1612


gunnarmorling:
ok to test

gsmet:
Hi @Maurice-Betzel ,

Thanks for your contribution! Minor nitpick: could we be consistent and not have leading capitals on some of the messages? I would fix it myself but I was wondering if maybe `Script` needed a capital in Dutch?

I just saw that the English validation messages are not consistent either, I suppose you followed what we did there, I'll go fix that.

Waiting for your feedback on this, will merge after that.

Thanks!

Maurice-Betzel:
Indeed, i just followed the default. Are you fixing the dutch props as well?

gsmet:
If you confirm me that `Script` can be lowercase too, I will amend your commit. Just checking as I know some languages have some weird rules about casing.

Maurice-Betzel:
No special rules apply, we can keep it analog to the default.

gsmet:
OK, cool, thanks for the confirmation.

gsmet:
Merged, thanks! I will make a global pass on case afterwards.

PR #956, opened 2018-05-04
HV-1611 Be consistent in the case of the validation messages

gsmet:
* https://hibernate.atlassian.net/browse/HV-1611

@marko-bekhta not the most interesting PR but would appreciate a quick check. Thanks!


marko-bekhta:
@gsmet the changed ones looks good. As for the rest - I suppose the ones in DE file should be with uppercase (e.g. Skriptausdruck), and also MN and RU ones have first words with uppercase chars. Do we need to change them ?

gsmet:
Yeah, German loves capitals. As for RU and MN, I'm not familiar with them so I refrained from doing any change.

Changes welcome if you're confident we should change them.

marko-bekhta:
well I don't know the Mongolian, but at least all the messages there start with a capital letter. So they are consistent. Updated a few lines from the Ru file. You can just squash them into your commit so it'll be cleaner.

gsmet:
OK, cool, thanks. Done and merged!

PR #957, opened 2018-05-07
HV-1534 <getter> property validation doesn't work when using <constraint-mappings> xml configuration

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1534

This doesn't seem to be a regression and I think it make sense to throw an exception in such cases. Changing `clazz. getMethod()` to `clazz.getDeclaredMethod()` does the trick. We also use `GetDeclaredMethod` which lookups the declared methods in case of xml method constraints. Hence having the same behavior for getters looks consistent. 
I've changed the message a bit to reflect the fact that the property should actually be defined for that particular type, rather than just containing it, which could happen through inheritance. 

Probably would be nice to mention such change of behavior somewhere in the notes later.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetMethodFromPropertyName.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetMethodFromPropertyName.java
@@ -10,8 +10,8 @@ import java.lang.reflect.Method;
 import java.security.PrivilegedAction;
 
 /**
- * Returns the method with the specified property name or {@code null} if it does not exist. This method will prepend
- * 'is' and 'get' to the property name and capitalize the first letter.
+ * Returns the method with the specified property name or {@code null} if it is not declared in the specified class.
+ * This method will prepend 'is' and 'get' to the property name and capitalize the first letter.
  *
  * @author Emmanuel Bernard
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetMethodFromPropertyName.java
```diff
@@ -31,19 +33,18 @@ private GetMethodFromPropertyName(Class<?> clazz, String property) {
...
-			string[0] = Character.toUpperCase( string[0] );
-			String fullMethodName = new String( string );
+		char[] string = property.toCharArray();
+		string[0] = Character.toUpperCase( string[0] );
+		String fullMethodName = new String( string );
+
+		for ( String prefix : ReflectionHelper.PROPERTY_ACCESSOR_PREFIXES ) {
 			try {
-				return clazz.getMethod( "get" + fullMethodName );
+				return clazz.getDeclaredMethod( prefix + fullMethodName );
```
gsmet: Maybe let's rename the class to `GetDeclaredMethodFromPropertyName`?

gsmet:
Closed in favor of https://github.com/hibernate/hibernate-validator/pull/958.

PR #958, opened 2018-05-09
HV-1534 Allow getter constraints to be specified for sub-classes in XML configuration.

RADickinson:
* https://hibernate.atlassian.net/browse/HV-1534

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyMetaData.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyMetaData.java
@@ -262,7 +262,14 @@ public class PropertyMetaData extends AbstractConstraintMetaData {
 
 			// fast track if it's a regular constraint
 			if ( !(constraint.getLocation() instanceof TypeArgumentConstraintLocation) ) {
-				converted = getterConstraintLocation;
+				// If the declaring class is a sub-class that has inherited the getter, ensure the location is preserved during conversion.
+				// RD: This is to address HV-1534
+				if ( requiresConversionToGetterLocation( getterConstraintLocation, constraint ) ) {
+					converted = getterConstraintLocation;
+				}
+				else {
+					converted = constraint.getLocation();
+				}
 			}
 			else {
 				Deque<ConstraintLocation> locationStack = new ArrayDeque<>();
@@ -296,6 +303,19 @@ public class PropertyMetaData extends AbstractConstraintMetaData {
 			}
 
 			return MetaConstraints.create( typeResolutionHelper, valueExtractorManager, constraint.getDescriptor(), converted );
+		}
+
+		/**
+		 * Used to determine if the location of the given constraint requires conversion to the given {@link ConstraintLocation getterConstraintLocation}.
+		 *
+		 * @param getterConstraintLocation The ConstraintLocation intended for converting the location of the existing constraint.
+		 * @param constraint The constraint to be converted for use by this PropertyMetaData
+		 * @return true if the constraint location requires conversion, otherwise false.
+		 */
+		private static boolean requiresConversionToGetterLocation(final ConstraintLocation getterConstraintLocation, final MetaConstraint<?> constraint) {
+			// If we're dealing with a getter constraint location for a different declaring class, do not convert a new getter location.
+			return !( getterConstraintLocation.getMember().equals( constraint.getLocation().getMember() )
+					&& !getterConstraintLocation.getDeclaringClass().equals( constraint.getLocation().getDeclaringClass() ) );
 		}
 
 		private String getPropertyName(ConstrainedElement constrainedElement) {
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyMetaData.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyMetaData.java
@@ -33,6 +33,7 @@ import org.hibernate.validator.internal.metadata.core.MetaConstraints;
 import org.hibernate.validator.internal.metadata.descriptor.PropertyDescriptorImpl;
 import org.hibernate.validator.internal.metadata.facets.Cascadable;
 import org.hibernate.validator.internal.metadata.location.ConstraintLocation;
+import org.hibernate.validator.internal.metadata.location.GetterConstraintLocation;
 import org.hibernate.validator.internal.metadata.location.TypeArgumentConstraintLocation;
 import org.hibernate.validator.internal.metadata.raw.ConstrainedElement;
 import org.hibernate.validator.internal.metadata.raw.ConstrainedElement.ConstrainedElementKind;
@@ -264,11 +265,11 @@ public class PropertyMetaData extends AbstractConstraintMetaData {
 			if ( !(constraint.getLocation() instanceof TypeArgumentConstraintLocation) ) {
 				// If the declaring class is a sub-class that has inherited the getter, ensure the location is preserved during conversion.
 				// RD: This is to address HV-1534
-				if ( requiresConversionToGetterLocation( getterConstraintLocation, constraint ) ) {
-					converted = getterConstraintLocation;
+				if ( constraint.getLocation() instanceof GetterConstraintLocation ) {
+					converted = constraint.getLocation();
 				}
 				else {
-					converted = constraint.getLocation();
+					converted = getterConstraintLocation;
 				}
 			}
 			else {
@@ -290,7 +291,14 @@ public class PropertyMetaData extends AbstractConstraintMetaData {
 				// 2. beginning at the root, transform each location so it references the transformed delegate
 				for ( ConstraintLocation location : locationStack ) {
 					if ( !(location instanceof TypeArgumentConstraintLocation) ) {
-						converted = getterConstraintLocation;
+						// If the declaring class is a sub-class that has inherited the getter, ensure the location is preserved during conversion.
+						// RD: This is to address HV-1534
+						if ( location instanceof GetterConstraintLocation ) {
+							converted = location;
+						}
+						else {
+							converted = getterConstraintLocation;
+						}
 					}
 					else {
 						converted = ConstraintLocation.forTypeArgument(
@@ -303,19 +311,6 @@ public class PropertyMetaData extends AbstractConstraintMetaData {
 			}
 
 			return MetaConstraints.create( typeResolutionHelper, valueExtractorManager, constraint.getDescriptor(), converted );
-		}
-
-		/**
-		 * Used to determine if the location of the given constraint requires conversion to the given {@link ConstraintLocation getterConstraintLocation}.
-		 *
-		 * @param getterConstraintLocation The ConstraintLocation intended for converting the location of the existing constraint.
-		 * @param constraint The constraint to be converted for use by this PropertyMetaData
-		 * @return true if the constraint location requires conversion, otherwise false.
-		 */
-		private static boolean requiresConversionToGetterLocation(final ConstraintLocation getterConstraintLocation, final MetaConstraint<?> constraint) {
-			// If we're dealing with a getter constraint location for a different declaring class, do not convert a new getter location.
-			return !( getterConstraintLocation.getMember().equals( constraint.getLocation().getMember() )
-					&& !getterConstraintLocation.getDeclaringClass().equals( constraint.getLocation().getDeclaringClass() ) );
 		}
 
 		private String getPropertyName(ConstrainedElement constrainedElement) {
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
@@ -43,12 +43,29 @@ public interface ConstraintLocation {
 		return new FieldConstraintLocation( field );
 	}
 
+	/**
+	 * Create a new {@link GetterConstraintLocation} for the given getter method.
+	 *
+	 * @param getter The getter method being constrained
+	 * @return A new GetterConstraintLocation
+	 */
 	static ConstraintLocation forGetter(Method getter) {
-		return new GetterConstraintLocation( getter , getter.getDeclaringClass() );
+		return new GetterConstraintLocation( getter.getDeclaringClass(), getter );
 	}
 
-	static ConstraintLocation forGetterOfClass(Method getter, Class<?> declaringClass) {
-		return new GetterConstraintLocation( getter, declaringClass );
+	/**
+	 * Create a new {@link GetterConstraintLocation} for the given declaringClass and getter method.
+	 * <p>
+	 * This provides an alternative to {@link ConstraintLocation#forGetter(Method)} where the given declaringClass is usually a sub-class of the
+	 * actual class on which the getter method is declared. This is provided to support XML mapping configurations used to specify constraints on
+	 * sub-classes for inherited getter methods.
+	 *
+	 * @param declaringClass The class on which the constraint is defined.
+	 * @param getter The getter method being constrained.
+	 * @return A new GetterConstraintLocation
+	 */
+	static ConstraintLocation forGetter(Class<?> declaringClass, Method getter ) {
+		return new GetterConstraintLocation( declaringClass, getter );
 	}
 
 	static ConstraintLocation forTypeArgument(ConstraintLocation delegate, TypeVariable<?> typeParameter, Type typeOfAnnotatedElement) {
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
@@ -44,7 +44,11 @@ public interface ConstraintLocation {
 	}
 
 	static ConstraintLocation forGetter(Method getter) {
-		return new GetterConstraintLocation( getter );
+		return new GetterConstraintLocation( getter , getter.getDeclaringClass() );
+	}
+
+	static ConstraintLocation forGetterOfClass(Method getter, Class<?> declaringClass) {
+		return new GetterConstraintLocation( getter, declaringClass );
 	}
 
 	static ConstraintLocation forTypeArgument(ConstraintLocation delegate, TypeVariable<?> typeParameter, Type typeOfAnnotatedElement) {
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/location/GetterConstraintLocation.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/location/GetterConstraintLocation.java
@@ -45,12 +45,14 @@ public class GetterConstraintLocation implements ConstraintLocation {
 	private final Type typeForValidatorResolution;
 
 	/**
-	 * The class declared for the method the constraint was defined on.
+	 * The class of the method for which the constraint was defined.
+	 * <p>
+	 * It is usually the same as the declaring class of the method itself, except in the XML case when a user could declare a constraint for a specific subclass.
 	 */
 	private final Class<?> declaringClass;
 
 
-	GetterConstraintLocation(Method method, Class<?> declaringClass) {
+	GetterConstraintLocation( Class<?> declaringClass, Method method ) {
 		this.method = method;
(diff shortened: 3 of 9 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyMetaData.java
```diff
@@ -262,7 +262,14 @@ else if ( constrainedElement.getKind() == ConstrainedElementKind.METHOD ) {
...
 			if ( !(constraint.getLocation() instanceof TypeArgumentConstraintLocation) ) {
-				converted = getterConstraintLocation;
+				// If the declaring class is a sub-class that has inherited the getter, ensure the location is preserved during conversion.
+				// RD: This is to address HV-1534
+				if ( requiresConversionToGetterLocation( getterConstraintLocation, constraint ) ) {
+					converted = getterConstraintLocation;
+				}
+				else {
+					converted = constraint.getLocation();
+				}
```
gsmet: OK, so this is an ugly hack so if it ends up to it, it's a no-no for me.

This method is used to convert other type of constraint location to a getter constraint location so if it is already one, we do not need to do anything.

So I would simply do:
```
if ( constraint.getLocation() instanceof GetterConstraintLocation ) {
    converted = constraint.getLocation();
}
else {
    converted = getterConstraintLocation;
}
```

and the same for the other spot you forgot at line 293 of your updated code.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyMetaData.java
```diff
@@ -298,6 +305,19 @@ else if ( constrainedElement.getKind() == ConstrainedElementKind.METHOD ) {
...
+		 *
+		 * @param getterConstraintLocation The ConstraintLocation intended for converting the location of the existing constraint.
+		 * @param constraint The constraint to be converted for use by this PropertyMetaData
+		 * @return true if the constraint location requires conversion, otherwise false.
+		 */
+		private static boolean requiresConversionToGetterLocation(final ConstraintLocation getterConstraintLocation, final MetaConstraint<?> constraint) {
+			// If we're dealing with a getter constraint location for a different declaring class, do not convert a new getter location.
+			return !( getterConstraintLocation.getMember().equals( constraint.getLocation().getMember() )
+					&& !getterConstraintLocation.getDeclaringClass().equals( constraint.getLocation().getDeclaringClass() ) );
+		}
```
gsmet: This would then be useless and could be removed.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
```diff
@@ -44,7 +44,11 @@ static ConstraintLocation forField(Field field) {
 	}
 
 	static ConstraintLocation forGetter(Method getter) {
-		return new GetterConstraintLocation( getter );
+		return new GetterConstraintLocation( getter , getter.getDeclaringClass() );
+	}
+
+	static ConstraintLocation forGetterOfClass(Method getter, Class<?> declaringClass) {
```
gsmet: I would put `declaringClass` first and we definitely need a comment here.

And I would simply name it `forGetter()` too.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/GetterConstraintLocation.java
```diff
@@ -44,17 +44,23 @@
...
 	private final Type typeForValidatorResolution;
 
+	/**
+	 * The class declared for the method the constraint was defined on.
+	 */
+	private final Class<?> declaringClass;
+
 
-	GetterConstraintLocation(Method method) {
+	GetterConstraintLocation(Method method, Class<?> declaringClass) {
```
gsmet: Same here about `declaringClass` being first.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/GetterConstraintLocation.java
```diff
@@ -44,17 +44,23 @@
 	 */
 	private final Type typeForValidatorResolution;
 
+	/**
+	 * The class declared for the method the constraint was defined on.
```
gsmet: What about that:
```
The class of the method for which the constraint was defined.
<p>
It is usually the same as the class of the method itself except in the XML case when a user could declare a constraint for a specific subclass.
```

engine/src/test/java/org/hibernate/validator/test/internal/metadata/location/ConstraintLocationTest.java
```diff
@@ -38,9 +38,33 @@ public void two_constraint_locations_for_the_same_member_should_be_equal() throw
...
+		assertEquals( location.getDeclaringClass(), Foo.class, "The owning class of a member should be the declaring class" );
+	}
+
+	@Test
+	@TestForIssue(jiraKey = "HV-1534")
+	public void bean_class_is_declaring_class() throws Exception {
+		Method getter = Bar.class.getMethod( "getBar" );
+		ConstraintLocation location = ConstraintLocation.forGetterOfClass( getter, Bar.class );
+
+		assertEquals( location.getDeclaringClass(), Bar.class, "The bean class should be the declaring class" );
```
gsmet: You can remove these two tests, they are useless and it's not a good idea to add tests for such trivial things.

engine/src/test/java/org/hibernate/validator/test/internal/xml/Child.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.test.internal.xml;
+
+/**
+ * Test class for HV-1534.
+ *
+ * @author robd
```
gsmet: Please use a real name as the author.

engine/src/test/java/org/hibernate/validator/test/internal/xml/Child.java
```diff
@@ -0,0 +1,26 @@
...
+	String childAttribute;
+
+	Child( String parentAttribute, String childAttribute ) {
+		super( parentAttribute );
+		this.childAttribute = childAttribute;
+	}
+
+	public final String getChildAttribute() {
+		return childAttribute;
+	}
```
gsmet: This `childAttribute` looks useless so let's get rid of it.

engine/src/test/java/org/hibernate/validator/test/internal/xml/Parent.java
```diff
@@ -0,0 +1,25 @@
...
+package org.hibernate.validator.test.internal.xml;
+
+/**
+ * Test class for HV-1534.
+ *
+ * @author robd
+ */
+public class Parent {
+
+	String parentAttribute = null;
```
gsmet: Let's make it private.

engine/src/test/java/org/hibernate/validator/test/internal/xml/Parent.java
```diff
@@ -0,0 +1,25 @@
...
+
+	Parent( String parentAttribute ) {
```
gsmet: Let's make it public.

engine/src/test/java/org/hibernate/validator/test/internal/xml/Parent.java
```diff
@@ -0,0 +1,25 @@
...
+		this.parentAttribute = parentAttribute;
+	}
+
+	public final String getParentAttribute() {
```
gsmet: Let's remove the final.

engine/src/test/java/org/hibernate/validator/test/internal/xml/Child.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Test class for HV-1534.
+ *
+ * @author robd
+ */
+public class Child extends Parent {
+
+	String childAttribute;
+
+	Child( String parentAttribute, String childAttribute ) {
```
gsmet: Let's make it public.

engine/src/test/java/org/hibernate/validator/test/internal/xml/Parent.java
gsmet: Please also add an attribute with a type argument constraint e.g. on String of `List<String>`. We need to have this working too.

Something like:
```
<getter name="parentListAttribute">
  <container-element-type type-argument-index="0">
    <constraint annotation="javax.validation.constraints.NotNull"/>
  </container-element-type>
</getter>
```
should do to declare the constraint.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyMetaData.java
RADickinson: Hi @gsmet thank you kindly for reviewing this PR so quickly. Appreciate this deviates from your original discussions. I will attempt to make changes and update the PR as soon as I'm able (hopefully tonight).

My usual preference is to move logic (along these lines) into a private static method in the class to improve readability, but I'm obviously happy to follow your direction on this.

I tried a couple of different checks to determine if the ConstraintLocation needed to be converted to the new GetterConstraintLocation and they caused other test failures. I will try again with your suggestion and if there are any failures (where fixes aren't apparent) I'll update in a comment here. Thanks.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
RADickinson: I initially overloaded the `forGetter()` method and it raised a lot of test failures as I think there are tests using reflection to inspect this interface. I'll rename as suggested and comment here with more details of failing tests and look for your direction. Thanks.

engine/src/test/java/org/hibernate/validator/test/internal/xml/Parent.java
RADickinson: If you're happy, I'd prefer this to remain final as it reflects the specific issue we have in our codebase that would be part of the problem we'd face if this fix was deemed unacceptable.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
gsmet: Mmmmh, that's weird, I wouldn't expect any issue here. But I haven't tried.

engine/src/test/java/org/hibernate/validator/test/internal/xml/Parent.java
gsmet: OK, makes sense.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyMetaData.java
gsmet: > My usual preference is to move logic (along these lines) into a private static method in the class to improve readability, but I'm obviously happy to follow your direction on this.

Agreed when it hides some complexity and allows to understand the logic better. If we end up with my proposal, I think it's better not to hide anything considering it's very simple and natural.

> I tried a couple of different checks to determine if the ConstraintLocation needed to be converted to the new GetterConstraintLocation and they caused other test failures. I will try again with your suggestion and if there are any failures (where fixes aren't apparent) I'll update in a comment here.

OK, let's try with my proposal. If it doesn't work, I'll have a look. But if we end up with complicated things, either we find a better solution or it will be a bit too hackyish for me I'm afraid.

RADickinson: Happy to proceed as you suggest.

marko-bekhta:
Hi @RADickinson @gsmet 
I was also thinking about this but in a slightly different direction, i.e. "how to achieve conditional validation of getters based on some subclass?". And I think this could be done using groups. So in case of example from the JIRA issue the corresponding xml could look like:
```xml
<constraint-mappings
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://jboss.org/xml/ns/javax/validation/mapping/validation-mapping-1.1.xsd"
        xmlns="http://jboss.org/xml/ns/javax/validation/mapping" version="1.1">

    <bean class="org.hibernate.validator.test.internal.xml.exception.XmlConfigurationExceptionTest$MyObject">
        <getter name="id">
            <constraint annotation="javax.validation.constraints.NotNull">
                <groups>
                    <value>org.hibernate.validator.test.internal.xml.exception.XmlConfigurationExceptionTest$MyClassGorup</value>
                </groups>
            </constraint>
        </getter>
    </bean>
    <bean class="org.hibernate.validator.test.internal.xml.exception.XmlConfigurationExceptionTest$MyClass">
        <class>
            <group-sequence>
                <value>org.hibernate.validator.test.internal.xml.exception.XmlConfigurationExceptionTest$MyClassGorup</value>
                <value>org.hibernate.validator.test.internal.xml.exception.XmlConfigurationExceptionTest$MyClass</value>
            </group-sequence>
        </class>
    </bean>
</constraint-mappings>
```
where you still define all constraints for classes where actual getters were defined, but also add group to it. And then for those subclasses that needs the particular constraint to work you add a group sequence.

I understand that this would still mean that your existing xmls should be changed, but wouldn't this solve the problem ?

The classes and the test for convenience: 
```java
public void testConstraintOnPropertyDefinedInSuperClass() {
    Configuration<?> configuration = ValidatorUtil.getConfiguration();
    configuration.addMapping(
            XmlConfigurationExceptionTest.class.getResourceAsStream(
                    "constraint-on-subclass-mapping.xml"
            )
    );

    Set<ConstraintViolation<MyClass>> violations = configuration.buildValidatorFactory().getValidator()
            .validate( new MyClass( null ) );
    
    assertThat( violations ).containsOnlyViolations(
            violationOf( NotNull.class )
    );
}

public static class MyObject {
    private final String id;

    public MyObject(String id) {
        this.id = id;
    }

    public String getId() {
        return id;
    }
}

public static class MyClass extends MyObject {
    public MyClass(String id) {
        super( id );
    }
}

public interface MyClassGorup {
}
```

gsmet:
@marko-bekhta nice trick but it's very verbose and you still need to have access to the code.

To be honest, my position is: if we can get it back with very minimal low risk changes, I would vote for going back to the 4.3 behavior. If not, then your workaround will be the way to go.

gsmet:
@RADickinson FYI, I plan to wrap the 6.0.10.Final release at the end of the week for announcement early next week. It would be nice if we could get this in.

RADickinson:
Hi @gsmet I'm just picking up the changes from the review comments now so hoping to get the PR updated before bed.
I'm having a little trouble with running the tests in my IDE now as I'm getting the error below. Is this something you've seen before and know how to fix? The tests appear to run OK from Maven so all being well I can complete the changes, however without the IDE working I can't debug anything. Any help gratefully received. Thanks
`AILED: bean_class_is_declaring_class
java.lang.ExceptionInInitializerError
	at org.hibernate.validator.internal.metadata.location.GetterConstraintLocation.<init>(GetterConstraintLocation.java:58)
	at org.hibernate.validator.internal.metadata.location.ConstraintLocation.forGetterOfClass(ConstraintLocation.java:51)
	at org.hibernate.validator.test.internal.metadata.location.ConstraintLocationTest.bean_class_is_declaring_class(ConstraintLocationTest.java:54)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:80)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:714)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:901)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1231)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:127)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:111)
	at org.testng.TestRunner.privateRun(TestRunner.java:767)
	at org.testng.TestRunner.run(TestRunner.java:617)
	at org.testng.SuiteRunner.runTest(SuiteRunner.java:334)
	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:329)
	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:291)
	at org.testng.SuiteRunner.run(SuiteRunner.java:240)
	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1198)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1123)
	at org.testng.TestNG.run(TestNG.java:1031)
	at org.testng.remote.AbstractRemoteTestNG.run(AbstractRemoteTestNG.java:114)
	at org.testng.remote.RemoteTestNG.initAndRun(RemoteTestNG.java:251)
	at org.testng.remote.RemoteTestNG.main(RemoteTestNG.java:77)
Caused by: java.lang.IllegalArgumentException: Invalid logger interface org.hibernate.validator.internal.util.logging.Log (implementation not found in sun.misc.Launcher$AppClassLoader@55f96302)
	at org.jboss.logging.Logger$1.run(Logger.java:2556)
	at org.jboss.logging.Logger.getMessageLogger(Logger.java:2529)
	at org.jboss.logging.Logger.getMessageLogger(Logger.java:2516)
	at org.hibernate.validator.internal.util.logging.LoggerFactory.make(LoggerFactory.java:22)
	at org.hibernate.validator.internal.util.ReflectionHelper.<clinit>(ReflectionHelper.java:50)`

gsmet:
Doesn't ring a bell. It seems to say that you have an error in the `GetterConstraintLocation` constructor but the stacktrace looks weird.

I think you should probably run a `mvn clean install -DskipTests` in your workspace directory then reload your classes in your IDE. It seems the `Log` implementation is not found (it's a generated class). You might also have to add the directory where it's generated to the build path of your IDE (don't remember the name exactly but it's in target and has something to do with apt or annotations).

If you can't get to debug it, try to apply the changes I mentioned and I'll get back to it on Friday.

RADickinson:
Hi @gsmet, 
PR now updated with (I believe) all suggestions and comments addressed. No further issues found after implementing the suggested changes, and the error I saw earlier disappeared after a few further changes, builds, refreshes etc.
I've not rebased my branch so changes are in a new commit, but happy to squash if you prefer.
I had to read up on constrained container-element-type but I think the new test should cover this adequately. My reading led me to the following which is slightly concerning https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/?v=6.0#section-method-constraints-inheritance-hierarchies

Hi @marko-bekhta ,
Thanks for your suggestion earlier. I hope my PR is acceptable to be merged as I believe it will allow our code to remain as is, however this is a good alternative to overriding the attributes in sub-classes. 
I think my main concern is the readability of the XML configuration file, and the fact you'd always need to define a new validation interface for each new sub-class requiring constraints, but we could probably live with this.
One question on your suggested approach, which I admit I've not looked into, is how we should handle validation of the sub-class when using the `validator.validate(T object, Class<?>... groups)` API. Would we need to ensure the classes specified in the `group-sequence` are included in the list of groups? If so, that may cause us further problems, otherwise this could be a viable alternative. 

Thanks
Rob

gsmet:
> My reading led me to the following which is slightly concerning

Yes, these are rules we won't change. but I think they are there since the very beginning of BV/HV.

RADickinson:
> Yes, these are rules we won't change. but I think they are there since the very beginning of BV/HV.

At risk of this PR being rejected, doesn't this change break these rules?

RADickinson:
Thanks Guilluame.

I understand this PR is in contention with the discussions you've had with Marko regarding the correctness of this behaviour in the XML mapping configuration. My intention was purely to provide a fix that will re-instate the original behaviour (albeit possibly incorrect) and leave the decision on whether it was acceptable to merge in your capable hands :)

Thanks for your help and guidance in completing this work. Very happy to contribute.
Rob

gsmet:
I editorialized a few comments and the commit message and squashed the 2 commits.

Merged and backported to 6.0. Will be part of the upcoming 6.0.10.Final.

Thanks a lot for taking the time to write some code. It always makes a difference.

PR #959, opened 2018-05-09
HV-1614 Unable to specify constraints at more than 1 nested parameter of a typed container

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1614

So it turned out that there actually was a bug. We need to check if the context was already created by previous calls prior to returning the new one. I've changed a couple of `Map#put()` calls to `Map#computeIfAbsent()` and it should be fine now.


marko-bekhta:
@gsmet looks good! Thanks for the mail on lambdas.

gsmet:
Squashed the 2 commits, merged and backported to 6.0.

Thanks @marko-bekhta .

PR #960, opened 2018-05-10
HV-1484 Unify JPMS module names and OSGi bundle names

gsmet:
* https://hibernate.atlassian.net/browse/HV-1484

Not the most interesting PR but, heh, it's one less issue to deal with :).


gsmet:
Merged!

PR #962, opened 2018-06-01
HV-667 Create the Bean Validation descriptor model lazily

gsmet:
@gunnarmorling an oldie (January 2013...) but goodie.

An easy way to reduce bootstrapping work and memory usage. Best reviewed commit per commit (one is just a simple method renaming)

https://hibernate.atlassian.net/browse/HV-667

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -364,7 +364,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 		}
 
 		BeanMetaData<U> beanMetaData = valueContext.getCurrentBeanMetaData();
-		if ( beanMetaData.defaultGroupSequenceIsRedefined() ) {
+		if ( beanMetaData.isDefaultGroupSequenceRedefined() ) {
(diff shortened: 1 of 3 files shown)


yrodiere:
To clarify: I don't *know* if performance will degrade, just saying I'm not sure it will improve or degrade, and depending on the context both scenarios seem plausible. If only there was some benchmark of Hibernate Validator... :)

gsmet:
Yes, so the descriptors should nearly never be used, that's why we decided to do that - it was Gunnar's idea originally and I concur with him. It's part of the Bean Validation spec to allow users to inspect the metadata but it's not something people should often do. Typically, HV doesn't use these descriptors internally at all.

So it's going to be a win for 99.99% of our users. And a probably not quantifiable loss for the rest.

Apart from the bootstrapping time it spares, it also spares some memory.

gunnarmorling:
Hey @gsmet, very nice. It looks good to me. A you say, it's going to be an improvement for the vast majority of users - in terms of bootstrap time and memory usage - and I don't think even make things worse for users of the metamodel. As before, they'll have to pay the init cost once, only now it's delayed. And well, a volatile read instead of non-volatile. I think that's ok :)

gsmet:
Rebased and merged, thanks to both of you.

PR #963, opened 2018-06-01
HV-1623 Build an abstraction over reflection in engine code

marko-bekhta:

this is an "abstraction" part of this PR https://github.com/hibernate/hibernate-validator/pull/938

there are a few more clean up commits added on top of the existing changes in the mentioned PR. 
As for the last commit that wraps `ElementType` - I am not very happy about it and I'll be looking into changing it. Initially I wanted to use `ConstraintLocation` to determine `ElementType` needed for calls like this - https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java#L1282-L1287 but then there's this usage https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/metadata/descriptor/ElementDescriptorImpl.java#L145 of `ElementType` from constraint descriptors which I didn't know about and it broke the initial approach on this :) But I plan to revisit this idea once more and try to clean things up in this area of `ElementType` usage.


marko-bekhta:
NOTE to myself: address this - https://github.com/hibernate/hibernate-validator/pull/949#discussion_r191736280

engine/src/main/java/org/hibernate/validator/internal/cfg/context/ConstructorConstraintMappingContextImpl.java
```diff
@@ -17,15 +16,15 @@
  */
 class ConstructorConstraintMappingContextImpl extends ExecutableConstraintMappingContextImpl implements ConstructorConstraintMappingContext {
 
-	<T> ConstructorConstraintMappingContextImpl(TypeConstraintMappingContextImpl<T> typeContext, Constructor<T> constructor) {
+	<T> ConstructorConstraintMappingContextImpl(TypeConstraintMappingContextImpl<T> typeContext, Callable constructor) {
```
gsmet: I still have my reservations about not making `Constructor` a thing. Typically, I did it in the java bean part in one of my commit. I think it wouldn't be bad to have this distinction.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/MethodConstraintMappingContextImpl.java
```diff
@@ -17,15 +16,15 @@
  */
 class MethodConstraintMappingContextImpl extends ExecutableConstraintMappingContextImpl implements MethodConstraintMappingContext {
 
-	MethodConstraintMappingContextImpl(TypeConstraintMappingContextImpl<?> typeContext, Method method) {
-		super( typeContext, method );
+	MethodConstraintMappingContextImpl(TypeConstraintMappingContextImpl<?> typeContext, Callable callable) {
```
gsmet: And I would also make `Method` a thing.

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -39,7 +39,6 @@
 	private final Map<String, Object> messageParameters;
 	private final Map<String, Object> expressionVariables;
 	private final Class<T> rootBeanClass;
-	private final ElementType elementType;
```
marko-bekhta: I've remembered what I wanted to ask about this commit :) I was just curious why we had `ElementType` in here in the first place ? Everything else in here looks fine.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/PropertyConstraintMappingContextImpl.java
```diff
@@ -64,17 +57,17 @@ protected PropertyConstraintMappingContextImpl getThis() {
 
 	@Override
 	public PropertyConstraintMappingContext constraint(ConstraintDef<?, ?> definition) {
-		if ( member instanceof Field ) {
+		if ( property instanceof JavaBeanField ) {
```
gsmet: Our abstraction doesn't work very well here.

That's why I'm not a big fan of merging field and getters so early and not being able to have a distinction between them at the base layer (by base, I mean `Property`, `Callable` and so on).

TBH, considering the behavior is entirely different, I'm not a big fan of this class. Note that the issue was preexisting, it's not something you introduced.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanAnnotatedElement.java
```diff
@@ -0,0 +1,34 @@
...
+
+	AnnotatedType getAnnotatedType();
+
+	Annotation[] getDeclaredAnnotations();
+
+	Type getGenericType();
+
+	TypeVariable<?>[] getTypeParameters();
+
+	<A extends Annotation> A getAnnotation(Class<A> annotationClass);
```
marko-bekhta: how about using Optional here? or would it be inconsistent with the rest of the code in, I suspect, `AnnotationMetaDataProvider` ?

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PropertyCascadable.java
```diff
@@ -6,37 +6,38 @@
...
 
-	GetterCascadable(Method method, CascadingMetaData cascadingMetaData) {
-		this.method = method;
-		this.propertyName = ReflectionHelper.getPropertyName( method );
-		this.cascadableType = ReflectionHelper.typeOf( method );
+	PropertyCascadable(Property property, CascadingMetaData cascadingMetaData) {
+		this.property = property;
+		this.cascadableType = property.getType();
 		this.cascadingMetaData = cascadingMetaData;
+		this.constraintLocationKind = property instanceof JavaBeanField ? ConstraintLocationKind.PROPERTY : ConstraintLocationKind.METHOD;
```
gsmet: We shouldn't have to rely on `JavaBeanField` here.

That will probably change if we reintroduce the difference between getters and properties. Frankly, I would really like to be able to follow the path of a getter and understand how it's specific.

If we end up still having to do that, I think we should have a `ConstrainedElementKind getConstrainedElementKind()` in `Constrainable` (instead of `Callable`). That might help for other spots too.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanGetter.java
```diff
@@ -88,6 +76,11 @@ public String getParameterName(ExecutableParameterNameProvider parameterNameProv
 		return declaringClass;
 	}
 
+	@Override
+	public ConstrainedElementKind getConstrainedElementKind() {
```
marko-bekhta: It seems that `JavaBeanMethod` already returns the same element kind. Do we need to override it here?

engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
```diff
@@ -112,8 +87,33 @@ static ConstraintLocation forParameter(Executable executable, int index) {
...
+	 * Returns the nature of the constraint location.
+	 */
+	ConstraintLocationKind getKind();
+
+	enum ConstraintLocationKind {
+		TYPE( ElementType.TYPE ),
+		CONSTRUCTOR( ElementType.CONSTRUCTOR ),
+		METHOD( ElementType.METHOD ),
+		PARAMETER( ElementType.PARAMETER ),
+		PROPERTY( ElementType.FIELD ),
```
gsmet: So it's used for getters + fields?

engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
```diff
@@ -112,8 +87,33 @@ static ConstraintLocation forParameter(Executable executable, int index) {
...
 	 * object array for a {@link ParameterConstraintLocation}.
 	 */
 	Object getValue(Object parent);
+
+	/**
+	 * Returns the nature of the constraint location.
+	 */
+	ConstraintLocationKind getKind();
+
+	enum ConstraintLocationKind {
```
gsmet: Wondering if we should move this one into a separate file as it's now a first class citizen.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanExecutable.java
```diff
@@ -83,60 +107,67 @@ public Type getType() {
 	}
 
 	@Override
-	public Class<?>[] getParameterTypes() {
```
marko-bekhta: This one was hard to read 😄

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedElement.java
```diff
@@ -47,7 +47,7 @@
 	 * @author Gunnar Morling
 	 */
 	enum ConstrainedElementKind {
-		TYPE, FIELD, CONSTRUCTOR, METHOD, PARAMETER
+		TYPE, CONSTRUCTOR, METHOD, PARAMETER, PROPERTY
```
gsmet: So that means `PROPERTY` is used for getters and fields, I suppose?

We would have to check this wrt the other changes regarding getter + field.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanField.java
```diff
@@ -136,7 +149,7 @@ private static Field getAccessible(Field original) {
 
 	/**
 	 * Runs the given privileged action, using a privileged block if required.
-	 * <p>
+	 *
```
gsmet: I don't think we should remove this.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanField.java
```diff
@@ -63,6 +66,31 @@ public String getPropertyName() {
 		return getName();
 	}
 
+	@Override
+	public AnnotatedType getAnnotatedType() {
```
marko-bekhta: I like these changes. We don't expose the reflection `Member` from these classes and annotation provider got cleaner. Only question left - should we iterate once more and hide the annotations as well ? I had such idea at first but discarded it till later times. I was thinking about moving all this logic of finding constraints into these `org.hibernate.validator.internal.properties.javabean.*` classes that would make the `AnnotationMetaDataProvider` much smaller. And also "hide" the reflection even more.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanGetter.java
```diff
@@ -146,7 +127,7 @@ private static Method getAccessible(Method original) {
 
 	/**
 	 * Runs the given privileged action, using a privileged block if required.
-	 * <p>
+	 *
```
gsmet: Same here. IMHO a quick check of the diff is in order :).

engine/src/main/java/org/hibernate/validator/internal/metadata/location/ParameterConstraintLocation.java
```diff
@@ -22,24 +22,24 @@
...
 	}
 
 	@Override
 	public Class<?> getDeclaringClass() {
-		return executable.getDeclaringClass();
+		return callable.getDeclaringClass();
 	}
 
 	@Override
 	public Constrainable getMember() {
```
marko-bekhta: Maybe let's rename the method as well? `getConstrainable()`?

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedElement.java
gunnarmorling: Hey, this ties in a bit with my recent comment: the "raw" model was not meant to do abstractions from actual elements (fields, getters etc.). Adding `PROPERTY` to `ConstraintElementKind` blurries that line. I didn't look into the entire PR, so perhaps it makes sense, but I wanted to bring up that it appears as quite some deviation from the existing abstractions.

marko-bekhta:
As for this part:

> the biggest issue being the property / field / getter thing

I agree. That's why I also left that ^ note to myself to fix things. It might be that we'd need to find a better name instead of a property. Right now we represent getters as methods all the way till the aggregation of the metadata. Hence using 'field' would be a better approach, but I'm not sure if these classes/enum values with property in the name are not also needed for non java bean types. I'll try to check that later today and will get back to you.

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedElement.java
marko-bekhta: +1 that's the same problem. Will work on it :)

gsmet:
Just force-pushed an update to fix the unused import issue.

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
gsmet: No idea tbh. Probably a leftover of some sort.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanAnnotatedElement.java
gsmet: I made it consistent with the reflection API. I think it's good enough as it is.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanField.java
gsmet: For now, I don't think we should add the reflection from `AnnotationMetaDataProvider`: it's tied to Java beans anyway. Maybe we will need to do it later but I think it can be kept for another day.

engine/src/main/java/org/hibernate/validator/internal/metadata/location/ParameterConstraintLocation.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanGetter.java
gsmet: Fixed.

gsmet:
@marko-bekhta fixed the issues you raised. I'm done with my changes now.

As for the naming, I think I would go with property = field + getter. And be consistent with it all the way. I would also try to make the getter case clearer than before. It's full of special cases everywhere about it being a method or a property, it's kinda confusing IMHO.

engine/src/main/java/org/hibernate/validator/internal/metadata/raw/ConstrainedElement.java
gsmet: @gunnarmorling yup, we are well aware of the issue. That's the big thing that still needs fixing in this PR.

gunnarmorling:
Ok, thanks! Maybe let me know once it's sorted out and I could take another
look.

gsmet:
@gunnarmorling I'll ping you before we merge this one so that you can take a look if you want.

gunnarmorling:
Excellent, thanks!

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ClassLevelMetaData.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Represents the constraint related meta data for a type i.e. class-level
+ * constraints.
+ *
+ * @author Gunnar Morling
+ * @author Guillaume Smet
+ * @author Marko Bekhta
+ */
+public class ClassLevelMetaData extends AbstractConstraintMetaData {
```
gsmet: +1 for this change.

I would call it `ClassMetaData` though, considering we don't have the `Level` part for the others.

engine/src/main/java/org/hibernate/validator/internal/metadata/descriptor/ClassLevelDescriptorImpl.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Set;
+
+import javax.validation.metadata.ElementDescriptor;
+
+/**
+ * Describes a validated type class-level constraints.
+ *
+ * @author Marko Bekhta
+ */
+public class ClassLevelDescriptorImpl extends ElementDescriptorImpl implements ElementDescriptor {
```
gsmet: Same here about removing the `Level` part.

engine/src/main/java/org/hibernate/validator/internal/metadata/descriptor/ClassLevelDescriptorImpl.java
```diff
@@ -0,0 +1,37 @@
...
+			Set<ConstraintDescriptorImpl<?>> constraints,
+			boolean defaultGroupSequenceRedefined,
+			List<Class<?>> defaultGroupSequence) {
+		super( beanType, constraints, defaultGroupSequenceRedefined, defaultGroupSequence );
+	}
+
+	@Override
+	public String toString() {
+		final StringBuilder sb = new StringBuilder();
+		sb.append( "ClassLevelDescriptorImpl" );
```
gsmet: Let's use `getClass().getSimpleName()` here to avoid future inconsistencies.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/FieldPropertyConstraintMappingContextImpl.java
```diff
@@ -0,0 +1,56 @@
...
+import org.hibernate.validator.internal.metadata.core.ConstraintHelper;
+import org.hibernate.validator.internal.metadata.descriptor.ConstraintDescriptorImpl.ConstraintType;
+import org.hibernate.validator.internal.metadata.raw.ConfigurationSource;
+import org.hibernate.validator.internal.metadata.raw.ConstrainedElement;
+import org.hibernate.validator.internal.metadata.raw.ConstrainedField;
+import org.hibernate.validator.internal.properties.Property;
+import org.hibernate.validator.internal.util.TypeResolutionHelper;
+
+/**
+ * Constraint mapping creational context implementation {@link PropertyConstraintMappingContextImpl} for
```
gsmet: There's a word missing or a typo here.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/GetterPropertyConstraintMappingContextImpl.java
```diff
@@ -0,0 +1,51 @@
...
+import org.hibernate.validator.internal.metadata.core.ConstraintHelper;
+import org.hibernate.validator.internal.metadata.raw.ConfigurationSource;
+import org.hibernate.validator.internal.metadata.raw.ConstrainedElement;
+import org.hibernate.validator.internal.metadata.raw.ConstrainedExecutable;
+import org.hibernate.validator.internal.properties.Callable;
+import org.hibernate.validator.internal.properties.Property;
+import org.hibernate.validator.internal.util.TypeResolutionHelper;
+
+/**
+ * Constraint mapping creational context implementation {@link PropertyConstraintMappingContextImpl} for
```
gsmet: Same here.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/PropertyConstraintMappingContextImpl.java
```diff
@@ -17,32 +17,50 @@
...
+		else if ( elementType == ElementType.METHOD ) {
+			return new GetterPropertyConstraintMappingContextImpl(
+					typeContext,
+					property
+			);
+		}
+		else {
+			throw LOG.getUnexpectedElementType( elementType, ElementType.FIELD, ElementType.METHOD );
+		}
+	}
```
gsmet: I wonder if we should deprecate `property(String property, ElementType type)` and introduce `field()` and `getter()`. That would be more in line with what we do in the XML.

Then I would instantiate directly `FieldPropertyConstraintMappingContextImpl` and `GetterPropertyConstraintMappingContextImpl` in those methods and disambiguate the types and throw the exception in the deprecated `property()`.

That if we take the road of having a specific DSL for JSON. My opinion is that it will be better. WDYT?

engine/src/main/java/org/hibernate/validator/internal/cfg/context/TypeConstraintMappingContextImpl.java
```diff
@@ -149,7 +150,7 @@ public MethodConstraintMappingContext method(String name, Class<?>... parameterT
 			throw LOG.getBeanDoesNotContainMethodException( beanClass, name, parameterTypes );
 		}
 
-		JavaBeanExecutable<?> javaBeanExecutable = JavaBeanExecutable.of( method );
+		JavaBeanMethod javaBeanExecutable = (JavaBeanMethod) JavaBeanExecutable.of( method );
```
gsmet: Mmmh, let's instantiate a `JavaBeanMethod` directly with the constructor instead of casting then :)

And let's call the variable `javaBeanMethod`.

marko-bekhta: as method here can be either method or getter we couldn't just use a constructor hence I've added an overloaded version of `JavaBeanExecutable#of()` one takes `Method` and returns `JavaBeanMethod` which might also be a getter, and the other one is the one that we already have that can also handle constructors.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/PropertyConstraintMappingContextImpl.java
marko-bekhta: Applied the change in the last commit. As for the JSON and all that "free-form validation" I'd say that as a lot of stuff there would be different and probably tagged as `@Incubating` we'd better separate these things. We already know that cascading would be different and for properties we need to pass a property type.

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
```diff
@@ -30,6 +30,45 @@
...
 	 * @return A creational context representing the selected property.
+	 *
+	 * @deprecated Since 6.1. Planned for removal. Use one of {@link PropertyTarget#field(String)},
+	 * {@link PropertyTarget#getter(String)} instead, based on a property kind.
 	 */
+	@Deprecated
 	PropertyConstraintMappingContext property(String property, ElementType type);
+
+	/**
+	 * Selects a filed to which the next operations shall apply.
```
gsmet: s/filed/field/

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
```diff
@@ -30,6 +30,45 @@
...
+	 * <p>
+	 * Until this method is called constraints apply on class level. After calling this method constraints
+	 * apply on the specified field property.
+	 * </p>
```
gsmet: Don't ask me why but in the javadoc, we don't close the `<p>`, we only open a new one when needed.

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
```diff
@@ -30,6 +30,45 @@
...
+	 * <p>
+	 * A given field may only be configured once.
+	 * </p>
+	 *
+	 * @param property The field name, that represents a property on which to apply the following constraints.
```
gsmet: Minor nitpicking: no comma here, `that` directly qualifies the `field name`.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/TypeConstraintMappingContextImpl.java
```diff
@@ -118,24 +117,58 @@
...
+		}
+		else {
+			return getter( property );
+		}
+	}
+
+	@Override
+	public PropertyConstraintMappingContext field(String property) {
+		Contracts.assertNotNull( property, "The property name must not be null." );
 		Contracts.assertNotEmpty( property, MESSAGES.propertyNameMustNotBeEmpty() );
```
gsmet: I think I would instead change `Contracts.assertNotEmpty()` to use `StringHelper.isNullOrEmptyString()`.

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
```diff
@@ -30,6 +30,45 @@
...
+	 *
+	 * @return A creational context representing the selected field property.
+	 *
```
gsmet: Let's remove this line.

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
```diff
@@ -30,6 +30,45 @@
...
+	 * </p>
+	 * <p>
+	 * A given getter may only be configured once.
+	 * </p>
+	 *
+	 * @param property The getter property name (Java Bean notation), that represents a property on which
+	 * to apply the following constraints.
+	 *
+	 * @return A creational context representing the selected getter property.
+	 *
```
gsmet: Let's remove this line.

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
```diff
@@ -30,6 +30,45 @@
...
+	 * <p>
+	 * Until this method is called constraints apply on class level. After calling this method constraints
+	 * apply on the specified getter property.
+	 * </p>
+	 * <p>
+	 * A given getter may only be configured once.
+	 * </p>
+	 *
+	 * @param property The getter property name (Java Bean notation), that represents a property on which
+	 * to apply the following constraints.
```
gsmet: I think we should give an example. Something like: "(using the JavaBean convention, e.g. {@code name} to address {@code getName()})".

engine/src/main/java/org/hibernate/validator/internal/cfg/context/TypeConstraintMappingContextImpl.java
```diff
@@ -133,8 +133,7 @@ public PropertyConstraintMappingContext property(String property, ElementType el
 
 	@Override
 	public PropertyConstraintMappingContext field(String property) {
-		Contracts.assertNotNull( property, "The property name must not be null." );
-		Contracts.assertNotEmpty( property, MESSAGES.propertyNameMustNotBeEmpty() );
+		Contracts.assertNeitherNullNorEmpty( property, MESSAGES.propertyNameMustBeNeitherNullNorEmpty() );
```
gsmet: Hum, sorry if I wasn't clear but I did expect you to only change the method content, not the method name or the message.

Being not empty implies being not null. See `NotEmptyValidatorForCharSequence` for instance :).

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -791,62 +776,62 @@ else if ( annotatedType instanceof AnnotatedParameterizedType ) {
...
+		return CascadingMetaDataBuilder.annotatedObject( annotatedElement.getType(), annotatedElement.isAnnotationPresent( Valid.class ),
+				containerElementTypesCascadingMetaData, getGroupConversions( annotatedElement.getAnnotatedType() ) );
 	}
 
 	/**
 	 * The location of a type argument before it is really considered a constraint location.
 	 * <p>
 	 * It avoids initializing a constraint location if we did not find any constraints. This is especially useful in
-	 * a Java 9 environment as {@link ConstraintLocation#forProperty(Member) tries to make the {@code Member} accessible
+	 * a Java 9 environment as {@link ConstraintLocation#forProperty(Property)} tries to make the {@code Member} accessible
```
marko-bekhta: This is actually not true anymore, as now we are making things accessible right away in `JavaBeanField` or `JavaBeanGetter` constructors. Hence two questions:
- do we need these `TypeArgumentLocation`s ?
- do we need to make things accessible lazily in those `JavaBean*` constructors?

engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
```diff
@@ -131,10 +131,24 @@ public boolean isMethod() {
 			return this == METHOD || this == GETTER;
 		}
 
-		public static ConstraintLocationKind of(Callable callable) {
-			return callable.getConstrainedElementKind() == ConstrainedElementKind.CONSTRUCTOR
-					? ConstraintLocationKind.CONSTRUCTOR
-					: callable instanceof JavaBeanGetter ? ConstraintLocationKind.GETTER : ConstraintLocationKind.METHOD;
+		public static ConstraintLocationKind of(ConstrainedElementKind constrainedElementKind) {
```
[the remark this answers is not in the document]
gsmet: Not a big fan to be honest: I prefer we keep them loosely coupled via the enums.

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
gsmet: Hum. That is indeed an interesting point. I missed that in the various refactorings.

This is annoying :/.

> do we need these ``TypeArgumentLocation``s ?

If you find a clean way to remove them, we can get rid of them. But let's do it in a follow-up PR.

> do we need to make things accessible lazily in those JavaBean* constructors?

Yes, we do. We shouldn't make the field accessible until we are sure there are constraints (or cascading) and we will need the value.

I'm not a big fan of using a double checked locking mechanism as it will be in a hot path so I tried to introduce a `PropertyAccessor` in a follow-up commit.

Another alternative would be to make `JavaBeanField` and `JavaBeanGetter` builders but I'm not sure it's going to be better.

A drawback of this approach is that we create 2 instances of the accessible fields. But we already did that.

WDYT?

gsmet:
@marko-bekhta I squashed your javadoc fix commit as I was worried it could break the javadoc build of the commit introducing this issue.

I also removed the commit I didn't like that much, I think it's better to keep the loose coupling.

As for your last concern, I tried to address it, let me know what you think.

gsmet:
@gunnarmorling I think we are near the conclusion of this PR so, if you want to take a look, now is the time :).

I plan to commit it on Tuesday morning. We should have converged on the last minute changes we are doing.

I recommend to review the whole diff as we took quite a lot of different directions for this one.

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
```diff
@@ -21,15 +21,46 @@
...
 	 *
 	 * @param property The property on which to apply the following constraints (Java Bean notation).
 	 * @param type The access type (field/property).
 	 *
 	 * @return A creational context representing the selected property.
+	 *
+	 * @deprecated Since 6.1. Planned for removal. Use either {@link PropertyTarget#field(String)} or
+	 * 		{@link PropertyTarget#getter(String)} instead.
 	 */
+	@Deprecated
```
gunnarmorling: @gsmet While I agree that the two new methods are nicer, I'm unsure whether it's worth the API change (hazzle with WildFly etc.).

gunnarmorling: Btw. one advantage of the original form was that it'll also better work with models that don't distinguish between fields/getters (HV-1000). In that case we could think of either accepting null for `type` (or simply ignoring whatever is passed) or add an overload `property(String name)`. So it feels we're limiting ourselves with this change a little bit.

engine/src/main/java/org/hibernate/validator/internal/engine/ConstraintViolationImpl.java
```diff
@@ -39,24 +38,22 @@
...
+			Map<String, Object> messageParameters,
+			Map<String, Object> expressionVariables,
+			String interpolatedMessage,
+			Class<T> rootBeanClass,
+			T rootBean,
+			Object leafBeanInstance,
+			Object value,
+			Path propertyPath,
+			ConstraintDescriptor<?> constraintDescriptor,
+			Object dynamicPayload) {
```
gunnarmorling: Thanks :)

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ClassMetaData.java
```diff
@@ -0,0 +1,111 @@
...
+				getType(),
+				asDescriptors( getDirectConstraints() ),
+				defaultGroupSequenceRedefined,
+				defaultGroupSequence
+		);
+	}
+
+	@Override
+	public String toString() {
+		return "ClassLevelMetaData [type=" + getType() + "]]";
```
gunnarmorling: s/ClassLevelMetaData/ClassMetaData/

engine/src/main/java/org/hibernate/validator/internal/metadata/core/AnnotationProcessingOptionsImpl.java
```diff
@@ -40,62 +40,62 @@
...
 	private final Map<ExecutableParameterKey, Boolean> annotationIgnoresForMethodParameter = newHashMap();
 
 	@Override
-	public boolean areMemberConstraintsIgnoredFor(Member member) {
-		Class<?> clazz = member.getDeclaringClass();
-		if ( annotationIgnoredForMembers.containsKey( member ) ) {
-			return annotationIgnoredForMembers.get( member );
+	public boolean areMemberConstraintsIgnoredFor(Constrainable constrainable) {
+		Class<?> clazz = constrainable.getDeclaringClass();
+		if ( annotationIgnoredForMembers.containsKey( constrainable ) ) {
```
gunnarmorling: I'd prefer to `get()` once and then check on null, so to avoid one lookup.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/AnnotationProcessingOptionsImpl.java
```diff
@@ -40,62 +40,62 @@
...
 			return areAllConstraintAnnotationsIgnoredFor( clazz );
 		}
 	}
 
 	@Override
-	public boolean areReturnValueConstraintsIgnoredFor(Member member) {
-		if ( annotationIgnoresForReturnValues.containsKey( member ) ) {
-			return annotationIgnoresForReturnValues.get( member );
+	public boolean areReturnValueConstraintsIgnoredFor(Constrainable constrainable) {
+		if ( annotationIgnoresForReturnValues.containsKey( constrainable ) ) {
```
gunnarmorling: Same here.

engine/src/main/java/org/hibernate/validator/internal/properties/Callable.java
```diff
@@ -0,0 +1,40 @@
...
+
+	Type getParameterGenericType(int index);
+
+	Class<?>[] getParameterTypes();
+
+	String getParameterName(ExecutableParameterNameProvider parameterNameProvider, int parameterIndex);
+
+	boolean isPrivate();
+
+	String getSignature();
```
gunnarmorling: I suppose that's for logging only?

engine/src/main/java/org/hibernate/validator/internal/properties/Constrainable.java
```diff
@@ -0,0 +1,32 @@
...
+package org.hibernate.validator.internal.properties;
+
+import java.lang.reflect.Type;
+
+import org.hibernate.validator.internal.metadata.raw.ConstrainedElement.ConstrainedElementKind;
+
+/**
+ * @author Marko Bekhta
+ */
+public interface Constrainable {
```
gunnarmorling: A class-level comment would be nice.

engine/src/main/java/org/hibernate/validator/internal/properties/Constrainable.java
```diff
@@ -0,0 +1,32 @@
...
+
+	String getName();
+
+	Class<?> getDeclaringClass();
```
gunnarmorling: Doesn't this break the "abstraction"? I.e. what to return for a JSON type?

engine/src/main/java/org/hibernate/validator/internal/properties/ConstrainableType.java
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.properties;
+
+/**
+ * @author Marko Bekhta
+ */
+public interface ConstrainableType {
```
gunnarmorling: What is this doing?

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
gsmet: > While I agree that the two new methods are nicer, I'm unsure whether it's worth the API change (hazzle with WildFly etc.).

The old method is still there, it's marked as deprecated.

> Btw. one advantage of the original form was that it'll also better work with models that don't distinguish between fields/getters (HV-1000).

The original form was misleading because there is indeed a difference in treatment between fields and getters and they are totally different beasts.

As for HV-1000 (which is one of the ultimate goals of this PR), there's a good chance the programmatic API entry points will be different (mostly because we will have to define the type for cascaded validation) so we will be able to have a `property()`.

If we end up reusing the same API classes, then we will unmark `property()` as deprecated but I still think keeping `getter()` and `field()` makes sense.

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ClassMetaData.java
gsmet: @marko-bekhta see, that's why I want us to use `getClass().getSimpleName()` :)

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
marko-bekhta: @gunnarmorling from what we have right now for json and maps, `PropertyTarget` for new programmatic mapping has just this method:
```java
public interface PropertyTarget {
    PropertyConstraintMappingContext property(String property, Class<?> propertyType);
}
```
as we need to pass the type of the property. Otherwise we would have all the info to create `Property`. Passing this type also serves as a type validation which I think is a nice bonus :).

So most likely we will end up adding another method anyway. 

Having those two shortcuts for field and getter is consistent with what we have for XML configuration (that's what we talked with @gsmet about). Hence if there might be issues with deprecation of `property()` maybe we should leave it as is also but keep the two shortcuts ?

engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/ClassMetaData.java
marko-bekhta: And I trusted IDEA to be good with such kind of refactorings!!! :smile: Thanks!

engine/src/main/java/org/hibernate/validator/internal/metadata/core/AnnotationProcessingOptionsImpl.java
gsmet: OK, this is a bit outside the scope of this PR as it's preexisting.

I opened https://hibernate.atlassian.net/browse/HV-1631. Will push an additional commit with this change.

engine/src/main/java/org/hibernate/validator/internal/properties/Callable.java
gsmet: Not only. The signature is used a mean to identify the executable in various parts of the code.

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
gunnarmorling: > The old method is still there, it's marked as deprecated.

Understood, but the goal of deprecation is removal, no? Feel free do it if you think it's worth it, to me it doesn't necessarily appear that way (given all the discussions we had in the past...).

> As for HV-1000 [...] entry points will be different

Mh, I see. I was hoping that wouldn't be the case.

> so we will be able to have a property().

I see.

engine/src/main/java/org/hibernate/validator/internal/properties/Constrainable.java
gsmet: Mmmh, yes it does. I think @marko-bekhta had in mind to provide a `PropertyHolder` class here or something similar.

The class is used in some portions of the code where it's not really easy to abstract from it.

Let's see how it goes with the JSON PR and act accordingly.

I created https://hibernate.atlassian.net/browse/HV-1632 so that we have another try at it once we got the JSON support into shape.

engine/src/main/java/org/hibernate/validator/internal/properties/ConstrainableType.java
gsmet: For now, nothing. I think Marko wanted a common interface for the `JavaBean` and the upcoming `PropertyHolder`.

Let's add a comment.

engine/src/main/java/org/hibernate/validator/cfg/context/PropertyTarget.java
gsmet: > Understood, but the goal of deprecation is removal, no?

Maybe at some point. But I agree we might just keep the method here forever.

To be honest, it's more to be clear than getters and fields should be different from the get to and to offer similar APIs for programmatic API and XML definition.

engine/src/main/java/org/hibernate/validator/internal/properties/Constrainable.java
gsmet: Done.

gsmet:
And it's merged!

@marko-bekhta thanks for the hard work on this one. @gunnarmorling thanks for the review.

PR #964, opened 2018-06-06
HV-1617 Make our pom files more consistent with WildFly and JBoss parent

gsmet:
* https://hibernate.atlassian.net/browse/HV-1617
 * https://hibernate.atlassian.net/browse/HV-1627
 * https://hibernate.atlassian.net/browse/HV-1626

Not the most interesting PR but all our projects are now consistent in the naming and more in line with what is done in WF parent and JBoss parent.

This should help for productization.

### DIFF
--- a/cdi/pom.xml
+++ b/cdi/pom.xml
@@ -214,7 +214,7 @@
             </activation>
             <properties>
                 <!-- required by Weld -->
-                <maven-surefire-plugin.argLine.add-opens>--add-opens java.base/java.lang=ALL-UNNAMED</maven-surefire-plugin.argLine.add-opens>
+                <surefire.jvm.add-opens>--add-opens java.base/java.lang=ALL-UNNAMED</surefire.jvm.add-opens>
             </properties>
         </profile>
     </profiles>
--- a/cdi/pom.xml
+++ b/cdi/pom.xml
@@ -214,7 +214,7 @@
             </activation>
             <properties>
                 <!-- required by Weld -->
-                <surefire.jvm.add-opens>--add-opens java.base/java.lang=ALL-UNNAMED</surefire.jvm.add-opens>
+                <surefire.jvm.args.add-opens>--add-opens java.base/java.lang=ALL-UNNAMED</surefire.jvm.args.add-opens>
             </properties>
         </profile>
     </profiles>
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -54,8 +54,8 @@
             <artifactId>log4j</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.hibernate.javax.persistence</groupId>
-            <artifactId>hibernate-jpa-2.1-api</artifactId>
+            <groupId>javax.persistence</groupId>
+            <artifactId>javax.persistence-api</artifactId>
         </dependency>
         <dependency>
             <groupId>joda-time</groupId>
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -93,13 +93,13 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>hibernate-validator-modules</artifactId>
-            <classifier>wildfly-${wildfly.version}-patch</classifier>
+            <classifier>wildfly-${version.wildfly}-patch</classifier>
             <type>zip</type>
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>hibernate-validator-modules</artifactId>
-            <classifier>wildfly-${wildfly-secondary.version}-patch</classifier>
+            <classifier>wildfly-${version.wildfly.secondary}-patch</classifier>
             <type>zip</type>
         </dependency>
     </dependencies>
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -40,7 +40,7 @@
 
         <!-- Skip artifact deployment -->
         <maven.deploy.skip>true</maven.deploy.skip>
-        <maven-surefire-plugin.argLine>-Duser.language=en</maven-surefire-plugin.argLine>
+        <surefire.jvm.args>-Duser.language=en</surefire.jvm.args>
 
         <forbiddenapis-junit.path>forbidden-allow-junit.txt</forbiddenapis-junit.path>
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
@@ -126,7 +126,7 @@
                                 <artifactItem>
                                     <groupId>org.hibernate.infra</groupId>
                                     <artifactId>hibernate-asciidoctor-theme</artifactId>
-                                    <version>${hibernate-asciidoctor-theme.version}</version>
+                                    <version>${version.org.hibernate.infra.hibernate-asciidoctor-theme}</version>
                                     <type>zip</type>
                                     <overWrite>true</overWrite>
                                     <outputDirectory>${project.build.directory}/</outputDirectory>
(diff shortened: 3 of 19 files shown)


pom.xml
```diff
@@ -95,131 +95,186 @@
...
-        <!-- URLs used in javadoc and documentation generation -->
-        <bv.spec.url>http://beanvalidation.org/2.0/spec/</bv.spec.url>
-        <java.api-docs.base-url>http://docs.oracle.com/javase/8/docs/api</java.api-docs.base-url>
-        <java.technotes.base-url>http://docs.oracle.com/javase/8/docs/technotes</java.technotes.base-url>
-        <javaee.api-docs.base-url>http://docs.oracle.com/javaee/7/api</javaee.api-docs.base-url>
-        <javafx.docs.url>http://docs.oracle.com/javase/8/javase-clienttechnologies.htm</javafx.docs.url>
-        <bv.api-docs.base-url>http://docs.jboss.org/hibernate/beanvalidation/spec/2.0/api</bv.api-docs.base-url>
-        <javamoney.api-docs.base-url>http://javamoney.github.io/apidocs</javamoney.api-docs.base-url>
+        <!-- JVM options required for Arquillian and WildFly -->
+        <arquillian.jvm.args></arquillian.jvm.args>
```
yrodiere: Note that Arquillian can be used with much more than WildFly, so it may be wise to at least put the term "wildfly" in there if you intend to put wildfly-specific JVM options.

pom.xml
```diff
@@ -95,131 +95,186 @@
...
-        <arquillian.javaVmArguments.add-opens></arquillian.javaVmArguments.add-opens>
+        <maven.compiler.argument.target>${maven.compiler.target}</maven.compiler.argument.target>
+        <maven.compiler.argument.source>${maven.compiler.source}</maven.compiler.argument.source>
+        <maven.compiler.argument.testTarget>${maven.compiler.testTarget}</maven.compiler.argument.testTarget>
+        <maven.compiler.argument.testSource>${maven.compiler.testSource}</maven.compiler.argument.testSource>
 
-        <!-- Forbidden API related properties -->
-        <forbiddenapis-junit.path>forbidden-junit.txt</forbiddenapis-junit.path>
+        <!-- Set empty default values to avoid Maven leaving property references (${...}) when it doesn't find a value -->
+        <surefire.jvm.args></surefire.jvm.args>
```
yrodiere: So, to explain the rationale... In Search, I used "jvm.args" for properties containing arguments that are specific to a version of the JVM, and "system.args" for the actual arguments passed to surefire/failsafe. Yeah, it's dumb.

If you don't use the exact same property names as Search, and thus are not constrained by consistency with Search, I would recommend to use clearer names, like what I proposed above.

pom.xml
```diff
@@ -414,38 +469,34 @@
...
     </dependencyManagement>
 
     <build>
-        <extensions>
-            <extension>
-                <groupId>org.apache.maven.wagon</groupId>
-                <artifactId>wagon-webdav</artifactId>
-                <version>1.0-beta-2</version>
-            </extension>
-        </extensions>
```
yrodiere: That looks dubious. Maybe move it to a separate commit with an explanation?

pom.xml
```diff
@@ -746,63 +801,50 @@
...
                 </plugin>
-                <plugin>
-                    <groupId>org.codehaus.mojo</groupId>
-                    <artifactId>exec-maven-plugin</artifactId>
-                    <!--
-                    1.6.0 has a bug that affects us, waiting for the next version
-                    https://github.com/mojohaus/exec-maven-plugin/issues/75
-                    -->
-                    <version>1.5.0</version>
-                </plugin>
```
yrodiere: Same here. Separate commit?

pom.xml
```diff
@@ -746,63 +801,50 @@
...
                     </dependencies>
                 </plugin>
                 <plugin>
                     <groupId>ch.mfrey.maven.plugin</groupId>
                     <artifactId>copy-maven-plugin</artifactId>
-                    <version>0.0.6</version>
-                </plugin>
-                <plugin>
-                    <artifactId>maven-project-info-reports-plugin</artifactId>
-                    <version>2.9</version>
```
yrodiere: Same here. Separate commit?

pom.xml
```diff
@@ -1140,12 +1175,11 @@
                 <jdk>[9,)</jdk>
             </activation>
             <properties>
-                <maven-surefire-plugin.argLine.add-modules>--add-modules=${maven-surefire-plugin.jigsaw.modules}</maven-surefire-plugin.argLine.add-modules>
```
yrodiere: Is the add-modules option not necessary anymore?

tck-runner/pom.xml
```diff
@@ -47,8 +47,8 @@
             <artifactId>javax.el</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.hibernate.javax.persistence</groupId>
-            <artifactId>hibernate-jpa-2.1-api</artifactId>
+            <groupId>javax.persistence</groupId>
+            <artifactId>javax.persistence-api</artifactId>
```
yrodiere: Same as above:

> Might want a separate commit for that? That has way more implications that just refactoring the properties.

cdi/pom.xml
```diff
@@ -214,7 +214,7 @@
             </activation>
             <properties>
                 <!-- required by Weld -->
-                <maven-surefire-plugin.argLine.add-opens>--add-opens java.base/java.lang=ALL-UNNAMED</maven-surefire-plugin.argLine.add-opens>
+                <surefire.jvm.add-opens>--add-opens java.base/java.lang=ALL-UNNAMED</surefire.jvm.add-opens>
```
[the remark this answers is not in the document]
gsmet: To be honest, I'm not sure. For now, I decided to keep it that way as the intent is clear. If at some point we need other things for future versions, I'll revisit.

I don't think it's easy to take the right decision completely in the dark and I fear we will lose the clarity we have now.

pom.xml
gsmet: Yes, we don't need it anymore as we removed our JAXB dependency: as it's totally removed from JDK 11, we needed to move to Stax (or add a specific dependency but that's not what we decided).

documentation/pom.xml
```diff
@@ -40,7 +40,7 @@
 
         <!-- Skip artifact deployment -->
         <maven.deploy.skip>true</maven.deploy.skip>
-        <maven-surefire-plugin.argLine>-Duser.language=en</maven-surefire-plugin.argLine>
+        <surefire.jvm.args>-Duser.language=en</surefire.jvm.args>
```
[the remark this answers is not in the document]
gsmet: Yes, that's what it did before so I'll keep it that way for now. Either we'll need a separate property or define the locale globally. Not decided for now.

pom.xml
```diff
@@ -95,131 +95,186 @@
...
+        <version.license.plugin>3.0</version.license.plugin>
+        <version.release.plugin>2.5.3</version.release.plugin>
+        <version.resources.plugin>3.0.2</version.resources.plugin>
+        <version.shade.plugin>3.1.0</version.shade.plugin>
+        <version.sigtest.plugin>1.0</version.sigtest.plugin>
+        <version.source.plugin>3.0.1</version.source.plugin>
+        <version.surefire.plugin>2.21.0</version.surefire.plugin>
+        <version.failsafe.plugin>${version.surefire.plugin}</version.failsafe.plugin>
+        <version.wildfly.plugin>1.2.1.Final</version.wildfly.plugin>
+        <version.org.wildfly.core>5.0.0.Final</version.org.wildfly.core>
```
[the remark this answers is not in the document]
gsmet: I decided it to put it there as it's tightly related to the WildFly Maven plugin. I did the same thing in OGM.

distribution/pom.xml
```diff
@@ -54,8 +54,8 @@
             <artifactId>log4j</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.hibernate.javax.persistence</groupId>
-            <artifactId>hibernate-jpa-2.1-api</artifactId>
+            <groupId>javax.persistence</groupId>
+            <artifactId>javax.persistence-api</artifactId>
```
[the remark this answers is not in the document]
gsmet: Done.

engine/pom.xml
```diff
@@ -279,7 +279,7 @@
                 <jdk>[9,)</jdk>
             </activation>
             <properties>
-                <maven-surefire-plugin.argLine>--illegal-access=deny</maven-surefire-plugin.argLine>
+                <surefire.jvm.args>--illegal-access=deny</surefire.jvm.args>
```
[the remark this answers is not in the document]
gsmet: I improved the situation slightly.

pom.xml
gsmet: Yeah, I had an hesitation here. Looks like you had the same so let's add `wildfly`.

pom.xml
gsmet: Done.

pom.xml
gsmet: Done.

pom.xml
gsmet: Done.

gsmet:
@yrodiere force-pushed an update that should address your concerns. The changes are in a fixup commit.

modules/pom.xml
```diff
@@ -314,7 +315,7 @@
                 <jdk>[9,)</jdk>
             </activation>
             <properties>
-                <patch-gen.system.args>--add-opens=java.base/java.lang=ALL-UNNAMED</patch-gen.system.args>
+                <patch-gen.jvm.args.add-opens>--add-opens=java.base/java.lang=ALL-UNNAMED</patch-gen.jvm.args.add-opens>
```
yrodiere: Looks like this is never used? You're probably missing a default value for `patch-gen.jvm.args`.

gsmet:
Merged! Thanks for the review.

PR #965, opened 2018-06-06
HV-1624 Remove the StaticFieldELResolver

gsmet:
* https://hibernate.atlassian.net/browse/HV-1624

### DIFF
--- a/documentation/src/main/asciidoc/ch06.asciidoc
+++ b/documentation/src/main/asciidoc/ch06.asciidoc
@@ -173,6 +173,19 @@ It is important to add each configured constraint violation by calling `addConst
 Only after that the new constraint violation will be created.
 ====
 
+[WARNING]
+====
+Note that the custom message template is passed directly to the Expression Language engine.
+
+Thus, you should be very careful when integrating user input in a custom message template as it will be interpreted
+by the Expression Language engine, which is usually not the behavior you expect and could cause a sensitive data leak.
+
+If you have this requirement, you should:
+
+ * either escape the user input by using the http://beanvalidation.org/2.0/spec/#validationapi-message-defaultmessageinterpolation[Bean Validation message interpolation escaping rules];
+ * or, even better, <<section-hibernateconstraintvalidatorcontext,pass the user input as message parameters or expression variables>> by unwrapping the context to `HibernateConstraintValidatorContext`.
+====
+
 Refer to <<section-custom-property-paths>> to learn how to use the `ConstraintValidatorContext` API to
 control the property path of constraint violations for class-level constraints.
 
--- a/documentation/src/main/asciidoc/ch12.asciidoc
+++ b/documentation/src/main/asciidoc/ch12.asciidoc
@@ -456,7 +456,8 @@ custom extensions for both of these interfaces.
 `HibernateConstraintValidatorContext` is a subtype of `ConstraintValidatorContext` which allows you to:
 
 * set arbitrary parameters for interpolation via the Expression Language message interpolation
-facility using `HibernateConstraintValidatorContext#addExpressionVariable(String, Object)`.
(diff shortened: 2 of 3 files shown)


gsmet:
@gunnarmorling this is what we talked about yesterday.

documentation/src/main/asciidoc/ch06.asciidoc
```diff
@@ -173,6 +173,19 @@ It is important to add each configured constraint violation by calling `addConst
 Only after that the new constraint violation will be created.
 ====
 
+[WARNING]
+====
+Note that the custom message template is passed directly to the Expression Language engine.
+
+Thus, you should be very careful when integrating user input in a custom message template as it will be interpreted
+by the Expression Language engine, which is usually not the behavior you expect and could cause a sensitive data leak.
```
yrodiere: My first thought reading this was "what is a sensitive leak?". I would use "leak of sensitive data" to avoid the confusion, or "could cause sensitive data to leak", or even "could allow malicious users to leak sensitive data", but it's your call.

documentation/src/main/asciidoc/ch06.asciidoc
```diff
@@ -173,6 +173,19 @@ It is important to add each configured constraint violation by calling `addConst
...
+
+If you have this requirement, you should:
```
yrodiere: Maybe "If you need to integrate user input"? I wasn't clear whether the requirement you were mentioning was "using custom message templates" or "allowing user input in custom message templates".

documentation/src/main/asciidoc/ch12.asciidoc
```diff
@@ -456,7 +456,8 @@ custom extensions for both of these interfaces.
 `HibernateConstraintValidatorContext` is a subtype of `ConstraintValidatorContext` which allows you to:
 
 * set arbitrary parameters for interpolation via the Expression Language message interpolation
-facility using `HibernateConstraintValidatorContext#addExpressionVariable(String, Object)`.
+facility using `HibernateConstraintValidatorContext#addExpressionVariable(String, Object)`
+or `HibernateConstraintValidatorContext#addMessageParameter(String, Object)`.
```
yrodiere: I didn't check the example, but do you explain somewhere what the difference between those two methods is?

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/SimpleELContext.java
```diff
@@ -25,7 +24,6 @@
 	private static final ELResolver DEFAULT_RESOLVER = new CompositeELResolver() {
 		{
 			add( new RootResolver() );
-			add( new StaticFieldELResolver() );
```
yrodiere: I understand this was the purpose of this PR, but it seems odd that a feature is removed, and no documentation is removed. Was this undocumented?

gsmet: Yes, it's related to the javax.el upgrade: it's a new feature in javax.el so it was exposed as we expose all the others. We don't detail what's possible with javax.el though.

documentation/src/main/asciidoc/ch06.asciidoc
gsmet: Your second proposal is very nice. Changed, thanks!

documentation/src/main/asciidoc/ch06.asciidoc
gsmet: Right, changed.

documentation/src/main/asciidoc/ch12.asciidoc
gsmet: Not really, I added more information.

gsmet:
Applied the changes and merged.

PR #966, opened 2018-06-08
HV-1628 Configure a stricter forbidden-apis policy and remove calls deprecated in Java 10

gsmet:
* https://hibernate.atlassian.net/browse/HV-1628

### DIFF
--- a/annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/Configuration.java
+++ b/annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/Configuration.java
@@ -6,8 +6,9 @@
  */
 package org.hibernate.validator.ap.internal.util;
 
-import java.text.MessageFormat;
+import java.util.Locale;
 import java.util.Map;
+
 import javax.annotation.processing.Messager;
 import javax.tools.Diagnostic.Kind;
 
@@ -97,11 +98,12 @@ public class Configuration {
 			}
 			catch (IllegalArgumentException e) {
 				messager.printMessage(
-						Kind.WARNING, MessageFormat.format(
-						"The given value {0} is no valid diagnostic kind. {1} will be used.",
-						diagnosticKindFromOptions,
-						DEFAULT_DIAGNOSTIC_KIND
-				)
+						Kind.WARNING, String.format(
+								Locale.ROOT,
+								"The given value %1$s is no valid diagnostic kind. %2$s will be used.",
+								diagnosticKindFromOptions,
+								DEFAULT_DIAGNOSTIC_KIND
+						)
 				);
 			}
 		}
@@ -118,10 +120,11 @@ public class Configuration {
 
 		if ( theValue ) {
 			messager.printMessage(
-					Kind.NOTE, MessageFormat.format(
-					"Verbose reporting is activated. Some processing information will be displayed using diagnostic kind {0}.",
-					Kind.NOTE
-			)
+					Kind.NOTE, String.format(
+							Locale.ROOT,
+							"Verbose reporting is activated. Some processing information will be displayed using diagnostic kind %1$s.",
+							Kind.NOTE
+					)
 			);
 		}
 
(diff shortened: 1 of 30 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/FieldCascadable.java
```diff
@@ -85,14 +84,9 @@ public FieldCascadable build() {
 		}
 
 		/**
-		 * Returns an accessible version of the given member. Will be the given member itself in case it is accessible,
-		 * otherwise a copy which is set accessible.
+		 * Returns an accessible copy of the given member.
 		 */
 		private Field getAccessible(Field original) {
-			if ( ( (AccessibleObject) original ).isAccessible() ) {
```
gunnarmorling: Hum, hum, a pity to see that go. Isn't there some new way of doing this, e.g. via `canAccess()`? It's a generally interesting question for our projects.

gsmet: As explained in the commit message, this call is not useful. Keep in mind that `isAccessible()` does not check if the method is accessible. It checks that the security permissions have previously been overridden using `setAccessible()`. That's a common misconception.

It's not the first time I'm tempted to remove these calls as they are useless, considering we call `setAccessible()` only once per element and we do that on a copy anyway.

As for calling `canAccess()`, I started to go the MR jar way and try to see how we could do that but, AFAICS, to use `canAccess()` on a method or a field, you need an instance of the object, which we don't have in this part of the code.

At some point, we might want to take that path but I'm pretty sure it's not going to be pretty and it's an orthogonal issue.

This commit only removes a useless call.

marko-bekhta: Just wanted to add that most likely this would go away later, when we switch to using handles. I remember when I was experimenting with the handles I also removed this `isAccessible` check, as it was causing problems.

annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/Configuration.java
```diff
@@ -97,11 +98,12 @@ private Kind getDiagnosticKindOption(Map<String, String> options, Messager messa
 			}
 			catch (IllegalArgumentException e) {
 				messager.printMessage(
-						Kind.WARNING, MessageFormat.format(
-						"The given value {0} is no valid diagnostic kind. {1} will be used.",
-						diagnosticKindFromOptions,
-						DEFAULT_DIAGNOSTIC_KIND
-				)
+						Kind.WARNING, String.format(
+								Locale.ROOT,
```
marko-bekhta: just as an idea, would be nice to hide these `Locale.ROOT` in some `StringHelper` util method. But as they are scattered all over the modules I cannot see a nice place for such method.

pom.xml
```diff
@@ -675,10 +676,6 @@
                             <phase>verify</phase>
                             <configuration>
                                 <bundledSignatures>
-                                    <!--
-                                      this will automatically choose the right
-                                      signatures based on 'maven.compiler.target'
-                                    -->
                                     <bundledSignature>jdk-deprecated</bundledSignature>
```
marko-bekhta: I cannot remember why we would allow system.out or things from other bundledSignature files in tests...Was there a specific reason for that, or should we make the rules stricter here as well ?

gunnarmorling:
> Was there a specific reason for that

Personally I dislike these kind of checks to be too strict, often it just
gets into your way. E.g. my build would fail when I temporarily added a
sysout call for debugging purposes. I think the frustration caused by this
sort of thing out-weighs the "risk" of accidentally committing such thing
to test code (can't remember it happening anyways).

Of course YMMV.

gsmet:
Agreed with Gunnar about the test code. No need to be too strict there.

annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/Configuration.java
gsmet: Yeah, for now, I don't think it's worth it. We don't have thousands of calls and we don't add them regularly, mostly thanks to JBoss Logging.

gunnarmorling:
Ok, thanks for the explanation. Makes perfect sense!

gsmet:
Rebased and merged, then.

PR #967, opened 2018-06-08
HV-1622 Make WildFly 13 our default WildFly target

gsmet:
* https://hibernate.atlassian.net/browse/HV-1622


gunnarmorling:
LGTM, merged.

PR #968, opened 2018-06-11
HV-1363 Support for non-standard Java beans

marko-bekhta:

Here we go again 😃a squashed, rebased and a bit more cleaned up version of a getter selection strategy PR.

@gsmet if you will not like the current behavior of `JavaBean`, the alternative would be to pass around the `GetterPropertySelectionStrategy` and have a factory method in `JavaBeanExecutable` to create correct objects for us.


engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
  * @author Marko Bekhta
  */
 public class JavaBean implements ConstrainableType {
 
-	private final Class<?> clazz;
+	private final String name;
+	private final List<JavaBeanField> beanFields;
+	private final List<JavaBeanMethod> beanMethods;
+	private final List<JavaBeanGetter> beanGetters;
+	private final List<JavaBeanConstructor> beanConstructors;
```
gsmet: So... this is the big issue I have with this patch. At one point, you were talking about using `ConstrainableType` as a vehicle for the class in various parts of the code (with the idea to replace `getDeclaringClass()`). From what I see of the `getDeclaringClass()` usage, we would have to keep it after the bootstrapping of the metadata. And I'm not sure I want to keep all these references. Keep in mind that we're not only talking about the constrained elements here but all the elements of a bean.

So I have a couple of comments:
- I think these elements should not be located here but rather in an object that has a limited lifespan limited in scope to the metadata discovery;
- The usage pattern for the XML and the programmatic API is very suboptimal. Now that I think about it, I have probably the answer of why the strategy is so simplified now, it's due to the removal of the reflection calls.

I think I would completely separate it from the `properties` hierarchy (I think we will need a `JavaBean` there but a simpler one). Maybe you can call this one `BeanMetaDataDiscoverer`? I wouldn't implement any interface as it's just something we will want to use as is.

I also think that we should use ``Map``s instead of ``List``s to avoid iterating again and again on the elements. Is there a reason why you didn't use ``Map``s?

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+		}
+	}
+
+	private boolean staticAndSyntheticFilter(Member member) {
+		// HV-172; ignoring synthetic members (inserted by the compiler), as they can't have any constraints
+		// anyway and possibly hide the actual member with the same signature in the built meta model
+		return Modifier.isStatic( member.getModifiers() ) || member.isSynthetic();
+	}
+
+	public Optional<JavaBeanMethod> getCallableByNameAndParameters(String methodName, Class<?>... parameterTypes) {
```
gsmet: Let's simplify all this. Here it should be `getDeclaredMethod()` (`Getter` is a specific case of `Method`).

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+			if ( beanMethod.getName().equals( methodName ) && Arrays.equals( beanMethod.getParameterTypes(), parameterTypes ) ) {
+				return Optional.of( beanMethod );
+			}
+		}
+
+		for ( JavaBeanGetter beanGetter : beanGetters ) {
+			if ( beanGetter.getName().equals( methodName ) ) {
+				return Optional.of( beanGetter );
+			}
+		}
```
gsmet: There is a logic error here: you have to check there are no `parameterTypes` before checking for the getters.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+		return Optional.empty();
+	}
+
+	public Optional<JavaBeanConstructor> getConstructorByParameters(Class<?>... parameterTypes) {
```
gsmet: `getDeclaredConstructor()`.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+		for ( JavaBeanConstructor beanConstructor : beanConstructors ) {
+			if ( Arrays.equals( beanConstructor.getParameterTypes(), parameterTypes ) ) {
+				return Optional.of( beanConstructor );
+			}
+		}
+		return Optional.empty();
+	}
+
+	public Optional<JavaBeanGetter> getCallablePropertyByName(String propertyName) {
```
gsmet: `getDeclaredGetter()`

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+		for ( JavaBeanGetter beanGetter : beanGetters ) {
+			if ( beanGetter.getPropertyName().equals( propertyName ) ) {
+				return Optional.of( beanGetter );
+			}
+		}
+		return Optional.empty();
+	}
+
+	public Optional<JavaBeanField> getFieldPropertyByName(String propertyName) {
```
gsmet: `getDeclaredField()`

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+		for ( JavaBeanField beanField : beanFields ) {
+			if ( beanField.getPropertyName().equals( propertyName ) ) {
+				return Optional.of( beanField );
+			}
+		}
+		return Optional.empty();
+	}
+
+	public List<JavaBeanField> getBeanFields() {
```
gsmet: `getDeclaredFields()`

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+		return beanFields;
+	}
+
+	public List<JavaBeanMethod> getBeanMethods() {
```
gsmet: `getDeclaredMethods()`

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+		return beanMethods;
+	}
 
-	public JavaBean(Class<?> clazz) {
-		this.clazz = clazz;
+	public List<JavaBeanGetter> getBeanGetters() {
```
gsmet: `getGetters()`

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+		return beanGetters;
 	}
 
-	public Stream<Property> getFieldProperties() {
-		return Arrays.stream( clazz.getDeclaredFields() )
-				.map( JavaBeanField::new );
+	public List<JavaBeanConstructor> getBeanConstructors() {
```
gsmet: `getDeclaredConstructors()`

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
```diff
@@ -6,32 +6,210 @@
...
+	private final List<JavaBeanMethod> beanMethods;
+	private final List<JavaBeanGetter> beanGetters;
+	private final List<JavaBeanConstructor> beanConstructors;
+
+	public JavaBean(GetterPropertySelectionStrategy getterPropertySelectionStrategy, Class<?> clazz) {
+		this( getterPropertySelectionStrategy, clazz, false );
+	}
+
+	public JavaBean(GetterPropertySelectionStrategy getterPropertySelectionStrategy, Class<?> clazz, boolean includeNonDeclaredMethods) {
+		this.name = clazz.getName();
```
gsmet: Maybe `includeMethodsFromHierarchy` would be better?

Even in this case, I think I would separate them in another `Map` and have a different method to access them (a method without the `Declared` part as in the reflection API).

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetDeclaredFieldValueMethodHandle.java
```diff
@@ -0,0 +1,61 @@
...
+import org.hibernate.validator.internal.util.logging.Log;
+import org.hibernate.validator.internal.util.logging.LoggerFactory;
+
+/**
+ * Returns the {@link MethodHandle} getter for declared field value with the specified name,
+ * throws an exception if it does not exist or cannot be accessed.
+ *
+ * @author Marko Bekhta
+ */
+public final class GetDeclaredFieldValueMethodHandle implements PrivilegedAction<MethodHandle> {
```
gsmet: I think this is dead code.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetJavaBeanExecutableProperty.java
```diff
@@ -0,0 +1,59 @@
...
+import java.security.PrivilegedAction;
+
+import org.hibernate.validator.internal.properties.javabean.JavaBeanExecutable;
+
+/**
+ * Returns the {@link Executable} property from the {@link JavaBeanExecutable} instance.
+ *
+ * @author Marko Bekhta
+ */
+public final class GetJavaBeanExecutableProperty implements PrivilegedAction<Executable> {
```
gsmet: I think this is dead code.

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetJavaBeanFieldProperty.java
```diff
@@ -0,0 +1,59 @@
...
+import java.security.PrivilegedAction;
+
+import org.hibernate.validator.internal.properties.javabean.JavaBeanField;
+
+/**
+ * Returns the {@link Field} property from the {@link JavaBeanField} instance.
+ *
+ * @author Marko Bekhta
+ */
+public final class GetJavaBeanFieldProperty implements PrivilegedAction<Field> {
```
gsmet: I think this is dead code.

engine/src/main/java/org/hibernate/validator/internal/xml/mapping/BeanStaxBuilder.java
```diff
@@ -174,25 +179,25 @@ void build(Set<Class<?>> processedClasses, Map<Class<?>, Set<ConstrainedElement>
...
 				constrainedElementsByType,
 				beanClass,
 				constrainedMethodStaxBuilders.stream()
-						.map( builder -> builder.build( beanClass, alreadyProcessedMethods ) )
+						.map( builder -> builder.build( javaBean, alreadyProcessedMethods ) )
 						.collect( Collectors.toList() )
 		);
 
-		List<Constructor<?>> alreadyProcessedConstructors = new ArrayList<>( constrainedConstructorStaxBuilders.size() );
+		List<Callable> alreadyProcessedConstructors = new ArrayList<>( constrainedConstructorStaxBuilders.size() );
```
gsmet: I think we could talk about `JavaBeanConstructor` here? Or can't we?

Same above about using `JavaBeanMethod`.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBean.java
gsmet: Ah and I would keep the boolean handy and throw an `IllegalStateException` if the method is called when it being false.

gsmet: Maybe hold on on this particular change as we might not need this feature if we decide to not use this thing for XML metadata.

gsmet:
OK, I thought about it a bit more and I think for now I would lean towards the conservative approach and keep the reflection calls for XML and the programmatic API.

It means one more method in our SPI, and it's a bit unfortunate, but I think it's still better than getting all the methods again and again.

AFAICS, it also removes the requirement for having maps in the `BeanMetaDataDiscoverer` (or whatever we name it) so you can discard the comments I made related to that.

cdi/src/main/java/org/hibernate/validator/cdi/internal/util/GetterPropertySelectionStrategyHelper.java
```diff
@@ -0,0 +1,75 @@
...
+import org.hibernate.validator.internal.properties.DefaultGetterPropertySelectionStrategy;
+import org.hibernate.validator.spi.properties.ConstrainableExecutable;
+import org.hibernate.validator.spi.properties.GetterPropertySelectionStrategy;
+
+/**
+ * A wrapper around {@link GetterPropertySelectionStrategy}.
+ *
+ * @author Marko Bekhta
+ */
+public class GetterPropertySelectionStrategyHelper {
```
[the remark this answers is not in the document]
marko-bekhta: Didn't want to put all the content of this class into `ValidationExtension`. But I'm open to suggestions :)

cdi/src/main/java/org/hibernate/validator/cdi/internal/util/GetterPropertySelectionStrategyHelper.java
```diff
@@ -0,0 +1,74 @@
...
+
+	private final GetterPropertySelectionStrategy getterPropertySelectionStrategy;
+
+	private GetterPropertySelectionStrategyHelper(GetterPropertySelectionStrategy getterPropertySelectionStrategy) {
+		this.getterPropertySelectionStrategy = getterPropertySelectionStrategy;
+	}
+
+	public boolean isProperty(Method method) {
```
gsmet: Ah sorry, forgot about this one: let's be consistent and call it `isGetter()`.

cdi/src/main/java/org/hibernate/validator/cdi/internal/util/GetterPropertySelectionStrategyHelper.java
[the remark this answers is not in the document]
gsmet: OK. I can live with it :).

engine/src/main/java/org/hibernate/validator/internal/engine/ConfigurationImpl.java
```diff
@@ -304,7 +315,17 @@ public MethodValidationConfiguration getMethodValidationConfiguration() {
 
 	@Override
 	public final DefaultConstraintMapping createConstraintMapping() {
-		return new DefaultConstraintMapping();
+		GetterPropertySelectionStrategy getterPropertySelectionStrategyToUse = null;
+		if ( getterPropertySelectionStrategy == null ) {
+			getterPropertySelectionStrategyToUse = new DefaultGetterPropertySelectionStrategy();
+			LOG.creatingConstraintMappingPriorToBuildingFactoryWithoutDefinedFilter();
```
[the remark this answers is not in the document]
gsmet: Let it be part of the documentation and make the documentation examples clear about it.

If we have users complaining, we will act on it.

engine/src/main/java/org/hibernate/validator/internal/properties/Constrainable.java
```diff
@@ -11,7 +11,7 @@
 import org.hibernate.validator.internal.metadata.raw.ConstrainedElement.ConstrainedElementKind;
 
 /**
- * An element on which it is possible to define constraints (e.g. a JavaBean property, a JavaBean method, a JSON
+ * An element on which it is possible to define constraints (e.g. a JavaBeanHelper property, a JavaBeanHelper method, a JSON
```
gsmet: Cough cough, I don't think you wanted to change that :).

engine/src/main/java/org/hibernate/validator/internal/properties/ConstrainableType.java
```diff
@@ -7,10 +7,10 @@
 package org.hibernate.validator.internal.properties;
 
 /**
- * A type that can be constrained. Will represent a Java class for a JavaBean or a named type for JSON.
+ * A type that can be constrained. Will represent a Java class for a JavaBeanHelper or a named type for JSON.
```
gsmet: Same here :).

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanGetter.java
```diff
@@ -33,16 +34,12 @@
 	 */
 	private final Class<?> declaringClass;
 
-	public JavaBeanGetter(Method method) {
+	public JavaBeanGetter(Class<?> declaringClass, Method method, String name) {
```
gsmet: Let's call the parameter `propertyName`.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -0,0 +1,162 @@
...
+			return Optional.of( new JavaBeanConstructor( constructor ) );
+		}
+	}
+
+	public JavaBeanExecutable<?> executable(Executable executable) {
+		if ( executable instanceof Constructor ) {
+			return new JavaBeanConstructor( (Constructor<?>) executable );
+		}
+
+		return executable( executable.getDeclaringClass(), ( (Method) executable ) );
```
gsmet: Can't you just call `executable(Class<?> declaringClass, Executable executable)` and remove the `if()`?

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -0,0 +1,162 @@
...
+		Contracts.assertNotNull( declaringClass, MESSAGES.classCannotBeNull() );
+
+		Field field = run( GetDeclaredField.action( declaringClass, property ) );
+		return Optional.ofNullable( field ).map( JavaBeanField::new );
+	}
+
+	public Optional<JavaBeanGetter> findDeclaredGetterProperty(Class<?> declaringClass, String property) {
+		Contracts.assertNotNull( declaringClass, MESSAGES.classCannotBeNull() );
+		for ( String possibleName : getterPropertySelectionStrategy.getPossibleMethodNamesForGetterProperty( property ) ) {
+			Method getter = run( GetDeclaredMethod.action( declaringClass, possibleName ) );
```
gsmet: I think it would be better to only call one secured method, similar to what we had before. The big difference would be you would pass the list of candidates to the action.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -0,0 +1,162 @@
...
+			if ( getter != null ) {
+				return Optional.of( new JavaBeanGetter( declaringClass, getter, getterPropertySelectionStrategy.getPropertyName( new JavaBeanConstrainableExecutable( getter ) ) ) );
```
gsmet: Can't we try to get our `Constrainable` executables implement `ConstrainableExecutable`? That would avoid creating one more object.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -0,0 +1,162 @@
...
+
+		if ( declaredGetterProperty.isPresent() ) {
+			return declaredGetterProperty;
+		}
+		else {
+			// look for getters from the superclasses and interfaces
+			for ( String possibleName : getterPropertySelectionStrategy.getPossibleMethodNamesForGetterProperty( property ) ) {
+				Method getter = run( GetMethod.action( declaringClass, possibleName ) );
+				if ( getter != null ) {
+					return Optional.of( new JavaBeanGetter( declaringClass, getter, getterPropertySelectionStrategy.getPropertyName( new JavaBeanConstrainableExecutable( getter ) ) ) );
```
gsmet: Obviously same remark as above.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -0,0 +1,162 @@
...
+				}
+			}
+			return Optional.empty();
+		}
+	}
+
+	public Optional<JavaBeanMethod> findDeclaredMethodByNameAndParameters(Class<?> declaringClass, String methodName, Class<?>... parameterTypes) {
```
gsmet: Just remove the `ByNameAndParameters`. It doesn't bring anything considering the parameters are very clear (and you only have one method anyway).

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -0,0 +1,162 @@
...
+		Method method = run( GetDeclaredMethod.action( declaringClass, methodName, parameterTypes ) );
+		if ( method == null ) {
+			return Optional.empty();
+		}
+		else {
+			return Optional.of( executable( declaringClass, method ) );
+		}
+	}
+
+	public <T> Optional<JavaBeanConstructor> findDeclaredConstructorByParameters(Class<T> declaringClass, Class<?>... parameterTypes) {
```
gsmet: Same here about `ByParameters`.

engine/src/main/java/org/hibernate/validator/spi/properties/GetterPropertySelectionStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+package org.hibernate.validator.spi.properties;
+
+import java.util.Set;
+
+import org.hibernate.validator.Incubating;
+import org.hibernate.validator.internal.properties.DefaultGetterPropertySelectionStrategy;
+
+/**
+ * <p>
```
gsmet: No `<p>` is needed here: it's the first sentence, `<p>` is just used to separate paragraphs.

engine/src/main/java/org/hibernate/validator/spi/properties/GetterPropertySelectionStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+ * Used to define JavaBeans property detection algorithm. The default
+ * implementation ({@link DefaultGetterPropertySelectionStrategy}) uses next definition
+ * of JavaBeans property:
+ * </p>
```
gsmet: No need to close the `</p>`.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -0,0 +1,162 @@
...
+import org.hibernate.validator.internal.util.Contracts;
+import org.hibernate.validator.internal.util.privilegedactions.GetDeclaredConstructor;
+import org.hibernate.validator.internal.util.privilegedactions.GetDeclaredField;
+import org.hibernate.validator.internal.util.privilegedactions.GetDeclaredMethod;
+import org.hibernate.validator.internal.util.privilegedactions.GetMethod;
+import org.hibernate.validator.spi.properties.ConstrainableExecutable;
+import org.hibernate.validator.spi.properties.GetterPropertySelectionStrategy;
+
+/**
+ * {@link ConstrainableType} that describes a JavaBeanHelper.
```
gsmet: The javadoc needs an update, it's not a `ConstrainableType` :)

engine/src/main/java/org/hibernate/validator/internal/cfg/context/TypeConstraintMappingContextImpl.java
```diff
@@ -75,8 +71,9 @@
 	private List<Class<?>> defaultGroupSequence;
 	private Class<? extends DefaultGroupSequenceProvider<? super C>> defaultGroupSequenceProviderClass;
 
-	TypeConstraintMappingContextImpl(DefaultConstraintMapping mapping, Class<C> beanClass) {
+	TypeConstraintMappingContextImpl(DefaultConstraintMapping mapping, Class<C> beanClass, GetterPropertySelectionStrategy getterPropertySelectionStrategy) {
 		super( mapping );
+		this.javaBeanHelper = new JavaBeanHelper( getterPropertySelectionStrategy );
```
gsmet: Just a question, not sure it's easy but could we create only one and share it?

engine/src/main/java/org/hibernate/validator/internal/properties/DefaultGetterPropertySelectionStrategy.java
```diff
@@ -0,0 +1,99 @@
...
+		for ( String prefix : GETTER_PREFIXES ) {
+			if ( methodName.startsWith( prefix ) ) {
+				return StringHelper.decapitalize( methodName.substring( prefix.length() ) );
+			}
+		}
+		throw LOG.getIllegalArgumentException( "Property name cannot be constructed for a given method " + executable.getName() );
+	}
+
+	@Override
+	public Set<String> getPossibleMethodNamesForGetterProperty(String propertyName) {
```
gsmet: Maybe `getGetterNameCandidates()`?

If so, let's make the core of the method consistent with the new name (e.g. variable names).

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -0,0 +1,162 @@
...
+ */
+public class JavaBeanHelper {
+
+	private final GetterPropertySelectionStrategy getterPropertySelectionStrategy;
+
+	public JavaBeanHelper(GetterPropertySelectionStrategy getterPropertySelectionStrategy) {
+		this.getterPropertySelectionStrategy = getterPropertySelectionStrategy;
+	}
+
+	public Optional<JavaBeanField> findDeclaredFieldProperty(Class<?> declaringClass, String property) {
```
gsmet: `findDeclaredField()`?

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -0,0 +1,162 @@
...
+		Contracts.assertNotNull( declaringClass, MESSAGES.classCannotBeNull() );
+
+		Field field = run( GetDeclaredField.action( declaringClass, property ) );
+		return Optional.ofNullable( field ).map( JavaBeanField::new );
+	}
+
+	public Optional<JavaBeanGetter> findDeclaredGetterProperty(Class<?> declaringClass, String property) {
```
gsmet: `findDeclaredGetter()`?

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -864,4 +863,22 @@ ConstraintDefinitionException getConstraintValidatorDefinitionConstraintMismatch
...
+	@LogMessage(level = INFO)
+	@Message(id = 246, value = "Using %s as getter property selection strategy.")
+	void usingGetterPropertySelectionStrategy(@FormatWith(ClassObjectFormatter.class) Class<? extends GetterPropertySelectionStrategy> getterPropertySelectionStrategyClass);
+
+	@Message(id = 247, value = "Unable to instantiate getter property selection strategy class %s.")
+	ValidationException getUnableToInstantiateGetterPropertySelectionStrategyClassException(String getterPropertySelectionStrategyClassName, @Cause Exception e);
+
+	@LogMessage(level = WARN)
+	@Message(id = 248, value = "Note that constraint mappings will be build with %s getter property selection strategy.")
+	void creatingConstraintMappingPriorToBuildingFactory(@FormatWith(ClassObjectFormatter.class) Class<? extends GetterPropertySelectionStrategy> getterPropertySelectionStrategyClass);
```
gsmet: So I think this one will go away but if not s/build/built/ in the message.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/TypeConstraintMappingContextImpl.java
marko-bekhta: moved the construction of the helper to `ValidatorFactoryImpl` but it is still also created in `ConfigurationImpl`

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
marko-bekhta: That's exactly what I was thinking initially. But as we need to know if a method is a getter or not to create a `Constrainable` we cannot use the same thing for the strategy methods. I also was thinking that maybe we should just use the plain reflection `Executable` for the `GetterPropertySelectionStrategy`, but I would want to expose more APIs tied to reflection directly....

documentation/src/main/asciidoc/ch12.asciidoc
```diff
@@ -677,35 +677,41 @@ validator factory still is referenced by application code.
...
-- a method name starts with `get`, it has a nonvoid return type and has no parameters
-- a method name starts with `is`, has a return type of `boolean` and has no parameters
-- a method name starts with `has`, has a return type of `boolean` and has no parameters (this
+- the method name starts with `get`, it has a non-void return type and has no parameters;
+- the method name starts with `is`, has a return type of `boolean` and has no parameters;
+- the method name starts with `has`, has a return type of `boolean` and has no parameters (this
 is a Hibernate Validator specific rule and is not mandated by the JavaBeans specification)
 
-If any of the statements above are true - method is considered a getter.
+While there rules are usually appropriate when following the classic JavaBeans convention, it might happen,
```
marko-bekhta: there -> these ?

documentation/src/main/asciidoc/ch12.asciidoc
```diff
@@ -677,35 +677,41 @@ validator factory still is referenced by application code.
...
+especially with code generators, that the JavaBeans naming convention is not followed and that the getters'
+name are following a different convention.
 
-When working with objects that do not follow JavaBean specification rules on getters. These rules
-should either be redefined or extended, in order to be able to fully validate the object. An example
-of such objects can be the ones constructed from classes that follow fluent design pattern (i.e. <<example-using-fluent-api-pattern>> ).
+In this case, the strategy for detecting getters should be either extended or redefined
+in order to fully validate the object.
+
+A classic example of this requirement is when the classes follows a fluent naming convention.
```
marko-bekhta: Maybe comma at the end or capitalize the first letter ('As')

gsmet: Good catch, it was supposed to be a comma!

documentation/src/main/asciidoc/ch12.asciidoc
gsmet: Fixed, thanks!

gsmet:
@marko-bekhta I made a final review of the PR and made a few fixes here and there.

There's one last thing that worries me a bit: when using the fluent strategy, we consider a **lot** of methods as possible getters such as `clone()`, `hashCode()`, `toString()`... I'm not sure it's such a good idea.

I'm wondering if we should extend `isGetter()` to also take a list of the existing fields in the class. I'm not totally sure about it as at the moment we don't check that a method with `get`, `is`... has a corresponding property. And even more, I don't think we would want to do that as we want to consider `getFoo()` as a property even if there is no corresponding field.

But with the fluent naming convention, having `toString()` has a getter doesn't seem right.

@gunnarmorling interested in your opinion about this issue.

marko-bekhta:
hmm but then in case if somebody has:
```java
class StrageThing{
    private final String toString;

    public StrageThing() {
        //other constructor stuff...
        this.tostring = "Some sort of string representation of the class";
    }

    @Override
    public String toString() {
        return tostring;
    }
}
```
the toString would be considered a getter.
I'd say we shouldn't check for fields. And also we could always do something like:
```java
public static class NoPrefixGetterPropertySelectionStrategy implements GetterPropertySelectionStrategy {

    private final Set<String> methodNamesToIgnore;

    public NoPrefixGetterPropertySelectionStrategy() {
        this.methodNamesToIgnore = Arrays.stream( Object.class.getDeclaredMethods() ).map( Method::getName )
                .collect( Collectors.toSet() );
    }

    @Override
    public boolean isGetter(ConstrainableExecutable executable) {
        String name = executable.getName();
        return executable.getReturnType() != void.class
                && executable.getParameterTypes().length == 0
                && !methodNamesToIgnore.contains( name );
    }

    @Override
    public String getPropertyName(ConstrainableExecutable method) {
        return method.getName();
    }

    @Override
    public Set<String> getGetterMethodNameCandidates(String propertyName) {
        return Collections.singleton( propertyName );
    }
}
```
where we ignore the methods we don't want to be getters. It doesn't look ideal, but .... WDYT?

engine/src/main/java/org/hibernate/validator/spi/properties/ConstrainableExecutable.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.reflect.Type;
+
+/**
+ * A class that describes JavaBeans method. Is used for determining if a given
+ * JavaBeans method is a property or not.
+ *
+ * @author Marko Bekhta
+ * @since 6.1.0
+ */
+public interface ConstrainableExecutable {
```
gunnarmorling: This should be `@Incubating`.

engine/src/main/java/org/hibernate/validator/spi/properties/ConstrainableExecutable.java
```diff
@@ -0,0 +1,34 @@
...
+
+	/**
+	 * @return the name of the method represented by this {@code ConstrainableExecutable} object
+	 */
+	String getName();
+
+	/**
+	 * @return the parameter types for the executable this object represents
+	 */
+	Type[] getParameterTypes();
```
gunnarmorling: Why is it `Type` here but `Class` for the return type?

engine/src/main/java/org/hibernate/validator/spi/properties/GetterPropertySelectionStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+package org.hibernate.validator.spi.properties;
+
+import java.util.Set;
+
+import org.hibernate.validator.Incubating;
+import org.hibernate.validator.internal.properties.DefaultGetterPropertySelectionStrategy;
+
+/**
+ * Used to define the strategy to detect the getter of a bean. The default
+ * implementation ({@link DefaultGetterPropertySelectionStrategy}) uses the JavaBeans
```
gunnarmorling: I wouldn't link to an implementation class, also if it's just JavaDoc.

engine/src/main/java/org/hibernate/validator/spi/properties/ConstrainableExecutable.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.spi.properties;
+
+import java.lang.reflect.Type;
+
+/**
+ * A class that describes JavaBeans method. Is used for determining if a given
```
gunnarmorling: It seems odd to explicitly mention "JavaBeans" here. Isn't the point of it exactly to go beyond the notion of JavaBeans? "Descriptor for a method of a Java class" might be a better alternative.

engine/src/main/java/org/hibernate/validator/spi/properties/GetterPropertySelectionStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * Gives a set of possible method names based on a property name. Usually, it means
+	 * a property name prefixed with something like "get", "is", "has" etc.
+	 *
+	 * @param propertyName a property name
+	 *
+	 * @return the {@link Set} of possible getter names
+	 */
+	Set<String> getGetterMethodNameCandidates(String propertyName);
```
gunnarmorling: How about just `Set<String> getGetterMethodPrefixes()` and calling it only once instead of individually for all properties?

engine/src/main/java/org/hibernate/validator/spi/properties/GetterPropertySelectionStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+public interface GetterPropertySelectionStrategy {
+
+	/**
+	 * Determines if a given {@link ConstrainableExecutable} is a getter method (property).
+	 *
+	 * @param executable a {@link ConstrainableExecutable}
+	 *
+	 * @return {@code true} if the given executable should be considered a getter, {@code false} otherwise
+	 */
+	boolean isGetter(ConstrainableExecutable executable);
```
gunnarmorling: No strong preference, but you could consider to bake the two methods into one:

    Optional<PropertyDescriptor> getPropertyDescriptor(ConstrainableExecutable executable);

    class PropertyDescriptor { name; ... }

gsmet:
@marko-bekhta your `methodNamesToIgnore` method is interesting. We talked with Gunnar about ignoring the class if the declaring class was Object.class but that would indeed not solve everything and it requires adding `getDeclaringClass()` to `ConstrainableExecutable` which is not ideal.

I integrated it in the example.

About the marker annotation, why not, but I think it's the responsibility of the developer to do so if he wants to do it like that. I don't feel comfortable providing an annotation in HV.

engine/src/main/java/org/hibernate/validator/spi/properties/ConstrainableExecutable.java
gsmet: Good catch, done.

engine/src/main/java/org/hibernate/validator/spi/properties/ConstrainableExecutable.java
gsmet: I thought it came from the reflection API but, indeed, both returns classes in the reflection API. I changed that.

engine/src/main/java/org/hibernate/validator/spi/properties/GetterPropertySelectionStrategy.java
gsmet: OK, done.

engine/src/main/java/org/hibernate/validator/spi/properties/ConstrainableExecutable.java
gsmet: I removed a couple in one of my latest commits but apparently missed this one. Fixed.

engine/src/main/java/org/hibernate/validator/spi/properties/GetterPropertySelectionStrategy.java
gsmet: I don't think it's a good idea as you then constrain the API to only consider prefixes (and possibly the absence of prefix for the fluent case).

I think it's better to be more flexible.

marko-bekhta:
>About the marker annotation, why not, but I think it's the responsibility of the developer to do so if he wants to do it like that. I don't feel comfortable providing an annotation in HV.

To allow them do that though, we would need to at least extend ConstrainableExecutable with something like:
```java
public interface ConstrainableExecutable {
    //.....

    boolean isMarkedWithAnnotation(Class<? extends Annotation> annotationClass);
}
```
but probably we should just leave it as is and see if anybody would ask for it.

gsmet:
@marko-bekhta @gunnarmorling I pushed two additional commits. I think it's ready to be merged now. Confirmation welcome.

gsmet:
> To allow them do that though, we would need to at least extend ConstrainableExecutable with something like:
> but probably we should just leave it as is and see if anybody would ask for it.

I took a look at what Lombok does and the annotations only has a retention of source so we wouldn't be able to find them.

I'd say let's be conservative for now and we can still adjust later.

gsmet:
Merged. Thanks everyone! Nice work @marko-bekhta !

PR #969, opened 2018-06-12
HV-1630 Introduce StringHelper.format()

gsmet:
* https://hibernate.atlassian.net/browse/HV-1630

@marko-bekhta this one is specially for you.

Note that I did not change the occurrences in the annotation processor and so on as we don't have a common artifact where we can share helpers. I don't think it's worth introducing it for that.

At least, in the engine, the helper is available.

### DIFF
--- a/annotation-processor/src/test/java/org/hibernate/validator/ap/testutil/CompilerTestHelper.java
+++ b/annotation-processor/src/test/java/org/hibernate/validator/ap/testutil/CompilerTestHelper.java
@@ -29,6 +29,7 @@ import javax.tools.StandardLocation;
 import org.hibernate.validator.ap.internal.util.CollectionHelper;
 import org.hibernate.validator.ap.internal.util.Configuration;
 import org.hibernate.validator.ap.util.DiagnosticExpectation;
+import org.hibernate.validator.internal.util.StringHelper;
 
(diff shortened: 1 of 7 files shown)


annotation-processor/src/test/java/org/hibernate/validator/ap/testutil/CompilerTestHelper.java
```diff
@@ -180,16 +181,16 @@ public boolean compile(Processor annotationProcessor,
 		List<String> options = new ArrayList<String>();
 
 		if ( diagnosticKind != null ) {
-			options.add( String.format( "-A%s=%s", Configuration.DIAGNOSTIC_KIND_PROCESSOR_OPTION, diagnosticKind ) );
+			options.add( StringHelper.format( "-A%s=%s", Configuration.DIAGNOSTIC_KIND_PROCESSOR_OPTION, diagnosticKind ) );
```
marko-bekhta: Just thinking out loud.. Is it OK that we are "leaking" internal util into another module? And I've checked, we have the `StringHelper` in AP which is almost a copy of the one from the engine. I'd probably would either keep the changes to the engine only, or add that same method to AP helper. But I'll leave it for you to decide :)

gsmet:
Fixed and merged.

PR #971, opened 2018-06-13
HV-1629 Fix compiler warnings

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1629

@gsmet I was going  through the review comments of the other PR and saw the one about the missing <>. And also I saw this JIRA ticket, hence I thought why not fix all of these at one go rather than adding fixes one by one. 

`IdentitySet` had some unchecked warnings and as it is not used anywhere I just dropped it. And after that there are two commits one with fixes of unchecked warnings and another for deprecation ones.


gsmet:
@marko-bekhta looks like the build is failing!

Thanks for the PR btw, very helpful :).

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter12/constraintapi/ConstraintApiTest.java
```diff
@@ -20,6 +20,7 @@
 import org.hibernate.validator.constraintvalidators.RegexpURLValidator;
 import org.junit.Test;
 
+@SuppressWarnings("deprecation")
```
gsmet: Maybe we should use the new `getter()` and `field()` and check the documentation is up to date?

(same comment for all the deprecation tags added in API)

engine/src/main/java/org/hibernate/validator/internal/cfg/context/TypeConstraintMappingContextImpl.java
```diff
@@ -117,6 +118,7 @@
 	}
 
 	@Override
+	@SuppressWarnings("deprecation")
 	public PropertyConstraintMappingContext property(String property, ElementType elementType) {
```
gsmet: I'm surprised by this one? Maybe we should mark this method as deprecated instead?

engine/src/main/java/org/hibernate/validator/constraints/Email.java
```diff
@@ -39,7 +37,7 @@
 @Constraint(validatedBy = { })
 @Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
 @Retention(RUNTIME)
-@Repeatable(List.class)
+@Repeatable(Email.List.class)
```
gsmet: Not sure I understand this one?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ModCheckValidator.java
```diff
@@ -24,7 +22,8 @@
  * @deprecated As of release 5.1.0, replaced by {@link Mod10CheckValidator} and {@link Mod11CheckValidator}
  */
 @Deprecated
-public class ModCheckValidator extends ModCheckBase implements ConstraintValidator<ModCheck, CharSequence> {
+@SuppressWarnings("deprecation")
+public class ModCheckValidator extends ModCheckBase implements ConstraintValidator<org.hibernate.validator.constraints.ModCheck, CharSequence> {
```
gsmet: Isn't the fact that the class is marked as deprecated sufficient?

I don't like to use the FQCN as a workaround. It's not pretty at all.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
```diff
@@ -209,6 +209,7 @@ protected ConstraintViolationBuilderImpl(String template, PathImpl path) {
 		}
 
 		@Override
+		@SuppressWarnings("deprecation")
```
gsmet: What's deprecated here?

engine/src/test/java/org/hibernate/validator/test/cfg/CascadingWithConstraintMappingTest.java
```diff
@@ -31,6 +31,7 @@
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
+@SuppressWarnings("deprecation")
```
gsmet: Again, let's use our brand new constructs :).

engine/src/test/java/org/hibernate/validator/test/cfg/ConstraintMappingContributorConfiguredInValidationXmlTest.java
```diff
@@ -92,9 +91,9 @@ private void runWithCustomValidationXml(String validationXmlName, Runnable runna
 		public void createConstraintMappings(ConstraintMappingBuilder builder) {
 			builder.addConstraintMapping()
 				.type( Marathon.class )
-					.property( "name", METHOD )
+					.getter( "name" )
 						.constraint( new NotNullDef() )
-					.property( "numberOfHelpers", FIELD )
+					.field( "numberOfHelpers" )
```
gsmet: See, you can do it :).

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ModCheckValidator.java
marko-bekhta: Just a quick answer on this one - the compiler was still not happy about the imports of deprecated stuff and as a workaround I've used the fqcn. But let me double check that as I might be mixing the warnings from the IDEA with the ones from the build.

engine/src/main/java/org/hibernate/validator/constraints/Email.java
gsmet: Agreed.

gsmet:
@marko-bekhta thanks, looks very nice.

About the deprecation warnings, it's unfortunate but I think using FQCN should be avoided. So let's leave with it for now, maybe the tooling will get improved at some point.

There's one thing missing though: I think a very simple test specifically testing that `property()` still works for a getter and a field constraint would be in order. WDYT?

marko-bekhta:
@gsmet there's this test - https://github.com/hibernate/hibernate-validator/pull/971/files#diff-dab88d4bebb8bbdf4c49e52217bbede3R535 but let me change the name, as looking at `testDepreFieldAndGetterMethodsForProgrammaticConstraintDefinition` it seems I was too lazy to finish typing "Deprecated" :)

gsmet:
Ah yeah sure, sorry. I was looking for an additional test and I missed it. Totally makes sense to reuse the one added for `getter()`/`field()`.

gsmet:
Rebased and merged, thanks!

PR #972, opened 2018-06-13
HV-1608 Some work related to JDK 11

gsmet:
* https://hibernate.atlassian.net/browse/HV-1608

I wouldn't close the issue for now as not everything is settled (I really would like to use a real Maven dependency instead of this hack downloading the jars from the java.net website).

But even if not perfect, it contains some improvements that should go in.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,7 +156,7 @@
         <!-- JavaMoney dependencies -->
         <version.javax.money>1.0.1</version.javax.money>
         <version.org.javamoney.moneta>1.1</version.org.javamoney.moneta>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -1217,6 +1222,44 @@
...
+            </modules>
+        </profile>
+        <profile>
+            <id>jdk11+</id>
+            <activation>
+                <jdk>[11,)</jdk>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
```
gunnarmorling: Nice approach!

gsmet:
Merged!

PR #973, opened 2018-06-14
HV-1636 Avoid instantiating unnecessary objects during constraint metadata creation

gsmet:
* https://hibernate.atlassian.net/browse/HV-1636

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
@@ -90,8 +90,6 @@ public class AnnotationMetaDataProvider implements MetaDataProvider {
 
 	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
-	private static final Annotation[] EMPTY_PARAMETER_ANNOTATIONS = new Annotation[0];
-
 	private final ConstraintHelper constraintHelper;
 	private final TypeResolutionHelper typeResolutionHelper;
 	private final AnnotationProcessingOptions annotationProcessingOptions;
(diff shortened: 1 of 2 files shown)


gsmet:
@marko-bekhta could you please take a look at this one? Thanks!

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -196,15 +194,18 @@ public AnnotationProcessingOptions getAnnotationProcessingOptions() {
...
+		List<ConstraintDescriptorImpl<?>> classLevelConstraintDescriptors = findConstraints( null, clazz.getDeclaredAnnotations(),
+				ConstraintLocationKind.TYPE );
 
-		// HV-262
-		List<ConstraintDescriptorImpl<?>> classMetaData = findClassLevelConstraints( clazz );
+		if ( classLevelConstraintDescriptors.isEmpty() ) {
+			return Collections.emptySet();
+		}
 
+		Set<MetaConstraint<?>> classLevelConstraints = newHashSet( classLevelConstraintDescriptors.size() );
```
marko-bekhta: Nothing big, but maybe we could slightly modify the existing `convertToMetaConstraints()` for field to receive list of descriptors and a `ConstraintLocation` instead of a field. This way we could use it for class, field and parameter constraints. And the code would look even more consistent for these cases. WDYT?

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -445,36 +438,60 @@ private ConstrainedExecutable findExecutableMetaData(Executable executable) {
...
+	 * @return A list of constraint descriptors for all constraint specified for the given member.
 	 */
-	private List<ConstraintDescriptorImpl<?>> findClassLevelConstraints(Class<?> beanClass) {
+	private List<ConstraintDescriptorImpl<?>> findConstraints(Constrainable constrainable, Annotation[] annotations,
+			ConstraintLocationKind kind) {
+		if ( annotations.length == 0 ) {
+			return Collections.emptyList();
+		}
+
 		List<ConstraintDescriptorImpl<?>> metaData = newArrayList();
```
gunnarmorling: Minor style question, mostly out of curiousity: did you actively decide against using streams and map() + collect()? I guess I'd have done that if I were touch the code anyways, but I suppose you have your reasons. It's an honest question, just interested :)

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
gsmet: So, one of the thing is that I want to avoid creating locations if we don't need it. Having to pass a location would defeat that or I would have to move the empty test above in the call stack. I don't think it will help.

We could probably do things a bit differently but I think making things consistent will be easier once we really use `JavaBean` to abstract the `getDeclaringClass()` thing. We will have a `Constrainable` to pass and everything will be quite consistent (but we would still need several `convertToMetaConstraints()` methods).

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
gsmet: The main reason is that using streams usually leads to use lambdas or method references which consumes memory with an additional class loader (Sanne sent us a few emails about this).

Using streams also has its own overhead and, while I'm willing to pay for it when it makes things lots easier, I'm not when a simple for loop does the trick.

In this example https://github.com/hibernate/hibernate-validator/pull/973/files#diff-a0952ad4963831a974f6b5fa78895a6dL762 , not using a stream also helps evacuating the empty case early.

So, my rule of thumb is why not use them when they make things really really nicer but avoid them when they are not a big improvement.

So when I change a code area containing a stream, I consider it carefully and remove it if I don't think it's a clear win.

gsmet:
@marko-bekhta thanks for the review.

> I've also noticed a very minor thing - we could omit creating JavaBeanField here https://github.com/hibernate/hibernate-validator/pull/973/files#diff-a0952ad4963831a974f6b5fa78895a6dR218 for static or synthetic fields.

Done.

To be honest, I considered for a while keeping the reflection API usage for `annotationProcessingOptions` to avoid creating the `JavaBeanField` but IIRC it does not make sense for all calls and it was messy.

gsmet:
And merged, thanks for the review!

PR #974, opened 2018-06-19
HV-1637 Fix variable name in @DecimalMax error message for German and Dutch

gsmet:
* https://hibernate.atlassian.net/browse/HV-1637


gsmet:
In https://beanvalidation.org/2.0/spec/#validationapi-message-defaultmessageinterpolation-expressionlanguage :

> If an exception occurs during message interpolation, e.g. due to invalid expressions or references to an unknown property, the message expression stays unchanged.

But in any case, it's working as expected: the exception is logged at the WARN level and is not thrown. The message is indeed returned un-interpolated.

gunnarmorling:
Ah, ok. All cool then, wasn't quite clear from the OP whether the exception
would actually be raised or not. Thanks for checking!

PR #975, opened 2018-06-20
HV-1226 Explore usage of method handles instead of core reflection in ReflectionHelper#getValue()

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1226

This PR is an attempt to allow users to switch between method handles and reflection when accessing properties.


gsmet:
OK, for now, let's give up on MethodHandles. (Just so that you don't waste time rebasing this)

PR #980, opened 2018-08-09
HV-1644 Avoid having JavaFX as a transitive dependency [6.0 backport]

gsmet:
Also remove the snapshot repository we temporarily added as the
artifacts have been published to Maven Central.

 * https://hibernate.atlassian.net/browse/HV-1644


gsmet:
Merged a different version of it with additional commits based on https://github.com/hibernate/hibernate-validator/pull/978.

PR #986, opened 2018-08-22
HV-1652 Fix a few theoretical null pointer dereference issues

gsmet:
* https://hibernate.atlassian.net/browse/HV-1652


gsmet:
Merged, thanks for the review.

PR #987, opened 2018-08-29
HV-1653 Improve the wrong doc of ParameterScriptAssert.

aristotll:
The ParameterScriptAssert doc example is not working.
arg0, arg1 ,arg2, etc are not passed to the script evaluations, at least in current implementation.  

See https://github.com/hibernate/hibernate-validator/blob/60e5c17bfbe709f32018e97a483f6011093587fe/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ParameterScriptAssertValidator.java#L61-L68

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraints/ParameterScriptAssert.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/ParameterScriptAssert.java
@@ -33,7 +33,10 @@ import org.hibernate.validator.constraints.ParameterScriptAssert.List;
  * ("Scripting for the Java<sup>TM</sup> Platform") compatible engine can be
  * found on the classpath. To refer to a parameter within the scripting
  * expression, use its name as obtained by the active
- * {@link javax.validation.ParameterNameProvider}. 
+ * {@link javax.validation.ParameterNameProvider}. The default provider will
+ * return the actual parameter names, if the -parameters compiler option
+ * has been enabled, and {@code arg0}, {@code arg1} etc. otherwise
+ * will be used as parameter names.
  * </p>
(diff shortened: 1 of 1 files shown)


gsmet:
ok to test

engine/src/main/java/org/hibernate/validator/constraints/ParameterScriptAssert.java
```diff
@@ -33,15 +33,14 @@
  * ("Scripting for the Java<sup>TM</sup> Platform") compatible engine can be
  * found on the classpath. To refer to a parameter within the scripting
  * expression, use its name as obtained by the active
- * {@link javax.validation.ParameterNameProvider}. By default, {@code arg0}, {@code arg1} etc.
- * will be used as parameter names.
+ * {@link javax.validation.ParameterNameProvider}.
```
gunnarmorling: How about that:

> The default provider will return the actual parameter names, if the _-parameters_ compiler option has been enabled, and {@code arg0}, {@code arg1} etc. otherwise.

gsmet: +1 to add that.

@aristotll could you take care of it? BTW, there is a trailing whitespace at the end of one of the lines you added and it makes our Checkstyle run fail.

Thanks!

aristotll: @gsmet  Thanks you all for suggestions, could you review my changes.

gsmet:
Merged, thanks! (and sorry for the delay!)

PR #989, opened 2018-10-22
Explore FreeForm validation v2

marko-bekhta:
Link to previous discussion on the same topic - https://github.com/hibernate/hibernate-validator/pull/949

Note: This is still a work in progress.


engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaDataBuilder.java
```diff
@@ -47,6 +50,8 @@
 	private static final CascadingMetaDataBuilder NON_CASCADING =
 			new CascadingMetaDataBuilder( null, null, null, null, false, Collections.emptyMap(), Collections.emptyMap() );
 
+	private final String mapping;
```
marko-bekhta: For now property holders and regular beans are using the same cascading metadata builder. When I've tried to break it into smaller reusable pieces, I've broke it so badly :smile: that I decided to revert the changes for now and just use the same builder. But this still is needed to be done.

engine/src/main/java/org/hibernate/validator/internal/metadata/core/MetaConstraintBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+import org.hibernate.validator.internal.properties.Constrainable;
+import org.hibernate.validator.internal.util.TypeResolutionHelper;
+
+/**
+ * @author Marko Bekhta
+ */
+public class MetaConstraintBuilder<A extends Annotation> {
+
+	private final ConstraintDescriptorImpl<A> constraintDescriptor;
+	private final ConstraintLocation.Builder locationBuilder;
```
marko-bekhta: The part about location builder in this commit can be ignored. It will be removed in later commits. And I think about removing this builder as well. I think that storing the ConstraintDefs will be enough and this logic will be moved elsewhere.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -453,42 +453,60 @@ private void validateConstraintsForCurrentGroup(BaseBeanValidationContext<?> val
...
-						}
-					}
+		boolean defaultGroupSequenceIsRedefined = beanMetaData.isDefaultGroupSequenceRedefined();
+		if ( defaultGroupSequenceIsRedefined ) {
+			validateConstraintsForRedefinedGroupSequence( validationContext, valueContext, validatedInterfaces, beanMetaData.getBeanClass(), beanMetaData );
+			validationContext.markCurrentBeanAsProcessed( valueContext );
+		}
+		else {
+			validateConstraintsForSingleDefaultGroupElement( validationContext, valueContext, validatedInterfaces, beanMetaData.getBeanClass(), beanMetaData.getDirectMetaConstraints(), Group.DEFAULT_GROUP );
+			List<Class<? super U>> classHierarchy = beanMetaData.getClassHierarchy();
```
marko-bekhta: I plan to make `BeanMetaData` an abstract class with regular and property holder impls. Then in case of property holder there will be nothing else but the property holder class in this list.

Also note the change itself. Before this we were accessing a map to get the metadata that we already have. With this we remove that one additional map access.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -453,42 +453,60 @@ private void validateConstraintsForCurrentGroup(BaseBeanValidationContext<?> val
...
+
+			validationContext.markCurrentBeanAsProcessed( valueContext );
+			// we start from the second element as first one (the most specific class) was already validated ^
+
+			// evaluating the constraints of a bean per class in hierarchy, this is necessary to detect potential default group re-definitions
+			for ( int i = 1; i < classHierarchy.size(); i++ ) {
+				Class<? super U> clazz = classHierarchy.get( i );
+
+				BeanMetaData<? super U> hostingBeanMetaData = constraintMetaDataManager.getBeanMetaData( clazz );
```
marko-bekhta: As for this part `constraintMetaDataManager.getBeanMetaData( clazz );` I don't think that property holders should have anything in their "hierarchy", but if we decide that we want to have something like that we would either need to move out this entire method into the metadata impls (which doesn't feels right) or have two different `ValidartorImpls` where we override this specific method to use either property holders or simple beans.

marko-bekhta:
From the notes that I've found on this I have:
- need to make BeanMetaData abstract. There's a comment about it above. There will be two impls one for regular bean and another for property holder.
- Store simple constraint defs instead of `Set<MetaConstraintBuilder<?>> constraints` in programmatic API
- figure out how users will provide "type argument" constraints.

engine/src/main/java/org/hibernate/validator/internal/cfg/context/CascadableConstraintMappingContextImplBase.java
```diff
@@ -160,19 +160,15 @@ public boolean isCascading() {
...
-		}
-
-		return CascadingMetaDataBuilder.annotatedObject( configuredType, isCascading, typeParametersCascadingMetaData, groupConversions );
+		return CascadingMetaDataBuilder.annotatedObject(
+				configuredType,
+				isCascading,
+				containerElementContexts.values().stream()
+						.map( ContainerElementConstraintMappingContextImpl::getContainerElementCascadingMetaDataBuilder )
+						.filter( Objects::nonNull )
+						.collect( Collectors.toList() ),
```
gsmet: I think I would prefer to keep a local variable for debugging.

BTW, I'm not convinced of the values of using streams + method references here. And as we're trying to reduce them, we could simply remove it.

Another thing is I wonder if we should try to merge the changes about cascading early to get them out of the way?

It might not be totally easy to isolate them but I think it would have value.

engine/src/main/java/org/hibernate/validator/cfg/propertyholder/Cascadable.java
```diff
@@ -0,0 +1,33 @@
...
+ * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
+ */
+public interface Cascadable<C extends Cascadable<C>> {
+
+	/**
+	 * Marks the current element (property, parameter etc.) as cascadable.
+	 *
+	 * @return The current creational context following the method chaining pattern.
+	 */
+	C valid(String mapping);
```
gsmet: Same remark about using a String.

engine/src/main/java/org/hibernate/validator/cfg/propertyholder/ContainerElementConstraintMappingContext.java
```diff
@@ -0,0 +1,21 @@
...
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.cfg.propertyholder;
+
+/**
+ * Constraint mapping creational context representing a type argument of a property, parameter or method return value
+ * with a generic (return) type. Allows to place constraints on that type argument, mark it as cascadable and to
```
gsmet: This is far more restricted in the case of a property holder (e.g. no methods for instance)

engine/src/main/java/org/hibernate/validator/cfg/propertyholder/Cascadable.java
```diff
@@ -0,0 +1,33 @@
...
+package org.hibernate.validator.cfg.propertyholder;
+
+/**
+ * Facet of a constraint mapping creational context which allows to mark the underlying
+ * element as to be validated in a cascaded way.
+ *
+ * @author Gunnar Morling
+ * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
+ */
+public interface Cascadable<C extends Cascadable<C>> {
```
gsmet: This is very generic. Maybe `CascadableProperty` would be better?

engine/src/main/java/org/hibernate/validator/cfg/propertyholder/Constrainable.java
```diff
@@ -0,0 +1,27 @@
...
+import org.hibernate.validator.cfg.ConstraintDef;
+
+/**
+ * Facet of a property holder constraint mapping creational context which allows to place
+ * constraints on the underlying element.
+ *
+ * @author Gunnar Morling
+ * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
+ */
+public interface Constrainable<C extends Constrainable<C>> {
```
gsmet: Same here about it being generic. Maybe `ConstrainableProperty` would be better?

engine/src/main/java/org/hibernate/validator/internal/metadata/location/ConstraintLocation.java
```diff
@@ -104,14 +104,6 @@ static ConstraintLocation forParameter(Callable callable, int index) {
...
 	ConstraintLocationKind getKind();
 
-	interface Builder {
-		ConstraintLocation build(Constrainable constrainable);
-
-		static Builder forPropertyHolderProperty() {
-			return new PropertyHolderPropertyConstraintLocationBuilder();
-		}
-	}
-
```
gsmet: Maybe it should be removed in the commit where it has been added? Or was it useful at the time?

engine/src/main/java/org/hibernate/validator/internal/metadata/manager/ConstraintMetaDataManager.java
```diff
@@ -14,7 +14,7 @@
 import org.hibernate.validator.internal.metadata.aggregated.BeanMetaData;
 import org.hibernate.validator.internal.metadata.core.ConstraintHelper;
 import org.hibernate.validator.internal.metadata.provider.MetaDataProvider;
-import org.hibernate.validator.internal.metadata.provider.PropertyHolderMetaDataProvider;
+import org.hibernate.validator.internal.metadata.provider.proeprtyholder.PropertyHolderMetaDataProvider;
```
gsmet: There's a typo in the package name: it should be propertyholder.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
marko-bekhta: Yeah ... Let me try to explain how I got here. The main "problem" that I was trying to "solve" was to reduce the number of places where `constraintMetaDataManager.getBeanMetaData( clazz );` is called. In case of property holders we cannot use such call And that's why for example in case of cascadebles I've moved such call into metadata class. In this particular place except the call for metadata itself we also access the metadata for the "most specific" class twice.

We already have the metadata from the context here - `BeanMetaData<U> beanMetaData = valueContext.getCurrentBeanMetaData()`. But then in the loop `for ( Class<? super U> clazz : beanMetaData.getClassHierarchy() )` we would get the beans class as a first element and try to get that same metadata again - `BeanMetaData<? super U> hostingBeanMetaData = constraintMetaDataManager.getBeanMetaData( clazz );` 

Hence to skip this additional metadata retrieval I've broken the method into two parts - first one will work with the existing "current" metadata `valueContext.getCurrentBeanMetaData()` and in the second part we iterate through the bean class hierarchy skipping the first element which is the bean class. 

I also remember running JMH tests for these changes and the results were almost identical.

In the end this change gives us:
- one additional call to metadata manager removed ( as metadata for that class is already cached it's basically removing one Map access call). Which doesn't give us much in terms of perfomance
- as we don't call the metadata manager to get metadata at this point, the code is working for the property holders. Because we build the correct metadata for them in corresponding value contexts instead of building it here in the `ValidtorImpl`

Now thinking about wrapping the `String mappingName` or beans class into wrapper ... we might end up not needing this at all. As we would then, most likely, have the `getClassHierarchy()` method return the wrappers. Or even maybe return the metadata right away, instead of simply returning classes.

PR #990, opened 2018-10-24
HV-1658 Replace a couple of stream usages

gsmet:
https://hibernate.atlassian.net/browse/HV-1658

@Sanne I think that's what you wanted?


Sanne:
Looks good! Thanks, I'll test it today to verify the impact on class definitions and let you know.

Sanne:
That worked fine. Thanks! I'll merge.

PR #993, opened 2018-12-03
HV-1665 Support incremental annotation processing in Gradle

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1665

As per Gradle [doc on incremental APs](https://docs.gradle.org/nightly/userguide/java_plugin.html#making_an_annotation_processor_incremental) it looks like it should be enough to register AP in META-INF/gradle/incremental.annotation.processors
with corresponding category separated by a comma.

Also added a small section to documentation on AP and Gradle. (that's how I "configured" the AP for testing the gradle builds and messages from AP). We have the doc sections for various build tools but nothing for gradle.

### DIFF
--- a/annotation-processor/src/main/resources/META-INF/services/gradle/incremental.annotation.processors
+++ b/annotation-processor/src/main/resources/META-INF/services/gradle/incremental.annotation.processors
@@ -1,0 +1,1 @@
+org.hibernate.validator.ap.ConstraintValidationProcessor,isolating
--- a/documentation/src/main/asciidoc/ch13.asciidoc
+++ b/documentation/src/main/asciidoc/ch13.asciidoc
@@ -116,6 +116,24 @@ For using the Hibernate Validator annotation processor with Maven, set it up via
 ----
 ====
 
+[[validator-annotationprocessor-gradle]]
+===== Gradle
+
+When using https://gradle.org[Gradle] it is enough to reference the annotation processor as an `annotationProcessor` dependency.
+
+.Using the annotation processor with Gradle
+====
+[source, groovy]
+[subs="verbatim,attributes"]
+----
+dependencies {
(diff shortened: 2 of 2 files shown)


gsmet:
Wondering if we should have a Gradle example somewhere but not sure where. Probably a separate project?

marko-bekhta:
Hi Guillaume, 
I thought it was about time I start opening some PRs if I wanted to get at least something done for free form validation till the new year 😄 
I've applied the suggested changes and force-pushed a new commit. As for the testing of this ... I've just created a simple clean gradle project and added all the things we have in the tests folder for AP module. This way I'd get all kinds of errors/warnings. And here's my build file from that project:

```groovy
plugins {
	id 'java'
}

group 'org.hibernate.validator'
version '1.0-SNAPSHOT'

sourceCompatibility = 1.8

repositories {
	mavenCentral()
	mavenLocal()
}

dependencies {
	annotationProcessor group: 'org.hibernate.validator', name: 'hibernate-validator-annotation-processor', version: '6.1.0-SNAPSHOT'
	
	// any other dependencies ...	
	
	compile group: 'org.hibernate.validator', name: 'hibernate-validator', version: '6.1.0-SNAPSHOT'
	compile group: 'org.glassfish', name: 'javax.el', version: '3.0.1-b09'

	testCompile group: 'junit', name: 'junit', version: '4.12'
}

compileJava {
	
}
```

It compiles and gives warnings/errors  😃 Not sure what else to check here ...

As for 

>Wondering if we should have a Gradle example somewhere but not sure where. Probably a separate project?

I was thinking about this as well. Usually we give maven examples but never mention gradle anywhere in the doc. Don't get me wrong - I love maven 😄. Hence I would agree that having a gradle example would be nice. But I'm not sure where to put it, maybe under https://github.com/hibernate/hibernate-demos/tree/master/hibernate-validator ? But it would also be nice to be able to reference some examples in the doc as well...

gsmet:
Merged, thanks!

PR #1005, opened 2018-12-18
HV-1682 Introduce BeanMetaDataClassNormalizer in predefined scope

gsmet:
The idea is to be able to normalize the class and get the parent in the
case of a generated proxy.

@gunnarmorling very much interested if you have other name suggestions. Couldn't figure out something better but not totally satisfied.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/PredefinedScopeBeanMetaDataManager.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/PredefinedScopeBeanMetaDataManager.java
@@ -31,17 +31,21 @@ import org.hibernate.validator.internal.util.CollectionHelper;
 import org.hibernate.validator.internal.util.ExecutableHelper;
 import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
 import org.hibernate.validator.internal.util.classhierarchy.ClassHierarchyHelper;
+import org.hibernate.validator.internal.util.classhierarchy.Filters;
 import org.hibernate.validator.internal.util.logging.Log;
 import org.hibernate.validator.internal.util.logging.LoggerFactory;
+import org.hibernate.validator.metadata.BeanMetaDataClassNormalizer;
 
 public class PredefinedScopeBeanMetaDataManager implements BeanMetaDataManager {
 
 	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
+	private final BeanMetaDataClassNormalizer beanMetaDataClassNormalizer;
+
 	/**
 	 * Used to cache the constraint meta data for validated entities
 	 */
-	private final Map<String, BeanMetaData<?>> beanMetaDataMap;
+	private final Map<Class<?>, BeanMetaData<?>> beanMetaDataMap;
 
 	public PredefinedScopeBeanMetaDataManager(ConstraintCreationContext constraintCreationContext,
 			ExecutableHelper executableHelper,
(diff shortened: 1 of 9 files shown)


gunnarmorling:
@gsmet, can you clarify what you mean by "normalize the class" in this context?

engine/src/main/java/org/hibernate/validator/metadata/BeanMetaDataClassNormalizer.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.metadata;
+
+import org.hibernate.validator.Incubating;
+
+/**
+ * Defines how the validated class is normalized before being used as the reference to get the bean metadata.
```
gunnarmorling: I don't really understand the purpose of this contract :( Given it's public I think this should be improved to make the usage more apparent. Perhaps an example added to the JavaDoc would help? Also the method should be documented.

gsmet:
So basically, I have:
- `MyBean`
- `MyBeanGeneratedProxy` extends `MyBean`

I'm able to register `MyBean` in the VF but `MyBeanGeneratedProxy` is generated and I can't easily register it. Thus the idea to be able to `normalize()` the class and work on the parent if it's a proxy.

The added test gives an example of how it can be used.

engine/src/main/java/org/hibernate/validator/internal/metadata/PredefinedScopeBeanMetaDataManager.java
```diff
@@ -30,13 +30,17 @@
...
 
 public class PredefinedScopeBeanMetaDataManager implements BeanMetaDataManager {
 
+	private final BeanMetaDataClassNormalizer beanMetaDataClassNormalizer;
+
 	/**
 	 * Used to cache the constraint meta data for validated entities
 	 */
-	private final Map<String, BeanMetaData<?>> beanMetaDataMap;
+	private final Map<Class<?>, BeanMetaData<?>> beanMetaDataMap;
```
gunnarmorling: My memory is blurry, but didn't we key this intentionally _not_ by class objects to avoid any CL leaks caused by a container-managed VF kept after an application got undeployed?

engine/src/main/java/org/hibernate/validator/metadata/BeanMetaDataClassNormalizer.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.metadata;
+
+import org.hibernate.validator.Incubating;
+
+/**
+ * Defines how the validated class is normalized before being used as the key to get the bean metadata.
+ * <p>
+ * In the case of the predefined scope validator factory, we have to register all the classes that will ever been
```
gunnarmorling:
```suggestion
 * In the case of the predefined scope validator factory, we have to register all the classes that will ever be
```

engine/src/main/java/org/hibernate/validator/metadata/BeanMetaDataClassNormalizer.java
```diff
@@ -0,0 +1,37 @@
...
+ * validated. To validate method calls, frameworks usually generates proxies to intercept the calls. Such proxies might
```
gunnarmorling:
```suggestion
 * validated. To validate method calls, frameworks usually generate proxies to intercept the calls. Such proxies might
```

engine/src/main/java/org/hibernate/validator/metadata/BeanMetaDataClassNormalizer.java
```diff
@@ -0,0 +1,37 @@
...
+ * be hard to register in the predefined scope validator factory as they are generated code.
+ * <p>
+ * This contract allows to normalize the class before obtaining the metadata from the PredefinedScopeBeanMetaDataManager
```
gunnarmorling:
```suggestion
 * This contract allows to normalize the class before obtaining the metadata from the {@code PredefinedScopeBeanMetaDataManager}
```

engine/src/main/java/org/hibernate/validator/internal/metadata/PredefinedScopeBeanMetaDataManager.java
gsmet: The standard BeanMetaDataManager is keyed by class. I had issues with the predefined scope one so I used strings for a while but it's now fixed so we can use classes too now.

gsmet:
Has been merged a while ago.

PR #1006, opened 2018-12-21
HV-1683 Upgrade javadoc plugin

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1683

This one was caused by the ambiguity of the null exception that the 3.0.0 plugin could throw. And  this helped while I was trying to make the other pr to pass the distribution build. So I've changed the version and made a few updates to the javadoc here and there.


gsmet:
Nice, thanks!

PR #1009, opened 2019-01-04
HV-1684 Fix value extraction logic to avoid stack overflows

gsmet:
So as Gunnar discovered it, the issue is only triggered with constraints
on Maps as otherwise CollectionHelper#toImmutableSet() avoids the issue
when there is only one element. Maps have 2 value extractors.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorResolver.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorResolver.java
@@ -47,8 +47,6 @@ public class ValueExtractorResolver {
 
 	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
-	private static final Object NON_CONTAINER_VALUE = new Object();
-
 	@Immutable
 	private final Set<ValueExtractorDescriptor> registeredValueExtractors;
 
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/valueextraction/ValueExtractorResolver.java
```diff
@@ -303,26 +303,27 @@ else if ( TypeHelper.isAssignable( descriptor.getContainerType(), candidate.getC
...
 
+		Set<ValueExtractorDescriptor> possibleValueExtractors = potentialValueExtractorDescriptors
+				.stream()
+				.filter( e -> TypeHelper.isAssignable( e.getContainerType(), runtimeType ) )
+				.collect( Collectors.toSet() );
+
+		valueExtractorDescriptors = getMaximallySpecificValueExtractors( possibleValueExtractors );
+
 		if ( valueExtractorDescriptors.isEmpty() ) {
 			nonContainerTypes.put( runtimeType, NON_CONTAINER_VALUE );
```
gunnarmorling: @gsmet, not directly related to this change, but can you check the beginning of the two methods:

```
if ( nonContainerTypes.contains( runtimeType ) ) {
```

Shouldn't this be `containsKey()` instead? It's not wrong per se, but it currently defeats the purpose of the `nonContainerTypes` cache.

Another question is the usage of the unbounded maps; in `BeanMetaDataManagerImpl` we're using `ConcurrentReferenceHashMap` instead, which would release the entries if there's need.

gsmet: Fixed the lookup issue by using a Set, thanks for the `newSetFromMap()` trick.

As discussed, all these elements are properly cleared when cleared when closing the VF so we are safe on this side.

gsmet:
Merged to master, will backport to 6.0.

PR #1011, opened 2019-01-05
HV-1687 Reduce bootstrap log verbosity

gsmet:
* https://hibernate.atlassian.net/browse/HV-1687

@gunnarmorling interested in your feedback on this one. Rationale explained in the issue.


marko-bekhta:
We probably can also change the level of `usingParameterNameProvider()` other than that it seems to be a good idea!

gsmet:
@marko-bekhta good catch, will fix and push.

PR #1012, opened 2019-01-09
WIP HV-823

dalibegovic:
First draft of the HV-823 feature.

What is included:
- `PropertyNodeNameProvider` contract that can be used to provide name for a property node
- Default implementation that returns property name
- Propagation of this new contract via: `HibernateValidatorConfiguration`, and `HibernateValidatorContext`
- Feature implementation (resolving name where nessecary, and adding it to `ConstraintViolation`)
- Simple tests showing how this contract can be used with external implementation


gsmet:
OK to test

dalibegovic:
Ok. We are looking forward for the review

engine/src/main/java/org/hibernate/validator/spi/nodenameprovider/PropertyNodeNameProvider.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.spi.nodenameprovider;
+
+public interface PropertyNodeNameProvider {
+	String getName(String propertyName, Object object);
```
marko-bekhta: hmm, that's an interesting choice of a contract. It's a nice way to not rely neither on reflection nor on our current `internal.properties.*` wrappers. I think though there might be a problem with JSON/Map validation case. An implementation of a `PropertyNodeNameProvider` would need to check the type of an object passed and determine how to proceed based on that. And as we plan to support any kind of property holders the implementations of `PropertyNodeNameProvider` would need to know about all of them.
Using `internal.properties.*` types would probably solve the problem but as @gsmet mentioned, these types are not yet ready to be public, and they might not become such.

engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
```diff
@@ -56,6 +56,7 @@
 	public static final String MAP_VALUE_NODE_NAME = "<map value>";
 
 	private final String name;
+	private final String resolvedName;
```
marko-bekhta: Do we need to keep both names resolved/original ones ? I was thinking that we could keep just one name. Then we also wouldn't need the `getResolvedPropertyPath()` method in `ConstraintViolationImpl`

engine/src/main/java/org/hibernate/validator/internal/engine/path/PathImpl.java
```diff
@@ -113,11 +113,11 @@ public PathImpl getPathWithoutLeafNode() {
 		return new PathImpl( nodeList.subList( 0, nodeList.size() - 1 ) );
 	}
 
-	public NodeImpl addPropertyNode(String nodeName) {
+	public NodeImpl addPropertyNode(String nodeName, String resolvedNodeName) {
```
marko-bekhta: keeping just one name as mentioned in one of the above comments would allow us to keep these methods unchanged.

engine/src/main/java/org/hibernate/validator/internal/util/ExecutablePropertyNodeNameProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.util;
+
+import java.io.Serializable;
+
+import org.hibernate.validator.spi.nodenameprovider.PropertyNodeNameProvider;
+
+public class ExecutablePropertyNodeNameProvider implements Serializable {
```
marko-bekhta: We probably should change the name of this class so it matches more to its purpose. Let's first get settled with the main interface contract before any changes here.

engine/src/test/java/org/hibernate/validator/test/spi/nodenameprovider/AnnotationPropertyNodeNameProvider.java
```diff
@@ -0,0 +1,66 @@
...
+
+	AnnotationPropertyNodeNameProvider(Class<? extends Annotation> annotationType) {
+		this( annotationType, VALUE );
+	}
+
+	@Override
+	public String getName(String propertyName, Object object) {
+		String resolvedName = propertyName;
+
+		Field field = getField( propertyName, object );
```
marko-bekhta: It's great that you've added this test implementation as we can uncover potentially interesting cases.
- How would it handle a case of getter properties? would you check both fields/getters ?
- Note that recently we merged in a new feature - https://hibernate.atlassian.net/browse/HV-1363 which would allow to define which methods should be considered as getter properties. Hence such `AnnotationPropertyNodeNameProvider` would need to know about currently used `GetterPropertySelectionStrategy`.

engine/src/main/java/org/hibernate/validator/spi/nodenameprovider/PropertyNodeNameProvider.java
dalibegovic: The idea we have is that we provide the contract, an extension point, for property name resolution, and let others implement it (not have the implementations part of the `hibernate-validator`). That way, it's up to the one who is implementing the contract to decide how to resolve the name (depending on the use case). One example of implementation is to use reflection and resolve the name from the annotation.

Could you please describe in more details your concern regarding JSON/Map validation case?

engine/src/test/java/org/hibernate/validator/test/spi/nodenameprovider/AnnotationPropertyNodeNameProvider.java
dalibegovic: This is just to show how it could look. There is more work around it.

The questions you have would be actually addressed by someone who is implementing the contract. It should't be hard to check both fields and getters.

As for `GetterPropertySelectionStrategy ` thanks for pointing it out. If we agree on everything else, we could try to create an implementation that would include `GetterPropertySelectionStrategy ` in consideration as a proof of concept.

dalibegovic: One more example is that someone who is implementing the contract to be used for name resolution from [jackson](https://github.com/FasterXML/jackson) annotations might want to take [property naming strategy](https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/PropertyNamingStrategy.html) into consideration when resolving the names. But again, that is up to the implementor to decide how to handle

engine/src/main/java/org/hibernate/validator/spi/nodenameprovider/PropertyNodeNameProvider.java
marko-bekhta: Yes sure. Let me try to show it with the example. Let's assume we have next bean:
```java
class SomeBean {
    @Valid 
    private SomeOtherBean someOtherBean;

    @NotNull
    private String string;

    @ValidPropertyHolder(...)
    private Map<String, ?> mapAsPropertyHolder;

    @ValidPropertyHolder(...)
    private JsonObject someJsonAsPropertyHolder;

    // .... anything else
}
```
fields annotated with `ValidPropertyHolder` are considered to be "property holders" and each their property could have own set of validation constraint rules. Hence `PropertyNodeNameProvider` would also be called on properties of these structures. By properties I mean something like:
```java
mapAsPropertyHolder.get("SomePropertyKey");
```
 And what I was concerned about is that in case of the current `PropertyNodeNameProvider` interface that you've provided, you could get in situation where object is either `Map` or some `JsonObject` and an implementation would need to check for all these possible cases before making any actions. 
Going to the example that you have in the tests, in case of the bean above, you would first need to check if passed object is a `Map`, if so - do some custom logic for it (or simply return the key) then check if maybe it's a JSON, and in the end if it doesn't fall under any "special" cases you would try to get a field or a getter matching the given property name. 
While it's all doable, the implementation of such `PropertyNodeNameProvider` could grow big and would be dependent on all these cases. Using something like `internal.properties.*`, in theory, should prevent this from happening as it all we be abstracted out.

engine/src/test/java/org/hibernate/validator/test/spi/nodenameprovider/AnnotationPropertyNodeNameProvider.java
marko-bekhta: Right, I'm just thinking how we could minimize the additional factors that such `PropertyNodeNameProvider` need to care about, and make implementation of them easier and smaller.

engine/src/test/java/org/hibernate/validator/test/spi/nodenameprovider/jackson/JacksonAnnotationPropertyNodeNameProvider.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.hibernate.validator.spi.nodenameprovider.Property;
+import org.hibernate.validator.spi.nodenameprovider.PropertyNodeNameProvider;
+
+import com.fasterxml.jackson.databind.BeanDescription;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.introspect.BeanPropertyDefinition;
+
+public class JacksonAnnotationPropertyNodeNameProvider implements PropertyNodeNameProvider {
```
dalibegovic: @marko-bekhta this is an example of `PropertyNodeNameProvider` for the [jackson](https://github.com/FasterXML/jackson) use-case (which will probably be the most used one). Here we can see that the implementor has the freedom of using any API that can assist him in implementing the correct behaviour. In this example, jackson provides all the necessary API-s that are needed for the implementation. If `internal.properties.*` becomes open one day, the implementor can use it.

The point is that this feature, in our opinion, is not related, or blocked by `internal.properties.*` being opened or closed.

[this](/engine/src/test/java/org/hibernate/validator/test/spi/nodenameprovider/jackson/JacksonAnnotationPropertyNodeNameProviderTest.java) is a test for `JacksonAnnotationPropertyNodeNameProvider `

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -150,6 +151,12 @@
 	 */
 	ResourceBundleLocator getDefaultResourceBundleLocator();
 
+	S propertyNodeNameProvider(PropertyNodeNameProvider propertyNodeNameProvider);
+
+	PropertyNodeNameProvider getDefaultPropertyNodeNameProvider();
```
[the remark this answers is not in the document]
dalibegovic: This is done, please review and resolve if the change is ok.

engine/src/main/java/org/hibernate/validator/spi/nodenameprovider/PropertyNodeNameProvider.java
dalibegovic: Thank you for the example.

I have updated the PR, and added a new comment addressing your concerns: https://github.com/hibernate/hibernate-validator/pull/1012#discussion_r250205593

engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
dalibegovic: This is done, please review and resolve if the change is ok.

dalibegovic:
Any updates? Should we continue with the implementation (do refactoring, code cleanup, add documentation, tests...), or there are still some concerns/doubts that we need to address?

engine/src/main/java/org/hibernate/validator/HibernateValidatorContext.java
```diff
@@ -44,6 +46,11 @@
 	@Override
 	HibernateValidatorContext parameterNameProvider(ParameterNameProvider parameterNameProvider);
 
+	/**
+	 * TODO
```
gsmet: You can add the `@since 6.1` tag here.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryScopedContext.java
```diff
@@ -35,6 +36,11 @@
 	 */
 	private final ExecutableParameterNameProvider parameterNameProvider;
 
+	/**
+	 *  Docs
```
gsmet: Looks simple enough to write the doc instead of having a TODO :). Wouldn't want to forget about this.

engine/src/test/java/org/hibernate/validator/test/spi/nodenameprovider/jackson/JacksonAnnotationPropertyNodeNameProvider.java
gsmet: So, indeed you can work around not having the properties as you did it.

But I think it will be very slow and probably not acceptable in the general case. Take a look at what you have to do in the `getName()` method. It does not look good performance-wise.

While I don't think we're ready to push `internal.properties.*` in SPIs, I think it might be possible to have a subset of it covering the most common use cases (e.g. property name and annotations). Could you maybe take a stab at it?

dalibegovic:
> Hi, @dalibegovic it's great that you've done some more changes. I didn't have much time to take a look at them though. Also I had an idea that I wanted to try out. I'll try to find some time this week and will let you know how it went. Thanks!
> 
> **EDIT**
> One quick question that I wanted to ask you. Is there any reason we would need a runtime object to determine a property path label? I mean is there a case when a property path would be different for the same class but different instances ?

@marko-bekhta good question.
For the use-case we are currently focused on, that is resolving the name from the annotation on a field, you are right, the class information should be enough.

On the other hand, if the name is resolved by using some different mechanism (e.g. by looking at a value in some field), then the property path would be different depending on the runtime instances. 

It is hard to predict if this will be a use-case, but since we have a runtime instance, and it provides more flexibility to the implementor, i don't see a reason why we shouldn't provide a runtime object.

Do you see any disadvantages in using a runtime object? Maybe some performance concerns?

engine/src/test/java/org/hibernate/validator/test/spi/nodenameprovider/jackson/JacksonAnnotationPropertyNodeNameProvider.java
julien-may: @gsmet: i agree that it might be a bit of a hassle for some the cases to implement something simple and that we should provide more property related information. but this is why we proposed to have the `Property` interface, to be flexible with new features that improve the experience when implementing `PropertyNodeNameProvider`.

I just wanted to indicate that e.g. `Jackson` is more complicated than it seems. Influencing how a property is resolved cannot only be done through an annotation but also through a globally defined naming strategy. Plus `Jackson` is doing a lot of caching, the call that is made (as an example) in the `JacksonAnnotationPropertyNodeNameProvider.getName` is actually quite fast.

So i would suggest to refactor `PropertyNodeNameProvider.getName` to:
```
String getName(Property property);
```

and the `Property` to:

```
public interface Property {
	String getName();
	Annotation[] getDeclaredAnnotations();
	Class<?> getDeclaringClass();
	Object getObject();
}
```

WDYT?

dalibegovic:
> @dalibegovic sorry for the delay. We had our Hibernate face to face meeting last week and I'm pretty swamped.
> 
> Tried to provide some feedback so that you can make progress on this.
> 
> Overall, it looks like a useful addition so let's drive this puppy home.

@marko-bekhta @gsmet we are ready to drive this puppy home, just waiting on some opinions and input from you guys :)

marko-bekhta:
Sorry for long response times. I didn't have a chance to spend much time on this one. But here's a few things that might give you more directions. 
- I would like us to resolve the property path name at the bootstrap time, i.e. when we create metadata rather than doing it at the time of the checks. I cannot find a good use case where a path would vary for the same type based on the state of the particular instance, that's why I was asking if you maybe have some examples.
- If we go with such approach the most logical place to store the path would be inside a property (https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/properties/Property.java). Maybe add another method - `String getPropertyPath()`?
- In this case we would need to have a property path label before we actually create an instance of `internal.properties.Property`
- Thus exposing part of `internal.properties` wouldn't work for us in this case. We probably would need to either have a separate hierarchy of property interfaces just for determining the property path label, or  have a set of various name providers for each and every case (field, method, "property" of a map, json property, etc). So I'm leaning more towards a hierarchy of property interfaces :wink: 
The interfaces based on the suggestion of one of your colleges could be something like
```java
interface Property {
    String getName();
}

interface JavaBeanProperty extends Property {
    Annotation[] getDeclaredAnnotations();
    Class<?> getDeclaringClass();
}
```
or instead of exposing the `Annotation` type we could go with something like:
```java
interface JavaBeanProperty extends Property {
    <T> T getParameterValueOfDeclaredAnnotation(Class<?> annotationClass, Class<T> valueClass );
    Class<?> getDeclaringClass();
}
```
It would be important to have this base `Property` interface in place and use it in the name provider. This would allow us to include other property types (map, json etc) in the future. And name provider could then check if a passed property is of some more specific type (e.x. `JavaBeanProperty`) and if it's of unknown type it could fallback to using the `Property.getName()`

dalibegovic:
If we are going to take that aproach (resolving the name at bootstrap time), I am going to need some more directions.

I am able to resolve the name in `internal.metadata.provider.AnnotationMetaDataProvider` and add it to `internal.properties.javabean.JavaBeanField` because this is the place where it's created.

What I am missing is how to propagate this meta-data, stored in `internal.properties.Property` to `javax.validation.metadata.ConstraintDescriptor.ConstraintViolation` that is used at the end by the client?

marko-bekhta:
There's 3 places where JavaBean properties are created - annotation provider (as you mentioned) then there's xml and programmatic API providers. Last two are using `JavaBeanHelper` to find/create instances of `Property`. 

> What I am missing is how to propagate this meta-data, stored in `internal.properties.Property` to `javax.validation.metadata.ConstraintDescriptor.ConstraintViolation` that is used at the end, by the client?

I think we actually don't need it that far. If we have a property-path-node-name inside of a property we should be able to use it right away for construction of paths. So for example `AbstractPropertyConstraintLocation#appendTo(ExecutableParameterNameProvider, PathImpl)` would change from current:
```java
	@Override
	public void appendTo(ExecutableParameterNameProvider parameterNameProvider, PathImpl path) {
		path.addPropertyNode( property.getPropertyName() );
	}
```
to something like (you probably can come up with a better getter name):
```java
	@Override
	public void appendTo(ExecutableParameterNameProvider parameterNameProvider, PathImpl path) {
		path.addPropertyNode( property.getPropertyNodePathName() );
	}
```
where you can use the resolved node name. The same would apply to `AbstractPropertyCascadable`. And, I think, the last place where the property can be appended to a path is in `ConstraintViolationBuilderImpl`. But I guess we shouldn't be worried about this one as it's up to the users to put what they want when using this builder.

This way when a `ConstraintViolation` is created you would get a `Path propertyPath` that is already using all expected node names.

Is this helpful ? Or did I misinterpreted the problem ?

dalibegovic:
Yes, it is, that was exactly what I was missing. Ok, I will try with this new approach and update the PR soon

engine/pom.xml
```diff
@@ -147,6 +147,16 @@
             <artifactId>moneta</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.9.8</version>
```
julien-may: add test scope

engine/pom.xml
```diff
@@ -147,6 +147,16 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>2.9.8</version>
```
julien-may: add test scope

engine/src/main/java/org/hibernate/validator/internal/engine/AbstractConfigurationImpl.java
```diff
@@ -208,6 +211,20 @@ public T parameterNameProvider(ParameterNameProvider parameterNameProvider) {
 		return thisAsT();
 	}
 
+	@Override
+	public T propertyNodeNameProvider(PropertyNodeNameProvider propertyNodeNameProvider) {
+		if ( LOG.isDebugEnabled() ) {
+			if ( propertyNodeNameProvider != null ) {
+				LOG.debug(
+						"Setting custom PropertyNodeNameProvider of type " + propertyNodeNameProvider.getClass()
```
julien-may: indentation

engine/src/main/java/org/hibernate/validator/spi/nodenameprovider/Property.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.spi.nodenameprovider;
+
+public interface Property {
+	Class get();
```
julien-may: `Class<?>`

engine/src/main/java/org/hibernate/validator/internal/engine/AbstractConfigurationImpl.java
dalibegovic: It's how the project formatter indented it, so i don't want to argue :)

engine/src/main/java/org/hibernate/validator/spi/nodenameprovider/PropertyNodeNameProvider.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.spi.nodenameprovider;
+
+public interface PropertyNodeNameProvider {
+	String getName(String propertyName, Property property);
```
marko-bekhta: I think we should just have
```java
String getName(Property property);
```
And property interface would be:
```java
public interface Property {
	String getName();
}
```
property itself should already contain information about its own name.

engine/src/main/java/org/hibernate/validator/spi/nodenameprovider/Property.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.spi.nodenameprovider;
+
+public interface Property {
+	Class<?> get();
```
marko-bekhta: It's unclear what's the purpose of this method. We should have javadoc for all public classes and their methods. Please also note that we should have an hierarchy of property interfaces as mentioned before.
```java
interface Property {
	String getName();
}

interface JavaBeanProperty extends Property {
	Annotation[] getDeclaredAnnotations();
	Class<?> getDeclaringClass();
}
```
It is important that we don't expose any methods in property interface that wouldn't make sense for non JavaBean properties.

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -147,6 +147,8 @@ public ValidatorFactoryImpl(ConfigurationState configurationState) {
 				configurationState.getMessageInterpolator(),
 				configurationState.getTraversableResolver(),
 				new ExecutableParameterNameProvider( configurationState.getParameterNameProvider() ),
+				// check this
```
marko-bekhta: ? :smiley:

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryImpl.java
```diff
@@ -169,7 +171,7 @@ public ValidatorFactoryImpl(ConfigurationState configurationState) {
 		this.constraintCreationContext = new ConstraintCreationContext( constraintHelper, constraintValidatorManager, typeResolutionHelper, valueExtractorManager );
 
 		this.executableHelper = new ExecutableHelper( typeResolutionHelper );
-		this.javaBeanHelper = new JavaBeanHelper( ValidatorFactoryConfigurationHelper.determineGetterPropertySelectionStrategy( hibernateSpecificConfig, properties, externalClassLoader ) );
+		this.javaBeanHelper = new JavaBeanHelper( ValidatorFactoryConfigurationHelper.determineGetterPropertySelectionStrategy( hibernateSpecificConfig, properties, externalClassLoader ), hibernateSpecificConfig.getPropertyNodeNameProvider() );
```
marko-bekhta: We probably should have another method in `ValidatorFactoryConfigurationHelper` to determine the property name provider. `determineGetterPropertySelectionStrategy()` is a nice example.

engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
```diff
@@ -220,7 +225,8 @@ public AnnotationProcessingOptions getAnnotationProcessingOptions() {
 				continue;
 			}
 
-			JavaBeanField javaBeanField = new JavaBeanField( field );
+			String name = new ExecutablePropertyNodeNameProvider( propertyNodeNameProvider, beanClass ).getName( field.getName() );
```
marko-bekhta: We don't really need to create this `ExecutablePropertyNodeNameProvider`. We should have a wrapper of field/getter which implements `Property` interface that was added in this PR and pass it to the propertyNodeNameProvider. We even shouldn't add `propertyNodeNameProvider` as an additional field to this provider. We already have it in `JavaBeanHelper` so we probably should use it. We can add a method in that helper that would resolve a name for us.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanField.java
```diff
@@ -25,13 +25,19 @@
 public class JavaBeanField implements org.hibernate.validator.internal.properties.Field, JavaBeanAnnotatedConstrainable {
 
 	private final Field field;
+	private final String resolvedName;
 	private final Type typeForValidatorResolution;
 	private final Type type;
 
 	public JavaBeanField(Field field) {
```
marko-bekhta: If this constructor is not used anymore we should probably remove it.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanGetter.java
```diff
@@ -47,6 +47,13 @@ public String getPropertyName() {
 		return propertyName;
 	}
 
+	@Override
+	public String getResolvedPropertyName() {
+		// or throw unsupported operation?
+		// or resolve the name
```
marko-bekhta: It's still a property, right? you could have annotated getters that would be converted to json properties. If we follow the approach of resolving the name based on a `Property` this case would get covered as well.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanHelper.java
```diff
@@ -47,7 +51,9 @@ public GetterPropertySelectionStrategy getGetterPropertySelectionStrategy() {
 		Contracts.assertNotNull( declaringClass, MESSAGES.classCannotBeNull() );
 
 		Field field = run( GetDeclaredField.action( declaringClass, property ) );
-		return Optional.ofNullable( field ).map( JavaBeanField::new );
+		String name = new ExecutablePropertyNodeNameProvider( propertyNodeNameProvider, declaringClass ).getName( property );
```
marko-bekhta: Same comment as in the `AnnotationMetadataProvider`

engine/src/main/java/org/hibernate/validator/internal/util/ExecutablePropertyNodeNameProvider.java
marko-bekhta: So it seems that we don't actually need this class.

dalibegovic:
I added a commit with the changes. I tried addressing the comments to the best of my understanding. If I misinterpreted something, please do not hesitate to leave a comment.

Note that this is still a WIP, some docs are missing, some things need to be renamed, etc, but the main part should be there.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanGetter.java
```diff
@@ -47,6 +47,11 @@ public String getPropertyName() {
 		return propertyName;
 	}
 
+	@Override
+	public String getResolvedPropertyName() {
+		return propertyName;
```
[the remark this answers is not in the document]
dalibegovic: It's missing. Because i forgot to put `@Valid` on the getter, the tests were passing, so i thought this was done. I will update it in the next commit

dalibegovic:
Any hints on where the documentation should be added? I was thinking of adding a new section in chapter 12: Hibernate Validator Specifics (hibernate-validator/documentation/src/main/asciidoc/ch12.asciidoc)

marko-bekhta:
> a new section in chapter 12: [Hibernate Validator Specifics](https://github.com/hibernate/hibernate-validator/blob/master/documentation/src/main/asciidoc/ch12.asciidoc)

Yes that would be a good place for it. Just add one more section in the end. Thanks!

dalibegovic:
I think that the feature is ready for a final code review, after that, I will:
- Adding the documentation
- Rebase and Squash

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryScopedContext.java
```diff
@@ -35,6 +36,11 @@
 	 */
 	private final ExecutableParameterNameProvider parameterNameProvider;
 
+	/**
+	 *  {@link PropertyNodeNameProvider} to be used for property name resolution when constructing a property path
+	 */
+	private final PropertyNodeNameProvider propertyNodeNameProvider;
```
marko-bekhta: we can remove it from here too.

marko-bekhta:
@dalibegovic that was quick! Nice. I think you can go ahead and squash+rebase it all. And add that documentation section if you have time for it.

dalibegovic:
@marko-bekhta thanks :) On it, I will add the documentation, since I think it's an important part

dalibegovic:
I was not sure how to do the squashing without messing up the history, so i created a new PR: https://github.com/hibernate/hibernate-validator/pull/1029. We could merge it, and close this one. The comments from this PR will not be lost, and the git history will be clean.

I hope this is fine.

marko-bekhta:
Yes, it even might be better. Having a clean PR without these existing comments would be easier to read. Thanks!

Closing this PR in favor of https://github.com/hibernate/hibernate-validator/pull/1029

NatasG:
I have an Object `Person`, and `Person` has a field: `List<Home>home`, also `home` has a field: `List<Room> rooms`, when a set a custom name for `rooms`, like: `@PropertyName("myRooms")`,

when do the validation with an empty rooms, I got: `JSR-303 validated property 'home[0].myRooms' does not have a corresponding accessor for Spring data binding - check your DataBinder's configuration (bean property versus direct field access)`

PR #1013, opened 2019-01-12
HV-1657 “propertyPath” available via the HibernateMessageInterpolatorContext

sjaakd:
https://hibernate.atlassian.net/browse/HV-1657

See HV-1657 for more details. I did not add the `@Since` tag to the new method (which I probably should) since I don't know which version to indicate. I'm using Hibernate Validation in JBOSS EAP 7.1 (5.3.5Final) so I hope to get this available in some maintenance release.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/MessageInterpolatorContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/MessageInterpolatorContext.java
@@ -11,6 +11,7 @@ import static org.hibernate.validator.internal.util.CollectionHelper.toImmutable
 import java.lang.invoke.MethodHandles;
 import java.util.Map;
 
+import javax.validation.Path;
 import javax.validation.metadata.ConstraintDescriptor;
 
 import org.hibernate.validator.internal.util.logging.Log;
@@ -33,19 +34,22 @@ public class MessageInterpolatorContext implements HibernateMessageInterpolatorC
 	private final ConstraintDescriptor<?> constraintDescriptor;
 	private final Object validatedValue;
 	private final Class<?> rootBeanType;
+	private final Path propertyPath;
 	@Immutable
 	private final Map<String, Object> messageParameters;
 	@Immutable
 	private final Map<String, Object> expressionVariables;
 
 	public MessageInterpolatorContext(ConstraintDescriptor<?> constraintDescriptor,
-			Object validatedValue,
-			Class<?> rootBeanType,
-			Map<String, Object> messageParameters,
-			Map<String, Object> expressionVariables) {
+					Object validatedValue,
+					Class<?> rootBeanType,
+					Path propertyPath,
+					Map<String, Object> messageParameters,
+					Map<String, Object> expressionVariables) {
 		this.constraintDescriptor = constraintDescriptor;
 		this.validatedValue = validatedValue;
 		this.rootBeanType = rootBeanType;
+		this.propertyPath = propertyPath;
 		this.messageParameters = toImmutableMap( messageParameters );
 		this.expressionVariables = toImmutableMap( expressionVariables );
 	}
@@ -73,6 +77,11 @@ public class MessageInterpolatorContext implements HibernateMessageInterpolatorC
 	@Override
 	public Map<String, Object> getExpressionVariables() {
 		return expressionVariables;
+	}
+
+	@Override
+	public Path getPropertyPath() {
+		return propertyPath;
 	}
 
 	@Override
@@ -122,6 +131,8 @@ public class MessageInterpolatorContext implements HibernateMessageInterpolatorC
 		sb.append( "MessageInterpolatorContext" );
 		sb.append( "{constraintDescriptor=" ).append( constraintDescriptor );
 		sb.append( ", validatedValue=" ).append( validatedValue );
+		sb.append( ", rootBeanType=" ).append( rootBeanType.getName() );
+		sb.append( ", propertyPath=" ).append( propertyPath );
 		sb.append( ", messageParameters=" ).append( messageParameters );
 		sb.append( ", expressionVariables=" ).append( expressionVariables );
 		sb.append( '}' );
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/AbstractValidationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/AbstractValidationContext.java
@@ -228,6 +228,7 @@ abstract class AbstractValidationContext<T> implements BaseBeanValidationContext
 				messageTemplate,
 				valueContext.getCurrentValidatedValue(),
 				descriptor,
+				constraintViolationCreationContext.getPath(),
 				constraintViolationCreationContext.getMessageParameters(),
 				constraintViolationCreationContext.getExpressionVariables()
 		);
@@ -293,12 +294,14 @@ abstract class AbstractValidationContext<T> implements BaseBeanValidationContext
 			String messageTemplate,
 			Object validatedValue,
 			ConstraintDescriptor<?> descriptor,
+			Path path,
 			Map<String, Object> messageParameters,
 			Map<String, Object> expressionVariables) {
 		MessageInterpolatorContext context = new MessageInterpolatorContext(
 				descriptor,
 				validatedValue,
 				getRootBeanClass(),
+				path,
 				messageParameters,
 				expressionVariables
 		);
--- a/engine/src/main/java/org/hibernate/validator/messageinterpolation/HibernateMessageInterpolatorContext.java
+++ b/engine/src/main/java/org/hibernate/validator/messageinterpolation/HibernateMessageInterpolatorContext.java
@@ -9,6 +9,7 @@ package org.hibernate.validator.messageinterpolation;
 import java.util.Map;
 
 import javax.validation.MessageInterpolator;
+import javax.validation.Path;
 
 /**
  * Extension to {@code MessageInterpolator.Context} which provides functionality
@@ -40,4 +41,11 @@ public interface HibernateMessageInterpolatorContext extends MessageInterpolator
 	 * @since 5.4.1
 	 */
 	Map<String, Object> getExpressionVariables();
+
+	/**
+	 * @return the path to the validated constraint starting from the root bean
+	 *
+	 * @since 6.1
+	 */
+	Path getPropertyPath();
 }
--- a/engine/src/test/java/org/hibernate/validator/test/internal/engine/messageinterpolation/ExpressionLanguageMessageInterpolationTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/internal/engine/messageinterpolation/ExpressionLanguageMessageInterpolationTest.java
@@ -67,9 +67,9 @@ public class ExpressionLanguageMessageInterpolationTest {
 				notNullDescriptor,
 				user,
 				null,
+				null,
 				Collections.<String, Object>emptyMap(),
-				Collections.<String, Object>emptyMap()
-		);
+				Collections.<String, Object>emptyMap() );
 
 		String expected = "18";
 		String actual = interpolatorUnderTest.interpolate( "${validatedValue.age}", context );
@@ -82,9 +82,9 @@ public class ExpressionLanguageMessageInterpolationTest {
 				notNullDescriptor,
 				new User(),
 				null,
+				null,
 				Collections.<String, Object>emptyMap(),
-				Collections.<String, Object>emptyMap()
-		);
+				Collections.<String, Object>emptyMap() );
 
 		String expected = "${validatedValue.foo}";
 		String actual = interpolatorUnderTest.interpolate( "${validatedValue.foo}", context );
@@ -172,9 +172,9 @@ public class ExpressionLanguageMessageInterpolationTest {
 				notNullDescriptor,
 				42.00000d,
 				null,
+				null,
 				Collections.<String, Object>emptyMap(),
-				Collections.<String, Object>emptyMap()
-		);
+				Collections.<String, Object>emptyMap() );
 
 		// german locale
 		String expected = "42,00";
@@ -232,9 +232,9 @@ public class ExpressionLanguageMessageInterpolationTest {
 				notNullDescriptor,
 				42.00000d,
 				null,
+				null,
 				Collections.<String, Object>emptyMap(),
-				Collections.<String, Object>emptyMap()
-		);
+				Collections.<String, Object>emptyMap() );
 
 		String expected = "${formatter.foo('%1$.2f', validatedValue)}";
 		String actual = interpolatorUnderTest.interpolate(
@@ -308,8 +308,8 @@ public class ExpressionLanguageMessageInterpolationTest {
 				descriptor,
 				null,
 				null,
+				null,
 				Collections.<String, Object>emptyMap(),
-				Collections.<String, Object>emptyMap()
-		);
+				Collections.<String, Object>emptyMap() );
 	}
 }
--- a/engine/src/test/java/org/hibernate/validator/test/internal/engine/messageinterpolation/MessageInterpolatorContextTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/internal/engine/messageinterpolation/MessageInterpolatorContextTest.java
@@ -7,35 +7,46 @@
 
 package org.hibernate.validator.test.internal.engine.messageinterpolation;
 
-import static org.easymock.EasyMock.createMock;
-import static org.easymock.EasyMock.expect;
-import static org.easymock.EasyMock.replay;
-import static org.easymock.EasyMock.verify;
-import static org.hibernate.validator.testutil.ConstraintViolationAssert.assertThat;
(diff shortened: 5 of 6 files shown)


gunnarmorling:
ok to test

gunnarmorling:
(sorry, closed accidentally)

Hey @sjaakd, very nice to see you here, thanks a lot for your first PR (@gsmet, Sjaak is a good friend and MapStruct hacker)! I remember you mentioned this issue before but still need to dive into the details before being able to say something meaningful :) Guillaume might have more feedback in the meantime.

gunnarmorling:
@sjaakd, could you add an actual (i.e. non-mocked) test that shows how this would be actually used (i.e. resembling the behavior of your current work-around you described in the JIRA issue)?

gunnarmorling:
Btw. how are you invoking validation? Are you in control of the validator (factory)? The way this is solved in JSF is that they define a specific resource key which takes a value like `{1}: {0}`. Here, {0} refers to the interpolated BV constraint violation message and {1} to the component label as obtained from JSF. I.e. the field name is injected only after the fact. I'm wondering whether you could do something similar even without any changes to HV, by post-processing the messages of the constraint violation similar to that?

sjaakd:
> @sjaakd, could you add an actual (i.e. non-mocked) test that shows how this would be actually used (i.e. resembling the behavior of your current work-around you described in the JIRA issue)?

Good one..  I need to think about that a little more and add a test for that..

sjaakd:
> Btw. how are you invoking validation? Are you in control of the validator (factory)?

Most of the times, yes. I can imagine REST scenarios + parameter validation that are a bit harder to control, but we create the validator from the factory..



> The way this is solved in JSF is that they define a specific resource key which takes a value like `{1}: {0}`. Here, {0} refers to the interpolated BV constraint violation message and {1} to the component label as obtained from JSF. I.e. the field name is injected only after the fact. I'm wondering whether you could do something similar even without any changes to HV, by post-processing the messages of the constraint violation similar to that?

In our case it is an object graph (in essence a complete model of a physical measurement, device or even a permit), which stems directly from an xm object received via a SOAP interface. 

The place in the object graph determines the exact name in the resource bundle (the naming of parameters refers to a [standard](https://www.basisregistratieondergrond.nl/documenten/publicaties/2017/06/27/catalogus-bodemkundige-boormonsterbeschrijving)). We validate the graph from its root (the root bean object). The path in the class determines the exact (short) name to be used in the message (and is a key in the resource bundle).

The problem (as such) is not related to giving a proper label a-priori (like a component label), which you typically do when you validate one component. . 



> by post-processing the messages of the constraint violation similar to that?

We do post-processing afterwards and use our own message interpolation. The drawback is that I've got to use some book-keeping and reflection to find the proper place in the model, which could be easy avoided if I've the proper information available. 

If it helps (for explanation, to get an idea) I can add the current solution to the Jira ticket.

sjaakd:
> Hey @sjaakd, very nice to see you here, thanks a lot for your first PR

My pleasure 😄

sjaakd:
So this is the class we currently use to intercept `ValidatorFactory ` and returned classes.. I only want to do message interpolation, but now I've got to correlate the original message via a map stored in a thread local in order to post-process message it just before its returned as `ConstraintViolation`. Its all a big ugly.

```java
public class BroValidation {

    private static final String PERCENT_BUT_NOT_ESCAPED_PERCENT = "(?:%(?<!\\\\%))";
    private static final Pattern PTRN = Pattern.compile( PERCENT_BUT_NOT_ESCAPED_PERCENT + "(.+?)" + PERCENT_BUT_NOT_ESCAPED_PERCENT );
    private static final String VALIDATED_ENTITY_ATTRIBUTE = "validatedEntityAttribute";
    private static final String BRO_VALIDATION_MESSAGES = "nl.bro.dto.common.BroValidationMessages";
    private static final String STANDARD_VALIDATION_MESSAGES = "ValidationMessages";

    private static class BroMessageInterpolator extends ResourceBundleMessageInterpolator implements MessageInterpolator {

        private ThreadLocal<Map<ConstraintDescriptor, Context>> constraintDescriptorMap = ThreadLocal.withInitial( HashMap::new );

        public BroMessageInterpolator(ResourceBundleLocator userResourceBundleLocator) {
            super( userResourceBundleLocator );
        }

        @Override
        public String interpolate(String messageTemplate, Context context) {
            constraintDescriptorMap.get().put( context.getConstraintDescriptor(), context );
            return super.interpolate( messageTemplate, context );
        }

        @Override
        public String interpolate(String messageTemplate, Context context, Locale locale) {
            constraintDescriptorMap.get().put( context.getConstraintDescriptor(), context );
            return super.interpolate( messageTemplate, context, locale );
        }

        public String interpolateSuper(String messageTemplate, Context context) {
            return super.interpolate( messageTemplate, context );
        }

        public void reset() {
            constraintDescriptorMap.get().clear();
        }

        public Context getContextForConstraintDescription( ConstraintDescriptor constraintDescriptor ) {
            return constraintDescriptorMap.get().get( constraintDescriptor );
        }

    }

    private abstract static class BroProxy<T> implements InvocationHandler {

        @Getter
        private final T proxiedObject;
        @Getter
        private final BroMessageInterpolator messageInterpolator;
        private final String proxyMethod;

        public BroProxy(T validatorFactory, BroMessageInterpolator messageInterpolator, String proxyMethod) {
            this.proxiedObject = validatorFactory;
            this.messageInterpolator = messageInterpolator;
            this.proxyMethod = proxyMethod;
        }

        @Override
        public Object invoke(Object proxy, Method method, Object[] args) throws ValidationException {

            try {
                if ( proxyMethod.equals( method.getName() ) ) {
                    return proxy( proxy, method, args );
                }
                else {
                    return method.invoke( proxiedObject, args );
                }
            }
            catch ( IllegalAccessException | IllegalArgumentException ex ) {
                throw new IllegalStateException( ex );
            }
            catch ( InvocationTargetException ex ) {
                // there's only one checked exception thrown in validation
                Throwable targetException = ex.getTargetException();
                if ( targetException instanceof ValidationException ) {
                    throw (ValidationException) ex.getTargetException();
                }
                else {
                    throw new IllegalStateException( ex );
                }
            }
        }

        protected abstract Object proxy(Object proxy, Method method, Object[] args) throws IllegalAccessException, IllegalArgumentException, InvocationTargetException;
    }

    private static class ValidatorFactoryProxy extends BroProxy<ValidatorFactory> {

        public ValidatorFactoryProxy(ValidatorFactory validatorFactory, BroMessageInterpolator messageInterpolator) {
            super( validatorFactory, messageInterpolator, "getValidator" );
        }

        @Override
        protected Object proxy(Object proxy, Method method, Object[] args) throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
            return createProxy( new ValidatorProxy( getProxiedObject().getValidator(), getMessageInterpolator() ), Validator.class );
        }
    }

    private static class ValidatorProxy extends BroProxy<Validator> {

        public ValidatorProxy(Validator validator, BroMessageInterpolator messageInterpolator) {
            super( validator, messageInterpolator, "validate" );
        }

        @Override
        protected Object proxy(Object proxy, Method method, Object[] args) throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
            getMessageInterpolator().reset();
            Set<ConstraintViolation> result = (Set<ConstraintViolation>) method.invoke( getProxiedObject(), args );
            Set<ConstraintViolation> postProcessedResult =  result.stream() //
                         .map( cv -> createProxy( new ConstraintViolationProxy( cv, getMessageInterpolator() ), ConstraintViolation.class ) ) //
                         .collect( Collectors.toSet() ); //
            getMessageInterpolator().reset();
            return postProcessedResult;
        }
    }

    private static class ConstraintViolationProxy extends BroProxy<ConstraintViolation> {

        private Context messageContext;

        public ConstraintViolationProxy(ConstraintViolation constraintViolation, BroMessageInterpolator messageInterpolator ) {
            super( constraintViolation, messageInterpolator, "getMessage" );
            this.messageContext = messageInterpolator.getContextForConstraintDescription( constraintViolation.getConstraintDescriptor() );
        }

        @Override
        protected Object proxy(Object proxy, Method method, Object[] args) throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
            String origMessage = (String) method.invoke( getProxiedObject(), args );
            String newMessage = replaceBroSpecificSubstitution( origMessage );
            return getMessageInterpolator().interpolateSuper( newMessage, messageContext );
        }

        private String replaceBroSpecificSubstitution(String message) {
            Matcher matcher = PTRN.matcher( message );
            StringBuffer buffer = new StringBuffer();

            while ( matcher.find() ) {
                String devoidOfEscapes =  matcher.group( 1 ).replace( "\\", "" );
                String replacement = getReplacement( devoidOfEscapes );
                if ( replacement != null ) {
                    matcher.appendReplacement( buffer, "" );
                    buffer.append( replacement );
                }
            }
            matcher.appendTail( buffer );
            return buffer.toString();
        }

        private String getReplacement(String key) {

            String validatedEntityAttribute = getEntityAttributePath();
            if ( VALIDATED_ENTITY_ATTRIBUTE.equals( key ) ) {
                return "{" + validatedEntityAttribute + "}";
            }
            else {
                return "{" + validatedEntityAttribute + "." + key + "}";
            }
        }

        private String getEntityAttributePath() {
            // create the key, agnostic from lists (indices)
            StringBuilder baseNodeBuilder = new StringBuilder( getProxiedObject().getRootBean().getClass().getSimpleName() );
            for ( Path.Node node : getProxiedObject().getPropertyPath() ) {
                if ( node.getName() != null ) {
                    baseNodeBuilder.append( "." ).append( node.getName() );
                }
            }
            return baseNodeBuilder.toString();
        }
    }

    private BroValidation() {
    }

    /**
     * Deze {@link ValidatorFactory} proxied de originele {@link javax.validation.ValidatorFactory}. Er worden custom
     * aanpassingen aan de boodschappen uitgevoerd middels post processing. De sleutel die wordt gebruikt is %(tag)%. Dit
     * wordt gedaan om de naam van de entiteit/attribuut combinatie vast te stellen in de object-graph (dto model). Deze
     * naam wordt vervangen door een sleutel die alsnog wordt geinterpoleerd. Volgorde van resource loading:
     * <ol>
     * <li>ValidationMessages: let op, bij testen wordt deze gelezen uit /test/resources</li>
     * <li>nl.bro.dto.common.BroValidationMessages</li>
     * <li>ValidationMessages: (nog een keertje, maar vermoedelijk uit /source/resources)</li>
     * <li>provider resources (voor de provider annotaties)</li>
     * </ol>
     *
     * @param bundlenames additional bundlenames, in prioritization order.
     * @return the wrapped ValidatorFactory
     */
    public static ValidatorFactory buildFactory(String... bundlenames) {
        List<String> bundleNames = new ArrayList<>( asList( bundlenames ) );
        bundleNames.addAll( asList( STANDARD_VALIDATION_MESSAGES, BRO_VALIDATION_MESSAGES ) );
        BroMessageInterpolator messageInterpolator = new BroMessageInterpolator( new AggregateResourceBundleLocator( bundleNames ) );
        ValidatorFactory validatorFactory = Validation.byDefaultProvider() //
                                                      .configure() //
                                                      .messageInterpolator( messageInterpolator )//
                                                      .buildValidatorFactory(); //
        return createProxy( new ValidatorFactoryProxy( validatorFactory, messageInterpolator ), ValidatorFactory.class );
    }

    private static <T> T createProxy(BroProxy<T> proxy, Class<T> proxiedInterface) {
        return (T) Proxy.newProxyInstance( proxiedInterface.getClassLoader(), new Class[] { proxiedInterface }, proxy );
    }

    public static String toBroEntityAttributes(Collection<String> expressionVariables) {
        StringBuilder buffer = new StringBuilder();
        Iterator<String> iterator = expressionVariables.iterator();
        while ( iterator.hasNext() ) {
            buffer.append( toBroEntityAttribute( iterator.next() ) );
            if ( iterator.hasNext() ) {
                buffer.append( ", " );
            }
        }
        return buffer.toString();
    }

    /**
     * A value expression is defined as ${}
     *
     * @param expressionVariable
     * @return
     */
    public static String toBroEntityAttribute(String expressionVariable) {
        if ( expressionVariable.startsWith( "${" ) && expressionVariable.endsWith( "}" ) ) {
            return "%" + expressionVariable.substring( 2, expressionVariable.length() - 1 ) + "%";
        }
        else {
            throw new IllegalArgumentException( String.format( "%s is not defined as ${expressionVariable}", expressionVariable ) );
        }
    }
}
```

sjaakd:
oops same mistake.. wrong button :smile:

gunnarmorling:
Hey @gsmet, any chance you could take a look at this?

gunnarmorling:
Yes, I think the PR is reasonable. Still I feel having a non-mocked test will be helpful to ensure that the API does the indented thing and everything fits together. Not sure though whether @sjaakd can add something or you want to do it? In any case it's your call. The other thing is to mention the new method in the docs.

gunnarmorling:
@sjaakd, btw. I got to say you seem to have quite some [bro culture](https://en.wikipedia.org/wiki/Bro_(subculture)) at your work place 🤣

sjaakd:
Yep.. @gsmet I can add a testcase. Seems that we are waiting on each other :smile: (I did not want to put a lot of effort in it if you guys don't see any value in the solution).

@gunnarmorling : it seems to be a Swedish cola brand as well... But in my work field it stands for Basis Registratie Ondergrond.. But then again.. you knew that already :wink:

gunnarmorling:
>  it seems to be a Swedish cola brand as well... 

LOL, didn't know that. I'll watch out for it when we go there this year :)

> But in my work field it stands for Basis Registratie Ondergrond.. But then again.. you knew that already

Yeah :) Just had to smile when I imagine all you bro's hacking on stuff in front of my inner eye :)

gunnarmorling:
Really happy to see this contribution, given TNO is such an extensive user of Bean Validation. Always nice to incorporate new features based on demand in the field. Thanks a lot!

gsmet:
@sjaakd I think the PR has value and it looks like something that I could add in the next alpha of 6.1.

As mentioned by Gunnar, apart from the test case, you need to take a look at the documentation and see if we need to add something there.

Feel free to invest as much time as you want in it, it won't be a waste of time :).

engine/src/main/java/org/hibernate/validator/internal/engine/MessageInterpolatorContext.java
```diff
@@ -37,17 +38,20 @@
...
 	public MessageInterpolatorContext(ConstraintDescriptor<?> constraintDescriptor,
-			Object validatedValue,
-			Class<?> rootBeanType,
-			Map<String, Object> messageParameters,
-			Map<String, Object> expressionVariables) {
+									  Object validatedValue,
+									  Class<?> rootBeanType,
+									  Map<String, Object> messageParameters,
+									  Map<String, Object> expressionVariables,
+									  Path propertyPath) {
```
gsmet: Let's not reindent things please.

And let's add `propertyPath` just after `rootBeanType`. It makes more sense.

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/AbstractValidationContext.java
```diff
@@ -294,13 +295,15 @@ private String interpolate(
 			Object validatedValue,
 			ConstraintDescriptor<?> descriptor,
 			Map<String, Object> messageParameters,
-			Map<String, Object> expressionVariables) {
+			Map<String, Object> expressionVariables,
+			Path path) {
```
gsmet: Let's add the path after the descriptor. It makes more sense.

engine/src/main/java/org/hibernate/validator/messageinterpolation/HibernateMessageInterpolatorContext.java
```diff
@@ -40,4 +41,9 @@
 	 * @since 5.4.1
 	 */
 	Map<String, Object> getExpressionVariables();
+
+	/**
+	 * @return the property path to the value from {@code rootBean}
```
gsmet: Please add `@since 6.1`

gsmet: Maybe `@return the property path to the validated constraint starting from {@code rootBean}` ?

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/AbstractValidationContext.java
sjaakd: fixed

engine/src/main/java/org/hibernate/validator/messageinterpolation/HibernateMessageInterpolatorContext.java
sjaakd: fixed

sjaakd:
@gunnarmorling , @gsmet I've processed your comments and added a test case. It is a simplified version of what we do in practice.

gsmet:
Made some additional editorialization but otherwise it looked good.

Just pushed to your branch a squashed version of it to get CI running. Will merge as soon as it's green.

gsmet:
Merged, thanks!

sjaakd:
@gsmet .. you too!

PR #1014, opened 2019-02-01
HV-1088 Making sure constraint validators are always properly released when using multiple custom constraint validator factories (backport to 5.4.X)

exbe:
As the title suggests, this PR back ports [same change ](https://github.com/hibernate/hibernate-validator/commit/9e2c0d99e9ed856b8b57cbac3aac8a3bbb1d47a6)as it was done for v6.

For more details, see #658 

@gunnarmorling FYI

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorManager.java
@@ -63,7 +63,13 @@ public class ConstraintValidatorManager {
 	/**
 	 * The most recently used non default constraint validator factory.
 	 */
-	private ConstraintValidatorFactory mostRecentlyUsedNonDefaultConstraintValidatorFactory;
+	private volatile ConstraintValidatorFactory mostRecentlyUsedNonDefaultConstraintValidatorFactory;
+
+	/**
+	 * Used for synchronizing access to {@link #mostRecentlyUsedNonDefaultConstraintValidatorFactory} (which can be
+	 * null itself).
+	 */
+	private final Object mostRecentlyUsedNonDefaultConstraintValidatorFactoryMutex = new Object();
 
 	/**
 	 * Cache of initialized {@code ConstraintValidator} instances keyed against validates type, annotation and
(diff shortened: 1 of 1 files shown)


gunnarmorling:
@exbe, it's our general strategy to support only one release line in community at a given time. Instead of maintaining older versions, we try to focus on adding new features in the current stable line and fixing bugs in there. 5.4 hasn't seen a release in more than a year, and that it's still mentioned on the downloads page is mainly for the reason that it is the last Bean Validation 1.1 compatible implementation (we probably should remove it there).

That said, I'd highly recommend the upgrade to HV 6, which does not only come with all the BV 2.0 features, but also many bug fixes applied since 5.4.x and a much better performance: http://in.relation.to/2018/03/20/bean-validation-benchmark-rerevisited/. The upgrade will be smooth (no-op really if you solely use the BV spec'ed APIs). Only if you use some of the experimental/incubating features in HV 5.x, you'd have to do some (not too large) adjustments in your code base.

I'll leave it to @gsmet (who took over as the HV project lead) to decide on your PR. Thanks for your efforts in any case.

gsmet:
@exbe so I agree there's an oversight on our part because it was supposed to be fixed in 5.4.1.Final and it apparently wasn't backported. AFAICS, it was my mistake, sorry about that.

The next question is the big question: do you have any valid reason to not upgrade to 6? It has been around for a while now and as Gunnar mentioned it, it's considerably better.

That being said, I'm not very comfortable with the latest release supporting BV 1.1 having an obvious concurrency issue so I think a backport/release is in order. I will see if the original commits are easily backportable. If so, I'll push a quick release but I won't do any public relation work about it.

gsmet:
Hum, in fact, it's a little more complicated than that. @gunnarmorling apparently, you forgot the last commit when you created the backport PR (I have the backported commits coming from contrib-gunnarmorling/HV-1088_5.4 but I don't have the last one) - or it was added afterwards and we forgot to push it too.

So we just have one commit missing: 9e2c0d99e9ed856b8b57cbac3aac8a3bbb1d47a6 .

I created https://hibernate.atlassian.net/browse/HV-1690 .

Will close this one as I will backport the original commit.

gsmet:
See https://github.com/hibernate/hibernate-validator/pull/1015 .

exbe:
> The next question is the big question: do you have any valid reason to not upgrade to 6? It has been around for a while now and as Gunnar mentioned it, it's considerably better.

Short answer: it is in progress actually, I was working on it in parallel.

Our use cases are little bit more complex and. Upgrade was almost smooth except one issue with validating objects with cyclic references, which I am busy investigating now.  Once this is resolved, it would pave the way to upgrade to BV 2.0, which would take months in corporate environment to certify upgrade, so having options would remove a lot of pressure on existing services in production.

amatiushkin:
@gsmet @gunnarmorling re:upgrade to v6

Here is the issue which prevents me from upgrading to latest and greatest: https://hibernate.atlassian.net/browse/HV-1692 ([test case](https://github.com/hibernate/hibernate-test-case-templates/pull/26))

gsmet:
Well, apparently not greatest...

I have seen the issue you opened. Didn’t have the time to take a look yet but will do as soon as time permits.

amatiushkin:
@gsmet it is greatest if you ask me :) and I am still yet to see a defect-less software with more than 1K lines of code.

feel free to ping me for a walkthrough. I am worry that my description of a problem is not clear enough.

PR #1016, opened 2019-02-16
HV-1689 : Added log error whenever exception occurs.

asutosh936:


asutosh936:
@gsmet @gunnarmorling - Please review

engine/src/main/java/org/hibernate/validator/internal/util/privilegedactions/GetInstancesFromServiceLoader.java
```diff
@@ -57,6 +63,7 @@ private GetInstancesFromServiceLoader(ClassLoader primaryClassLoader, Class<T> c
 				// ignore, because it can happen when multiple
 				// services are present and some of them are not class loader
 				// compatible with our API.
+				LOG.error( "Error while loading Instances {}",e );
```
gsmet: This message should be included in our `Log` class (with a parameter for the service name and a `@Cause`, you have tons of examples in the code).

The message should also probably be something like "An error occurred while loading an instance of service <service class>.".

asutosh936: @gsmet - Thanks for review. Have tried implementing the review comments and updated code.

gsmet:
I squashed the commits, made minor changes and merged. Thanks.

PR #1017, opened 2019-02-17
HV-1692 Custom group sequence might cause StackOverflowError on objects with cycles

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1692

Change when beans are marked as processed
- in case of redefined group sequences only the last group was marked as processed while all other groups were ignored. Marking bean as processed in the loop allows to catch all the groups in the sequence

As it turned out we were missing some of the processed units when dealing with group sequences. I've added a test from the ticket as a separate one, as I didn't find a better place for it.


gsmet:
Merged and backported to 6.0.

Will release a 6.0.15 and a new 6.1 alpha soon.

PR #1018, opened 2019-02-17
HV-1691 Rename PathImpl#getPathWithoutLeafNode to createCopyWithoutLeafNode

gsmet:
* https://hibernate.atlassian.net/browse/HV-1691

So this won't solve the whole issue of `PathImpl` and `NodeImpl` not being serializable by Jackson but it will solve the immediate issue by making the `getPathWithoutLeafNode()` method purpose more clear and not suitable for serialization.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
@@ -1277,7 +1277,7 @@ public class ValidatorImpl implements Validator, ExecutableValidator {
 			return true;
 		}
 
-		Path pathToObject = path.getPathWithoutLeafNode();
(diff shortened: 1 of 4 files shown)


gsmet:
@marko-bekhta could you check this one? TBH I wouldn't have made the change if I didn't think the name was better suggesting the intent of the method.

marko-bekhta:
Hi @gsmet
The change looks good. I wonder if maybe we should make this method static to match the style of all those create methods from `NodeImpl`?
```java
	public static PathImpl createCopyWithoutLeafNode(PathImpl path) {
		return new PathImpl( path.nodeList.subList( 0, path.nodeList.size() - 1 ) );
	}
```

gsmet:
@marko-bekhta good idea, I pushed an update. I let you check everything is OK? I will push a new alpha probably today or tomorrow.

marko-bekhta:
Looks good. Thanks!

PR #1022, opened 2019-03-01
HV-1700 Remove reference of absent valueextractor from test resources

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1700

Remove ValueExtractor service file as not needed
- as the value extractors and tests for it from this file was moved to BVTCK the file became redundant and can be removed to prevent tests from logging messages about troubles with missing VE


gsmet:
Merged, thanks!

PR #1023, opened 2019-03-01
HV-1698 Try to make ConstraintValidator constructors accessible if not public

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1698

Make constraint validator constructors accessible:
- update `NewInstance` action with makeAccessible switch
- use new action switch when instantiating constraint validators


gsmet:
I missed Gunnar's opening this issue. I can't say I'm excited about it. The spec references a few times explicitly the fact that we use the `public no-arg constructor`. And we also require it for a few other classes (for some the spec even requires we throw an exception if there is no public no-arg constructor).

@gunnarmorling do you really think it's worth not being consistent?

gsmet:
As mentioned in the issue, I'm not convinced this is an improvement.

PR #1024, opened 2019-03-03
HV-1701 Add Max/MinValidatorForInteger

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1701

- Add specific validators for min/max constraints and byte/short/integer types as the existing common number one makes additional actions that can be omitted in this case.
- Make checks in Number validator for DecimalMin/DecimalMax to check for double/float to capture infinity situation correctly. Also direct a check of Big* types to corresponding methods
- Add specific validators for DecimalMin/DecimalMax constraints and byte/short/integer types as the existing common number one makes additional actions that can be omitted in this case.

So it got me thinking about the DecimalMin/max as there's the same situation with Number validator. Hence I've made similar changes for that validator hierarchy as well.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/number/bound/NumberComparatorHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/number/bound/NumberComparatorHelper.java
@@ -33,8 +33,9 @@ final class NumberComparatorHelper {
 	}
 
 	public static int compare(Number number, long value, OptionalInt treatNanAs) {
-		// In case of comparing numbers before we compare them as two longs:
-		// 1. We need to check for floating point number as it should be treated differently in such case:
+		// In case of comparing numbers we need to check for special cases:
+		// 1. Floating point numbers should consider nan/infinity as values hence they should
+		// be directed to corresponding overloaded methods:
 		if ( number instanceof Double ) {
 			return compare( (Double) number, value, treatNanAs );
 		}
@@ -42,7 +43,7 @@ final class NumberComparatorHelper {
 			return compare( (Float) number, value, treatNanAs );
 		}
 
-		// 2. We need to check for big numbers so that we don't lose data when converting them to long:
+		// 2. For big numbers we don't want to lose any data so we just cast them and call corresponding methods:
 		if ( number instanceof BigDecimal ) {
 			return compare( (BigDecimal) number, value );
 		}
@@ -50,7 +51,14 @@ final class NumberComparatorHelper {
 			return compare( (BigInteger) number, value );
 		}
 
-		return Long.compare( number.longValue(), value );
+		// 3. For any integer types we convert them to long as we would do that anyway
+		// when comparing with a long value. And use corresponding method for longs:
+		if ( number instanceof Byte || number instanceof Integer || number instanceof Long || number instanceof Short ) {
+			return compare( number.longValue(), value );
+		}
+
+		// 4. As a fallback we convert the number to double:
+		return compare( number.doubleValue(), value, treatNanAs );
 	}
 
 	public static int compare(Double number, long value, OptionalInt treatNanAs) {
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/number/bound/decimal/DecimalNumberComparatorHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/number/bound/decimal/DecimalNumberComparatorHelper.java
@@ -32,8 +32,33 @@ final class DecimalNumberComparatorHelper {
 		return BigDecimal.valueOf( number ).compareTo( value );
 	}
 
-	public static int compare(Number number, BigDecimal value) {
-		return BigDecimal.valueOf( number.doubleValue() ).compareTo( value );
+	public static int compare(Number number, BigDecimal value, OptionalInt treatNanAs) {
+		// In case of comparing numbers we need to check for special cases:
+		// 1. Floating point numbers should consider nan/infinity as values hence they should
+		// be directed to corresponding overloaded methods:
+		if ( number instanceof Double ) {
+			return compare( (Double) number, value, treatNanAs );
+		}
+		if ( number instanceof Float ) {
+			return compare( (Float) number, value, treatNanAs );
+		}
+
+		// 2. For big numbers we don't want to lose any data so we just cast them and call corresponding methods:
+		if ( number instanceof BigDecimal ) {
+			return compare( (BigDecimal) number, value );
+		}
+		if ( number instanceof BigInteger ) {
+			return compare( (BigInteger) number, value );
+		}
+
+		// 3. For any integer types we convert them to long as we would do that anyway
+		// to create a BigDecimal instance. And use corresponding method for longs:
+		if ( number instanceof Byte || number instanceof Integer || number instanceof Long || number instanceof Short ) {
+			return compare( number.longValue(), value );
+		}
+
+		// 4. As a fallback we convert the number to double:
+		return compare( number.doubleValue(), value, treatNanAs );
 	}
 
 	public static int compare(Double number, BigDecimal value, OptionalInt treatNanAs) {
--- a/engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/bv/BaseMinMaxValidatorForNumberTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/bv/BaseMinMaxValidatorForNumberTest.java
@@ -75,6 +75,51 @@ public class BaseMinMaxValidatorForNumberTest {
 		assertEquals( constraint.isValid( BigInteger.valueOf( 1560000000 ), null ), !isMax );
 	}
 
+	protected void testValidatorByte(ConstraintValidator<?, Byte> constraint, boolean inclusive, boolean isMax) {
+		if ( inclusive ) {
+			assertTrue( constraint.isValid( (byte) 15, null ) );
+		}
+		else {
+			assertFalse( constraint.isValid( (byte) 15, null ) );
+		}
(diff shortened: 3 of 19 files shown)


engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/number/bound/decimal/DecimalNumberComparatorHelper.java
```diff
@@ -32,8 +32,26 @@ public static int compare(Long number, BigDecimal value) {
...
+		// 2. We need to check for big numbers so that we don't lose data when converting them to long:
+		if ( number instanceof BigDecimal ) {
+			return compare( (BigDecimal) number, value );
+		}
+		if ( number instanceof BigInteger ) {
+			return compare( (BigInteger) number, value );
+		}
+
+		// if we reached this step only integer numbers remains hence we convert to long:
+		return BigDecimal.valueOf( number.longValue() ).compareTo( value );
```
gsmet: Hmmm. I'm not sure it's such a good idea. People could have their specific decimal implementation of `Number`.

I think you could have special cases but at the end you should use `doubleValue()`. BTW, if we do that, all the comments of the method needs updating.

marko-bekhta: Honestly if someone have their own number I would expect them to add their own validator :smiley: 
Thinking more about it I think we should do 
```java
new BigDecimal( number.toString() ).compareTo( value )
```
And then there's the same situation in `NumberComparatorHelper` where we convert to long. We should do the same `BigDecimal` check there as well then. WDYT ?

gsmet: I don't think using `toString()` will be reliable enough. What I would do is special case all the known types and end up with a `double` based comparison for both `Max/Min` and `DecimalMax/Min`.

Does it make sense to you?

marko-bekhta: yes, I'll update the PR soon

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/bv/BaseMinMaxValidatorForNumberTest.java
```diff
@@ -74,6 +74,44 @@ protected void testValidatorBigInteger(ConstraintValidator<?, BigInteger> constr
 		assertEquals( constraint.isValid( BigInteger.valueOf( -1560000000 ), null ), isMax );
 		assertEquals( constraint.isValid( BigInteger.valueOf( 1560000000 ), null ), !isMax );
 	}
+	protected void testValidatorByte(ConstraintValidator<?, Byte> constraint, boolean inclusive, boolean isMax) {
+		if ( inclusive ) {
+			assertTrue( constraint.isValid( (byte) 15, null ) );
+		}
+		else {
+			assertFalse( constraint.isValid( (byte) 15, null ) );
+		}
```
gsmet: Wondering if it would be worth it to add a few more tests with the value not being the limit (one lower, one higher)?

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/bv/BaseMinMaxValidatorForNumberTest.java
```diff
@@ -74,6 +74,44 @@ protected void testValidatorBigInteger(ConstraintValidator<?, BigInteger> constr
 		assertEquals( constraint.isValid( BigInteger.valueOf( -1560000000 ), null ), isMax );
 		assertEquals( constraint.isValid( BigInteger.valueOf( 1560000000 ), null ), !isMax );
 	}
+	protected void testValidatorByte(ConstraintValidator<?, Byte> constraint, boolean inclusive, boolean isMax) {
```
gsmet: You need a new line above.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/bv/MinValidatorForNumberTest.java
```diff
@@ -92,6 +98,18 @@ private void testDecimalMin(DecimalMin m, boolean inclusive) {
...
+		constraint.initialize( m );
+		testValidatorByte( constraint, inclusive, false );
+
+		constraint = new DecimalMinValidatorForShort();
+		constraint.initialize( m );
+		testValidatorShort( constraint, inclusive, false );
+
+		constraint = new DecimalMinValidatorForInteger();
+		constraint.initialize( m );
+		testValidatorInteger( constraint, inclusive, false );
```
gsmet: Do we have tests validating a decimal value for the new validators (in the case of `DecimalMin/Max`)?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/number/bound/NumberComparatorHelper.java
```diff
@@ -33,24 +33,32 @@ public static int compare(Long number, long value) {
...
+		// be directed to corresponding overloaded methods:
 		if ( number instanceof Double ) {
 			return compare( (Double) number, value, treatNanAs );
 		}
 		if ( number instanceof Float ) {
 			return compare( (Float) number, value, treatNanAs );
 		}
 
-		// 2. We need to check for big numbers so that we don't lose data when converting them to long:
+		// 2. For big numbers we don't want to loose any data so we just cast them and call corresponding methods:
```
gsmet: s/loose/lose/ again :)

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/number/bound/NumberComparatorHelper.java
```diff
@@ -33,24 +33,32 @@ public static int compare(Long number, long value) {
...
 		}
 
-		return Long.compare( number.longValue(), value );
+		// 3. For any integer types we convert them to long as we would do that anyway
+		// when comparing with a long value. And use corresponding method for longs:
+		if ( number instanceof Byte || number instanceof Integer || number instanceof Long || number instanceof Short ) {
+			return compare( number.longValue(), value );
+		}
+
+		// 4. As a fallback we convert the numnber to double:
```
gsmet: s/numnber/number/

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/number/bound/decimal/DecimalNumberComparatorHelper.java
```diff
@@ -32,8 +32,33 @@ public static int compare(Long number, BigDecimal value) {
...
+		// 1. Floating point numbers should consider nan/infinity as values hence they should
+		// be directed to corresponding overloaded methods:
+		if ( number instanceof Double ) {
+			return compare( (Double) number, value, treatNanAs );
+		}
+		if ( number instanceof Float ) {
+			return compare( (Float) number, value, treatNanAs );
+		}
+
+		// 2. For big numbers we don't want to loose any data so we just cast them and call corresponding methods:
```
gsmet: s/loose/lose/

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/number/bound/decimal/DecimalNumberComparatorHelper.java
```diff
@@ -32,8 +32,33 @@ public static int compare(Long number, BigDecimal value) {
...
+			return compare( (BigInteger) number, value );
+		}
+
+		// 3. For any integer types we convert them to long as we would do that anyway
+		// to create a BigDecimal instance. And use corresponding method for longs:
+		if ( number instanceof Byte || number instanceof Integer || number instanceof Long || number instanceof Short ) {
+			return compare( number.longValue(), value );
+		}
+
+		// 4. As a fallback we convert the numnber to double:
```
gsmet: s/numnber/number/

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/bv/BaseMinMaxValidatorForNumberTest.java
marko-bekhta: I've added a test for 14/16 as 15 is a value set in the constraint.

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/bv/MinValidatorForNumberTest.java
marko-bekhta: Both Decimal and regular Min/Max tests are using same asserts. So we should have the test from the original issue as well for Decimal validators.

gsmet: Yup, that's what worried me a bit: do they test a decimal limit properly?

gsmet: @marko-bekhta not sure if you have seen this one? If you think we are safe enough, I trust you, just wanted to share my doubts :).

marko-bekhta: @gsmet  yes, thanks. I just got distracted with other work. I've just pushed an additional commit. We should squash it. I just made it as a separate so it'll be easier to see what I've added.

gsmet:
I merged it, thanks!

PR #1025, opened 2019-03-05
HV-1680 Avoid reflection by using instrumentation - build the enhancer

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1680

First commit just adds the interface and uses that interface when possible to get the values from the beans.
Second one adds a test with byte-buddy instrumentation. There are two cases one when the beans super class is object and the other one when it's a sublass of an enhanced bean. The difference is should we throw an exception when we reached the end of a method or call super method.

I'm also thinking that it might be a good idea to add another pair of methods to the interface which would give the list of "supported" properties (fields/getters). Having them in place would help us decide if we can use the enhanced bean interface or if we should fallback to regular reflection access for a particular property.

And a quick benchmark that I've put together :
```bash
Benchmark                                           Mode  Cnt    Score    Error   Units
NotSoSimpleValidation.testSimpleBeanValidationBar  thrpt   20  325.398 ±  4.597  ops/ms
NotSoSimpleValidation.testSimpleBeanValidationFoo  thrpt   20  357.233 ± 79.813  ops/ms
```
`Bar` is regular bean, `Foo` is enhanced one.

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -157,6 +157,11 @@
             <artifactId>jackson-annotations</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>net.bytebuddy</groupId>
+            <artifactId>byte-buddy</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/engine/src/main/java/org/hibernate/validator/engine/HibernateValidatorEnhancedBean.java
+++ b/engine/src/main/java/org/hibernate/validator/engine/HibernateValidatorEnhancedBean.java
@@ -8,8 +8,8 @@ package org.hibernate.validator.engine;
 
 /**
  * Hibernate Validator specific marker interface. Beans implementing this interface
- * would use corresponding {@link HibernateValidatorEnhancedBean#getFieldValue(String)}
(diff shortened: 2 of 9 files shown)


gsmet:
@marko-bekhta sorry it took so long but I will experiment to get this in Quarkus.

gsmet:
I made some adjustments (rebasing, fixing a few typos, added a `@since` and made ByteBuddy a `test` dependency).

I will now try to port the BB code to Gizmo and integrate everything in Quarkus.

gsmet:
@marko-bekhta could you have a look at the additional commits? They should be simple enough but better have another pair of eyes.

marko-bekhta:
@gsmet sure :smiley: 
looks good. I thought about adding another test to make sure that constants match the methods, but BB test should fail if they don't. So should be good to go. 

How is the integration with quarkus going ?

gsmet:
Well, it was going well but in the end it didn't improve the situation I wanted to improve so I didn't include it yet.

The heavy lifting is done but I would also like to generate only the needed fields and methods instead of all of them. And this, I still have to do.

Will take another look as soon as this part released.

gsmet:
Merged, thanks!

PR #1027, opened 2019-03-19
[HV-1704] use valid file URI in Windows

simkam:
https://hibernate.atlassian.net/browse/HV-1704


gsmet:
Good catch, merged, thanks!

PR #1030, opened 2019-04-01
HV-1706 ISBN-13 algorithm does not handle checksum 10

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1706

Add special case for '0' checksum of ISBN13
- need to make sure that '0' checksum digit is handled correctly for ISBN13. '0' could represent both 0 and 10 checksum.


gsmet:
Hmm, I'm wondering if we have the same issue for the ISBN-10?

Wikipedia says: `Therefore, the check digit is (11 minus the remainder of the sum of the products modulo 11) modulo 11. Taking the remainder modulo 11 a second time accounts for the possibility that the first remainder is 0. Without the second modulo operation the calculation could end up with 11 – 0 = 11 which is invalid.`

marko-bekhta:
Let me think about this one. It seems that for ISBN10 the algo is different. There's no `11 - smth`.

marko-bekhta:
I've changed the check algo for both cases. Instead of trying to find a checksum digit I just calculate the whole sum. This requires less operations. I've also added a few more valid ISBN tests to cover more cases.

gsmet:
Looks good to me. Good idea to take a step back, it's much cleaner now!

gsmet:
Merged and backported to 6.0! Thanks!

PR #1031, opened 2019-04-01
HV-1328 Add an option to validate class-level constraints only if all property constraints are valid

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1328

First commit has a change for what looked like a strange behavior. the value only for the last group was considered....

Any ideas on naming of the "mode" is also welcomed. For now PR is missing the doc.

### DIFF
--- a/documentation/src/main/asciidoc/ch12.asciidoc
+++ b/documentation/src/main/asciidoc/ch12.asciidoc
@@ -1001,3 +1001,57 @@ should be used as a default way of resolving the name (see how it is used in <<e
 - `org.hibernate.validator.spi.nodenameprovider.JavaBeanProperty` is an interface that holds metadata about a bean property. It
 extends `org.hibernate.validator.spi.nodenameprovider.Property` and provide some additional methods like `Class<?> getDeclaringClass()`
 which returns the class that is the owner of the property.
+
+[[section-fail-fast-on-property-violation]]
+=== Fail fast on property violation mode
+
+[WARNING]
+====
+This mode is marked as `@Incubating`, so it might be subject to change in the future.
+====
+
+Fail fast on property violation mode allows to perform class level constraint validation only if all property level constrints have passed without any violations.
+Enabling such mode should make writing class level constraint validators easier and cleaner, as the user will be working with valid properties.
+Hence, there will be no need in any additional `null` checks or data sanity checks.
+
+This mode can be enabled using one of the Hibernate Validator common ways.
+It can be done either programmatically (<<example-programmatic-fail-fast-on-property-violation>>), or through the property (<<example-property-fail-fast-on-property-violation>>).
+
+[[example-programmatic-fail-fast-on-property-violation]]
+.Using the fail fast on property violation validation mode
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter12/failfastonpropertyviolation/Book.java[tags=include]
+----
+
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter12/failfastonpropertyviolation/FailFastOnPropertyViolationTest.java[tags=include]
+----
+====
+
+Using this mode allows us to omit any additional checks in our class level validator.
+We don't need to be worried about any of the properties being `null` as we've dealt with it already, using standard constraints.
+
+[WARNING]
+====
+It is important to mention that only simple property constraints will be checked before going to class level ones. Any cascading constraints will be delayed till later.
+====
+
+.Implementation of `NonSelfPublishing` class level constrain validator
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter12/failfastonpropertyviolation/NonSelfPublishing.java[tags=include]
+----
+====
+
+[[example-property-fail-fast-on-property-violation]]
+.Setting the fail fast on property violation validation mode using an XML property
+====
+[source, XML, indent=0]
+----
+include::{resourcesdir}/org/hibernate/validator/referenceguide/chapter12/fail-fast-on-property-violation-validation.xml[]
+----
+====
--- a/engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
+++ b/engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
@@ -177,6 +177,15 @@ public interface BaseHibernateValidatorConfiguration<S extends BaseHibernateVali
 	String SHOW_VALIDATED_VALUE_IN_TRACE_LOGS = "hibernate.validator.show_validated_value_in_trace_logs";
 
 	/**
+	 * Property corresponding to the {@link #failFastOnPropertyViolation} method.
+	 * Accepts {@code true} or {@code false}. Defaults to {@code false}.
+	 *
+	 * @since 9.0
+	 */
+	@Incubating
+	String FAIL_FAST_ON_PROPERTY_VIOLATION = "hibernate.validator.fail_fast_on_property_violation";
+
+	/**
 	 * <p>
 	 * Returns the {@link ResourceBundleLocator} used by the
 	 * {@link Configuration#getDefaultMessageInterpolator() default message
@@ -501,4 +510,18 @@ public interface BaseHibernateValidatorConfiguration<S extends BaseHibernateVali
 	 */
 	@Incubating
 	S showValidatedValuesInTraceLogs(boolean enabled);
+
+	/**
+	 * En- or disables the skipping of class level constraints based on validation of property level ones. When this
+	 * mode is enabled the validation of class level constraints will not be performed if any of the property level
+	 * constraints generated a violation.
+	 *
+	 * @param failFastOnPropertyViolation {@code true} to enable the skipping mode, {@code false} otherwise.
+	 *
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 9.0
+	 */
+	@Incubating
+	S failFastOnPropertyViolation(boolean failFastOnPropertyViolation);
 }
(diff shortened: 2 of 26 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -471,7 +471,7 @@ private void validateConstraintsForCurrentGroup(BaseBeanValidationContext<?> val
 
 						for ( Group defaultSequenceMember : groupOfGroups ) {
 							validationSuccessful = validateConstraintsForSingleDefaultGroupElement( validationContext, valueContext, validatedInterfaces, clazz,
-									metaConstraints, defaultSequenceMember );
+									metaConstraints, defaultSequenceMember ) && validationSuccessful;
```
gsmet: Could you open a separate issue for that one? A test case might be in order too as it looks a lot like a bug!

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -119,6 +119,16 @@
 	@Incubating
 	String GETTER_PROPERTY_SELECTION_STRATEGY_CLASSNAME = "hibernate.validator.getter_property_selection_strategy";
```
gsmet: typo in commit message: s/thorough/through/

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -119,6 +119,16 @@
...
 
+
+	/**
+	 * Property corresponding to the {@link #ignoreClassLevelOnPropertyFail} method.
+	 * Accepts {@code true} or {@code false}. Defaults to {@code false}.
+	 *
+	 * @since 6.1.0
+	 */
+	@Incubating
+	String IGNORE_CLASS_LEVEL_ON_PROPERTY_FAIL = "hibernate.validator.ignore_class_level_on_property_fail";
```
gsmet: I wonder if we should try to find a naming similar to the fail fast mode as it's really similar to it.

Maybe `fail_fast_on_property_violation`?

It's not easy to find a right name for this sort of configuration knob.

Happy to discuss it further (maybe it will be easier on Zulip).

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java
```diff
@@ -463,15 +463,19 @@ private void validateConstraintsForCurrentGroup(BaseBeanValidationContext<?> val
...
 				while ( defaultGroupSequence.hasNext() ) {
 					for ( GroupWithInheritance groupOfGroups : defaultGroupSequence.next() ) {
 						boolean validationSuccessful = true;
 
 						for ( Group defaultSequenceMember : groupOfGroups ) {
-							validationSuccessful = validateConstraintsForSingleDefaultGroupElement( validationContext, valueContext, validatedInterfaces, clazz,
-									metaConstraints, defaultSequenceMember ) && validationSuccessful;
+							boolean propertyValidationResult = validateConstraintsForSingleDefaultGroupElement( validationContext, valueContext, validatedInterfaces, clazz, propertyMetaConstraints, defaultSequenceMember );
+							validationSuccessful = propertyValidationResult && validationSuccessful;
+							if ( !validationContext.isIgnoreClassLevelOnPropertyFailModeEnabled() || propertyValidationResult ) {
```
gsmet: I'm wondering if having these two calls could slow down things. I wonder if we should have a shortcut if this mode is disabled.

Also wondering if we should test if the set is empty (can't say if it will be a win though but it looks like something we should do - and bench).

In any case, we should test it.

engine/src/test/java/org/hibernate/validator/test/internal/engine/groups/sequence/SequenceTest.java
```diff
@@ -0,0 +1,89 @@
...
+		Validator validator = ValidatorUtil.getValidator();
+
+		Set<ConstraintViolation<Foo>> violations = validator.validate( new Foo( null, "", null ) );
+
+		assertThat( violations ).containsOnlyViolations(
+				violationOf( NotNull.class ).withProperty( "str1" )
+		);
+	}
+
+	interface Group1 extends Group11, Group12, Group13, Group14, Group15, Group16, Group17, Group18, Group19 {
```
marko-bekhta: As that loop in which `validationSuccessful` is set is based on a `Set` it actually doesn't give consistent results. To be able to reproduce it more often I've added a lot of groups, which increase the probability that `true` will override the result in the end.

marko-bekhta:
And here's the benchmark results that I've got from yesterday:
```bash
6.1
Benchmark                                   Mode  Cnt     Score     Error   Units
SimpleValidation.testSimpleBeanValidation  thrpt   20  1595.544 ± 127.297  ops/ms

CascadedValidation.testCascadedValidation                                                  thrpt   20   377.533 ±  6.882  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1012.516 ± 12.189   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1147.559 ± 13.166   ops/s

current:
Benchmark                                   Mode  Cnt     Score     Error   Units
SimpleValidation.testSimpleBeanValidation  thrpt   20  1628.147 ± 185.557  ops/ms

CascadedValidation.testCascadedValidation                                                  thrpt   20   332.512 ±  5.129  ops/ms
CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems  thrpt   20  1039.538 ± 12.060   ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                    thrpt   20  1127.804 ± 13.872   ops/s
```

gsmet:
@marko-bekhta do we have a benchmark with property constraints + class level constraints? Because that would be an interesting one :).

I'm not sure the others are significant in this case. Although we should probably have a closer look to `CascadedValidation.testCascadedValidation`.

marko-bekhta:
jenkins retest this please

marko-bekhta:
I've added another benchmark that has both class and property level constraints. The results are:
**no changes:**
```
Benchmark                                                Mode  Cnt     Score    Error   Units
SimpleClassPropertyValidation.testSimpleBeanValidation  thrpt   20  1452.835 ± 21.071  ops/ms
```
**with changes:**
```
Benchmark                                                Mode  Cnt     Score    Error   Units
SimpleClassPropertyValidation.testSimpleBeanValidation  thrpt   20  1438.992 ± 216.247  ops/ms
```
I also tried to run the cascading benchmarks again. The results are all around the place :roll_eyes: . I'll try to run these on my laptop later today to see how they would behave there.

marko-bekhta:
While looking through the PRs I've realized that I had not added those benchmarks.... So here they are:
```bash
#nochange (current)
Benchmark                                                Mode  Cnt     Score    Error   Units
SimpleClassPropertyValidation.testSimpleBeanValidation  thrpt   20  1694.371 ± 22.571  ops/ms

#chagne (changes in this PR)
Benchmark                                                Mode  Cnt     Score    Error   Units
SimpleClassPropertyValidation.testSimpleBeanValidation  thrpt   20  1518.953 ± 29.393  ops/ms

# update (not in this PR, additional if to check if the option is enabled and if not use the "current" all constraints collections)
Benchmark                                                Mode  Cnt     Score    Error   Units
SimpleClassPropertyValidation.testSimpleBeanValidation  thrpt   20  1679.028 ± 33.595  ops/ms
```

gsmet:
@marko-bekhta so I think it makes it clear we need the additional test. Any chance you could push this to the PR (your report above says it wasn't pushed)?

gsmet:
I need to do a quick release so I'm merging the first commit to master and 6.0 as it looks like a bug to me.

marko-bekhta:
@gsmet yes, it totally make sense to break this into two pieces. I'll look for that additional commit and push it here when I find it :)

gsmet:
@marko-bekhta hey Marko. I tried to contact you by email but I didn't hear from you? Did you get my email from December?

documentation/src/main/asciidoc/ch12.asciidoc
```diff
@@ -855,3 +855,52 @@ should be used as a default way of resolving the name (see how it is used in <<e
 - `org.hibernate.validator.spi.nodenameprovider.JavaBeanProperty` is an interface that holds metadata about a bean property. It
 extends `org.hibernate.validator.spi.nodenameprovider.Property` and provide some additional methods like `Class<?> getDeclaringClass()`
 which returns the class that is the owner of the property.
+
+[[section-fail-fast-on-property-violation]]
+=== Fail fast on property violation mode
+
+Fail fast on property violation mode allows to perform class level constraint validation only if all property level constrints have passed without any violations.
+Enabling such mode should make writing class level constraint validators easier and cleaner, as the user will be working with valid properties.
+Hance there will be no need in any additional `null` checks or data sanity checks.
```
yrodiere, line 864:
```suggestion
Hence there will be no need in any additional `null` checks or data sanity checks.
```

marko-bekhta: thanks for noticing the typo :wink: updated the text and rebased the branch as it was quite outdated

marko-bekhta:
I've rebased this PR and tried the benchmarks:
Changes:
```java
Benchmark                                                Mode  Cnt      Score    Error   Units
SimpleClassPropertyValidation.testSimpleBeanValidation  thrpt   20  12765.571 ± 84.695  ops/ms
```
main:
```java
Benchmark                                                Mode  Cnt      Score    Error   Units
SimpleClassPropertyValidation.testSimpleBeanValidation  thrpt   20  12672.389 ± 11.129  ops/ms
```

The longer a PR is opened, the better the numbers get (https://github.com/hibernate/hibernate-validator/pull/1031#issuecomment-521964665) 😆 😄 

Jokes aside, I'm not sure whether we should merge it or just close it, as the community has shown little interest in it all this time. 

I'll let you decide what to do about it @yrodiere

yrodiere:
I see the value in this feature, though I'm not sure it's a silver bullet. I suspect that anyone doing a `if ( bean instanceof ) { ... } else if ... ` in their class-level validation constraint will discover the hard way that subclass property constraints haven't been validated at that point.

Anyway... I'd vote for marking this feature as incubating, merging it and moving on :)

marko-bekhta:
> Anyway... Assuming the effect is minimal when the feature isn't enabled, I'd vote for marking this feature as incubating, merging it and moving on :)

thanks 😃. It's marked as incubating. I'll merge it into 9.0, and we can see if we get any feedback on it. 🤞🏻

PR #1032, opened 2019-04-27
Fix bug that Regon validation may accept incorrect length Regon

michaldo:
Current Regon validator implementation has a bug: it accepts for example
1234567845 - but Regon may have only 9 or 14 digits.

Regon validator is implemented from scratch, as existing design follows
yo-yo antipattern and is impossimble to use.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/pl/REGONValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/pl/REGONValidator.java
@@ -6,46 +6,62 @@
  */
 package org.hibernate.validator.internal.constraintvalidators.hv.pl;
 
-import java.util.List;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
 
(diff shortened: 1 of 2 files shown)


gsmet:
ok to test

gsmet:
@marko-bekhta IIRC you were the original implementor of these validators? Could you take a close look at this PR?

marko-bekhta:
@michaldo thanks for reporting this issue and for your work. I've opened A JIRA ticket based on it - https://hibernate.atlassian.net/browse/HV-1709 
As it turned out all 3 polish identification numbers - NIP/PESEL/REGON are suffering from the same issue. Both shorter and longer values would pass the validation. 
Because of this and additional logic of accepting/rejecting non-digits I'd prefer we keep the original implementation in place and add a check for the length of validated values, rather than creating 3 implementations with the similar logic.

I'll open a PR to address all 3 constraints shorty. It would be great if you could take a look and provide any additional feedback. Thanks!

gsmet:
Superseded by https://github.com/hibernate/hibernate-validator/pull/1034 .

Thanks anyway!

PR #1035, opened 2019-05-07
HV-1707 Add ISBN.Type.ANY

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1707

Add ISBN.Type.ANY. This new type allows to validate values that are either ISBN10 or ISBN13

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraints/ISBN.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/ISBN.java
@@ -67,9 +67,13 @@ public @interface ISBN {
 	/**
 	 * Defines the ISBN length. Valid lengths of ISBNs are {@code 10} and {@code 13}
 	 * which are represented as {@link Type#ISBN_10} and {@link Type#ISBN_13} respectively.
+	 * <p>
+	 * Using {@link Type#ANY} allows to validate values that could either be ISBN10 or ISBN13.
+	 * In such case, ISBN type would be determined by the length of the corresponding value.
 	 */
 	enum Type {
 		ISBN_10,
-		ISBN_13
+		ISBN_13,
+		ANY,
 	}
 }
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ISBNValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ISBNValidator.java
@@ -6,13 +6,13 @@
  */
 package org.hibernate.validator.internal.constraintvalidators.hv;
 
-import java.util.function.Function;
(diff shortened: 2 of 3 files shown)


yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

yrodiere:
Jenkins, retest this please

engine/src/main/java/org/hibernate/validator/constraints/ISBN.java
```diff
@@ -67,9 +67,13 @@
...
 	 * which are represented as {@link Type#ISBN_10} and {@link Type#ISBN_13} respectively.
+	 * <p>
+	 * Using {@link Type#ANY} allows to validate values that could either be ISBN10 or ISBN13.
+	 * In such case, ISBN type would be determined by the length of the corresponding value.
 	 */
 	enum Type {
 		ISBN_10,
-		ISBN_13
+		ISBN_13,
+		ANY,
```
yrodiere: Since I was working on the CI failures, I couldn't help but notice this change.

So, my two cents: allowing an array of types would be more future-proof than adding an `ANY` type.
If, one day, we want to add the `SBN` type (basically the ancestor of ISBN, 9 digits, you just need to prefix it with `0` to convert to an ISBN10 number), users won't be able to pick "just SBN or ISBN 10", for example, unless we add yet another type, `SBN_OR_ISBN_10`.

That was my two cents, I'll be on my way...

gsmet: I think it's a bit unfortunate we already have a default or I would have made the default ANY.

I like the fact that we have a shortcut for `ANY` as I would expect it to be often used.

We could make it an array later if we support more.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ISBNValidator.java
```diff
@@ -54,37 +43,101 @@ public boolean isValid(CharSequence isbn, ConstraintValidatorContext context) {
...
+				int length = isbn.length();
+				if ( length == 10 ) {
+					return checkChecksumISBN10( isbn );
+				}
+				else if ( length == 13 ) {
+					return checkChecksumISBN13( isbn );
+				}
+				throw new IllegalStateException( "Invalid/unsupported isbn value length" );
+			}
+		},
```
gsmet: I would remove the `,`. I know modern JVM doesn't have a problem with it but I'm not a big fan of it.

marko-bekhta:
merged.

PR #1036, opened 2019-05-14
Build fixes

gsmet:
@marko-bekhta can you take a look at it? On my laptop, the first commit solves the Javadoc issue with OpenJDK **11.0.3** (the micro is important).

I also fixed an unrelated OSGi issue noticed while building.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -211,19 +211,13 @@
             </properties>
         </profile>
         <profile>
-            <id>jdk12+</id>
-            <activation>
-                <jdk>[12,)</jdk>
-            </activation>
-            <properties>
-                <javadoc.additional.options>-html5 -source 8</javadoc.additional.options>
-            </properties>
-        </profile>
-        <profile>
             <id>jdk11+</id>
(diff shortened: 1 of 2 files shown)


marko-bekhta:
hmm, tried to build it locally as well - still fails for me with this version:
```bash
java -version
openjdk version "11.0.3" 2019-04-16
OpenJDK Runtime Environment (build 11.0.3+7-Ubuntu-1ubuntu218.04.1)
OpenJDK 64-Bit Server VM (build 11.0.3+7-Ubuntu-1ubuntu218.04.1, mixed mode, sharing)
```
I also tried to use `<source>8</source>` in parent pom configuration (https://bugs.openjdk.java.net/browse/JDK-8212233) instead of `-source 8` and it worked.
Also there's `3.1.0` version of javadoc plugin. Maybe it's worth to update it as well ?

gsmet:
Ah cool, I was wondering what was the issue :).

gsmet:
Will wait for CI before backporting the JDK 11 fix.

PR #1038, opened 2019-06-05
HV-1718 Support validating unconstrained beans with PredefinedScopeValidatorFactory

gsmet:
@marko-bekhta could you take a look at that one, please?

I have been a bit too pedantic considering that only initialized classes could be validated as, well, unconstrained bean can also be validated...

This fixes it. The main drawback is that the framework really needs to be thorough in initializing everything, otherwise the constraints are not validated. But I don't think we have other options.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/PredefinedScopeBeanMetaDataManager.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/PredefinedScopeBeanMetaDataManager.java
@@ -8,44 +8,57 @@ package org.hibernate.validator.internal.metadata;
 
 import static org.hibernate.validator.internal.util.CollectionHelper.newArrayList;
 
-import java.lang.invoke.MethodHandles;
+import java.lang.annotation.ElementType;
+import java.lang.reflect.Executable;
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import javax.validation.metadata.BeanDescriptor;
+import javax.validation.metadata.ConstraintDescriptor;
+import javax.validation.metadata.ConstructorDescriptor;
+import javax.validation.metadata.ElementDescriptor.ConstraintFinder;
+import javax.validation.metadata.MethodDescriptor;
+import javax.validation.metadata.MethodType;
+import javax.validation.metadata.PropertyDescriptor;
+import javax.validation.metadata.Scope;
 
 import org.hibernate.validator.internal.engine.ConstraintCreationContext;
 import org.hibernate.validator.internal.engine.MethodValidationConfiguration;
+import org.hibernate.validator.internal.engine.groups.Sequence;
 import org.hibernate.validator.internal.engine.groups.ValidationOrderGenerator;
 import org.hibernate.validator.internal.metadata.aggregated.BeanMetaData;
 import org.hibernate.validator.internal.metadata.aggregated.BeanMetaDataBuilder;
 import org.hibernate.validator.internal.metadata.aggregated.BeanMetaDataImpl;
+import org.hibernate.validator.internal.metadata.aggregated.ExecutableMetaData;
+import org.hibernate.validator.internal.metadata.aggregated.PropertyMetaData;
 import org.hibernate.validator.internal.metadata.core.AnnotationProcessingOptions;
 import org.hibernate.validator.internal.metadata.core.AnnotationProcessingOptionsImpl;
+import org.hibernate.validator.internal.metadata.core.MetaConstraint;
+import org.hibernate.validator.internal.metadata.facets.Cascadable;
 import org.hibernate.validator.internal.metadata.provider.AnnotationMetaDataProvider;
 import org.hibernate.validator.internal.metadata.provider.MetaDataProvider;
 import org.hibernate.validator.internal.metadata.raw.BeanConfiguration;
 import org.hibernate.validator.internal.properties.javabean.JavaBeanHelper;
-import org.hibernate.validator.internal.util.CollectionHelper;
 import org.hibernate.validator.internal.util.ExecutableHelper;
 import org.hibernate.validator.internal.util.ExecutableParameterNameProvider;
 import org.hibernate.validator.internal.util.classhierarchy.ClassHierarchyHelper;
 import org.hibernate.validator.internal.util.classhierarchy.Filters;
-import org.hibernate.validator.internal.util.logging.Log;
-import org.hibernate.validator.internal.util.logging.LoggerFactory;
 import org.hibernate.validator.metadata.BeanMetaDataClassNormalizer;
 
 public class PredefinedScopeBeanMetaDataManager implements BeanMetaDataManager {
 
-	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
-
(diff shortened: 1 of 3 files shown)


engine/src/test/java/org/hibernate/validator/test/predefinedscope/PredefinedScopeValidatorFactoryTest.java
```diff
@@ -190,11 +191,26 @@ public void testUninitializedLocale() {
...
+		Set<Class<?>> beanMetaDataToInitialize = new HashSet<>();
+		beanMetaDataToInitialize.add( Bean.class );
+
+		ValidatorFactory validatorFactory = Validation.byProvider( PredefinedScopeHibernateValidator.class )
+				.configure()
+				.initializeBeanMetaData( beanMetaDataToInitialize )
+				.initializeLocales( Collections.singleton( Locale.ENGLISH ) )
+				.buildValidatorFactory();
+
+		// As we don't have any metadata for BeanProxy, we consider it is not constrained at all.
```
marko-bekhta: that's interesting. So in case something will create a proxy that we don't know about everything would "break". Maybe instead of storing the "final" metadata we could create "descriptors" for the beans that were requested for initialization and then create "final" metadata based only on them ?

marko-bekhta: basically instead of predefined metadata manager, have predefined metadata provider ... not sure it would work though

engine/src/main/java/org/hibernate/validator/internal/metadata/PredefinedScopeBeanMetaDataManager.java
```diff
@@ -71,32 +84,37 @@ public PredefinedScopeBeanMetaDataManager(ConstraintCreationContext constraintCr
...
 	@Override
 	public <T> BeanMetaData<T> getBeanMetaData(Class<T> beanClass) {
-		BeanMetaData<T> beanMetaData = (BeanMetaData<T>) beanMetaDataMap.get( beanMetaDataClassNormalizer.normalize( beanClass ) );
+		Class<?> normalizedBeanClass = beanMetaDataClassNormalizer.normalize( beanClass );
+		BeanMetaData<T> beanMetaData = (BeanMetaData<T>) beanMetaDataMap.get( normalizedBeanClass );
 		if ( beanMetaData == null ) {
-			throw LOG.uninitializedBeanMetaData( beanClass );
+			// note that if at least one element of the hierarchy is constrained, the child classes should really be initialized
+			// otherwise they will be considered unconstrained.
+			beanMetaData = (BeanMetaData<T>) beanMetaDataMap.computeIfAbsent( normalizedBeanClass, UninitializedBeanMetaData::new );
```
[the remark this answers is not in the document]
gsmet: No, it will happen often. I know a few applications where it's the most common path for the use of HV.

engine/src/main/java/org/hibernate/validator/internal/metadata/PredefinedScopeBeanMetaDataManager.java
```diff
@@ -170,4 +188,179 @@ private static AnnotationProcessingOptions getAnnotationProcessingOptionsFromNon
...
+
+		private final BeanDescriptor beanDescriptor;
+
+		private final List<Class<? super T>> classHierarchy;
+
+		@SuppressWarnings("unchecked")
+		private UninitializedBeanMetaData(Class<T> beanClass) {
+			this.beanClass = beanClass;
+			this.classHierarchy = (List<Class<? super T>>) (Object) ClassHierarchyHelper.getHierarchy( beanClass, Filters.excludeInterfaces() );
+			this.beanDescriptor = new UninitializedBeanDescriptor( beanClass );
```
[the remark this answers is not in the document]
gsmet: Yeah, maybe. I'm not sure it's worth it to be honest.

Let's leave it at that and you will be able to tell me "I told you so" in a few months :).

engine/src/main/java/org/hibernate/validator/internal/metadata/PredefinedScopeBeanMetaDataManager.java
```diff
@@ -170,4 +188,179 @@ private static AnnotationProcessingOptions getAnnotationProcessingOptionsFromNon
...
+		}
+
+		@Override
+		public List<Class<?>> getDefaultGroupSequence(T beanState) {
+			throw new IllegalStateException( "Metadata has not been initialized for bean of type " + beanClass.getName() );
+		}
+
+		@Override
+		public Iterator<Sequence> getDefaultValidationSequence(T beanState) {
+			throw new IllegalStateException( "Metadata has not been initialized for bean of type " + beanClass.getName() );
```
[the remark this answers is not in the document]
gsmet: This one really shouldn't be called, otherwise it's that we started somehow to validate the bean and it shouldn't happen.

I think it's safe.

engine/src/test/java/org/hibernate/validator/test/predefinedscope/PredefinedScopeValidatorFactoryTest.java
gsmet: Yes, I agree it's not ideal and that throwing an exception was more comfortable.

I thought about this a bit and I think that would be a significant amount of work as we would need to check all the operations we do when we assemble the metadata and see if they are doing reflection operations.

It's probably something we should do at some point.

marko-bekhta: yeah I agree that it'll take much time. So I think we should be ok as it is now. This is only going into the alpha, right? so we might still have time to do it differently till the final version :) or at least try and see if it's possible..

gsmet: @marko-bekhta I thought a bit about it. I don't think we can reuse the standard path of merging metadata as I'm pretty sure it requires reflection (for instance to determine if a method overrides another) and we don't want any at this point.

What I was wondering is if we should scan the class hierarchy and take into account the first class with existing metadata and then forge a valid metadata instance with the hypothesis that the current class doesn't have any added constraint. I'm not entirely sure it's doable to be honest as I think we will still have the issue with potentially overriding methods. Even if unconstrained, we will have to determine if they override a constrained method.

marko-bekhta: yeah.. the behavior of overridden methods would become unpredictable. What bothered me was that we just "silently" make a bean "unconstrained". I'm not sure that it's worth to make a config knob to fail on uninitialized bean. But maybe let's at least keep a log message instead of removed `throw LOG.uninitializedBeanMetaData( beanClass );` . It'll make one log entry when initializing a metadata for an "unexpected" bean.

gsmet: Unfortunately, I don't think that would fly. People would have a lot of warnings if they validate unconstrained beans.

I think we need to have faith in the implementation as I don't see any other satisfying solution.

marko-bekhta: That's right.... And if we make it a debug, nobody will see it anyway :smile: 

Can't think of anything else we can do here. Should I merge it in or are we waiting on something else ?

gsmet: Hum. Apparently, I already merged it :). Closing then.

PR #1040, opened 2019-06-21
HV-1720  Support bounded wildcard types in container value unwrapping

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1720

Add wildcard type to allowed reference types list.

As it was mentioned on SO this `isReferenceType()` was only used prior to the call to `getErasedType()` which can handle wildcard types. hence it seems appropriate to include the wildcard type in the assert as well.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/util/TypeHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/util/TypeHelper.java
@@ -538,6 +538,7 @@ public final class TypeHelper {
 	 * <li>a parameterized type</li>
 	 * <li>a type variable</li>
 	 * <li>the null type</li>
+	 * <li>a wildcard type</li>
 	 * </ul>
 	 *
 	 * @param type the type to check
@@ -551,7 +552,8 @@ public final class TypeHelper {
 				|| type instanceof Class<?>
 				|| type instanceof ParameterizedType
 				|| type instanceof TypeVariable<?>
-				|| type instanceof GenericArrayType;
+				|| type instanceof GenericArrayType
+				|| type instanceof WildcardType;
 	}
 
 	private static boolean isArraySupertype(Class<?> type) {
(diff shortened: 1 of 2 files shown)


gsmet:
I’m not totally sure but I think we should check if it has an upper bound defined because I think the point of this method is to be sure we get a real type. WDYT?

marko-bekhta:
@gsmet I've fixed a typo in the test and added 2 more cases. So now we will check upper/lower bounded wildcards + just a wildcard. I cannot think of a case where there's no upper bound. Wouldn't it be at least `Object` ? If such case is possible I suppose we can do something like:
```java
|| ( type instanceof WildcardType && ( (WildcardType) type ).getUpperBounds().length > 0 );
```
instead of the current change

gsmet:
Could you check what happens with `List` or `List<?>`? I'm not totally sure you end up with a proper `Object` type that you can use afterwards.

marko-bekhta:
Sorry for the delay in replying. 
In case of `List` we just get a regular `Class` so we don't have the "issue". And in case of ` List<@NotNull ?> list` we have a `WildcardType` with empty lower bound and `Object` as an upper bound. I've added a test that covers this case.

gsmet:
@marko-bekhta last question about this one: if we have a lower bound defined, is the upper bound still defined and still `Object`?

marko-bekhta:
@gsmet yes, I've double checked it just right now. If we have only limit on lower bound, the upper bound would be `Object`.

gsmet:
@marko-bekhta I'm wondering if we should backport this one. What would be your inclination?

marko-bekhta:
We probably have to add it to 6.0 unless we are going to switch to 6.1 any time soon. Seems like this is not a very common case though... as nobody complied till now :smiley:

gsmet:
Merged to master and 6.0. Thanks!

PR #1041, opened 2019-07-01
HV-1721 Take into account HV-specific config with predefined scope VF

gsmet:
https://hibernate.atlassian.net/browse/HV-1721

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/predefinedscope/PredefinedScopeConstraintDefinitionContributorTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/predefinedscope/PredefinedScopeConstraintDefinitionContributorTest.java
@@ -1,0 +1,94 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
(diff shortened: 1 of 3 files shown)


engine/src/test/java/org/hibernate/validator/test/predefinedscope/PredefinedScopeConstraintDefinitionContributorTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.testng.annotations.Test;
+
+/**
+ * Test related to constraint validator discovery in the predefined scope case.
+ *
+ * @author Hardy Ferentschik
+ * @author Guillaume Smet
+ */
+@TestForIssue(jiraKey = "HV-1721")
+public class PredefinedScopeConstraintDefinitionContributorTest {
```
marko-bekhta: Maybe we should do this test the same way we have `AbstractConstructorValidationTest` and 2 impl for annotations and xml. In this case it'll be predefined and regular impls.

gsmet: Yeah, the thing is I don't like to change preexisting tests but if we end up having a lot more, I agree we should do something about it.

In general, most of the tests should be tested for both but that would require changing so many things and I don't think we have the bandwidth.

My estimated tolerance is of 3 copied tests :).

PR #1042, opened 2019-07-13
HV-1723 Provide a DigitsValidatorForMonetaryAmount to support @Digits on MonetaryAmounts

darioseidl:
Like the money validators for `DecimalMin`/`DecimalMax`, this new validator for `Digits` obtains a `BigDecimal` from the `MonetaryAmount` and validates that.

https://hibernate.atlassian.net/browse/HV-1723

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/money/DigitsValidatorForMonetaryAmount.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/money/DigitsValidatorForMonetaryAmount.java
@@ -1,0 +1,62 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.constraintvalidators.bv.money;
+
+import java.lang.invoke.MethodHandles;
+import java.math.BigDecimal;
+import javax.money.MonetaryAmount;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.constraints.Digits;
+
+import org.hibernate.validator.internal.util.logging.Log;
+import org.hibernate.validator.internal.util.logging.LoggerFactory;
+
+/**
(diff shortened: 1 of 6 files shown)


engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/money/DigitsValidatorForMonetaryAmount.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.constraintvalidators.bv.money;
```
marko-bekhta: @gsmet I saw you mentioning hv/bv packages on the old money related PR. It was probably for this one though :smiley: 
I'd keep this one in bv. This way we would have all money related validators in one package.

gsmet: Yeah, it was for this PR, not sure how I ended up on the other one.

In this case, I don't think consistency is a good idea: it's important to know that these validators are not part of Bean Validation. So that we can think about moving them to the spec at some point.

gsmet:
I would just add one other comment: let's make it one atomic commit for the whole thing.

Thanks for your contribution.

darioseidl:
Thanks for the feedback. I updated the PR with accordingly, except for moving classes to the `hv` package. Do you mean to move all money related classes or just the new digits validator and test?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/money/DigitsValidatorForMonetaryAmount.java
gsmet: OK, I took a closer look and, AFAICS, as long as it is a BV constraint, the validators are in the `bv` package.

So let's leave it at that. Even if I'm not entirely convinced it's a good idea.

gsmet:
Merged, thanks! I also added you to our `copyright.txt` file.

PR #1043, opened 2019-07-17
HV-1726 Make PredefinedScopeHibernateValidatorFactory extend HibernateValidatorFactory

gsmet:
* https://hibernate.atlassian.net/browse/HV-1726

I think I have been too pedantic on this one.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/PredefinedScopeValidatorFactoryImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/PredefinedScopeValidatorFactoryImpl.java
@@ -240,8 +240,9 @@ public class PredefinedScopeValidatorFactoryImpl implements PredefinedScopeHiber
 
 	@Override
 	public <T> T unwrap(Class<T> type) {
-		//allow unwrapping into public super types
-		if ( type.isAssignableFrom( HibernateValidatorFactory.class ) ) {
+		// allow unwrapping into public super types
+		if ( type.isAssignableFrom( PredefinedScopeHibernateValidatorFactory.class )
+				|| type.isAssignableFrom( HibernateValidatorFactory.class ) ) {
 			return type.cast( this );
 		}
 		throw LOG.getTypeNotSupportedForUnwrappingException( type );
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/PredefinedScopeValidatorFactoryImpl.java
```diff
@@ -240,8 +240,9 @@ public boolean isTraversableResolverResultCacheEnabled() {
 
 	@Override
 	public <T> T unwrap(Class<T> type) {
-		//allow unwrapping into public super types
-		if ( type.isAssignableFrom( HibernateValidatorFactory.class ) ) {
+		// allow unwrapping into public super types
+		if ( type.isAssignableFrom( PredefinedScopeHibernateValidatorFactory.class )
+				|| type.isAssignableFrom( HibernateValidatorFactory.class ) ) {
```
marko-bekhta: I'm probably missing something important here... But why the unwrap impl from the start wasn't something like:
```java
public <T> T unwrap(Class<T> type) {
	Contracts.assertNotNull( type, "type cannot be null" );
	if ( type.isAssignableFrom( this.getClass() ) ) {
		return type.cast( this );
	}
	throw LOG.getTypeNotSupportedForUnwrappingException( type );
}
```
With the current change we might get into a situation when we only have the `HibernateValidatorFactory` but try to unwrap to `PredefinedScopeHibernateValidatorFactory`. Which will pass the if-check but fail on the cast part. 

Other than that all looks good.

gsmet: I'm not sure I follow you?

The idea of this check is to only allow to unwrap to a public interface. Any other unwrapping should fail with the exception.

Did I miss something?

marko-bekhta: > unwrap to a public interface

this explains what I was missing :smile: 

As for the second part .. I thought that it was possible to get into such situation:
```java
Validation.byProvider( HibernateValidator.class )
		.configure()
		.buildValidatorFactory()
		.unwrap( PredefinedScopeHibernateValidatorFactory.class )
```

Apparently I've missed the fact that it's a method on `PredefinedScopeValidatorFactoryImpl` and was thinking that both factories are using the same method ... Hence just ignore the comment :smiley: It's good to go.

gsmet: OK cool, thanks for checking :).

PR #1048, opened 2019-08-02
HV-1729 Skip allocation of an action for each need to access the con…

Sanne:
…text classloader

 - https://hibernate.atlassian.net/browse/HV-1729


gsmet:
Merged to 6.0, 6.2 and master branches.

gsmet:
Thanks!

gsmet:
BTW, is this something we use at runtime (runtime definition is a bit loose here, I mean once the metadata have been initialized for a given bean class)? Because I would have expected this to only be called when bootstrapping the metadata and never after that.

The metadata initialization is far from being optimized. Typically, we do a lot of map allocations when calling reflection methods repeatedly (we should call the method once and pass the result everywhere). That's an area of improvement for when time permits.

Sanne:
I haven't seen evidence of it being used at runtime - so not sure.

I was actually checking something else, then couldn't help fixing this :)

PR #1049, opened 2019-08-05
HV-1730 JavaBeanExecutable fails to initialize for enum type

marko-bekhta:
- https://hibernate.atlassian.net/browse/HV-1730

Check if generic parameter for executable is present before accessing it.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanExecutable.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanExecutable.java
@@ -205,19 +205,30 @@ public abstract class JavaBeanExecutable<T extends Executable> implements Callab
 			int explicitlyDeclaredParameterIndex = 0;
 
 			for ( int i = 0; i < parameterArray.length; i++ ) {
-				if ( parameterArray[i].isSynthetic() || parameterArray[i].isImplicit() ) {
-					// in this case, the parameter is not present in genericParameterTypes
-					parameters.add( new JavaBeanParameter( i, parameterArray[i], parameterTypes[i], parameterTypes[i] ) );
-				}
-				else {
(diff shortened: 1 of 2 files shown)


gsmet:
I was unable to reproduce the issue with your test. I reverted the code change and I just kept the test but it seems to run, either with JDK 8 or 11.

I also included the original enum:
```
public enum Gender {
    MALE,
    FEMALE;

    public static Gender fromCode(@NotNull String code) {
        return Gender.valueOf(code);
    }
}
```
but still no luck.

Will push what I did to your branch in a separate commit so that you can check what's going on. I'm testing with `mvn clean test -Dtest=PredefinedScopeValidatorFactoryTest#variousObjectTypes`.

marko-bekhta:
I remember running these from the IDE. I'll try to test it later today.

gsmet:
@marko-bekhta have you been able to determine what the real problem is?

doctau:
I am experiencing what I think is the same problem.

In AnnotationMetaDataProvider.getConstructorMetaData(), the constructor that is found has parameterTypes [String,int,String] but the signature is "(Ljava/lang/String;)V", and this matches what is in the class file.

private example.MyEnum(java.lang.String);
    descriptor: (Ljava/lang/String;ILjava/lang/String;)V

for this:
```enum MyEnum {
  A("a"),
  B("b");
  private string value;

  MyEnum(String value) {
    this.value = value;
  }
}```


With a simple nullary-constructor enum MyEnum{A,B;} it results in the javap output:
private MyEnum();
    descriptor: (Ljava/lang/String;I)V

The decompiled bytecode shows it using the first two parameters to call the Enum supertype constructor.



This seems like it is related https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8024694, and it sounds like javac generates Signature attributes that match what is in the source code which does not in fact match the bytecode level method parameters.

yrodiere:
Superseded by #1055

PR #1050, opened 2019-08-13
HV-1731 SigTest updates;

gunnarmorling:
https://hibernate.atlassian.net/browse/HV-1731

* Moving to tck-runner module
* Upgrading to latest version
* Enabling strict check

@gsmet Here's the minor update around SigTest we were talking about.

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -307,57 +307,5 @@
                 </dependency>
             </dependencies>
         </profile>
-        <profile>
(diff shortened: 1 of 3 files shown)


gsmet:
Merged and backported to 6.0. I added a commit to restore the profile as sigtest can't be run with JDK 11+ (an annotation has been added to `Object#finalize()`).

I also adjusted the CI jobs.

PR #1052, opened 2019-09-16
HV-1733 Fix locale settings of PredefinedScopeValidatorFactoryTest

gsmet:
* https://hibernate.atlassian.net/browse/HV-1733


marko-bekhta:
Thanks! I though this was failing only for me.

gsmet:
Well, it started to fail for me too, whereas it did work before. No idea why.

PR #1054, opened 2019-10-03
HV-1736 Cache regex Pattern for PatternValidator

sandynz:
Just cache java.util.regex.Pattern for better performance.


gsmet:
Did you really have a performance issue with this? Because `initialize()` is just called once at startup so it shouldn't be an issue.

sandynz:
Thanks.
I didn't debug enough. You are right, I find the cache logic in ConstraintValidatorManager now.

gsmet:
Ok thanks anyway!

PR #1055, opened 2019-10-08
HV-1730 JavaBeanExecutable fails to initialize for enum type

yrodiere:
https://hibernate.atlassian.net/browse/HV-1730

Supersedes #1049 .

It turns out the problem is really caused by the lack of a `-parameters` flag when model classes are compiled. When the flag is missing, we don't have any information about which parameter is implicit/synthetic, and the existing code assumes we always have that information. As a result, it inspects all the parameters without finding the implicit/synthetic ones, and ends up triggering an `ArrayIndexOutOfBoundsException`.

The solution provided in #1049 solves the out-of bound exception, and goes a little further by trying to "guess" which type is implicit/synthetic or not based on type erasure. This will work in some cases, but will lead to incorrect results in other cases.

This PR just takes the fix from #1049, and adds the following changes:

* Splits marko's commits in two, so that it's easier to revert the fix and only keep the test, to reproduce the problem.
* Adds a warning when parameter metadata (`isImplicit`/`isSynthetic`) is missing, because that may lead to incorrect generic type information being detected for method parameters.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanExecutable.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanExecutable.java
@@ -205,19 +205,30 @@ public abstract class JavaBeanExecutable<T extends Executable> implements Callab
 			int explicitlyDeclaredParameterIndex = 0;
 
 			for ( int i = 0; i < parameterArray.length; i++ ) {
-				if ( parameterArray[i].isSynthetic() || parameterArray[i].isImplicit() ) {
-					// in this case, the parameter is not present in genericParameterTypes
-					parameters.add( new JavaBeanParameter( i, parameterArray[i], parameterTypes[i], parameterTypes[i] ) );
-				}
-				else {
+				if ( explicitlyDeclaredParameterIndex < genericParameterTypes.length // we might already be out of the bounds of generic params array
+						&& isExplicit( parameterArray[i] )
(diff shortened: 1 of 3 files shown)


yrodiere:
I also checked that Marko's changes fix https://github.com/quarkusio/quarkus/issues/3284

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanExecutable.java
```diff
@@ -202,24 +207,53 @@ public String toString() {
...
 		return CollectionHelper.toImmutableList( parameters );
 	}
 
+	private static boolean isAnyParameterCarryingMetadata(Parameter[] parameterArray) {
+		for ( Parameter parameter : parameterArray ) {
+			if ( parameter.isSynthetic() || parameter.isImplicit() ) {
+				return true;
+			}
+		}
+		return false;
```
gsmet: Hmmm, what if the method is fully static? We won't have any synthetic or implicit parameters or will we?

gsmet: Ah, yes, I missed that from the diff visible on GH, sorry :)

marko-bekhta:
Hi @yrodiere , @gsmet . Just want to say thanks for picking this up! Have almost no free time these past few weeks.

gsmet:
Hey @marko-bekhta ,

Glad to hear from you! We'll be there when you have more time :).

gsmet:
Merged, thanks to both of you!

PR #1059, opened 2019-10-22
[WIP] HV-1755 WFLY-11566 ConstraintDeclarationException on JAX-RS/EJB Methods with List/Set query parameter

yrodiere:
https://hibernate.atlassian.net/browse/HV-1755
https://issues.jboss.org/browse/WFLY-11566

This does not solve the problem, but adds tests and a beginning of solution.

## Problem

Essentially the problem is caused by WildFly creating proxies that extend user-constrained classes, but that override methods incorrectly (from the HV point of view). The overrides are incorrect because they change the constraints: constraint annotations are preserved, but the type of parameters is not (it is erased), and thus the constraint annotations have a different meaning, or at the very least become ambiguous.

For example `@Max(value = 2, payload = Unwrapping.Unwrap.class) List<Integer> param` would be executed differently than  `@Max(value = 2, payload = Unwrapping.Unwrap.class) List<String> param`. Thus, when WildFly generates a parameter with types erased such as `@Max(value = 2, payload = Unwrapping.Unwrap.class) List param`, we can't decide what the constraint means.

## Solutions

There are two possible solutions:

1. Change what HV considers as "incorrect method overrides".
2. Ignore proxies when processing bean metadata, and consider that their metadata is the metadata of the type they are proxying. This would involve the `BeanMetaDataClassNormalizer` introduced in 6.1.

I started some work on solution number 2 in this PR.

## Requirements

Whatever the solution, we will have to get some information from the environment:

1. For solution 1, I don't think we can change the default behavior, because it is mandated by the BV spec. We would have to alter the behavior when a specific configuration option is changed.
2. For solution 2, we need to execute Wildfly-specific code, testing whether the class is a proxy class or not. This would likely be done by checking wether the class implements a specific interface (e.g. `org.jboss.something.SomeProxyTaggingInterface`) that only proxy classes would implement. Since that class would be specific to WildFly, we'd better implement the check in WildFly and pass it to HV somehow. This would mean implementing `BeanMetaDataClassNormalizer` in WildFly.

Thus we will need changes in WF in order to pass that information.

We will also need changes in HV, because we currently don't expose any hook that WF could take advantage of. The `ValidatorFactory` is created by HV itself through the `ValidationExtension` (a CDI extension), and WF doesn't get to configure it. The only way to configure the `ValidatorFactory` seems to be by defining a `validation.xml` or a custom `ValidatorFactory` bean, but both of those options are aimed at applications, not integrators.

Since the default `ValidatorFactory` in WildFly is created through CDI, it's likely that whatever information we need to pass to the `ValidatorFactory` will need to go through CDI, be it a bean of type `BeanMetaDataClassNormalizer` or a map of properties.

Going forward, we will need to investigate how to pass this information. We will want to involve the WildFly team before we make a decision.

### DIFF
--- a/cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidatorFactoryBean.java
+++ b/cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidatorFactoryBean.java
@@ -142,7 +142,7 @@ public class ValidatorFactoryBean implements Bean<ValidatorFactory>, Passivation
 			if ( beanMetaDataClassNormalizerInstance.isResolvable() ) {
 				BeanMetaDataClassNormalizer normalizer = beanMetaDataClassNormalizerInstance.get();
 				destructibleResources.add( new DestructibleBeanInstance<>( beanManager, normalizer ) );
-				
+
 				hvConfig.beanMetaDataClassNormalizer( normalizer );
 			}
 		}
--- a/cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidatorFactoryBean.java
+++ b/cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidatorFactoryBean.java
@@ -22,6 +22,8 @@ import java.util.stream.Collectors;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.inject.Instance;
+import javax.enterprise.inject.literal.NamedLiteral;
 import javax.enterprise.inject.spi.Bean;
 import javax.enterprise.inject.spi.BeanManager;
 import javax.enterprise.inject.spi.InjectionPoint;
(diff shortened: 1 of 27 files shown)


yrodiere:
I added a few commits to handle the retrieval of the bean metadata class normalizer defined as a CDI bean.

I used the code from https://github.com/tadamski/wildfly/tree/WFLY-11566 to test that the retrieval works correctly.

I had to make a few changes to that code, though:

* The implementation `org.hibernate.validator.metadata.BeanMetaDataClassNormalizer#normalize` had the wrong signature, I adapted it.
* `Bean#getType` returned an empty collection, which is invalid. I changed it to return the bean type.
* `org.hibernate.validator.test.cdi.internal.beanmetadataclassnormalizer.CustomProxyBeanMetaDataClassNormalizer#getQualifiers` returned an empty collection, meaning I had no way of retrieving the bean. I changed it to retrieve a Named qualifier with the correct type.
* I changed the bean name, and defined a constant in Hibernate Validator that you can use: see `org.hibernate.validator.cdi.spi.BeanNames#BEAN_META_DATA_CLASS_NORMALIZER`.

gsmet:
@yrodiere I rebased on top of the Maven Central fix but there are test failures.

yrodiere:
@gsmet That's actually expected: the failing test is the reproducer for WFLY-11566, which won't be fixed until we upgrade to a version of Wildfly that takes advantage of the SPI we are adding.

Do you want me to add `@Ignore` and a comment to that test, or should I remove it completely?

gsmet:
Ah OK, yes, better ignore it for now and open an issue affected to 6.1-next.

We support two consecutive versions of WildFly so it will take a while before we will be able to run that test.

yrodiere:
@gsmet Created https://hibernate.atlassian.net/browse/HV-1754 and pushed a commit to ignore this test.

gsmet:
Merged, thanks!

PR #1063, opened 2019-12-20
HV-1748 HV-1747 HV-1749 Improve localization support

gsmet:
* https://hibernate.atlassian.net/browse/HV-1747
 * https://hibernate.atlassian.net/browse/HV-1748
 * https://hibernate.atlassian.net/browse/HV-1749

@marko-bekhta if you're still with us (hi :) ), I'm interested in your feedback: the idea of this patch is to allow some more localization greatness for the `PredefinedScopeValidatorFactory`.

For now, the improvements are targeting this one as it's the only one with a proper list of locale. I have allowed to override the default locale for the good old `ValidatorFactory` too.

@irenakezic you can use this branch to finish your localization work on Quarkus. You now have a method to pass the ``LanguageRange``s to. Note that you need to check if the message interpolator is an `HibernateMessageInterpolator`, otherwise you need to default to the message using the default locale as you do right now if you don't have any header. I also incorporated your commit in the PR and added a test.

### DIFF
--- a/documentation/src/main/asciidoc/ch12.asciidoc
+++ b/documentation/src/main/asciidoc/ch12.asciidoc
@@ -51,6 +51,9 @@ Note that when a package is part of the public API this is not necessarily true 
 `org.hibernate.validator.spi.constraintdefinition`::
 			An SPI for registering additional constraint validators programmatically, see <<section-constraint-definition-contribution>>.
 
+`org.hibernate.validator.spi.messageinterpolation`::
+			An SPI that can be used to tweak the resolution of the locale when interpolating the constraint violation messages. See <<section-locale-resolver>>.
+
 `org.hibernate.validator.spi.nodenameprovider`::
 			An SPI that can be used to alter how the names of properties will be resolved when the property path is constructed. See <<section-property-node-name-provider>>.
 
@@ -442,6 +445,55 @@ With `ResourceBundleLocator`, Hibernate Validator provides an additional SPI whi
 error messages from other resource bundles than _ValidationMessages_ while still using the actual
 interpolation algorithm as defined by the specification. Refer to
 <<section-resource-bundle-locator>> to learn how to make use of that SPI.
+
+[[section-locale-resolver]]
+=== Customizing the locale resolution
+
+[WARNING]
+====
+These contracts are marked as `@Incubating` so they might be subject to change in the future.
+====
+
+Hibernate Validator provides several extension points to build a custom locale resolution strategy.
+The resolved locale is used when interpolating the constraint violation messages.
+
+The default behavior of Hibernate Validator is to always use the system default locale (as obtained via `Locale.getDefault()`).
+This might not be the desired behavior if, for example, you usually set your system locale to `en-US` but want your application to provide messages in French.
+
+The following example shows how to set the Hibernate Validator default locale to `fr-FR`:
+
+[[example-configure-default-locale]]
+.Configure the default locale
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter12/localization/LocalizationTest.java[tags=default-locale]
+----
+====
+
+While this is already a nice improvement, in a fully internationalized application, this is not sufficient:
+you need Hibernate Validator to select the locale depending on the user context.
+
+Hibernate Validator provides the `org.hibernate.validator.spi.messageinterpolation.LocaleResolver` SPI
+which allows to fine-tune the resolution of the locale.
+Typically, in a JAX-RS environment, you can resolve the locale to use from the `Accept-Language` HTTP header.
+
+In the following example, we use a hardcoded value but, for instance, in the case of a RESTEasy application,
+you could extract the header from the `ResteasyContext`.
+
+[[example-locale-resolver]]
+.Fine tune the locale used to interpolate the messages via a `LocaleResolver`
+====
+[source, JAVA, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter12/localization/LocalizationTest.java[tags=locale-resolver]
+----
+====
+
+[NOTE]
+====
+When using the `LocaleResolver`, you have to define the list of supported locales via the `locales()` method.
+====
 
 === Custom contexts
 
--- a/engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
+++ b/engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
@@ -22,6 +22,7 @@ import javax.validation.valueextraction.ValueExtractor;
 import org.hibernate.validator.cfg.ConstraintMapping;
 import org.hibernate.validator.constraints.ParameterScriptAssert;
 import org.hibernate.validator.constraints.ScriptAssert;
+import org.hibernate.validator.spi.messageinterpolation.LocaleResolver;
 import org.hibernate.validator.spi.nodenameprovider.PropertyNodeNameProvider;
 import org.hibernate.validator.spi.properties.GetterPropertySelectionStrategy;
 import org.hibernate.validator.spi.resourceloading.ResourceBundleLocator;
@@ -130,6 +131,15 @@ public interface BaseHibernateValidatorConfiguration<S extends BaseHibernateVali
 	 */
 	@Incubating
 	String PROPERTY_NODE_NAME_PROVIDER_CLASSNAME = "hibernate.validator.property_node_name_provider";
+
+	/**
+	 * Property for configuring the locale resolver, allowing to select an implementation of {@link LocaleResolver}
+	 * which will be used for locale resolution when interpolating a message.
+	 *
+	 * @since 6.1.1
+	 */
+	@Incubating
+	String LOCALE_RESOLVER_CLASSNAME = "hibernate.validator.locale_resolver";
 
 	/**
 	 * <p>
@@ -371,4 +381,16 @@ public interface BaseHibernateValidatorConfiguration<S extends BaseHibernateVali
 	 */
 	@Incubating
 	S defaultLocale(Locale defaultLocale);
+
+	/**
+	 * Allows setting a locale resolver, defining how the locale will be resolved when interpolating the message of a constraint violation.
+	 *
+	 * @param localeResolver the {@link LocaleResolver} to be used
+	 *
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 6.1.1
+	 */
+	@Incubating
+	S localeResolver(LocaleResolver localeResolver);
 }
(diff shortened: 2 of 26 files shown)


gsmet:
Note: this shouldn't be merged right away, it needs documentation and a few more tests.

gunnarmorling:
IIRC, locale retrieval was originally intended to be done by `ResourceBundleLocator` implementations, so there might be an overlap with this new implementation? Might well be justified, just wanted to bring up this background from back in the day.

Most importantly, Merry Christmas, @gsmet et al. :)

gsmet:
@gunnarmorling hey. So, this patch is not about locale retrieval but about locale resolution.

The whole idea of this work is to provide a way for Hibernate Validator to resolve the locale from the `Accept-Language` header in a RESTEasy context.

I don't think it overlaps with what `ResourceBundleLocator` is supposed to do. We call it the same way as before with the resolved locale but we need another SPI to be able to resolve the locale from the RESTEasy context.

Merry Christmas to you and your family too :).

gunnarmorling:
> The whole idea of this work is to provide a way for Hibernate Validator to resolve the locale from the Accept-Language header in a RESTEasy context.

Gotcha; I thought that'd be handled at the spec level already, but this [open issue](https://github.com/eclipse-ee4j/jaxrs-api/issues/544) indicates it's only logged as a suggestion but not done yet. Makes sense, thanks!

> Merry Christmas to you and your family too :)

Thanks a lot my friend. Would love to have a chat in the new year, has been a while :)

gsmet:
> Gotcha; I thought that'd be handled at the spec level already, but this open issue indicates it's only logged as a suggestion but not done yet. Makes sense, thanks!

Yes, it's been a long standing issue. Quarkus gives us the opportunity to play with it as we have the list of the supported locales.

I will probably make some adjustments to the normal path soon enough.

> Thanks a lot my friend. Would love to have a chat in the new year, has been a while :)

Sure! And we still need to finalize the Jakarta migration (which will be a bit less fun) ;)

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryConfigurationHelper.java
```diff
@@ -318,6 +320,32 @@ static PropertyNodeNameProvider determinePropertyNodeNameProvider(AbstractConfig
...
+			try {
+				@SuppressWarnings("unchecked")
+				Class<? extends LocaleResolver> clazz = (Class<? extends LocaleResolver>) run( LoadClass.action( localeResolverFqcn, externalClassLoader ) );
+				LocaleResolver localeResolver = run( NewInstance.action( clazz, "locale resolver class" ) );
+				LOG.usingLocaleResolver( clazz );
+
+				return localeResolver;
+			}
+			catch (Exception e) {
+				throw LOG.getUnableToInstantiatePropertyNodeNameProviderClassException( localeResolverFqcn, e );
```
yrodiere: I think you need a dedicated logging method here; this one was copy-pasted.

engine/src/test/java/org/hibernate/validator/test/predefinedscope/LocaleResolverTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+@TestForIssue(jiraKey = "HV-1749")
+public class LocaleResolverTest {
+
+	@Test
+	public void testLanguageRangeSupport() throws NoSuchMethodException, SecurityException {
+		ValidatorFactory validatorFactory = getValidatorFactoryWithInitializedLocales( Locale.FRANCE, new Locale( "es", "ES" ) );
+		HibernateMessageInterpolator messageInterpolator = (HibernateMessageInterpolator) validatorFactory.getMessageInterpolator();
+
+		StaticFieldLocaleResolver.acceptLanguage = "fr-FR,fr;q=0.9";
```
yrodiere: Knowing you have this solution, is the `HibernateMessageInterpolator` interface still useful? I'm not quite understanding how that interface is supposed to be used, to be honest.

engine/src/main/java/org/hibernate/validator/messageinterpolation/ResourceBundleMessageInterpolator.java
```diff
@@ -115,8 +117,10 @@ public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundl
 			Set<Locale> localesToInitialize,
```
yrodiere: Looks like you missed this "localesToInitialize", and a few others below. Maybe try a search and replace?

engine/src/main/java/org/hibernate/validator/resourceloading/AggregateResourceBundleLocator.java
```diff
@@ -88,12 +89,14 @@ public AggregateResourceBundleLocator(List<String> bundleNames, ResourceBundleLo
 	 * contain all entries from all source bundles. In case a key occurs
 	 * in multiple source bundles, the value will be taken from the
 	 * first bundle containing the key.
-	 * @param localesToInitialize The set of locales to initialize at bootstrap.
+	 * @param preloadResourceBundles if resource bundles should be initialized when initializing the locator
+	 * @param localesToInitialize The set of locales to initialize at bootstrap
```
yrodiere: In message interpolators, you pass the "preloadResourceBundles" parameter _after_ the "locales" parameter.
Here, you pass the "preloadResourceBundles" parameter _before_ the "localesToInitialize" parameter.

Not sure it's on purpose?

documentation/src/main/asciidoc/ch12.asciidoc
```diff
@@ -443,6 +446,55 @@ error messages from other resource bundles than _ValidationMessages_ while still
...
+[WARNING]
+====
+These contracts are marked as `@Incubating` so they might be subject to change in the future.
+====
+
+Hibernate Validator provides several extension points to build a custom locale resolution strategy.
+The resolved locale is used when interpolating the constraint violation messages.
+
+The default behavior of Hibernate Validator is to always use the system default locale (as obtained via `Locale.getDefault()`).
+This might not be the desired behavior if you usually set your system locale to `en-US` but want your application to provide messages in French.
```
yrodiere: Maybe a "for example" would be a good idea in this sentence.

documentation/src/main/asciidoc/ch12.asciidoc
```diff
@@ -443,6 +446,55 @@ error messages from other resource bundles than _ValidationMessages_ while still
...
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter12/localization/LocalizationTest.java[tags=default-locale]
+----
+====
+
+While this is already a nice improvement, in a fully internationalized application, this is not sufficient:
+you need Hibernate Validator to select the locale depending on the user context.
+
+Hibernate Validator provides the `org.hibernate.validator.spi.messageinterpolation.LocaleResolver` SPI
+which allows to finely tune the resolution of the locale.
```
yrodiere: `finely tune` => [`fine-tune`](https://www.wordreference.com/enfr/fine-tune)?

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -385,6 +387,23 @@
 	@Incubating
 	S locales(Set<Locale> locales);
 
+	/**
+	 * Allows to set the list of the locales supported by this ValidatorFactory.
```
yrodiere: "Allows to set" => "Allows setting"

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -372,6 +372,19 @@
 	@Incubating
 	S propertyNodeNameProvider(PropertyNodeNameProvider propertyNodeNameProvider);
 
+	/**
+	 * Allows to set the list of the locales supported by this ValidatorFactory.
```
yrodiere: "Allows to set" => "Allows setting"

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -385,6 +387,23 @@
...
+	 * <p>
+	 * If not set, defaults to a singleton containing {@link Locale#getDefault()}.
+	 *
+	 * @since 6.1.1
+	 */
+	@SuppressWarnings("unchecked")
+	@Incubating
+	default S locales(Locale... locales) {
+		locales( new HashSet<>( Arrays.asList( locales ) ) );
+		return (S) this;
```
yrodiere: You could do this instead, and avoid the unchecked cast:

```
return locales( new HashSet<>( Arrays.asList( locales ) ) );
```

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryConfigurationHelper.java
gsmet: It was already there, I forgot to change the code.

engine/src/main/java/org/hibernate/validator/messageinterpolation/AbstractMessageInterpolator.java
```diff
@@ -166,61 +175,75 @@ public AbstractMessageInterpolator(ResourceBundleLocator userResourceBundleLocat
...
 	 * {@code MessageInterpolator} using the default resource bundle locators.
 	 *
-	 * @param localesToInitialize The set of locales to initialize at bootstrap.
+	 * @param localesToInitialize the set of locales to initialize at bootstrap
+	 * @param defaultLocale the default locale
 	 *
-	 * @since 6.1
+	 * @since 6.1.1
 	 */
-	public AbstractMessageInterpolator(Set<Locale> localesToInitialize) {
```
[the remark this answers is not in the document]
gsmet: Yeah, it should have been made `@Incubating` from the beginning but I missed it and it's only used for the `PredefinedScope` stuff and I thought it would be better to drop it rather than stacking all these constructors.

The new ones are properly marked as `@Incubating`.

engine/src/main/java/org/hibernate/validator/resourceloading/AggregateResourceBundleLocator.java
gsmet: Yes it is.

engine/src/test/java/org/hibernate/validator/test/predefinedscope/PredefinedScopeValidatorFactoryTest.java
```diff
@@ -182,24 +182,6 @@ public void testValidOnUnknownBean() {
 		assertNoViolations( violations );
 	}
 
-	@Test(expectedExceptions = ValidationException.class, expectedExceptionsMessageRegExp = "HV000250:.*")
-	public void testUninitializedLocale() {
```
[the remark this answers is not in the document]
gsmet: Yeah, I got rid of this test a bit too quickly. It was not passing anymore because of the default locale always being added. I changed it a bit to be able to trigger the error.

PR #1069, opened 2020-01-30
HV-1756 Fix variable names in the newly added translations

gsmet:
https://hibernate.atlassian.net/browse/HV-1756


gsmet:
@yrodiere I know it's not the most interesting PR but could you have a look?

I think I will probably release something tomorrow (so today when you read these lines) as it's quite bad.

gsmet:
Thanks! An automated test spotted a couple more issues. It should be all fixed now.

PR #1075, opened 2020-04-03
HV-1762 Avoid cloning overhead from Method.getParameterTypes()

dreis2211:
See https://hibernate.atlassian.net/browse/HV-1762

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/provider/AnnotationMetaDataProvider.java
@@ -172,9 +172,8 @@ public class AnnotationMetaDataProvider implements MetaDataProvider {
 			Class<? extends DefaultGroupSequenceProvider<? super T>> providerClass) {
 		Method[] providerMethods = run( GetMethods.action( providerClass ) );
 		for ( Method method : providerMethods ) {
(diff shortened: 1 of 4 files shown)


gsmet:
ok to test

gsmet:
@dreis2211 BTW, we made a lot of improvements to the runtime performances of HV but I suspect there's a lot of room for improvements in the bootstrap part. When I mean bootstrap, I also mean the part when we initialize the metadata of a given bean (which might happen at runtime but only once for a given bean class).

If you're interested, you could add a JMH benchmarks for the bootstrap in our performance module and profile that. Creating a `PredefinedScopeValidatorFactory` is probably the best way to work on that. I remember we had quite a lot of map initializations due to some reflection calls. I discovered that when working on something unrelated and didn't get to the bottom of it.

Given we are all about improving startup time right now, that would be a cool pet project.

dreis2211:
@gsmet I doubt that I will have time for that, but thanks for letting me know. My startup time improvement efforts these days go into the Spring ecosystem, but maybe I'll find some time in between.

dreis2211:
But for completeness reasons: Could you elaborate on the `PredefinedScopeValidatorFactory` thought a bit more? Is that the hotspot that needs improvement?

gsmet:
No, it's just that if you use that one you are sure all the metadata have been properly initialized. Whereas with the standard one, metadata for a class are generated at the first validation of a bean of this class. This is the one we use in Quarkus.

PR #1076, opened 2020-04-09
HV-1763 Improve performances of ExecutableHelper#getSignature()

gsmet:


engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
```diff
@@ -181,10 +179,26 @@ public static String getSignature(Executable executable) {
 		return getSignature( getSimpleName( executable ), executable.getParameterTypes() );
 	}
 
+	/**
+	 * That method needs to be fast.
```
Sanne, line 183: "That" ? You mean "this" ?

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
```diff
@@ -181,10 +179,26 @@ public static String getSignature(Executable executable) {
...
+	 * <p>
+	 * The purpose of it is to get uniqueness.
+	 */
 	public static String getSignature(String name, Class<?>[] parameterTypes) {
-		return Stream.of( parameterTypes )
-			.map( t -> t.getName() )
-			.collect( Collectors.joining( ",", name + "(", ")" ) );
+		StringBuilder signature = new StringBuilder( name.length() + 2 + parameterTypes.length * 25 );
+		signature.append( name ).append( "(" );
```
Sanne, line 189: could use `'('`

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
```diff
@@ -181,10 +179,26 @@ public static String getSignature(Executable executable) {
...
+		boolean separator = false;
+		for ( Class<?> parameterType : parameterTypes ) {
+			if ( separator ) {
+				signature.append( "," );
```
Sanne, line 193: use ','

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
Sanne: (I mean single quotes so to use the specialized appender method for a single char)

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
```diff
@@ -181,10 +179,26 @@ public static String getSignature(Executable executable) {
...
+			}
+			else {
+				separator = true;
+			}
+			signature.append( parameterType.getName() );
+		}
+		signature.append( ")" );
```
Sanne, line 200: ')'

Sanne:
and let's remember to blame @gunnarmorling a bit :)  A good old fashioned "we told you so.." :-P

performance/pom.xml
```diff
@@ -21,6 +21,9 @@
     <description>Hibernate Validator performance tests.</description>
 
     <properties>
+        <!-- Disable enforcer as we are using banned dependencies for old versions -->
+        <enforcer.skip>true</enforcer.skip>
```
Sanne, line 25: Did you mean to commit this? Not a problem but wondering, as I see no direct relation.

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
```diff
@@ -197,9 +211,20 @@ public static String getSignature(String name, Class<?>[] parameterTypes) {
...
 	public static String getExecutableAsString(String name, Class<?>... parameterTypes) {
-		return Stream.of( parameterTypes )
-			.map( t -> t.getSimpleName() )
-			.collect( Collectors.joining( ", ", name + "(", ")" ) );
+		StringBuilder signature = new StringBuilder( name.length() + 2 + parameterTypes.length * 25 );
+		signature.append( name ).append( '(' );
+		boolean separator = false;
+		for ( Class<?> parameterType : parameterTypes ) {
+			if ( separator ) {
+				signature.append( ", " );
```
Sanne, line 219: if the goal was merely uniqueness, you won't need the extra space and could still benefit from the single char invocation by just using ',' as separator.

gunnarmorling:
> and let's remember to blame @gunnarmorling a bit :)

Sure, no problem, I can handle it ;)

That said, I hope you didn't choose 25 as type name length factor lightly, but only after careful evaluation and benchmarking with a representative set of class names ;)

engine/src/main/java/org/hibernate/validator/internal/properties/Signature.java
```diff
@@ -0,0 +1,51 @@
...
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.properties;
+
+import java.util.Arrays;
+
+public class Signature {
```
Sanne, line 11: make the class `final` ?

engine/src/main/java/org/hibernate/validator/internal/properties/Signature.java
```diff
@@ -0,0 +1,51 @@
...
+
+	private String name;
```
Sanne, line 13: I think the two main fields need to be `final` (even if there's a cost): as these keys are getting entered in concurrent maps I'd not try to play smart games with this.

The "hashCode" is different; it's a bit special as it's not a reference so you _could_ avoid the final qualifier, but since you're already writing the other two fields as final you're already going to flush all state so I suppose there's no benefit in keeping that lazy.

I'm not sure if it's worth using that field though? Are you sure the hashcode is always going to be used? There's also a tradeoff with it consuming a precious field space.

engine/src/main/java/org/hibernate/validator/internal/properties/Signature.java
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if ( obj == null ) {
+			return false;
+		}
+
+		// we can safely assume the type will always be the right one
+		Signature other = (Signature) obj;
```
Sanne, line 32: ok :)

engine/src/main/java/org/hibernate/validator/internal/properties/Signature.java
Sanne: BTW just checked with jol: that `int` field is costing you 8 bytes because of alignemnt. I'd say it's not worth it.

gsmet: Yes, it's going to be used and it has been proven very useful in HV in general.

gsmet: OK for making the fields final.

engine/src/main/java/org/hibernate/validator/internal/properties/Signature.java
gsmet: Sure.

engine/src/main/java/org/hibernate/validator/internal/properties/Signature.java
Sanne: Regarding the `int`, I don't have data to make a strong suggestion but consider one thing:

in this case you're trading off a little CPU cost for 8 bytes. The 8 bytes are "free" in the context of a micro-benchmark, but they are not in a more complex real-world benchmark.

So maybe you're right and it's the better choice, but don't fool yourself in thinking that JMH is giving you this certainty.

gsmet: OK, I've done the changes, will run a benchmark later today.

engine/src/main/java/org/hibernate/validator/internal/util/ExecutableHelper.java
gsmet: Yeah so that one is about having a nice string representation. That's why it's still there and is in pretty print mode.

performance/pom.xml
gsmet: Yes. Otherwise the benchmarks can't be run as we have banned the old dependencies used by the benchmarks from HV (the javax ones) but we still need them for the old versions benchmarks. It's in a specific commit tweaking the benchmarks.

engine/src/main/java/org/hibernate/validator/internal/properties/Signature.java
gsmet: Benchmarks are not better but not worse either so let's get rid of the field.

gsmet:
@Sanne I think I made all the changes you requested.

PR #1080, opened 2020-04-15
HV-1767 Reduce the overhead of ConstraintHelper initialization in the predefined scope case

gsmet:
https://hibernate.atlassian.net/browse/HV-1767

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/core/BuiltinConstraint.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/core/BuiltinConstraint.java
@@ -1,0 +1,139 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+package org.hibernate.validator.internal.metadata.core;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * The list of built-in constraints.
+ * <p>
+ * We are not using the class directly to avoid loading the class if possible.
+ * <p>
+ * In the case of composing constraints, you need to define the dependencies manually.
+ * <p>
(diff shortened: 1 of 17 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
```diff
@@ -1092,16 +1097,28 @@ public void clear() {
...
 	}
 
-	private static boolean isJodaTimeInClasspath() {
-		return isClassPresent( JODA_TIME_CLASS_NAME );
+	private boolean isJodaTimeInClasspath() {
+		if ( jodaTimeInClassPath != null ) {
+			return jodaTimeInClassPath.booleanValue();
+		}
+		jodaTimeInClassPath = isClassPresent( JODA_TIME_CLASS_NAME );
+		return jodaTimeInClassPath.booleanValue();
```
yrodiere, line 1105: Some nitpicking (for a change!); I usually see this pattern written that way:

```
		if ( jodaTimeInClassPath == null ) {
			jodaTimeInClassPath = isClassPresent( JODA_TIME_CLASS_NAME );
		}
		return jodaTimeInClassPath.booleanValue(); 
```

engine/src/main/java/org/hibernate/validator/internal/metadata/core/BuiltinConstraint.java
```diff
@@ -0,0 +1,139 @@
...
+				Set<BuiltinConstraint> constraints = new HashSet<>();
+				constraints.add( constraint );
+				constraints.addAll( constraint.composingConstraints );
+				CONSTRAINT_MAPPING.put( constraint.constraint, constraints );
+			}
+		}
+	}
+
+	private String constraint;
+	private List<BuiltinConstraint> composingConstraints;
```
yrodiere, line 110: `requiredConstraints` or `constraintDependencies` or something like that? Feel free to ignore, but I had a hard time wrapping my head around this, so a more precise name might help.

gsmet: This is the technical term in HV but yeah I also thought about changing it to something along the lines of what you suggest so let's do it.

gsmet:
@yrodiere I addressed your comments.

PR #1085, opened 2020-04-28
HV-1771 Fix DecimalMin message German translation

gsmet:
https://hibernate.atlassian.net/browse/HV-1771


Bittersweetme:
> https://hibernate.atlassian.net/browse/HV-1771

PR #1101, opened 2020-05-27
Updated ValidationMessages_de.properties removing . in Max message

focbenz:
Removed the ending "." in the jakarta.validation.constraints.Max.message matching the other messages that have no ending "."

https://hibernate.atlassian.net/browse/HV-1782


focdanisch:
The original change is here: https://github.com/hibernate/hibernate-validator/commit/b723ee78c02e8c536150d4a2091d755ec191f510#diff-1d325c4df27f2892904f4b8f6553e485

As you can see, Max has been "forgotten" in that change for the german validation messages.

PR #1107, opened 2020-06-30
HV-1761 fix interpolation of primitive arrays

agatsenko:
https://hibernate.atlassian.net/browse/HV-1761

This is a fix for the 'HV-1328' bugs. The issue is that invoking `ParameterTermResolver.interpolate (context, expression)` throws the exception `java.lang.ClassCastException` if the value of the expression is a primitive array.

I can merge these changes into `6.1`, `6.2`, `master` brunches if needed.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/ParameterTermResolver.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/ParameterTermResolver.java
@@ -18,6 +18,7 @@ import org.hibernate.validator.messageinterpolation.HibernateMessageInterpolator
  * @author Hardy Ferentschik
  * @author Adam Stawicki
  * @author Guillaume Smet
+ * @author Alexander Gatsenko
  */
 public class ParameterTermResolver implements TermResolver {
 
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/ParameterTermResolver.java
```diff
@@ -52,4 +48,41 @@ private Object getVariable(Context context, String parameter) {
 	private String removeCurlyBraces(String parameter) {
 		return parameter.substring( 1, parameter.length() - 1 );
 	}
+
+	private String resolveExpressionByNotNullVariable(Object variable) {
```
gsmet, line 52: I would just rename that method to `resolveExpression`. No need to make it extra verbose.

agatsenko: fixed

agatsenko:
Could you clarify if you will release 6.1.x with this fix?

gsmet:
Yup, I will include it in 6.0, 6.1 and master. I'm just unsure of when I will do a release.

Do you need a quick 6.1 release?

gsmet:
I squashed all the commits into one. I'll merge as soon as CI is green and port it to 6.1 and master.

agatsenko:
> Yup, I will include it in 6.0, 6.1 and master. I'm just unsure of when I will do a release.

This bug is not blocker for me.

Thank you for the great work.

gsmet:
Merged, thanks a lot! Will work on pushing it to the other versions now.

gsmet:
Done. Pushed to 6.0, 6.1 and master. Thanks again for the fix!

PR #1110, opened 2020-07-15
HV-1789 - Upgrade to Bean Validation 3.0.0 and update the XML namespaces

gsmet:
The PR is complete but we have to wait for a new TCK release.

### DIFF
--- a/documentation/src/main/asciidoc/ch08.asciidoc
+++ b/documentation/src/main/asciidoc/ch08.asciidoc
@@ -9,12 +9,7 @@ approach via annotations. Let's have a look at these two document types.
 
 [NOTE]
 ====
-The XSD files are available via
-https://xmlns.jakarta.ee/xml/ns/validation/configuration and
-https://xmlns.jakarta.ee/xml/ns/validation/mapping.
-
-More information about the XML schemas can be found on the
-http://beanvalidation.org/xml/ns/validation/[Jakarta Bean Validation website].
+The XSD files are available on the https://jakarta.ee/xml/ns/validation/ page.
 ====
 
 [[section-configuration-validation-xml]]
--- a/documentation/src/main/asciidoc/ch12.asciidoc
+++ b/documentation/src/main/asciidoc/ch12.asciidoc
@@ -132,9 +132,9 @@ _hibernate.validator.allow_parallel_method_parameter_constraint_ in _validation.
 ----
 <?xml version="1.0" encoding="UTF-8"?>
 <validation-config
-        xmlns="https://xmlns.jakarta.ee/xml/ns/validation/configuration"
+        xmlns="https://jakarta.ee/xml/ns/validation/configuration"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xsi:schemaLocation="https://xmlns.jakarta.ee/xml/ns/validation/configuration validation-configuration-3.0.xsd"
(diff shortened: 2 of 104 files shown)


documentation/src/main/asciidoc/ch08.asciidoc
```diff
@@ -10,8 +10,8 @@ approach via annotations. Let's have a look at these two document types.
 [NOTE]
 ====
 The XSD files are available via
-https://xmlns.jakarta.ee/xml/ns/validation/configuration and
-https://xmlns.jakarta.ee/xml/ns/validation/mapping.
+https://jakarta.ee/xml/ns/validation/configuration and
+https://jakarta.ee/xml/ns/validation/mapping.
```
starksm64, line 14: The schemas are online under the https://jakarta.ee/xml/ns/validation/ page. On that page is a table linking to the validation-mapping-3.0.xsd and validation-configuration-3.0.xsd files.

documentation/src/main/asciidoc/ch12.asciidoc
```diff
@@ -132,9 +132,9 @@ _hibernate.validator.allow_parallel_method_parameter_constraint_ in _validation.
 ----
 <?xml version="1.0" encoding="UTF-8"?>
 <validation-config
-        xmlns="https://xmlns.jakarta.ee/xml/ns/validation/configuration"
+        xmlns="https://jakarta.ee/xml/ns/validation/configuration"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xsi:schemaLocation="https://xmlns.jakarta.ee/xml/ns/validation/configuration validation-configuration-3.0.xsd"
+        xsi:schemaLocation="https://jakarta.ee/xml/ns/validation/configuration validation-configuration-3.0.xsd"
```
starksm64, line 137: The xsi:schemaLocation here is not correct as the "validation-configuration-3.0.xsd" relative URL would not resolve correctly as the actual location on the jakarta.ee website is https://jakarta.ee/xml/ns/validation/validation-configuration-3.0.xsd.

The validation-configuration-3.0.xsd documentation example uses the correct form:

            <validation-config
                xmlns="https://jakarta.ee/xml/ns/validation/configuration"
                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                xsi:schemaLocation="
                    https://jakarta.ee/xml/ns/validation/configuration
                    https://jakarta.ee/xml/ns/validation/validation-configuration-3.0.xsd"
                version="3.0">

gunnarmorling: Hum, hum. HV should never actually fetch the XSD from there, but rather use its own local copy of the file. I.e. the value here shouldn't make any actual difference; it should be more taken as a visual cue than as a resolveable resource. I don't think we ever had the full path there in examples.

starksm64: Sure, HV does not need it, but if people are configuring an IDE or other xml aware parser it helps to have the actual resolvable location. For example, in Intellij if I use the full form of the schemaLocation and it shows up as red I can simply use the resolve external location hint to register the schema file with the namespace.

documentation/src/main/asciidoc/ch08.asciidoc
gsmet: Fixed.

documentation/src/main/asciidoc/ch12.asciidoc
gsmet: Fixed.

PR #1111, opened 2020-07-15
HV-1790 Remove the SafeHtml constraint and the jsoup dependency

gsmet:
There is no replacement for this constraint.
It has been deprecated and planned for removal in both 6.0 and 6.1.


gsmet:
Jenkins, retest this please

PR #1117, opened 2020-07-23
HV-1793 Upgrade to JavaMoney 1.1 and Moneta 1.4.1 - 6.1

gsmet:
In passing drop OSGi support for Java Money as I got some issues and I'm
leaning towards giving up on OSGi.


gsmet:
Closing as WildFly integration doesn't work and we also had some issues with Karaf features so not worth it for now.

PR #1130, opened 2020-10-01
[Branch 6.1] HV-1805 + HV-1806 + HV-1807 + HV-1809 + HV-1810  Jenkinsfile + various build improvements

yrodiere:
* [HV-1810](https://hibernate.atlassian.net/browse/HV-1810): Upgrade to Arquillian 1.6
* [HV-1809](https://hibernate.atlassian.net/browse/HV-1809): Upgrade to TestNG 6.14.3
* [HV-1807](https://hibernate.atlassian.net/browse/HV-1807): Add Maven profile to improve contributor experience on Intellij IDEA
* [HV-1806](https://hibernate.atlassian.net/browse/HV-1806): Use the local repository from the Maven build in OSGi/WildFly integration tests
* [HV-1805](https://hibernate.atlassian.net/browse/HV-1805): Add a Jenkinsfile to test multiple JDKs from a single job

This is mostly a backport of #1129. The main difference is that this upgrades to Arquillian 1.6: we were using 1.1 on this branch, and 1.7.Alpha1 on master.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -309,6 +309,8 @@ stage('Default build') {
 							install \
 					"} \
 					-Pdist \
+					-Psigtest \
+					-Pjqassistant \
 					${enableDefaultBuildIT ? '' : '-DskipITs'} \
 					${toTestJdkArg(environments.content.jdk.default)} \
 			"""
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,542 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+
+import groovy.transform.Field
+
+/*
+ * See https://github.com/hibernate/hibernate-jenkins-pipeline-helpers
+ */
+@Library('hibernate-jenkins-pipeline-helpers@1.3')
+import org.hibernate.jenkins.pipeline.helpers.job.JobHelper
+import org.hibernate.jenkins.pipeline.helpers.alternative.AlternativeMultiMap
+import org.hibernate.jenkins.pipeline.helpers.version.Version
+
+/*
+ * WARNING: DO NOT IMPORT LOCAL LIBRARIES HERE.
+ *
+ * By local, I mean libraries whose files are in the same Git repository.
+ *
+ * The Jenkinsfile is protected and will not be executed if modified in pull requests from external users,
+ * but other local library files are not protected.
+ * A user could potentially craft a malicious PR by modifying a local library.
+ *
+ * See https://blog.grdryn.me/blog/jenkins-pipeline-trust.html for a full explanation,
+ * and a potential solution if we really need local libraries.
+ * Alternatively we might be able to host libraries in a separate GitHub repo and configure
+ * them in the GUI: see https://ci.hibernate.org/job/hibernate-validator/configure, "Pipeline Libraries".
+ */
+
+/*
+ * See https://github.com/hibernate/hibernate-jenkins-pipeline-helpers for the documentation
+ * of the helpers library used in this Jenkinsfile,
+ * and for help writing Jenkinsfiles.
+ *
+ * ### Jenkins configuration
+ *
+ * #### Jenkins plugins
+ *
+ * This file requires the following plugins in particular:
+ *
+ * - everything required by the helpers library (see the org.hibernate.(...) imports for a link to its documentation)
+ * - https://plugins.jenkins.io/pipeline-github for the trigger on pull request comments
+ *
+ * #### Script approval
+ *
+ * If not already done, you will need to allow the following calls in <jenkinsUrl>/scriptApproval/:
+ *
+ * - everything required by the helpers library (see the org.hibernate.(...) imports for a link to its documentation)
+ *
+ * ### Integrations
+ *
+ * #### Nexus deployment
+ *
+ * This job includes two deployment modes:
+ *
+ * - A deployment of snapshot artifacts for every non-PR build on "primary" branches (master and maintenance branches).
+ * - A full release when starting the job with specific parameters.
+ *
+ * In the first case, the name of a Maven settings file must be provided in the job configuration file
+ * (see below).
+ *
+ * #### Gitter (optional)
+ *
+ * You need to enable the Jenkins integration in your Gitter room first:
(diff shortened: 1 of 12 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,542 @@
...
+	this.environments = AlternativeMultiMap.create([
+			jdk: [
+					// This should not include every JDK; in particular let's not care too much about EOL'd JDKs like version 9
+					// See http://www.oracle.com/technetwork/java/javase/eol-135779.html
+					new JdkBuildEnvironment(version: '8', buildJdkTool: 'OracleJDK8 Latest',
+							condition: TestCondition.BEFORE_MERGE,
+							isDefault: true),
+					new JdkBuildEnvironment(version: '11', buildJdkTool: 'OpenJDK 11 Latest',
+							condition: TestCondition.AFTER_MERGE),
+					new JdkBuildEnvironment(version: '14', buildJdkTool: 'OpenJDK 14 Latest',
```
gsmet, line 139: Yes, let's keep 14 until the issue with 15 is fixed.

Jenkinsfile
```diff
@@ -0,0 +1,542 @@
...
+							condition: TestCondition.AFTER_MERGE),
+					new JdkBuildEnvironment(version: '15', buildJdkTool: 'OpenJDK 15 Latest',
+							condition: TestCondition.AFTER_MERGE),
+					new JdkBuildEnvironment(version: '16', buildJdkTool: 'OpenJDK 16 Latest',
+							condition: TestCondition.AFTER_MERGE)
```
gsmet, line 144: I was wondering if we should include the sigtest, jqassistant and the WildFly TCK container build (for JDK 8 and 14).

I'm not exactly sure they make sense for the main build as at least the TCK builds are quite slow but if they can be parallelized, maybe it makes sense.

This is an honest question, happy to have your feedback/benefit from your experience on this.

Jenkinsfile
```diff
@@ -142,6 +142,12 @@ stage('Configure') {
 							condition: TestCondition.AFTER_MERGE),
 					new JdkBuildEnvironment(version: '16', buildJdkTool: 'OpenJDK 16 Latest',
 							condition: TestCondition.AFTER_MERGE)
+			],
+			wildflyTck: [
+					new WildFlyTckBuildEnvironment(javaVersion: '8', buildJdkTool: 'OracleJDK8 Latest',
+							condition: TestCondition.AFTER_MERGE),
+					new WildFlyTckBuildEnvironment(javaVersion: '11', buildJdkTool: 'OpenJDK 11 Latest',
+							condition: TestCondition.AFTER_MERGE)
```
gsmet, line 150: OK, apparently, you partially answered that. What about the sigtest and jqassistant?

I suppose `TestCondition.AFTER_MERGE` means this will be only built when a commit is merged?

Jenkinsfile
```diff
@@ -0,0 +1,542 @@
...
+stage('Configure') {
+	this.environments = AlternativeMultiMap.create([
+			jdk: [
+					// This should not include every JDK; in particular let's not care too much about EOL'd JDKs like version 9
+					// See http://www.oracle.com/technetwork/java/javase/eol-135779.html
+					new JdkBuildEnvironment(version: '8', buildJdkTool: 'OracleJDK8 Latest',
+							condition: TestCondition.BEFORE_MERGE,
+							isDefault: true),
+					new JdkBuildEnvironment(version: '11', buildJdkTool: 'OpenJDK 11 Latest',
+							condition: TestCondition.AFTER_MERGE),
```
gsmet, line 138: If they are parallelized, I think it makes sense to have both 8 and 11 running in parallel now.

gsmet (changes requested):
Looks very good to me.

I added a few questions/comments!

Jenkinsfile
yrodiere: So... there is a default build, on JDK8. Other builds only run *after* that, and only if the default build succeeded. The goal is to avoid consuming too much resources if even the default build doesn't pass.

We can consider changing this behavior, but I'd rather do it as a second step. I don't know how complicated this will be, and the current behavior is the one in this Jenkinsfile anyway.

Jenkinsfile
yrodiere: > What about the sigtest and jqassistant?

Personally on Search, I just run JQAssistant as part of the default build (JDK8). Maybe we could do that?

About sigtest, no idea, I don't use it. But sure, we could include it somewhere in the Jenkinsfile.

> I suppose TestCondition.AFTER_MERGE means this will be only built when a commit is merged?

Yes. There are comments further down, on the enum, to explain the exact meaning. And I think I explained the general idea in the PR description.

tck-runner/pom.xml
```diff
@@ -156,8 +156,6 @@
                             <value>${validation.provider}</value>
                         </property>
                     </systemProperties>
-                    <parallel>methods</parallel>
-                    <threadCount>4</threadCount>
```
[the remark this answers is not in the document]
yrodiere: I was about to say "but it only takes 20s [here](https://ci.hibernate.org/blue/organizations/jenkins/hibernate-validator-personal-yoann/detail/HV-1805-6.1/20/pipeline/65)", but it seems no tests are actually executed... I'm looking into this.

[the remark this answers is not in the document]
yrodiere: I solved the problem mentioned in my previous comment: wildfly TCK are now correctly executed. Parallel execution is still disabled. It takes about 5 minutes.

Jenkinsfile
yrodiere: As per our discussion on Zulip, I added sigtest and JQAssistant to the default build. It will run for every build, including PRs. We can talk about moving this to separate stages in another PR.

yrodiere (commented):
@gsmet I think I answered all your comments.

Here is a full build: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-validator/detail/PR-1130/5/pipeline/20

JDK15 and JDK16 fail (as expected, because of https://bugs.openjdk.java.net/browse/JDK-8253566). Do you want me to disable them for now?

Some of the other builds appear yellow even though they completed successfully; this seems to be a display bug in Jenkins, but it should matter most of the time, since failing builds are a rare occurrence, right? Right? :]

yrodiere:
> JDK15 and JDK16 fail (as expected, because of https://bugs.openjdk.java.net/browse/JDK-8253566). Do you want me to disable them for now?

No answer, but I suppose you do. I added a commit to disable these builds (temporarily).

gsmet (approved):

gsmet:
Merged, thanks!

PR #1132, opened 2020-10-14
HV-1812 Avoid reflection-based metadata extraction for built-in value extractors - 6.2

gsmet:
https://hibernate.atlassian.net/browse/HV-1812

@yrodiere not the most exciting PR but I would appreciate a second pair of eyes.

The idea is to avoid calling a method not present in the Android JDK for the built-in value extractors.


yrodiere (approved):
Can't say if this will make HV usable in Android, since there are no tests. There might be other missing methods in their JDK?

Anyway, the replacements look correct.

gsmet:
I asked the OP to test it. TBH, that's the maximum amount of work I'm willing to put into this :).

PR #1134, opened 2020-10-20
[7.0] Re-enable JDK16 testing in Jenkinsfile

yrodiere:
The fix for JDK-8253566 is in JDK 16 b20. Not in JDK 15 yet.


yrodiere:
Tested here, tests pass on JDK 16: https://ci.hibernate.org/blue/organizations/jenkins/hibernate-validator-personal-yoann/detail/jdk16/2/pipeline/63

gsmet (approved):

PR #1138, opened 2020-11-25
HV-1816 Disable Expression Language by default for custom constraint violations

gsmet:
* https://hibernate.atlassian.net/browse/HV-1816

### DIFF
--- a/documentation/src/main/asciidoc/ch06.asciidoc
+++ b/documentation/src/main/asciidoc/ch06.asciidoc
@@ -173,35 +173,12 @@ It is important to add each configured constraint violation by calling `addConst
 Only after that the new constraint violation will be created.
 ====
 
-[[el-injection-caution]]
-[CAUTION]
-====
-**Be aware that the custom message template is passed directly to the Expression Language engine.**
+By default, Expression Language is not enabled for custom violations created in the `ConstraintValidatorContext`.
 
-Thus, you should be very careful when integrating user input in a custom message template as it will be interpreted
-by the Expression Language engine, which is usually not the behavior you want and **could allow malicious users to leak
-sensitive data or even execute arbitrary code**.
+However, for some advanced requirements, using Expression Language might be necessary.
 
-If you need to integrate user input in your message, you must <<section-hibernateconstraintvalidatorcontext,pass it as an expression variable>>
-by unwrapping the context to `HibernateConstraintValidatorContext`.
-
-The following validator is very unsafe as it includes user input in the violation message.
-If the validated `value` contains EL expressions, they will be executed by the EL engine.
-
-[source, JAVA, indent=0]
-----
-include::{sourcedir}/org/hibernate/validator/referenceguide/chapter06/elinjection/UnsafeValidator.java[tags=include]
-----
-
-The following pattern must be used instead:
-
-[source]
-----
-include::{sourcedir}/org/hibernate/validator/referenceguide/chapter06/elinjection/SafeValidator.java[tags=include]
-----
-
-By using expression variables, Hibernate Validator properly handles escaping and EL expressions won't be executed.
-====
+In this case, you need to unwrap the `HibernateConstraintValidatorContext` and enable Expression Language explicitly.
+See <<section-hibernateconstraintvalidatorcontext>> for more information.
 
 Refer to <<section-custom-property-paths>> to learn how to use the `ConstraintValidatorContext` API to
 control the property path of constraint violations for class-level constraints.
--- a/engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
+++ b/engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
@@ -13,6 +13,7 @@ import java.util.Locale;
 import java.util.Set;
 
 import javax.validation.Configuration;
+import javax.validation.ConstraintValidatorContext;
 import javax.validation.ConstraintViolation;
 import javax.validation.TraversableResolver;
 import javax.validation.constraints.Future;
@@ -24,6 +25,7 @@ import javax.validation.valueextraction.ValueExtractor;
 import org.hibernate.validator.cfg.ConstraintMapping;
 import org.hibernate.validator.constraints.ParameterScriptAssert;
 import org.hibernate.validator.constraints.ScriptAssert;
+import org.hibernate.validator.messageinterpolation.ExpressionLanguageFeatureLevel;
 import org.hibernate.validator.spi.messageinterpolation.LocaleResolver;
 import org.hibernate.validator.metadata.BeanMetaDataClassNormalizer;
 import org.hibernate.validator.spi.nodenameprovider.PropertyNodeNameProvider;
@@ -143,6 +145,28 @@ public interface BaseHibernateValidatorConfiguration<S extends BaseHibernateVali
 	 */
 	@Incubating
 	String LOCALE_RESOLVER_CLASSNAME = "hibernate.validator.locale_resolver";
+
+	/**
+	 * Property for configuring the Expression Language feature level for constraints, allowing to define which
+	 * Expression Language features are available for message interpolation.
+	 * <p>
+	 * This property only affects the EL feature level of "static" constraint violation messages. In particular, it
+	 * doesn't affect the default EL feature level for custom violations. Refer to
+	 * {@link #CUSTOM_VIOLATION_EXPRESSION_LANGUAGE_FEATURE_LEVEL} to configure that.
+	 *
+	 * @since 6.2
+	 */
+	@Incubating
+	String CONSTRAINT_EXPRESSION_LANGUAGE_FEATURE_LEVEL = "hibernate.validator.constraint_expression_language_feature_level";
+
+	/**
+	 * Property for configuring the Expression Language feature level for custom violations, allowing to define which
+	 * Expression Language features are available for message interpolation.
+	 *
+	 * @since 6.2
+	 */
+	@Incubating
+	String CUSTOM_VIOLATION_EXPRESSION_LANGUAGE_FEATURE_LEVEL = "hibernate.validator.custom_violation_expression_language_feature_level";
 
 	/**
 	 * <p>
@@ -427,4 +451,33 @@ public interface BaseHibernateValidatorConfiguration<S extends BaseHibernateVali
 
 	@Incubating
 	S beanMetaDataClassNormalizer(BeanMetaDataClassNormalizer beanMetaDataClassNormalizer);
+
+	/**
+	 * Allows setting the Expression Language feature level for message interpolation of constraint messages.
+	 * <p>
+	 * This is the feature level used for messages hardcoded inside the constraint declaration.
+	 * <p>
+	 * If you are creating custom constraint violations, Expression Language support needs to be explicitly enabled and
+	 * use the safest feature level by default if enabled.
+	 *
+	 * @param expressionLanguageFeatureLevel the {@link ExpressionLanguageFeatureLevel} to be used
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 6.2
+	 */
+	@Incubating
+	S constraintExpressionLanguageFeatureLevel(ExpressionLanguageFeatureLevel expressionLanguageFeatureLevel);
+
+	/**
+	 * Allows setting the Expression Language feature level for message interpolation of custom violation messages.
+	 * <p>
+	 * This is the feature level used for messages of custom violations created by the {@link ConstraintValidatorContext}.
+	 *
+	 * @param expressionLanguageFeatureLevel the {@link ExpressionLanguageFeatureLevel} to be used
+	 * @return {@code this} following the chaining method pattern
+	 *
+	 * @since 6.2
+	 */
+	@Incubating
+	S customViolationExpressionLanguageFeatureLevel(ExpressionLanguageFeatureLevel expressionLanguageFeatureLevel);
 }
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintViolationBuilder.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintViolationBuilder.java
@@ -10,18 +10,34 @@ package org.hibernate.validator.constraintvalidation;
 import javax.validation.ConstraintValidatorContext.ConstraintViolationBuilder;
 
 import org.hibernate.validator.Incubating;
+import org.hibernate.validator.messageinterpolation.ExpressionLanguageFeatureLevel;
 
 public interface HibernateConstraintViolationBuilder extends ConstraintViolationBuilder {
 
 	/**
-	 * Enable Expression Language for the constraint violation created by this builder if the chosen
-	 * {@code MessageInterpolator} supports it.
+	 * Enable Expression Language with the default Expression Language feature level for the constraint violation
+	 * created by this builder if the chosen {@code MessageInterpolator} supports it.
 	 * <p>
-	 * If enabling this, you need to make sure your message template does not contain any unescaped user input (such as
+	 * If you enable this, you need to make sure your message template does not contain any unescaped user input (such as
 	 * the validated value): use {@code addExpressionVariable()} to inject properly escaped variables into the template.
 	 *
 	 * @since 6.2
 	 */
 	@Incubating
-	HibernateConstraintViolationBuilder enableExpressionLanguage();
+	default HibernateConstraintViolationBuilder enableExpressionLanguage() {
+		return enableExpressionLanguage( ExpressionLanguageFeatureLevel.DEFAULT );
+	};
+
+	/**
+	 * Enable Expression Language for the constraint violation created by this builder if the chosen
+	 * {@code MessageInterpolator} supports it.
+	 * <p>
+	 * If you enable this, you need to make sure your message template does not contain any unescaped user input (such as
+	 * the validated value): use {@code addExpressionVariable()} to inject properly escaped variables into the template.
+	 *
+	 * @param level The Expression Language features level supported.
+	 * @since 6.2
+	 */
+	@Incubating
+	HibernateConstraintViolationBuilder enableExpressionLanguage(ExpressionLanguageFeatureLevel level);
 }
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintViolationBuilder.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintViolationBuilder.java
@@ -1,0 +1,27 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
+ */
+
+package org.hibernate.validator.constraintvalidation;
+
+import javax.validation.ConstraintValidatorContext.ConstraintViolationBuilder;
+
+import org.hibernate.validator.Incubating;
+
+public interface HibernateConstraintViolationBuilder extends ConstraintViolationBuilder {
+
+	/**
+	 * Enable Expression Language for the constraint violation created by this builder if the chosen
+	 * {@code MessageInterpolator} supports it.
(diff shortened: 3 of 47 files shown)


engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
```diff
@@ -189,9 +193,14 @@ public ConstraintValidatorContext addConstraintViolation() {
 			if ( constraintViolationCreationContexts == null ) {
 				constraintViolationCreationContexts = CollectionHelper.newArrayList( 3 );
 			}
+			if ( !(expressionVariables == null || expressionVariables.isEmpty()) && !expressionLanguageEnabled ) {
+				LOG.expressionVariablesDefinedWithExpressionLanguageNotEnabled(
```
yrodiere, line 201: Maybe add a test to check that this warning actually occurs?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
```diff
@@ -59,8 +61,10 @@ public ConstraintValidatorContextImpl(
 			ClockProvider clockProvider,
 			PathImpl propertyPath,
 			ConstraintDescriptor<?> constraintDescriptor,
-			Object constraintValidatorPayload) {
+			Object constraintValidatorPayload,
+			ExpressionLanguageFeatureLevel expressionLanguageFeatureLevel) {
 		this.clockProvider = clockProvider;
+		this.defaultExpressionLanguageFeatureLevel = expressionLanguageFeatureLevel;
```
yrodiere, line 67: Isn't this rather `defaultConstraintExpressionLanguageFeatureLevel`? For non-default constraints, this isn't the default feature level... unless I misunderstand?

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/ElTermResolver.java
```diff
@@ -79,7 +85,26 @@ public String interpolate(MessageInterpolator.Context context, String expression
...
+	private ELContext getElContext(MessageInterpolator.Context context) {
+		if ( !( context instanceof HibernateMessageInterpolatorContext ) ) {
+			return new SafestELContext( expressionFactory );
+		}
+
+		switch ( ( (HibernateMessageInterpolatorContext) context ).getExpressionLanguageFeatureLevel() ) {
+			case NONE:
+				throw LOG.expressionsNotResolvedWhenExpressionLanguageFeaturesDisabled();
+			case SAFEST:
+				return new SafestELContext( expressionFactory );
```
yrodiere, line 97: If you want to be consistent will `INSECURE`, this should be named `SECURE`. Safety and security not being the same thing and all that...

Also, technically, the most secure ELContext is no ELContext... If there is a way to leak data from resource bundles, or to access a property of a variable, we cannot really say "it's absolutely secure".

To avoid the problem, isn't there a name that translates what the context allows, rather than our opinion of it? `VARIABLES`, maybe?

engine/src/main/java/org/hibernate/validator/messageinterpolation/ExpressionLanguageFeatureLevel.java
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * Only allows access to the variables injected via
+	 * {@link HibernateConstraintValidatorContext#addExpressionVariable(String, Object)}, the Jakarta Bean
+	 * Validation-defined {@code formatter} and the {@code ResourceBundle}s.
+	 */
+	SAFEST,
+
+	/**
+	 * Only allows to access bean properties and the Jakarta Bean Validation-defined {@code formatter}.
```
yrodiere, line 33: No access to the expression variables and resource bundle? In either case, I think this should be stated explicitly.

engine/src/main/java/org/hibernate/validator/messageinterpolation/ExpressionLanguageFeatureLevel.java
```diff
@@ -0,0 +1,49 @@
...
+	 * <p>
+	 * This is the minimal level to have a specification-compliant implementation.
+	 */
+	BEAN_PROPERTIES,
+
+	/**
+	 * This level allows bean method execution and can lead to serious security issues if not very carefully handled.
```
yrodiere, line 40:
```suggestion
	 * This level allows bean method execution and can lead to serious security issues, including arbitrary code execution, if not very carefully handled.
```

engine/src/main/java/org/hibernate/validator/messageinterpolation/ParameterMessageInterpolator.java
```diff
@@ -52,7 +52,7 @@ public ParameterMessageInterpolator(Set<Locale> locales, Locale defaultLocale, L
 	}
 
 	@Override
-	public String interpolate(Context context, Locale locale, String term) {
+	protected String interpolate(Context context, Locale locale, String term) {
```
yrodiere, line 55: Isn't this technically an API change? I'm not sure it will impact anyone, but still...

engine/src/main/java/org/hibernate/validator/messageinterpolation/ResourceBundleMessageInterpolator.java
```diff
@@ -154,7 +154,7 @@ public ResourceBundleMessageInterpolator(ResourceBundleLocator userResourceBundl
 	}
 
 	@Override
-	public String interpolate(Context context, Locale locale, String term) {
+	protected String interpolate(Context context, Locale locale, String term) {
```
yrodiere, line 157: Isn't this technically an API change? I'm not sure it will impact anyone, but still...

engine/src/main/java/org/hibernate/validator/messageinterpolation/ExpressionLanguageFeatureLevel.java
```diff
@@ -0,0 +1,49 @@
...
+	 * <p>
+	 * If using this level, you need to be sure you are not injecting user input in an expression without properly
+	 * escaping it using {@link HibernateConstraintValidatorContext#addExpressionVariable(String, Object)}.
+	 */
+	INSECURE;
+
+	public static final ExpressionLanguageFeatureLevel DEFAULT = ExpressionLanguageFeatureLevel.BEAN_PROPERTIES;
```
yrodiere, line 47: That's pretty neat, but will it not lead to confusion when calling `enableExpressionLanguage(ExpressionLanguageFeatureLevel.DEFAULT)`? One can wonder if you're resetting to the application-configured defaults or to the HV defaults.

Alternatively you could add a `DEFAULT` value to the enum, and handle it wherever necessary.

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintViolationBuilder.java
```diff
@@ -10,18 +10,34 @@
...
 import org.hibernate.validator.Incubating;
+import org.hibernate.validator.messageinterpolation.ExpressionLanguageFeatureLevel;
 
 public interface HibernateConstraintViolationBuilder extends ConstraintViolationBuilder {
 
+	/**
+	 * Enable Expression Language with the default Expression Language feature level for the constraint violation
+	 * created by this builder if the chosen {@code MessageInterpolator} supports it.
+	 * <p>
+	 * If enabling this, you need to make sure your message template does not contain any unescaped user input (such as
```
yrodiere, line 21:
```suggestion
	 * If you enable this, you need to make sure your message template does not contain any unescaped user input (such as
```

Or

```suggestion
	 * When enabling this, you need to make sure your message template does not contain any unescaped user input (such as
```

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -427,4 +428,15 @@ default S locales(Locale... locales) {
 
 	@Incubating
 	S beanMetaDataClassNormalizer(BeanMetaDataClassNormalizer beanMetaDataClassNormalizer);
+
+	/**
+	 * Allows setting the Expression Language feature level.
```
yrodiere, line 433: So... is this the default for all constraints? Or only for default (built-in) constraints? Is it used as the default for custom constraint violations?

I find this a bit confusing...

gsmet: I clarified it in the Javadoc.

I know it's a bit confusing but the defaults are different and I don't want the global default to be the one used by custom violations as they are more sensitive to security issues.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
gsmet: Yes, exactly. I think I will keep it simple though.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
gsmet: Yes, I need to add a bunch of tests.

engine/src/main/java/org/hibernate/validator/messageinterpolation/ExpressionLanguageFeatureLevel.java
gsmet: That's a good idea.

engine/src/main/java/org/hibernate/validator/messageinterpolation/ParameterMessageInterpolator.java
gsmet: It is but I think it was a mistake.

One information you might miss is I won't merge this to 6.1, I will do a 6.2 and also include this in 7.

documentation/src/main/asciidoc/ch06.asciidoc
```diff
@@ -173,35 +173,12 @@ It is important to add each configured constraint violation by calling `addConst
...
-The following validator is very unsafe as it includes user input in the violation message.
-If the validated `value` contains EL expressions, they will be executed by the EL engine.
-
-[source, JAVA, indent=0]
-----
-include::{sourcedir}/org/hibernate/validator/referenceguide/chapter06/elinjection/UnsafeValidator.java[tags=include]
-----
-
-The following pattern must be used instead:
+By default, Expression Language is not enabled for custom constraints created in the `ConstraintValidatorContext`.
```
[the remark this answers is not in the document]
gsmet: Fixed it locally as there was an additional space in your suggestion.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
gsmet: Done.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
gsmet: In the end, I adjusted it since I introduced another one for custom violations.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
```diff
@@ -218,8 +227,8 @@ protected ConstraintViolationBuilderImpl(String template, PathImpl path) {
 		}
 
 		@Override
-		public HibernateConstraintViolationBuilder enableExpressionLanguage() {
-			expressionLanguageEnabled = true;
+		public HibernateConstraintViolationBuilder enableExpressionLanguage(ExpressionLanguageFeatureLevel expressionLanguageFeatureLevel) {
+			this.expressionLanguageFeatureLevel = ExpressionLanguageFeatureLevel.interpretDefaultForCustomViolations( expressionLanguageFeatureLevel );
```
yrodiere, line 231: When you pass `DEFAULT` to this method, shouldn't it set the feature level to `defaultCustomViolationExpressionLanguageFeatureLevel`?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
```diff
@@ -181,7 +189,7 @@ private ConstraintViolationCreationContext getDefaultConstraintViolationCreation
 	private abstract class NodeBuilderBase {
 
 		protected final String messageTemplate;
-		protected boolean expressionLanguageEnabled;
+		protected ExpressionLanguageFeatureLevel expressionLanguageFeatureLevel = defaultCustomViolationExpressionLanguageFeatureLevel;
```
yrodiere, line 192: Since EL may now be enabled by default depending on configuration, shouldn't `enableExpressionLanguage` be renamed to `expressionLanguageFeatureLevel`?
On second thought, it also feels weird that we can call `enableExpressionLanguage(NONE)`.

engine/src/main/java/org/hibernate/validator/internal/engine/messageinterpolation/el/VariablesELContext.java
```diff
@@ -0,0 +1,55 @@
...
+
+	private static final ELResolver DEFAULT_RESOLVER = new CompositeELResolver() {
+
+		{
+			add( new RootResolver() );
+			add( new ArrayELResolver( true ) );
+			add( new ListELResolver( true ) );
+			add( new MapELResolver( true ) );
+			add( new ResourceBundleELResolver() );
+			add( new NoOpElResolver() );
```
yrodiere, line 31: Maybe a comment to explain why this is necessary? I'm sure there's a reason, but it seems weird.

engine/src/main/java/org/hibernate/validator/messageinterpolation/ExpressionLanguageFeatureLevel.java
```diff
@@ -0,0 +1,90 @@
...
+	/**
+	 * The default Expression Language feature level.
+	 * <p>
+	 * Depends on the context.
+	 * <p>
+	 * For standard constraint messages, it is {@link ExpressionLanguageFeatureLevel#BEAN_PROPERTIES}.
+	 * <p>
+	 * For custom violations, the default is {@link ExpressionLanguageFeatureLevel#NONE} and, if Expression Language is
+	 * enabled for a given custom violation via the API, the default becomes
+	 * {@link ExpressionLanguageFeatureLevel#VARIABLES} in the context of this given custom violation.
```
yrodiere, line 29: It's not quite the full story, though.

From what I can see:

* If not enabled explicitly, the default is whatever the user configured through properties, falling back to `NONE`.
* If enabled explicitly, the default is `VARIABLES`, and whatever the user configured through properties is ignored.

Which seems... complex?

engine/src/test/java/org/hibernate/validator/test/internal/engine/messageinterpolation/ExpressionLanguageMessageInterpolationTest.java
```diff
@@ -69,13 +70,52 @@ public void testExpressionLanguageGraphNavigation() {
...
+				ExpressionLanguageFeatureLevel.BEAN_PROPERTIES,
+				false );
+
+		String expected = "18";
+		String actual = interpolatorUnderTest.interpolate( "${validatedValue.age}", context );
+		assertEquals( actual, expected, "Wrong substitution" );
+	}
+
+	@Test
+	public void testExpressionLanguageGraphNavigationSafest() {
```
yrodiere, line 101:
```suggestion
	public void testExpressionLanguageGraphNavigationVariables() {
```

engine/src/test/java/org/hibernate/validator/test/internal/engine/messageinterpolation/ExpressionLanguageMessageInterpolationTest.java
```diff
@@ -221,13 +263,31 @@ public void testNoParametersToFormatter() {
...
-		MessageInterpolator.Context context = createMessageInterpolatorContext( sizeDescriptor );
+		MessageInterpolator.Context context = createMessageInterpolatorContextELInsecure( sizeDescriptor );
 
 		String expected = "foo";
 		String actual = interpolatorUnderTest.interpolate( "${'foobar'.substring(0,3)}", context );
 		assertEquals( actual, expected, "Calling of String#substring should work" );
 	}
 
+	@Test
+	public void testNonFormatterFunctionSafest() {
```
yrodiere, line 274:
```suggestion
	public void testNonFormatterFunctionVariables() {
```

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
gsmet: I think I prefer it the way it is.

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -427,4 +446,36 @@ default S locales(Locale... locales) {
 
 	@Incubating
 	S beanMetaDataClassNormalizer(BeanMetaDataClassNormalizer beanMetaDataClassNormalizer);
+
+	/**
+	 * Allows setting the Expression Language feature level for message interpolation of constraint messages.
+	 * <p>
+	 * This is the feature level used for messages hardcoded inside the constraint declaration.
+	 * <p>
+	 * If you are creating custom constraint violations, Expression Language support needs to be explicitly enabled and
```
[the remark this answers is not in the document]
gsmet: Yeah, so I'm not a big fan of cross linking APIs in javadoc from everywhere. It's a nightmare to maintain.

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -427,4 +446,36 @@ default S locales(Locale... locales) {
...
+	@Incubating
+	S constraintExpressionLanguageFeatureLevel(ExpressionLanguageFeatureLevel expressionLanguageFeatureLevel);
+
+	/**
+	 * Allows setting the Expression Language feature level for message interpolation of custom violation messages.
+	 * <p>
+	 * This is the feature level used for messages hardcoded inside the constraint declaration.
+	 * <p>
+	 * If you are creating custom constraint violations, Expression Language support needs to be explicitly enabled and
+	 * use the safest feature level by default if enabled.
```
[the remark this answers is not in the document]
gsmet: The Javadoc was incorrect, I fixed it.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ConstraintValidatorContextImpl.java
gsmet: Hmmm, I think I prefer it being a static default. I'll see if people complain about it.

gsmet:
/cc @pwntester this might be of interest to you. I plan to include this in a 6.2 and a 7.

I gave up on hoping that users will read the documentation and also the most problematic way is that there is no spec-compliant way to be safe.

The way I did it passes the Bean Validation TCK while reducing the EL Features even for standard constraints.

For custom violations, it's completely disabled by default and you can enable EL on a case per case basis. We also have a global settings for compatibility reasons but the idea is that people should enable EL for specific custom violations and thus be careful when doing so.

pwntester:
Secure by default, way to go!

gsmet:
Well, secure by default is easy.

The hard part is to still offer the flexibility of EL without compromising the whole thing (i.e. avoid pushing people to use a global switch which IMHO is equally as bad) and still passing the TCK.

I think we need to push `addExpressionVariable()` to the spec.

PR #1152, opened 2021-01-18
HV-1828 ContributorValidationMessages compatible with UTF-8 properties

jpep-in:
fix proposal from https://hibernate.atlassian.net/browse/HV-1828


gsmet:
Sorry it took me so long to have a closer look at this.

I pushed an additional commit to your branch but it doesn't appear in the PR. I created https://github.com/hibernate/hibernate-validator/pull/1154 , let's continue the discussion there.

PR #1153, opened 2021-01-30
Possibility of unique mansion implemented

MonthAlcantara:
I would like to apologize in advance if this is not the correct way to open a PR. It's my first PR.

I was unable to run locally but I would like to open this PR to present the idea.

When using more than one constraint all error messages are displayed. However, the idea would be to be able to use a single message for all validations.

This message () attribute must be optional, so when not declaring it, the messages of the annotated validations will be used.


gsmet:
Sorry for the late reply. I think you're looking for `@ReportAsSingleViolation`. As soon as one of the constraints is invalid, it will report only one single violation with the message of the constraint having the annotation.

PR #1154, opened 2021-02-03
HV-1828 ContributorValidationMessages compatible with UTF-8 properties

gsmet:
https://hibernate.atlassian.net/browse/HV-1828

Supersedes https://github.com/hibernate/hibernate-validator/pull/1152 .

@jpep-in could you have a look at my changes? I think it should make things a bit better. I also fixed a couple of preexisting issues.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/resourceloading/AggregateResourceBundle.java
+++ b/engine/src/main/java/org/hibernate/validator/resourceloading/AggregateResourceBundle.java
@@ -1,0 +1,71 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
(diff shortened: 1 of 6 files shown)


gsmet:
@yrodiere I'm interested in a review of this one. The original issue is explained here: https://hibernate.atlassian.net/browse/HV-1828 .

yrodiere (approved):
LGTM, but I'd like to point out there is another aggregate bundle class, `org.hibernate.validator.resourceloading.AggregateResourceBundleLocator.AggregateBundle`, that seems to have the exact same purpose, with the same implementation.

yrodiere (approved):

gsmet:
Pushed an additional commit to clean things up a bit.

@jpep-in any chance you could have a look and check things are good for you? I also prepared a 6.2 backport here if it's more practical for you: https://github.com/hibernate/hibernate-validator/pull/1155

gsmet:
@jpep-in ping?

PR #1157, opened 2021-02-24
HV-1831 Unfinished experiments

gsmet:
This is a very rough experiment based on the 6.2 branch. It's not finished at all and I have no idea if we can make it work in the end.


engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/ParameterExecutableValidationContext.java
```diff
@@ -85,13 +87,16 @@ public Executable getExecutable() {
...
+	private static boolean buildProcessedBeansTrackingEnabled(ProcessedBeansTrackingStrategy processedBeansTrackingStrategy,
+			Executable executable,
+			Optional<ExecutableMetaData> executableMetaData) {
 		if ( !executableMetaData.isPresent() ) {
 			// the method is unconstrained so there's no need to worry about the tracking
 			return false;
 		}
 
-		return !executableMetaData.get().getValidatableParametersMetaData().hasCascadables();
+		return processedBeansTrackingStrategy.isEnabledForReturnValue( executable,
```
gbadner, line 98: Should the above line be:

`return processedBeansTrackingStrategy.isEnabledForParameters( executable,`
?

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/AbstractValidationContext.java
```diff
@@ -369,12 +370,13 @@ private boolean isAlreadyValidatedForCurrentGroup(Object value, Class<?> group)
...
+		ProcessedBean processedBean = new ProcessedBean( bean );
+		ArrayList<PathImpl> processedPaths = processedPathsPerBean.get( processedBean );
 		if ( processedPaths == null ) {
-			processedPaths = new HashSet<>();
-			processedPathsPerBean.put( bean, processedPaths );
+			processedPaths = new ArrayList<>();
+			processedPathsPerBean.put( processedBean, processedPaths );
 		}
 
 		processedPaths.add( PathImpl.createCopy( path ) );
```
gbadner, line 382: While debugging `ProcessedBeansTrackingCycles1Test`, I see that `ValidatorImpl#validateConstraintsForDefaultGroup`  executes this code (in `#markCurrentBeanAsProcessedForCurrentPath`) twice for the same `Parent` bean ; first time for `Parent.class`, and second time for `Object.class`. Both times `#markCurrentBeanAsProcessedForCurrentPath` is called with the same `Parent` bean and the same `PathImpl` object as arguments.

Since `processedPaths` is an `ArrayList<PathImpl>`, it contains 2 copies of the same `PathImpl`. Was this intended?

I see that  `processedPaths` was changed from a `Set<PathImpl>` to an `ArrayList<PathImpl>`, so before this change, (I think) there would have been just one `PathImpl` in `Set`.

I don't know if this is a real problem; it's just something I noticed.

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/ParameterExecutableValidationContext.java
gsmet: You're absolutely right, I copy/pasted things a bit too fast.

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/AbstractValidationContext.java
gsmet: Hmmm. That's unfortunate. It was part of another experiment and shouldn't have been committed with this one.

How do you want us to proceed? Should I fix things and force push? Or we fix it later? Basically, everything in this class should be reverted, except for the change of meaning of the boolean.

gsmet:
Closing in favor of: https://github.com/hibernate/hibernate-validator/pull/1331

PR #1172, opened 2021-06-11
HV-1840 Drop/replace deprecated CDI API usages

manovotn:
I was testing WFLY with latest CDI API (4.0.0.Alpha1) where there was a number of removals of long deprecated methods.
See eclipse-ee4j/cdi#472

This causes hibernate validator to crash during TCK runs so I had to create a custom artifact and replace it within WFLY - this PR a sum of changes I needed to make it work.

Since you are bound to eventually bounce into the very same issue, I thought I might as well contribute. But whether you accept it or not is OFC completely up to you :-)

Oh and I should note - the current state of this PR will work with CDI 2/3 as well as CDI 4 so there should be no harm in having it applied right away.

### DIFF
--- a/cdi/src/main/java/org/hibernate/validator/cdi/internal/DestructibleBeanInstance.java
+++ b/cdi/src/main/java/org/hibernate/validator/cdi/internal/DestructibleBeanInstance.java
@@ -41,7 +41,7 @@ public class DestructibleBeanInstance<T> {
 
 	private InjectionTarget<T> createInjectionTarget(BeanManager beanManager, Class<T> type) {
 		AnnotatedType<T> annotatedType = beanManager.createAnnotatedType( type );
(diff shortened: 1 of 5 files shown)


gsmet:
Thanks Matej! I'll wait for CI and merge.

gsmet:
@manovotn let me know if you need a release.

manovotn:
Actually, if that release would then bubble all the way up to wfly, it would be a nice step towards testing some EE 10 changes in-container.
So +1.
There is no need to rush it though, so whenever you aren't too busy :-)

gsmet:
@manovotn cough cough. Sorry I only thought about this one again recently and saw your message then. I released 7.0.2.Final today and it contains these changes.

manovotn:
Haha, no problem @gsmet :)

PR #1179, opened 2021-07-30
HV-1848 Set up SSH config as part of the Jenkinsfile

yrodiere:
https://hibernate.atlassian.net/browse/HV-1848


yrodiere:
This is only useful for releases. Not sure you're using the Jenkinsfile for that, but better stay up-to-date anyway.

gsmet (approved):

PR #1183, opened 2021-09-29
HV-1853 Fix inconsistency of getter method detection in GetterPropertySelectionStrategy

rsvasilyev:
https://hibernate.atlassian.net/browse/HV-1853

### DIFF
--- a/documentation/src/test/java/org/hibernate/validator/referenceguide/chapter12/getterselectionstrategy/FluentGetterPropertySelectionStrategy.java
+++ b/documentation/src/test/java/org/hibernate/validator/referenceguide/chapter12/getterselectionstrategy/FluentGetterPropertySelectionStrategy.java
@@ -7,6 +7,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Optional;
 import java.util.Set;
+import java.util.List;
 import java.util.stream.Collectors;
 
 import org.hibernate.validator.spi.properties.ConstrainableExecutable;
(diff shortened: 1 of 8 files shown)


yrodiere:
ok to test

yrodiere:
@rsvasilyev Thank you for your PR. This seems reasonable, but since it breaks backwards compatibility of an SPI, I'd rather wait for the project lead's opinion. @gsmet is currently away, but he will have a look when he gets back in a few weeks.

marko-bekhta:
Hi ! I've just went trough suggested changes as well as this discussion and though I'll add my 2 cents to it 😃 

- If f I still remember it ... we usually tend to use `Set`s in HV codebase as a default collection unless something really requires to preserve the order of added elements. 
- the design idea behind the `GetterPropertySelectionStrategy` was that there's one getter per property so there is no need to have an ordered collection of possible getter names, as the first one that will be found is "the only one". Which apparently is not the case.... 
- as in different scenarios we might want different getter to get a priority it's hard to make a good choice here with the order... and user needs to make sure that the correct getters are selected.
- if we still want to provide a more predictable order I'd vote for something like `TreeSet` and keep the contract as is. 
- Just by looking at this particular scenario I almost want to see something similar to `@JsonIgnore` annotation (maybe `@GetterIgnore`?) so that only the getters that we care about remain. If we all like this idea, we can add it as a part of default strategy functionality ... But that would only work if there's access to the beans and they can be modified, and not in case all is configured via XML and unmodifiable bean classes...

gsmet:
I will merge this one. We might tweak the implementation later but IMO it makes sense to have the list properly ordered. TBH, I was always against our decision to use Set everywhere. Getting consistent ordering is better even if not ideal.

gsmet (approved):
I'm merging this one as I think in the long run, we better have this list ordered.

We can change the implementation as we want but the SPI change looks good and releasing 8 looks like a good

gsmet:
Merged, thanks!

PR #1184, opened 2021-11-02
HV-1854 Improve PESEL validator conditions

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-1854

With PESEL structure being YYMMDDZZZXQ only MM and DD parts make sense to have some additional checks applied. Year can be 00-99, ZZZ just "random digits", X - sex part is even/odd but ultimately 0-9 as well. 
Check digits in the new testcases would've allowed these numbers to be considered as valid without this additional date/month logic.

I'm actually glad that someone is using these specific validators and reported a problem with that more simple approach. 
But not as glad as being able to finally do some work here 😉

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/pl/PESELValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/pl/PESELValidator.java
@@ -6,6 +6,8 @@
  */
 package org.hibernate.validator.internal.constraintvalidators.hv.pl;
 
+import java.time.DateTimeException;
+import java.time.LocalDate;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/pl/PESELValidator.java
```diff
@@ -36,13 +36,30 @@ public void initialize(PESEL constraintAnnotation) {
...
+
+		//check the month part of the PESEL **XX*******: should be between 01 and 12 after % 20
+		// depending on the century 0/20/40/60/80 can be added to the month value
+		int month = doubleDigitNumberFromSubList( digits, 2 ) % 20;
+		if ( month > 12 || month == 0 ) {
+			return false;
+		}
+
+		// check the date part of the PESEL ****XX*****: should be between 01 abd 31
+		int date = doubleDigitNumberFromSubList( digits, 4 );
```
turo-90, line 55: you should check if a given month has this number of days

gsmet (changes requested):
Hey Marko, glad to see you here :). I spotted a few minor things. 

The question of @turo-90 is interesting: I have no idea if PESEL enforces the fact that the day must exist for it to be valid. That's probably something to check before enforcing it as I could see people relying on this sort of weird behavior.

turo-90:
> 
> 
> Hey Marko, glad to see you here :). I spotted a few minor things.
> 
> The question of @turo-90 is interesting: I have no idea if PESEL enforces the fact that the day must exist for it to be valid. That's probably something to check before enforcing it as I could see people relying on this sort of weird behavior.

The PESEL number must have a correct and existing date.
This is an identification number for people in Poland that includes their date of birth.

turo-90 (approved):

turo-90 (approved):

gsmet (approved):
Nice work.

marko-bekhta:
Hi Guillaume!!! Thanks! Glad to see that you all are still around! 
And thanks for taking a look at this one.

Will need to fix my email notifications as I've missed your original comment initially 🙈

PR #1196, opened 2021-12-15
HV-1864 Prepare for JDK 17 releasing - 6.2

gsmet:
https://hibernate.atlassian.net/browse/HV-1864

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -107,6 +107,8 @@ import org.hibernate.jenkins.pipeline.helpers.version.Version
 
 @Field boolean enableDefaultBuild = false
 @Field boolean enableDefaultBuildIT = false
+@Field boolean enableDefaultBuildJQAssistant = false
+@Field boolean enableDefaultBuildSigtest = false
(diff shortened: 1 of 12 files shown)


Jenkinsfile
```diff
@@ -302,8 +299,8 @@ stage('Default build') {
 							install \
 					"} \
 					-Pdist \
-					-Psigtest \
-					-Pjqassistant \
+					${enableDefaultBuildSigtest ? '-Psigtest' : ''} \
+					${enableDefaultBuildJQAssistant ? '-Pjqassistant' : ''} \
```
yrodiere, line 303: Not sure it matters, but you don't set `enableDefaultBuildSigtest`/`enableDefaultBuildJQAssistant` anywhere. They will always be `false`.

gsmet: Yes I know :). I plan to improve on that to get the info from the default build env but I didn't have the time to figure out how to do it properly and we don't really care for now as it doesn't work with 17 at all.

PR #1199, opened 2022-01-09
HV-1867 Add UUID validation

dheid:
https://hibernate.atlassian.net/browse/HV-1867


dheid:
@gsmet Could you have a look at it, please? I think, a character sequence UUID validation could be useful for many users.

engine/src/main/java/org/hibernate/validator/constraints/UUID.java
```diff
@@ -0,0 +1,144 @@
...
+	 */
+	int minimumVariant() default 0;
+
+	/**
+	 * Must not be lower than variant 0
+	 *
+	 * @return the maximum UUID variant number
+	 * Per default variant 2 should be maximum
+	 */
+	int maximumVariant() default 2;
```
yrodiere, line 102: Considering there are very few valid versions (1-5) and variants (0-2), wouldn't it be better to have `int[] version() default {1, 2, 3, 4, 5};` and `int[] variant() default {0, 1, 2}`?

That way a user can allow version 1 and 5 but not 2, 3, 4, for example.

dheid: Good idea! I will change that

dheid: @yrodiere Done! :smile:

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/UUIDValidator.java
```diff
@@ -0,0 +1,195 @@
...
+		if ( values.length == 0 ) {
+			throw new IllegalArgumentException( String.format(
+					Locale.ROOT,
+					"Parameter %s must not be empty",
+					parameterName
+			) );
+		}
+		for ( int value : values ) {
+			if ( value < minimum ) {
+				throw new IllegalArgumentException( String.format(
```
marko-bekhta, line 72: Could you please update these parameter verification checks to match this style:
```
Contracts.assertTrue(  value < minimum, MESSAGES.parameterShouldBeAtLeastEqualTo( "Minimum version", 1 ) );
```
see here for an example - https://github.com/hibernate/hibernate-validator/blob/main/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ParameterScriptAssertValidator.java#L72-L73

Essentially it is the same but would be more inline with the rest of the code. Thanks!

dheid: You're welcome :sweat_smile:

dheid:
@yrodiere @marko-bekhta Something else I can do? Have you decided to merge it yet?

yrodiere:
@yrodiere I don't decide that kind of things on this project :) I just saw your patch and thought I'd suggest an improvement.

You'll have to wait for @gsmet to have a look, and this can take time since he has a lot to do (on this project and others) these days.

dheid:
@gsmet Do you have time to look at this pull request?

gsmet:
Rebased to get the latest CI fixes.

gsmet:
Hey @dheid 

Sorry for the delay, I just got back from PTO and things were crazy before and obviously even crazier after.

This is on my TODO list. I definitely think it has value and it looks like a very nice work.

I'll ping you as soon as I'll get to it.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/UUIDValidator.java
```diff
@@ -0,0 +1,174 @@
...
+				if ( letterCase == LetterCase.UPPER_CASE && numericValue > 9 && !Character.isUpperCase( ch ) ) {
+					return false;
+				}
+				checksum += numericValue;
+				version = extractVersion( version, charIndex, numericValue );
+				variant = extractVariant( variant, charIndex, numericValue );
+
+			}
+
+		}
```
beikov, line 121: This is not a request for a change, but I just wanted to share how I would have expected this to look like in case you want to change the implementation.
I would have expected a constant loop from 0 to 35 with a big switch covering all indices and asserting a valid char is used. Something like this:

```suggestion
		LOOP: for ( int charIndex = 0; charIndex < 36; charIndex++ ) {

			char ch = value.charAt( charIndex );

			switch ( charIndex ) {
				// Handle hyphens
				case 9:
				case 14:
				case 19:
				case 24:
					if ( ch != '-' ) {
						return false;
					}
					continue LOOP;
				// Handle M and N parts
				case 15:
					version = extractVersion( ch );
					break;
				case 20:
					variant = extractVariant( ch );
					break;
			}

			int numericValue = Character.digit( ch, 16 );
			if ( numericValue == -1 ) {
				// not a hex digit
				return false;
			}
			switch ( letterCase ) {
				case LOWER_CASE:
					if ( numericValue > 9 && !Character.isLowerCase( ch ) ) {
						return false;
					}
					break;
				case UPPER_CASE :
					if ( numericValue > 9 && !Character.isUpperCase( ch ) ) {
						return false;
					}
					break;
			}
			checksum += numericValue;
		}
```

beikov: Since the chars have to be in ASCII encoding, you could go even a step further if you like by introducing an array for the character code points.

```java
private static final byte[] DIGITS = new byte[] {
        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
        -1, -1, -1, -1, -1, -1, -1, -1,  0,  1,  2,  3,  4,  5,  6,  7,  8,  9, -1, -1,
        -1, -1, -1, -1, -1, 10, 11, 12, 13, 14, 15, -1, -1, -1, -1, -1, -1, -1, -1, -1,
        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 10, 11, 12,
        13, 14, 15
};
int charAsInt = ch;
int numericValue = charAsInt >= 0 && charAsInt < DIGITS.length ? DIGITS[charAsInt] : -1;
if ( numericValue == -1 ) {
	// not a hex digit
	return false;
}

switch ( letterCase ) {
	case LOWER_CASE:
		if ( ch > 'A' && ch < 'F' ) {
			return false;
		}
		break;
	case UPPER_CASE :
		if ( ch > 'a' && ch < 'f' ) {
			return false;
		}
		break;
}
```

beikov (approved):
Other than my suggestion the code looks good to me.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/UUIDValidator.java
jrenaat: Tbh Christian, I find Daniel's approach more concise and I don't see how your suggestion actually improves upon it?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/UUIDValidator.java
```diff
@@ -0,0 +1,174 @@
...
+		if ( checksum == 0 ) {
+			return allowNil;
+		}
+		else {
+			if ( Arrays.binarySearch( this.version, version ) == -1 ) {
+				return false;
+			}
+			return Arrays.binarySearch( this.variant, variant ) != -1;
+		}
+
```
jrenaat, line 132: Shouldn't there be some sort of check in here that the version belongs to {1,2,3,4,5} ?
In the UUIDValidatorTest you do verify that versions match, but you're using versions that are (IIUC) are not allowed?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/UUIDValidator.java
dheid: I like this creative solution! But if you don't mind I'll stick to my solution, because it's easier to understand from my point of view.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/UUIDValidator.java
dheid: I hope that I understand you correctly: The version can have a value that is not yet officially existing. But maybe in the future there will be a new kind of UUID and the implementation will support it.

jrenaat: Ok, in general I'm somewhat more in favor of not foreseeing everything that could potentially happen in the future, but it's not a big issue, I'll leave it up to you. I do like your validation implementation, I think it's quite elegant.

dheid: Thank you! Really cheers me up. ☺️

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/UUIDValidatorTest.java
```diff
@@ -0,0 +1,287 @@
...
+		assertFalse( uuidValidator.isValid( "2d5614ff-891e-07a8-b49e-3758506a9bab", null ) );
+		assertFalse( uuidValidator.isValid( "2d5614ff-891e-07a8-b49e-4758506a9bab", null ) );
+		assertFalse( uuidValidator.isValid( "2d5614ff-891e-07a8-b49e-5758506a9bab", null ) );
+		assertFalse( uuidValidator.isValid( "2d5614ff-891e-07a8-b49e-6758506a9bab", null ) );
+		assertFalse( uuidValidator.isValid( "2d5614ff-891e-07a8-b49e-7758506a9bab", null ) );
+		assertFalse( uuidValidator.isValid( "2d5614ff-891e-07a8-b49e-8758506a9bab", null ) );
+		assertFalse( uuidValidator.isValid( "2d5614ff-891e-07a8-b49e-9758506a9bab", null ) );
+		assertFalse( uuidValidator.isValid( "2d5614ff-891e-07a8-b49e-a758506a9bab", null ) );
+		assertFalse( uuidValidator.isValid( "2d5614ff-891e-07a8-b49e-b758506a9bab", null ) );
+
```
jrenaat, line 283: Minor detail: is there a point in doing 12 invalid variant checks if only the last group is different?

dheid: Haha, that's a mistake. I'll fix that.

dheid:
@jrenaat @yrodiere @beikov @gsmet @marko-bekhta I extracted the letter case check to a separate method for enhanced readability

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/UUIDValidatorTest.java
jrenaat: yep, thanks

gsmet (approved):
Thanks for the various reviews everyone.

This looks good and will go into 8.0.0 (note that 8.0.0 is targeting Jakarta EE 10 and will be released when we have all the Jakarta EE 10 specs in their final versions). I adjusted the `@since` accordingly.

gsmet:
@dheid I will merge as soon as CI is green. Thanks for this work!

gsmet:
And merged! Thanks.

dheid:
Thanks everyone so much!

PR #1209, opened 2022-02-08
HV-1877 Move the release job to a separate Jenkinsfile

yrodiere:
https://hibernate.atlassian.net/browse/HV-1877

As explained in [HSEARCH-4472](https://hibernate.atlassian.net/browse/HSEARCH-4472), the main purpose of these changes is the third commit, which makes sure we abort previous builds when a new one starts, thereby saving on CI resources. 

However, I think it's also an improvement to separate the release job from the main one, because what happens on releases becomes more clear, and the main job becomes slightly simpler (there's more work there, but it's a start).

I created a pipeline job for this: https://ci.hibernate.org/view/Validator/job/hibernate-validator-release/
I moved the legacy job that we used in 5.x there: https://ci.hibernate.org/view/Validator/job/hibernate-validator-release-legacy/

### DIFF
--- a/jenkins/release.groovy
+++ b/jenkins/release.groovy
@@ -18,6 +18,7 @@ pipeline {
 		jdk 'OpenJDK 17 Latest'
 	}
 	options {
+		buildDiscarder logRotator(daysToKeepStr: '30', numToKeepStr: '10')
 		disableConcurrentBuilds(abortPrevious: false)
 	}
 	parameters {
--- a/jenkins/release.groovy
+++ b/jenkins/release.groovy
@@ -1,0 +1,92 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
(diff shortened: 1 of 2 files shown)


jenkins/release.groovy
```diff
@@ -0,0 +1,93 @@
...
+
+import org.hibernate.jenkins.pipeline.helpers.version.Version
+
+pipeline {
+	agent {
+		label 'Worker&&Containers'
+	}
+	tools {
+		maven 'Apache Maven 3.8'
+		jdk 'OpenJDK 11 Latest'
```
gsmet, line 18: I recently switched the release build to Java 17 to prepare for the support of records.

gsmet (changes requested):
I added a comment that applies to both `main` and `6.2`.

jenkins/release.groovy
yrodiere: Fixed for both `main` and `6.2`.

PR #1219, opened 2022-02-08
HV-1882 Use a different JDK for running Maven than for compiling/launching tests in JDK CI jobs

yrodiere:
https://hibernate.atlassian.net/browse/HV-1882

To run the tests against JDK8, do this:

```
mvn clean install -Djava-version.test.launcher.java_home=<path to JDK 8> -Djava-version.test.release=8
```

For JDK11:

```
mvn clean install -Djava-version.test.compiler.java_home=<path to JDK 11> -Djava-version.test.release=11
```

And so on.

Advantages over the previous solution:

1. Maven plugins always run on JDK17. If a Maven plugin isn't compatible with JDK 18/19/etc., we simply don't care.
2. The main code always get compiled with JDK17 into Java 8 bytecode, just as it would during releases, even when testing JDK8/11/18/19.
    Previously we would use JDK 8/11/18/19 to compile the main code, which is different from what we'd do during releases.
    Thus, the tests became that much closer to really using the published binaries.
3. When testing a particular JDK, *test* code is compiled into bytecode for the version of the JDK being tested. For example when we test JDK 18, we actually produce Java 18 bytecode for tests then run that bytecode with JDK 18.
    Previously we would produce Java 8 bytecode for tests, even when testing with JDK 11/17/18/19.
    Thus, the tests became that much closer to what a user of JDK 11/17/18/19 would do.


Jenkinsfile
```diff
@@ -117,29 +118,23 @@ stage('Configure') {
 			jdk: [
 					// This should not include every JDK; in particular let's not care too much about EOL'd JDKs like version 9
 					// See http://www.oracle.com/technetwork/java/javase/eol-135779.html
-					new JdkBuildEnvironment(version: '17', buildJdkTool: 'OpenJDK 17 Latest',
+					new JdkBuildEnvironment(testJavaVersion: '17', testCompilerTool: 'OpenJDK 17 Latest',
 							condition: TestCondition.BEFORE_MERGE,
 							isDefault: true),
-					new JdkBuildEnvironment(version: '8', buildJdkTool: 'OracleJDK8 Latest',
+					new JdkBuildEnvironment(testJavaVersion: '8', testLauncherTool: 'OracleJDK8 Latest',
```
yrodiere, line 124: Before you ask... no, I don't know why we only override the launcher, and not the compiler, when using JDK8. I just know that's how it's done in Hibernate Search, so I didn't take any risk and did the same here.

yrodiere: I cleared that up: using a JDK8 compiler with `--release` (which is what we use in pom.xml now) leads to fatal errors. So, it's just simpler to compile with JDK11 to Java 8 bytecode.

engine/pom.xml
```diff
@@ -99,6 +99,18 @@
...
+                 because OracleJDK 8 already bundles JavaFX. -->
+            <groupId>org.openjfx</groupId>
+            <artifactId>javafx-base</artifactId>
+            <!-- The JavaFX dependency is optional (not pulled transitively by our consumers) for two reasons:
+                 1. HV should work fine without it, as long as you don't use JavaFX.
+                 2. We don't want to force the dependency on our users who are still on OracleJDK 8,
+                    since OracleJDK 8 already bundles JavaFX.
+              -->
+            <optional>true</optional>
+        </dependency>
```
yrodiere, line 113: Be aware of this particular change.

I don't think it changes much for consumers of HV: the dependency is marked as optional, which means it won't transitively "contaminate" consuming projects.

However, the dependency is now necessary when testing with JDK8 as well as when testing with JDK11, because we never compile with JDK8 (we just run the tests with JDK8). So it's just simpler to always include the dependency, and add classpath exclusions to remove it at runtime when testing with JDK8 (see the maven-surefire-plugin configuration changes in the root POM).

If this optional dependency is a problem, I suppose we could just add it to both profiles (`testWithJdk8` and `testWithJdk11+`). Then consuming projects would not be affected at all then. But... this seems unnecessary.

yrodiere:
Note that we had to introduce this `.mvn/jvm.config` file to the 6.2 branch, whose build is now failing because (obviously) the JVM options such as `--add-opens` won't work when running Maven with JDK 8. See https://ci.hibernate.org/blue/organizations/jenkins/hibernate-validator/detail/6.2/44/pipeline/61

So... we might want to backport this to branch 6.2 after all.

One alternative would be to tweak the Jenkinsfile on branch 6.2 to remove `.mvn/jvm.config` just before we do a build with JDK8. Not sure that's much better.

gsmet (approved):

PR #1223, opened 2022-02-10
HV-1883 Upgrade to asciidoctor-maven-plugin 2.2.2

yrodiere:
https://hibernate.atlassian.net/browse/HV-1883


yrodiere:
Note the build failure is on the JDK 8 build, which also fails on the main branch. This is an unrelated problem which should be fixed by #1219 .

gsmet (approved):

gsmet:
Thanks!

PR #1230, opened 2022-02-24
HV-1882 follow-up: Make sure to copy javafx.api modules to WildFly when running the TCK with JDK8

yrodiere:
https://hibernate.atlassian.net/browse/HV-1882

Follows up on #1219, #1229

I made the mistake of removing this config in #1219, and did not notice it because we don't run the TCK in WildFly (yet) for branch `main`.

### DIFF
--- a/tck-runner/pom.xml
+++ b/tck-runner/pom.xml
@@ -401,6 +401,41 @@
             </build>
         </profile>
         <profile>
+            <id>testWithJdk8</id>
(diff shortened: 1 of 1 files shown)


yrodiere:
See #1231 for the tests, since we don't currently run the TCK in WidFly on this branch.

gsmet (approved):

PR #1264, opened 2022-05-06
HV-1903: Make flexible log tracing in the ConstraintTree class

RJayz:
Add configuration option to hide values currently being validated at TRACE level logging

see https://hibernate.atlassian.net/browse/HV-1903

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
+++ b/engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
@@ -169,6 +169,13 @@ public interface BaseHibernateValidatorConfiguration<S extends BaseHibernateVali
 	String CUSTOM_VIOLATION_EXPRESSION_LANGUAGE_FEATURE_LEVEL = "hibernate.validator.custom_violation_expression_language_feature_level";
 
 	/**
+	 * Property to hide values from trace level logging when validating constraints
+	 * @since 6.2
+	 */
+	@Incubating
+	String HIDE_VALIDATION_VALUE_FROM_TRACE_LOGGING = "hibernate.validator.hide_validation_value_from_trace_logging";
+
+	/**
 	 * <p>
 	 * Returns the {@link ResourceBundleLocator} used by the
 	 * {@link Configuration#getDefaultMessageInterpolator() default message
@@ -480,4 +487,10 @@ public interface BaseHibernateValidatorConfiguration<S extends BaseHibernateVali
 	 */
 	@Incubating
 	S customViolationExpressionLanguageFeatureLevel(ExpressionLanguageFeatureLevel expressionLanguageFeatureLevel);
+
+	/*
+	 * Hide sensitive values
+	 */
+	@Incubating
+	S hideValidationValuesFromTraceLogging(boolean enabled);
 }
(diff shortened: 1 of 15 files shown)


marko-bekhta:
Hey, @RJayz thanks for submitting the PR. It looks fine overall. I'm not 100% convinced if we should add the configuration or not. It would be up to @gsmet to decide. In the meantime... wouldn't configuring the logger do this for us without additional changes? For example, in case of log4j2 that is used in HV for tests, we can have something like:
```
....
rootLogger.level = trace
....
logger.simpleconstrainttree.name = org.hibernate.validator.internal.engine.constraintvalidation.SimpleConstraintTree
logger.simpleconstrainttree.level = info
logger.composingconstrainttree.name = org.hibernate.validator.internal.engine.constraintvalidation.ComposingConstraintTree
logger.composingconstrainttree.level = info
```
This way we'd have the application logging at `trace` level and our constraint tree logs will stay on `info` level and wouldn't expose the validated bean state into the logs. Wouldn't that cover the use case?

If we'd decide to get this one in, then I would suggest we add a section to the documentation under chapter 9  `Configuring a ValidatorFactory` see `ch09.asciidoc`. And it would be nice to have a test for it. We have `CustomViolationExpressionLanguageFeatureLevelTest` or `ParameterMessageInterpolatorTest` that are using `ListAppender` for testing logging messages, in case an example of such test is needed.

gsmet:
Hey there. Just a quick note that I'm thinking about this. I wonder if not showing the values wouldn't be a safer default. I will let you know in a few days.

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -168,6 +168,13 @@
 	@Incubating
 	String CUSTOM_VIOLATION_EXPRESSION_LANGUAGE_FEATURE_LEVEL = "hibernate.validator.custom_violation_expression_language_feature_level";
 
+	/**
+	 * Property to hide values from trace level logging when validating constraints
+	 * @since 6.2
+	 */
+	@Incubating
+	String HIDE_VALIDATION_VALUE_FROM_TRACE_LOGGING = "hibernate.validator.hide_validation_value_from_trace_logging";
```
gsmet, line 176: I thought more about it and I think we should hide the values by default.
Could you adjust your patch to make it `show**` (we prefer positive intent) being false by default.

engine/src/main/java/org/hibernate/validator/BaseHibernateValidatorConfiguration.java
```diff
@@ -168,6 +168,13 @@
 	@Incubating
 	String CUSTOM_VIOLATION_EXPRESSION_LANGUAGE_FEATURE_LEVEL = "hibernate.validator.custom_violation_expression_language_feature_level";
 
+	/**
+	 * Property to hide values from trace level logging when validating constraints
+	 * @since 6.2
```
gsmet, line 173:
```suggestion
	 * @since 8.0
```

gsmet (changes requested):
I think we should pursue this but I think we should make hiding values the default behavior.

I added suggestions along these lines.

Are you interested in pursuing this and adjusting the patch?

gsmet:
Closing in favor of https://github.com/hibernate/hibernate-validator/pull/1270 .

gsmet:
Thanks for pushing this, it was indeed an issue and making this behavior the default seems even better.

PR #1267, opened 2022-05-22
HV-1905 Add more tests for record validation

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-1905

- add composing constraints tests with correct/incorrect targets
- add cascading constraint test

While checking the reported issue I've added a couple more tests, and it probably wouldn't hurt to do a bit more testing - hence this PR 😄


gsmet (approved):
Thanks!

PR #1289, opened 2022-09-12
HV-1927 Use Executable#getAnnotatedParameterTypes() instead of Executable#getGenericParameterTypes()

gsmet:
getGenericParameterTypes() is problematic with synthetic and implicit parameters and we used to have ugly workarounds and warnings. Using getAnnotatedParameterTypes() solves the issue.

I was extremely conservative in the changes as I want to be able to backport it.

https://hibernate.atlassian.net/browse/HV-1927


gsmet:
@yrodiere could you have a look? We could probably clean up a few other things but I prefer not breaking anything. I did a few more changes at first and ended up with various test failures so I reverted back to only safe changes.

yrodiere:
@gsmet Merged. Will you port this to 7.x/8.x? I'll leave the JIRA issue open until you confirm.

PR #1306, opened 2023-02-23
HV-1939 Annotations on constructors of named classes declared in method get ignored

yrodiere:
https://hibernate.atlassian.net/browse/HV-1939

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanExecutable.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanExecutable.java
@@ -14,6 +14,7 @@ import java.lang.reflect.Modifier;
 import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanParameter.java
```diff
@@ -63,10 +67,14 @@ public AnnotatedType getAnnotatedType() {
 	@Override
 	public Annotation[] getDeclaredAnnotations() {
 		try {
+			if ( annotationsForJDK8303112 != null ) {
+				// Working around https://bugs.openjdk.org/browse/JDK-8303112
+				return annotationsForJDK8303112.clone();
```
yrodiere, line 72: We can remove the `.clone()` if you're absolutely certain nobody will ever try to mutate the array.

engine/src/main/java/org/hibernate/validator/internal/properties/javabean/JavaBeanExecutable.java
```diff
@@ -183,14 +183,39 @@ private static List<JavaBeanParameter> getParameters(Executable executable) {
...
+					annotationsForJDK8303112[i] =
+							parameterAnnotationsArray[nonImplicitNorSyntheticParamIndex];
+					++nonImplicitNorSyntheticParamIndex;
+				}
+			}
+		}
+
 		for ( int i = 0; i < parameterArray.length; i++ ) {
-			parameters.add( new JavaBeanParameter( i, parameterArray[i], parameterTypes[i], annotatedTypes[i] ) );
+			parameters.add( new JavaBeanParameter( i, parameterArray[i], parameterTypes[i], annotatedTypes[i],
```
marko-bekhta, line 217: Maybe it's worth just removing the `Parameter parameter` from `JavaBeanParameter` and always constructing this array of annotations? These annotations are iterated in any case when metadata is created.

yrodiere: Maybe... I went for the approach that affects as few applications as possible to avoid introducing regressions, but we can always pile up performance improvements on to of that.

yrodiere:
Marking as draft since I need to add a safeguard for when people don't use `-parameters`, because then `isImplicit`/`isSynthetic` always return `false` and we may hit an `IndexOutOfBoundsException`.

yrodiere:
Here, now if the class wasn't compiled with `-parameters` we'll assume the implicit/synthetic parameters are the first ones, because that's true in the only problematic case that I know of: method-local classes.
If you prefer I guess we could just ignore all annotations instead, but that seems worse to me.

I didn't add any tests for the case where the class wasn't compiled with `-parameters`, because I don't think we test this anywhere? If I'm wrong please point me to the right place and I'll add tests.

gsmet (approved):

PR #1318, opened 2023-05-26
HV-1950 Suppress java.lang.NoClassDefFoundError: com/sun/el/ExpressionFactoryImpl

quaff:
It will mislead user that hibernate validator will rely on particular EL implementation.

see https://hibernate.atlassian.net/browse/HV-1950.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/messageinterpolation/ResourceBundleMessageInterpolator.java
+++ b/engine/src/main/java/org/hibernate/validator/messageinterpolation/ResourceBundleMessageInterpolator.java
@@ -38,6 +38,7 @@ import jakarta.el.ExpressionFactory;
  * @author Kevin Pollet &lt;kevin.pollet@serli.com&gt; (C) 2011 SERLI
  * @author Adam Stawicki
  * @author Guillaume Smet
+ * @author Yanming Zhou
  */
 public class ResourceBundleMessageInterpolator extends AbstractMessageInterpolator {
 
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/hibernate/validator/messageinterpolation/ResourceBundleMessageInterpolator.java
```diff
@@ -200,11 +199,16 @@ private static ExpressionFactory buildExpressionFactory() {
 
 			// Finally we try the CL of the EL implementation itself. This is necessary for OSGi now that the
 			// implementation is separated from the API.
-			run( SetContextClassLoader.action( ExpressionFactoryImpl.class.getClassLoader() ) );
-			if ( canLoadExpressionFactory() ) {
-				ExpressionFactory expressionFactory = ELManager.getExpressionFactory();
-				LOG.debug( "Loaded expression factory via com.sun.el classloader" );
-				return expressionFactory;
+			try {
+				run( SetContextClassLoader.action( Class.forName( "com.sun.el.ExpressionFactoryImpl" ).getClassLoader() ) );
```
marko-bekhta, line 203: Hey, thanks for submitting this PR.

```java
    @CallerSensitive
    public static Class<?> forName(String className)
                throws ClassNotFoundException {
        Class<?> caller = Reflection.getCallerClass();
        return forName0(className, true, ClassLoader.getClassLoader(caller), caller);
    }
```
Since that's how the `Class.forName` is implemented wouldn't we just end up having the same class loader as in the very first try of this `buildExpressionFactory()`? This part:
```java
if ( canLoadExpressionFactory() ) {
			ExpressionFactory expressionFactory = ELManager.getExpressionFactory();
			LOG.debug( "Loaded expression factory via original TCCL" );
			return expressionFactory;
		}
```

It would also be nice if the changes were accompanied by a test case; thanks!

gsmet: I would refrain from changing the logic in there. It is very brittle and we worked around multiple issues with OSGi and the modular classpath from WildFly.

quaff: I've revert to `ExpressionFactoryImpl.class.getClassLoader()`, just catch `NoClassDefFoundError` and ignore it.

marko-bekhta:
With dependency on Expressly removed in https://hibernate.atlassian.net/browse/HV-2020 I think we can close this PR. Thanks!

PR #1331, opened 2023-09-15
HV-1831 Optimize cascading validation for large lists

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-1831

this is only a version of https://github.com/hibernate/hibernate-validator/pull/1157 for a main branch with javax->jakarta moved so far.

----------------------

----------------------


engine/src/main/java/org/hibernate/validator/internal/engine/tracking/PredefinedScopeProcessedBeansTrackingStrategy.java
```diff
@@ -0,0 +1,368 @@
...
+public class PredefinedScopeProcessedBeansTrackingStrategy implements ProcessedBeansTrackingStrategy {
+
+	private final Map<Class<?>, Boolean> trackingEnabledForBeans;
+
+	// TODO: signature is just name and parameters so that can clash between different beans.
+	//  with that.. do we even need to track it per signature or since
+	//  we already built the `trackingEnabledForBeans` we can just "inspect" the cascadable as we go
+	//  and check against this `trackingEnabledForBeans` to see if tracking is required ?
+	private final Map<Signature, Boolean> trackingEnabledForReturnValues;
+	private final Map<Signature, Boolean> trackingEnabledForParameters;
```
marko-bekhta, line 36: I'm not sure we need to prebuild these maps, but I wanted to check with you first before dropping them...
As per that TODO, it seems that we would only check an executable once when building the metadata and I think I got an impl that would rely only on the `trackingEnabledForBeans` ...  
and since the signatures could clash ... I guess we can drop these maps, right?

engine/src/main/java/org/hibernate/validator/internal/engine/tracking/PredefinedScopeProcessedBeansTrackingStrategy.java
```diff
@@ -0,0 +1,368 @@
...
+			this.rawBeanMetaDataMap = rawBeanMetaDataMap;
+			this.classToBeanTrackingEnabled = CollectionHelper.newHashMap( rawBeanMetaDataMap.size() );
+			this.subtypesMap = CollectionHelper.newHashMap( rawBeanMetaDataMap.size() );
+			for ( Class<?> beanClass : rawBeanMetaDataMap.keySet() ) {
+				for ( Class<?> otherBeanClass : rawBeanMetaDataMap.keySet() ) {
+					if ( beanClass.isAssignableFrom( otherBeanClass ) ) {
+						subtypesMap.computeIfAbsent( beanClass, k -> new HashSet<>() )
+								.add( otherBeanClass );
+					}
+				}
```
marko-bekhta, line 69: This subtype map is here to help identify subtypes that can potentially, at runtime, introduce a cycle, and if that's so, we would mark that path as requiring the bean tracking.

engine/src/main/java/org/hibernate/validator/internal/engine/tracking/PredefinedScopeProcessedBeansTrackingStrategy.java
```diff
@@ -0,0 +1,368 @@
...
+		//
+		//
+		// We also need to account for the case when the subtype is used at runtime that may change the cycles:
+		//  4) A -> B -> C -> D
+		//     And C1 extends C where C1 -> A
+		//     Hence, at runtime we "may" get:
+		//     A -> B -> C1 -> D
+		//     ^          |
+		//     |          |
+		//     -----------
```
marko-bekhta, line 125: this example is for that "sub-type" map...

engine/src/main/java/org/hibernate/validator/internal/engine/tracking/PredefinedScopeProcessedBeansTrackingStrategy.java
```diff
@@ -0,0 +1,368 @@
...
+			final ContainerCascadingMetaData containerCascadingMetaData = cascadingMetaData.as( ContainerCascadingMetaData.class );
+			processContainerCascadingMetaData( containerCascadingMetaData, directCascadedBeanClasses );
+		}
+		else if ( cascadingMetaData instanceof PotentiallyContainerCascadingMetaData potentiallyContainerCascadingMetaData ) {
+			// if it's a potentially container cascading one, we are "in trouble" as thing can be "almost anything".
+			// TODO: would it be enough to just take the type as defined ?
+			//  directCascadedBeanClasses.add( (Class<?>) cascadable.getCascadableType() );
+			//
+			// TODO: or be much more cautious and just assume that it can be "anything":
+			directCascadedBeanClasses.add( Object.class );
```
marko-bekhta, line 232: My thoughts here are that if we are unsure if the tracking could be required or not we should stay on the safe side and assume it would be... 
these potentially-container-cascading cases are trouble anyway so I'd hope users would stay away from them ... 

btw, this also reminds me of the other "problem", once we drop the support for `@Valid List<MyBean>`, things with these potentially cascading cases won't really work. I've started adding some thoughts on that here: https://github.com/jakartaee/validation/issues/266

engine/src/main/java/org/hibernate/validator/internal/engine/tracking/PredefinedScopeProcessedBeansTrackingStrategy.java
```diff
@@ -0,0 +1,368 @@
...
+
+		Set<Class<?>> directCascadedBeanClasses = new HashSet<>();
+		for ( Cascadable cascadable : returnValueMetaData.getCascadables() ) {
+			processSingleCascadable( cascadable, directCascadedBeanClasses );
+		}
+		for ( Class<?> directCascadedBeanClass : directCascadedBeanClasses ) {
+			if ( trackingEnabledForBeans.get( directCascadedBeanClass ) ) {
+				return true;
+			}
+		}
```
marko-bekhta, line 329: the idea here was to reuse the logic from bean tracking detection ^ and go through cascadables, collect what "top level" classes we can cascade into and then use the `trackingEnabledForBeans` map to see if tracking is required ...

engine/src/main/java/org/hibernate/validator/resourceloading/PlatformResourceBundleLocator.java
```diff
@@ -149,7 +149,12 @@ public PlatformResourceBundleLocator(String bundleName,
 		if ( !localesToInitialize.isEmpty() ) {
 			Map<Locale, ResourceBundle> tmpPreloadedResourceBundles = CollectionHelper.newHashMap( localesToInitialize.size() );
 			for ( Locale localeToPreload : localesToInitialize ) {
-				tmpPreloadedResourceBundles.put( localeToPreload, doGetResourceBundle( localeToPreload ) );
+				ResourceBundle bundle = doGetResourceBundle( localeToPreload );
+				if ( bundle == null ) {
+					LOG.resourceBundleNotPreLoaded( localeToPreload );
+					continue;
+				}
+				tmpPreloadedResourceBundles.put( localeToPreload, bundle );
```
marko-bekhta, line 157: had to do this, as with the change to `Map.copyOf` `null`s are not allowed anymore and the startup would fail ... (noticed it when running the benchmarks)

marko-bekhta (commented):
I've added an implementation for detecting return value/parameters tracking detection and tried to go through more cases for bean tracking detection and adjusted the steps to account for more scenarios I could think of. Instead of looking at the validated type I tried to leverage the cascading metadata instead... my thinking was that we may have something like:
```java
Map<KeyThatHasCycles, @Valid ValueNoCycles> prop;
``` 
and if we just inspect the map we'd think that we need tracking, but since there is no `@Valid` for the key we actually don't. + this seems to simplify the case when the `@Valid` is nested much deeper in the type arguments. 

And some numbers:

with the tracking changes:
```
Benchmark                                                                                                     Mode  Cnt      Score     Error  Units
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                                      thrpt   20  16303.592 ± 205.026  ops/s
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems:async                                thrpt             NaN              ---
PredefinedScopeCascadedWithLotsOfItemsValidation.testPredefinedScopeCascadedValidationWithLotsOfItems        thrpt   20  24974.083 ± 276.476  ops/s
PredefinedScopeCascadedWithLotsOfItemsValidation.testPredefinedScopeCascadedValidationWithLotsOfItems:async  thrpt             NaN              ---
Benchmark                                                                                               Mode  Cnt      Score     Error  Units
CascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems                       thrpt   20  10825.431 ± 154.286  ops/s
CascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems:async                 thrpt             NaN              ---
PredefinedScopeCascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems        thrpt   20  10500.602 ± 139.836  ops/s
PredefinedScopeCascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems:async  thrpt             NaN              ---
```

^ this seems reasonable, as when there are no cycles and we detect that tracking can be skipped in a predefined scope case things look much better, while with cycles in place the results are similar between the predefine scope and regular validator.

btw I also tried to do the tracking-without-maps:
```
tracking+no maps:

Benchmark                                                                                               Mode  Cnt      Score     Error  Units
CascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems                       thrpt   20  11101.882 ± 113.787  ops/s
CascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems:async                 thrpt             NaN              ---
PredefinedScopeCascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems        thrpt   20  11104.529 ± 109.809  ops/s
PredefinedScopeCascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems:async  thrpt             NaN              ---
```

and the results are only slightly better than when the maps are used ... will take a closer look at it to see if it can be improved

gsmet:
I had a look at the numbers but I'm not sure to what I should compare them to? Could you clarify?

gsmet:
I think I understand them: could you check if in the case with a cycle, we are not consistently slower? I would have expected things to be mostly transparent?

gsmet:
Also both versions tested include your previous optimization, right?

marko-bekhta:
> Also both versions tested include your previous optimization, right?

yes, all the runs are on top of main, so the previous improvements are already included. 



> I had a look at the numbers but I'm not sure to what I should compare them to? Could you clarify?

I haven't initially compared to results from main, just predefined/default, as my thought was that the change is really about disabling the tracking ... here are the result from running the benchmark on main:


```
Benchmark                                                                                                     Mode  Cnt      Score     Error  Units
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                                      thrpt   20  15547.480 ± 187.809  ops/s
PredefinedScopeCascadedWithLotsOfItemsValidation.testPredefinedScopeCascadedValidationWithLotsOfItems        thrpt   20  16213.816 ± 180.518  ops/s
```

so with that compared to the case when we disable tracking:

```
Benchmark                                                                                                     Mode  Cnt      Score     Error  Units
CascadedWithLotsOfItemsValidation.testCascadedValidationWithLotsOfItems                                      thrpt   20  16303.592 ± 205.026  ops/s
PredefinedScopeCascadedWithLotsOfItemsValidation.testPredefinedScopeCascadedValidationWithLotsOfItems        thrpt   20  24974.083 ± 276.476  ops/s
```

default case matches the results from main (no tracking improvements), and the predefined is much better as we disable tracking entirely there.

let me also run the one with cycles on main too...

marko-bekhta:
Here are the results for the case with cycles:

main:
```
Benchmark                                                                                               Mode  Cnt      Score     Error  Units
CascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems                       thrpt   20  10976.153 ± 139.115  ops/s
PredefinedScopeCascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems        thrpt   20  10829.914 ± 130.121  ops/s
```

this patch:
```
Benchmark                                                                                               Mode  Cnt      Score     Error  Units
CascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems                       thrpt   20  10850.865 ± 103.808  ops/s
PredefinedScopeCascadedWithLotsOfItemsAndCyclesValidation.testCascadedValidationWithLotsOfItems        thrpt   20  10835.561 ± 100.276  ops/s
```
 It's slower than `CascadedWithLotsOfItemsValidation`, but I guess the reason is that we have to try to cascade into the object that creates a cycle, and that means at minimum getting the value, which involves reflection and so on.

engine/src/main/java/org/hibernate/validator/internal/engine/path/ModifiableNode.java
```diff
@@ -218,97 +216,47 @@ public static NodeImpl createReturnValue(NodeImpl parent) {
...
-				node.valueSet,
-				node.containerClass,
-				node.typeArgumentIndex
-		);
+	public static ModifiableNode makeIterable(ModifiableNode node) {
+		node.isIterable = true;
+		node.index = null;
+		node.key = null;
+
+		return node;
```
marko-bekhta, line 224: now with the different approach to tracking, we don't need to have paths unless we need to create a constraint violation (and that's when it gets "materialized"). With that in mind, the path can now be fully mutable, and we can get pretty fast if there are no constraint violations:

9.0
```
Benchmark                                                     Mode  Cnt      Score     Error   Units
SimpleSingleElementValidation.invalidObjectValidation        thrpt   10   6433.561 ±  51.257  ops/ms
SimpleSingleElementValidation.validObjectValidation          thrpt   10  17108.253 ± 255.899  ops/ms
```

All changes:
```
Benchmark                                                     Mode  Cnt      Score     Error   Units
SimpleSingleElementValidation.invalidObjectValidation        thrpt   10   7312.547 ±  24.840  ops/ms
SimpleSingleElementValidation.validObjectValidation          thrpt   10  31838.404 ± 156.586  ops/ms
```

gsmet:
Congrats @marko-bekhta ! This is awesome work!

marko-bekhta:
thanks @gsmet 🥳🙂!
I'll try to get a few more things merged and maybe do an alpha 9.1 some time next week 🤞🏻 🙂

PR #1334, opened 2023-11-01
[HV-1831] filters @Valid annotations from jvm and native types

thst71:
https://hibernate.atlassian.net/browse/HV-1831

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
@@ -78,6 +78,7 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import jakarta.validation.Valid;
 import org.hibernate.validator.constraints.CodePointLength;
 import org.hibernate.validator.constraints.ConstraintComposition;
 import org.hibernate.validator.constraints.CreditCardNumber;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
```diff
@@ -1172,6 +1175,30 @@ private static <T> T run(PrivilegedAction<T> action) {
...
+	 * @param <A> type of annotation
+	 */
+	public <A extends Annotation> boolean isNonApplicableValidAnnotation(A annotation, Constrainable constrainable) {
+		if ( !( annotation instanceof Valid ) ) {
+			return false;
+		}
+
+		return constrainable.getType().getTypeName().startsWith( "java." )
+				|| constrainable.getType().getTypeName().startsWith( "javax." )
+				|| constrainable.getType().getTypeName().matches( BUILTIN_TYPE_NAMES );
```
marko-bekhta, line 1199: yeah ... I can see how it can be tempting to ignore these types 😃.
this probably will work for the primitives, but we cannot safely discard any classes. The user may define a model as:
```java
class SomeClass {
    @Valid
    Object object;
}
```
which is a valid use case and the type of object field is checked at the runtime when `SomeClass` instance is passed for validation. So even if the class from these packages will not have any constraints, someone might extend the class and apply the constraints in that extended class... 
Another problem is that even though these classes won't have any constraints declared as annotations, a user can always define a constraint through a programmatic DSL, or via XML...
That's why I'm leaning more towards this being a problem of the code generators that are adding redundant `@Valid` annotations ...

thst71: Ok, good point. I thought about making this configurable, but I need to look at other config properties and how these are wired. I will remove the java check.

PR #1335, opened 2023-11-07
HV-1968 Force user.country for testing

thst71:
… number formats

The CarTest uses an assertion expecting an US number format (decimal point). Running the build on a german locale, the CarTest fails

### DIFF
--- a/documentation/pom.xml
+++ b/documentation/pom.xml
@@ -39,7 +39,7 @@
         <!-- Skip artifact deployment -->
         <maven.deploy.skip>true</maven.deploy.skip>
         <gpg.skip>true</gpg.skip>
-        <surefire.jvm.args.additional>-Duser.language=en</surefire.jvm.args.additional>
+        <surefire.jvm.args.additional>-Duser.language=en -Duser.country=US</surefire.jvm.args.additional>
 
         <forbiddenapis-junit.path>forbidden-allow-junit.txt</forbiddenapis-junit.path>
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -22,7 +22,7 @@
 
     <properties>
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
-        <surefire.jvm.args.additional>-Duser.language=en</surefire.jvm.args.additional>
+        <surefire.jvm.args.additional>-Duser.language=en -Duser.country=US</surefire.jvm.args.additional>
     </properties>
 
     <distributionManagement>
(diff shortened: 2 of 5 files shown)


yrodiere:
I think the right fix would be to force the locale when running tests, no? And I thought that was exactly what we did...

yrodiere:
> I think the right fix would be to force the locale when running tests, no? And I thought that was exactly what we did...

@gsmet would know

gsmet:
Yeah, that's what we are supposed to do: https://github.com/hibernate/hibernate-validator/blob/main/documentation/pom.xml#L42

Are you running a full Maven build or are you running the test from the IDE?

thst71:
I ran this from fish with `mvn clean install`, also I reproduced it several times. The build only completes, if I enforce the locale. Maybe you can enforce it by setting the locale to DE_de?Am 08.11.2023 um 09:33 schrieb Guillaume Smet ***@***.***>:﻿
Yeah, that's what we are supposed to do: https://github.com/hibernate/hibernate-validator/blob/main/documentation/pom.xml#L42
Are you running a full Maven build or are you running the test from the IDE?

thst71:
> Yeah, that's what we are supposed to do: https://github.com/hibernate/hibernate-validator/blob/main/documentation/pom.xml#L42
> 
> Are you running a full Maven build or are you running the test from the IDE?

Ok, this is the problem, the user.language=en only sets the display language, it will not overwrite the number formats: 
```
jshell -R -Duser.language=en
|  Welcome to JShell -- Version 17.0.9
|  For an introduction type: /help intro

jshell> System.out.println(java.util.Locale.getDefault());
en_DE

jshell> System.out.println(java.text.NumberFormat.getInstance().format(400.234));
400,234
```

thst71:
I changed the PR, it now adds a user.country to every location where a user.language is set.

documentation/pom.xml
```diff
@@ -39,7 +39,7 @@
         <!-- Skip artifact deployment -->
         <maven.deploy.skip>true</maven.deploy.skip>
         <gpg.skip>true</gpg.skip>
-        <surefire.jvm.args.additional>-Duser.language=en</surefire.jvm.args.additional>
+        <surefire.jvm.args.additional>-Duser.language=en -Duser.country=EN</surefire.jvm.args.additional>
```
gsmet, line 42: I don't think `EN` is a valid country for English or it is an obscure one. Maybe you should make it `US` instead?

Other than that, I agree we should merge this.

gsmet: Ideally, when you are done with the `US` change, please squash your comments and prefix the commit message with `HV-1968 `.

Thanks!

thst71: done :-)

gsmet (approved):
Perfect thanks!

PR #1338, opened 2024-01-12
HV-1970 Add Korean specific RRN annotation

ing9990:
[Jira-(HV-1970)](https://hibernate.atlassian.net/browse/HV-1970)

Hai I tried to add Korean specific @RRN annotation.
I added annotation, validator, testcases, messages, etc.


engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/kor/KorRRNValidator.java
```diff
@@ -0,0 +1,29 @@
...
+import jakarta.validation.ConstraintValidatorContext;
+
+import java.util.regex.Pattern;
+
+import org.hibernate.validator.constraints.kor.KorRRN;
+
+public class KorRRNValidator implements ConstraintValidator<KorRRN, CharSequence> {
+
+	private static final Pattern KOR_RRN_REGEX = Pattern.compile(
+			"\\d{2}(0[1-9]|1[0-2])(0[1-9]|[12]\\d|3[01])-[1234]\\d{6}$" );
```
marko-bekhta, line 20: Do you think it is worth to check the check digit for the older numbers?

> c, the 13th digit, is a check digit, used to verify that the number has been transcribed correctly. It is generated from the rest of the digits using the following algorithm (digits lettered a through m in order):
m = [11 − {(2a + 3b + 4c + 5d + 6e + 7f + 8g + 9h + 2i + 3j + 4k + 5l) mod 11}] mod 10

marko-bekhta (commented):
Thanks for your pull request. 

This post on adding new constraints may be interesting to you, if you haven't seen it before: https://in.relation.to/2018/01/04/adding-new-constraint-to-engine/ . See the annotation processor and documentation steps in particular. 
Also, maybe you could update the `ValidationMessages_ko.properties` file with the new message as well?

engine/src/main/java/org/hibernate/validator/constraints/kor/KorRRN.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Checks that the annotated character sequence is a valid Korean resident registration number.
+ *
+ * @author Taewoo Kim
+ * @see <a href="https://ko.wikipedia.org/wiki/주민등록번호">Korean resident registration number</a>
```
marko-bekhta, line 28:
```suggestion
 * @see <a href="[https://ko.wikipedia.org/wiki/주민등록번호](https://ko.wikipedia.org/wiki/%EC%A3%BC%EB%AF%BC%EB%93%B1%EB%A1%9D%EB%B2%88%ED%98%B8)">Korean resident registration number</a>
```

Let's use an encoded url here as the current one seems to cause javadoc generation failure

ing9990: Thanks for your suggestion. I fixed it as per your suggestion

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/kor/KorRRNValidator.java
ing9990: What does "old number" mean?

ing9990: What is meaning "olrder number"?

marko-bekhta: From what  I understand, for RRNs issued prior to October 2020 the last digit in the number was a check-digit. After that date, it became a random number. So I was asking if it would make sense to validate the chec-digit for numbers issued prior to October 2020.

ing9990: As far as I know, RRNs are still using check digits after October 2020.  Let's say "RRN" is "ABCDEF-GHIJKLM", then
Since October 2020, 'HIJKL' has been changed to a randomized order, but check digits are valid ('M')

[Configuration of RRNs](https://namu.wiki/w/주민등록번호#구성)

ing9990: I checked and the law is ambiguous. 

See [Government homepage 2020.05](https://www.law.go.kr/LSW/lsInfoP.do?lsiSeq=222265&ancYd=20201005&ancNo=00204&efYd=20201005&nwJoYnInfo=N&efGubun=Y&chrClsCd=010202&ancYnChk=0#0000)

See [Government homepage 2020.05](https://www.law.go.kr/LSW/lsInfoP.do?lsiSeq=222265&ancYd=20201005&ancNo=00204&efYd=20201005&nwJoYnInfo=N&efGubun=Y&chrClsCd=010202&ancYnChk=0#0000)

So, I'll check with the government tomorrow.

marko-bekhta: Thank you! that would be nice.

Also as an idea: we could try adding an attribute to the constraint annotation:
```java
public @interface KorRRN {
	// ... 
	boolean validateCheckDigit() default true;
}
```
to make the validation optional. Or take it a step further and:
```java
public @interface KorRRN {
        //...
	ValidateCheckDigit validateCheckDigit() default ValidateCheckDigit.ALWAYS;
	
	enum ValidateCheckDigit {
		NEVER,
		ALWAYS,
		BEFORE_OCTOBER_2020_ONLY
	}
}
```

ing9990: I called the Korean goverment and they said you're right. 😂 This seems like a good idea. So I Updated KorRRN to include "validateCheckDigit"attribute and i added a few test cases.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/kor/KorRRNValidator.java
```diff
@@ -0,0 +1,60 @@
...
+package org.hibernate.validator.internal.constraintvalidators.hv.kor;
+
+import java.util.regex.Pattern;
+
+import jakarta.validation.ConstraintValidator;
+import jakarta.validation.ConstraintValidatorContext;
+
+import org.hibernate.validator.constraints.kor.KorRRN;
+
+public class KorRRNValidator implements ConstraintValidator<KorRRN, CharSequence> {
```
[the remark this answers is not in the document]
ing9990: Thanks. I updated Validate check-digit logic to use ModUtil

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/kor/KorRRNValidatorTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+	@BeforeMethod
+	void setUp() {
+		validator = new KorRRNValidator();
+	}
+
+	// 유효한 주민등록번호의 예시입니다.
+	@Test
+	void validRRN() {
+		assertValidRRN( "861224-2567484" );
```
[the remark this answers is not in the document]
ing9990: I updated to succeed or fail based on validity regardless of hyphen

https://github.com/hibernate/hibernate-validator/pull/1338/commits/a3f45bcb0fc107122ada531a09a72f9a679df258

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/kor/KorRRNValidatorTest.java
```diff
@@ -0,0 +1,118 @@
...
+public class KorRRNValidatorTest {
+
+	private KorRRNValidator validator;
+
+	@BeforeMethod
+	void setUp() {
+		validator = new KorRRNValidator();
+	}
+
+	// 유효한 주민등록번호의 예시입니다.
```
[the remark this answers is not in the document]
ing9990: done 🤗

engine/src/main/java/org/hibernate/validator/constraints/kor/KorRRN.java
```diff
@@ -0,0 +1,63 @@
...
+		NEVER,
+		/**
+		 * Inspect the following lines
+		 * 1. Length of RRN
+		 * 2. Validity of Gender-Digit in RRN
+		 * 3. Validity of date in RRN
+		 * 4. Validity of Check-digitin in RRN
+		 */
+		ALWAYS
+	}
```
marko-bekhta, line 62: Usually, we have this list annotation so that the constraint can be defined as repeatable:
```suggestion
	}
	
	/**
	 * Defines several {@code @KorRRN} annotations on the same element.
	 */
	@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
	@Retention(RUNTIME)
	@Documented
	public @interface List {
		KorRRN[] value();
	}
```
and then on the annotation itself `@Repeatable(KorRRN.class)`. See for example this constraint: https://github.com/hibernate/hibernate-validator/blob/729cd018599d046aa5db1dbe35700521dd596c4e/engine/src/main/java/org/hibernate/validator/constraints/pl/PESEL.java#L33-L56 

While it probably won't be used as a repeatable in most cases, it may be that someone would want to have this constraint added with different `validateCheckDigit` depending on a validation groups, e.g.:

```java
@KorRRN( validateCheckDigit = NEVER, groups = {MyGroup1.class})
@KorRRN( validateCheckDigit = ALWAYS, groups = {MyGroup2.class})
String number;
```

engine/src/main/java/org/hibernate/validator/constraints/kor/KorRRN.java
```diff
@@ -0,0 +1,63 @@
...
+
+	String message() default "{org.hibernate.validator.constraints.kor.KorRRN.message}";
+
+	Class<?>[] groups() default {};
+
+	Class<? extends Payload>[] payload() default {};
+
+	ValidateCheckDigit validateCheckDigit() default NEVER;
+
+	enum ValidateCheckDigit {
```
marko-bekhta, line 46: Maybe let's add a bit more documentation to this? Something like:
```suggestion
        /**
         * Defines the validation rules this constraint applies to the Korean resident registration number.
         * <p>
         * Each type specifies which particular checks will be applied.
         */
	enum ValidateCheckDigit {
```

engine/src/main/java/org/hibernate/validator/constraints/kor/KorRRN.java
```diff
@@ -0,0 +1,63 @@
...
+		/**
+		 * Inspect the following lines
+		 * 1. Length of RRN
+		 * 2. Validity of Gender-Digit in RRN
+		 * 3. Validity of date in RRN
+		 */
```
marko-bekhta, line 52: Javadoc will render this to a single line, so maybe let's update it to the list as: 
```suggestion
		/**
		 * Perform the checks on:
		 * <ul>
		 *     <li>The length of an RRN</li>
		 *     <li>The validity of the Gender-Digit in an RRN</li>
		 *     <li>The validity of the date in an RRN</li>
		 * </ul>
		 */
```
and do the same for the other constant

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/kor/KorRRNValidator.java
```diff
@@ -0,0 +1,150 @@
...
+	public void initialize(KorRRN constraintAnnotation) {
+		this.rrnValidationAlgorithm = RRNValidationAlgorithm.from( constraintAnnotation.validateCheckDigit() );
+	}
+
+	@Override
+	public boolean isValid(CharSequence rrnValue, ConstraintValidatorContext context) {
+		if ( rrnValue == null ) {
+			return true;
+		}
+		return rrnValidationAlgorithm.isValid( REPLACE_HYPEN.apply( rrnValue::toString ) );
```
marko-bekhta, line 50: We can do a simple tostring-replace, to avoid unnecessary lambda functions.
```suggestion
		return rrnValidationAlgorithm.isValid( rrnValue.toString().replace( "-", "" ) );
```

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/kor/KorRRNValidator.java
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	private interface RRNValidationAlgorithm {
```
marko-bekhta, line 53: I'd suggest moving the methods from this interface into the enum. You'd add a `abstract boolean isValid(String rrn);` and other methods without the `default` keyword.

engine/src/test/java/org/hibernate/validator/test/cfg/ProgrammaticConstraintDefinitionsTest.java
```diff
@@ -64,6 +66,9 @@ public void countrySpecificProgrammaticDefinition() {
 
 		doProgrammaticTest( INN.class, new INNDef().type( INN.Type.INDIVIDUAL ), "127530851622", "127530851623", "invalid Russian taxpayer identification number (INN)" );
 		doProgrammaticTest( INN.class, new INNDef().type( INN.Type.JURIDICAL ), "8606995694", "8606995695", "invalid Russian taxpayer identification number (INN)" );
+
+		doProgrammaticTest( KorRRN.class, new KorRRNDef().validateCheckDigit( KorRRN.ValidateCheckDigit.NEVER ), "030205-1000004", "010199-1063015", "invalid Korean resident registration number (KorRRN)" );
```
marko-bekhta, line 70: let's also test that a programmatic constraint will work fine with the default value of `validateCheckDigit`
```suggestion
		doProgrammaticTest( KorRRN.class, new KorRRNDef(), "030205-1000004", "010199-1063015", "invalid Korean resident registration number (KorRRN)" );
		doProgrammaticTest( KorRRN.class, new KorRRNDef().validateCheckDigit( KorRRN.ValidateCheckDigit.NEVER ), "030205-1000004", "010199-1063015", "invalid Korean resident registration number (KorRRN)" );
```

engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/kor/KorRRNValidatorTestHelper.java
```diff
@@ -0,0 +1,26 @@
...
+import org.hibernate.validator.internal.constraintvalidators.hv.kor.KorRRNValidator;
+import org.hibernate.validator.internal.util.annotation.ConstraintAnnotationDescriptor;
+
+/**
+ * Test helper for provide {@link ValidateCheckDigit} into {@link KorRRNValidator}.
+ *
+ * @author Taewoo Kim
+ */
+public abstract class KorRRNValidatorTestHelper {
+	protected KorRRN initAnootation(ValidateCheckDigit validateCheckDigit) {
```
marko-bekhta, line 20:
```suggestion
	protected KorRRN initAnnotation(ValidateCheckDigit validateCheckDigit) {
```
🙈

ing9990: All feedback applied, @marko-bekhta :)

marko-bekhta: thanks!

marko-bekhta (approved):

PR #1339, opened 2024-01-14
HV-1971 fix CNPJ validation when all digits are the same

edurbs:
https://hibernate.atlassian.net/browse/HV-1971

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
@@ -56,10 +56,19 @@ public class CNPJValidator implements ConstraintValidator<CNPJ, CharSequence> {
 			return true;
 		}
 
-		// Check for repeated digits
-		if ( value.toString().chars()
-				.filter( Character::isDigit )
-				.distinct().count() == 1 ) {
+		if ( value.length() < 2 ) {
+			return false;
+		}
+
+		char firstDigit = value.charAt( 0 );
+		char otherDigit = value.charAt( 1 );
+		for ( int i = 2; i < value.length(); i++ ) {
+			char c = value.charAt( i );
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
```diff
@@ -55,6 +56,13 @@ public boolean isValid(CharSequence value, ConstraintValidatorContext context) {
 			return true;
 		}
 
+		// Check for repeated digits
+		if ( value.toString().chars()
```
marko-bekhta, line 60: Maybe let's replace this with a simple for loop? Something along the lines:
```java
// the number is too short, hence it's invalid anyway:
if ( value.length() < 2 ) {
	return false;
}

char firstDigit = value.charAt( 0 );
char otherDigit = value.charAt( 1 );
for ( int i = 2; i < value.length(); i++ ) {
	char c = value.charAt( i );
	if ( Character.isDigit( c ) && firstDigit != c ) {
		otherDigit = c;
	}
}
if ( firstDigit == otherDigit ) {
	return false;
}
```

From what I see `distinct` may not be optimized:
```java
public final IntStream distinct() {
	// While functional and quick to implement, this approach is not very efficient.
	// An efficient version requires an int-specific map/set implementation.
	return boxed().distinct().mapToInt(i -> i);
}
```

marko-bekhta (commented):
Hey, thanks for reporting the problem and submitting a PR. Please see an inline comment.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/br/CNPJValidator.java
edurbs: perfect, I'll change it. Thanks!

marko-bekhta (approved):
Looks good, thanks!

PR #1342, opened 2024-04-08
Update asciidoctor configuration and theme

marko-bekhta:
While this is not really needed for Validator, let's keep it updated and using the latest theme so we won't forget to address this change once we actually would need to use a newer version of the theme for Validator.


yrodiere:
Superseded by #1346

PR #1345, opened 2024-04-17
HV-1975 Remove Security Manager usage

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-1975

I've kept the actions as utils classes with static action methods.

There's also `HibernateValidatorPermission`, which will not be affected by the removal of the SM itself but, at the same time, becomes useless... since it's in the API, I've only marked it as deprecated and for removal.


gsmet:
I'm probably +1 on this one but we need to make sure we won't have to update HV to this version in a component that still requires Security Manager support.
So I think @yrodiere need to approve the overall strategy for the Hibernate projects.

yrodiere:
Agreed on the intent, but this is a feature removal and thus needs to be in the next major version. We'll need a HV 9.

yrodiere (approved):
Now that `main` is HV 9, I think we can rebase and merge @marko-bekhta

yrodiere (approved):

PR #1348, opened 2024-05-27
HV-1978 Propagated EL level to iterable nodes

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-1978


yrodiere (approved):
Sorry for the delay, but next time try to ping someone: there's no way everyone will follow every repo :)

Thanks!

PR #1357, opened 2024-06-17
HV-1994 Stop publishing and remove hibernate-validator-modules / HV-1995 Test against latest Wildfly 32.0.1

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-1975
https://hibernate.atlassian.net/browse/HV-1994
https://hibernate.atlassian.net/browse/HV-1995

I've ended up with moving/slightly updating Maven plugins to remove the `hibernate-validator-modules`. I've tried looking at things like WildFly Glow, feature packs etc, but that turned out to be more complex than we need (or maybe I took a wrong turn somewhere 😃). Anyway, having a few simple maven plugins to do what we need seems more maintainable to me ...


yrodiere (commented):
I wouldn't have expected that using WildFly Glow would be so hard... I just found how to do it with ORM 5.3: https://docs.jboss.org/hibernate/orm/5.3/topical/html_single/wildfly/Wildfly.html
It does require having feature packs though, that part I probably misremembered.

Anyway, if it really is hard, fine by me, but you'll need to make sure someone can still use the HV build to test against a random version of WildFly (e.g. one that lives locally only, for productization). In Hibernate Search we had [something like this](https://github.com/hibernate/hibernate-search/blob/d723b9332912c2d896a3f1e36f2d51e3adb656ce/integrationtest/wildfly/pom.xml#L471-L475); not sure what's relevant here.

yrodiere:
BTW FWIW Galleon apparently supports [patches](https://docs.wildfly.org/galleon/#_patches) and [feature pack overrides](https://docs.wildfly.org/galleon/#_dependency_version_convergence)

marko-bekhta:
I've split the unpacking: one execution for the groovy script and the other for Wildfly, so that we can skip the one for the Wildlfly if a container is provided. I also double-checked that running a build:

```
mvn clean package -Dincontainer -Dincontainer-provided -Dwildfly.target-dir=/home/marko/development/tools/wildfly-32.0.1.Final
```
works ok, meaning the Wildfly instance provided in the property gets patched. 

Now as to the Galleon, Glow, patching and all that 😃. I've tried creating the feature pack, but while doing so, I hit some confusing exceptions, which turned out to be solved by adding an additional Maven repo 🙈. 
Maybe it was just my experience 😄 but because of the way it went, it convinced me even more that having these couple of groovy scripts (that were already written) to do all the patching we need is much simpler and easier to maintain 🙈. + From what it seems, we've dropped all this "Wildfly" integration/modules from other Hibernate projects.

If you think that we'd be better with the feature-packed-glowing-galleon ⛵✨ 🙈 😃 I can give it another try, I should still have the feature-pack patch somewhere locally...

documentation/src/main/asciidoc/ch01.asciidoc
```diff
@@ -86,69 +86,6 @@ Note that adding this dependency is usually not required for applications runnin
 application server. You can learn more about the integration of Jakarta Bean Validation and CDI in
 <<section-integration-with-cdi>>.
 
-==== Updating Hibernate Validator in WildFly
```
yrodiere, line 89: Maybe keep the section, but mention it's no longer possible as of Hibernate Validator 9.0?

By the way, is there a migration guide you should be updating?

yrodiere (approved):
> works ok, meaning the Wildfly instance provided in the property gets patched.

I think the point is just to run integration tests against another WildFly instance, not to patch it :thinking: We'd have to reach out to product teams to know for sure though.

> Maybe it was just my experience 😄 but because of the way it went, it convinced me even more that having these couple of groovy scripts (that were already written) to do all the patching we need is much simpler and easier to maintain 🙈

Fair enough.

> + From what it seems, we've dropped all this "Wildfly" integration/modules from other Hibernate projects.

Note other Hibernate projects had a very limited set of tests, and in the case of Hibernate Search we moved them directly to WildFly.
For Hibernate Validator, we're running (almost?) the entire test suite, so the value provided by this test run is much higher.

> If you think that we'd be better with the feature-packed-glowing-galleon ⛵✨ 🙈 😃 I can give it another try, I should still have the feature-pack patch somewhere locally...

Let's see how it goes with what you did, we can always reconsider later.

documentation/src/main/asciidoc/ch01.asciidoc
marko-bekhta: Hmm, ok 😃 will do.
I found a migration guide here https://github.com/hibernate/hibernate.org/blob/production/validator/documentation/migration-guide.adoc
I guess ... I'll open a draft PR, and we can keep adding things as we go and merge it on the release day.

marko-bekhta:
> I think the point is just to run integration tests against another WildFly instance, not to patch it 🤔 We'd have to reach out to product teams to know for sure though.

After looking at the profiles before the change, if I read it correctly, there's `incontainer-prepared` that will "prepare" the Wildfly instance (patch it) and then `incontainer-provided` that'll use whatever instance is provided. That's for the TCK run. As for the integration tests, those always unpack and patch the Wildfly instance the way the tests need.

I'll double-check that this profile logic is preserved and update a PR if needed. 😃

marko-bekhta:
I've tried running the builds from the main and this branch and seem to get the same results (in terms of updates to WF), so I'll merge this one. Thanks, Yoann, for taking a look!

PR #1362, opened 2024-06-20
HV-1989 Update documentation examples to use the min required Java version where applicable

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-1989
https://hibernate.atlassian.net/browse/HV-2009

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@
 
 ## What is it?
 
-This is the reference implementation of [Jakarta Bean Validation 3.0](http://beanvalidation.org/).
-Jakarta Bean Validation defines a metadata model and API for JavaBean as well as method validation.
+This is the reference implementation of [Jakarta Validation 3.1](https://jakarta.ee/specifications/bean-validation/3.1/).
+Jakarta Validation defines a metadata model and API for JavaBean as well as method validation.
 The default metadata source are annotations, with the ability to override and extend
 the metadata through the use of XML validation descriptors.
 
@@ -20,7 +20,7 @@ The full list of changes for this release can be found in changelog.txt.
 
 ## System Requirements
 
-JDK 8 or above.
+JDK 17 or above.
 
 ## Using Hibernate Validator
 
@@ -47,7 +47,7 @@ Logging will delegate any log requests to that provider.
            <version>5.0.0</version>
         </dependency>
 
-* Jakarta Bean Validation defines integration points with [CDI](http://jcp.org/en/jsr/detail?id=346). If your application runs
+* Jakarta Validation defines integration points with [CDI](http://jcp.org/en/jsr/detail?id=346). If your application runs
 in an environment which does not provide this integration out of the box, you may use the Hibernate Validator CDI portable
 extension by adding the following dependency:
 
@@ -63,13 +63,13 @@ documentation](https://docs.jboss.org/hibernate/stable/validator/reference/en-US
 
 ## Licensing
 
-Hibernate Validator itself as well as the Jakarta Bean Validation API and TCK are all provided and distributed under
+Hibernate Validator itself as well as the Jakarta Validation API and TCK are all provided and distributed under
 the Apache Software License 2.0. Refer to license.txt for more information.
 
 ## Build from Source
 
 You can build Hibernate Validator from source by cloning the git repository `git://github.com/hibernate/hibernate-validator.git`.
-You will also need a JDK 17+ and Maven 3 (>= 3.3.1). With these prerequisites in place you can compile the source via:
+You will also need a JDK 17+ and Maven 3 (>= 3.9.6). With these prerequisites in place you can compile the source via:
 
     mvn clean install
 
@@ -84,7 +84,8 @@ We provide a `.travis.yml` file so that you can enable CI for your GitHub fork b
 ## Hibernate Validator URLs
 
 * [Home Page](http://hibernate.org/validator/)
-* [Jakarta Bean Validation Home](http://beanvalidation.org/)
+* [Jakarta Validation Home](http://beanvalidation.org/)
+* [Jakarta Validation Specs](https://jakarta.ee/specifications/bean-validation/)
 * [Downloads](http://hibernate.org/validator/downloads/)
 * [Mailing Lists](http://hibernate.org/community/)
 * [Issue Tracking](https://hibernate.atlassian.net/browse/HV)
--- a/cdi/pom.xml
+++ b/cdi/pom.xml
@@ -173,7 +173,7 @@
                             <archive>
                                 <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
                                 <manifestEntries>
-                                    <Specification-Title>Jakarta Bean Validation</Specification-Title>
+                                    <Specification-Title>Jakarta Validation</Specification-Title>
                                     <Specification-Version>2.0</Specification-Version>
                                     <Automatic-Module-Name>${hibernate-validator-cdi.module-name}</Automatic-Module-Name>
                                 </manifestEntries>
--- a/cdi/pom.xml
+++ b/cdi/pom.xml
@@ -174,7 +174,7 @@
                                 <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
                                 <manifestEntries>
                                     <Specification-Title>Jakarta Validation</Specification-Title>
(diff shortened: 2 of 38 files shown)


documentation/src/main/asciidoc/ch01.asciidoc
```diff
@@ -94,9 +94,9 @@ Starting with Hibernate Validator 9.0, the patch to update WildFly Jakarta Valid
 You can learn more about the WildFly patching infrastructure in general link:https://github.com/wildfly/wildfly-feature-pack-template[here].
 
 [[validator-gettingstarted-java9]]
-==== Running on Java 11+
+==== Running on Java {jdkVersion}+
```
yrodiere, line 97: Seems weird that the title mentions a Java version, but the content is about JPMS.

Should probably be "Running in the modulepath" or "Running as a Java module"?

marko-bekhta: +1 done 👍🏻 😃 

BTW, while we are at it (updating the docs)... should we also update `Jakarta Bean Validation` to `Jakarta Validation` to match what was done in the spec?

yrodiere: > BTW, while we are at it (updating the docs)... should we also update `Jakarta Bean Validation` to `Jakarta Validation` to match what was done in the spec?

I think we should :/

marko-bekhta: let me know if you'd want to take another look at this one 🙈 😏

cdi/pom.xml
```diff
@@ -170,7 +170,7 @@
                             <archive>
                                 <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
                                 <manifestEntries>
-                                    <Specification-Title>Jakarta Bean Validation</Specification-Title>
+                                    <Specification-Title>Jakarta Validation</Specification-Title>
                                     <Specification-Version>2.0</Specification-Version>
```
yrodiere, line 174: This is out of date, isn't it?

README.md
```diff
@@ -20,7 +20,7 @@ The full list of changes for this release can be found in changelog.txt.
 
 ## System Requirements
 
-JDK 8 or above.
+JDK 17 or above.
```
yrodiere, line 23: PR description says Java 11... ?

documentation/src/main/asciidoc/ch02.asciidoc
```diff
@@ -95,7 +95,7 @@ It is possible to specify constraints directly on the type argument of a
 parameterized type: these constraints are called container element constraints.
 
 This requires that `ElementType.TYPE_USE` is specified via `@Target`
-in the constraint definition. As of Jakarta Bean Validation 2.0, built-in Jakarta Bean Validation as well as
+in the constraint definition. As of Jakarta Bean Validation 2.0, built-in Jakarta Validation as well as
```
yrodiere, line 98:
```suggestion
in the constraint definition. As of Jakarta Validation 2.0, built-in Jakarta Validation as well as
```

And is the version correct?

engine/pom.xml
```diff
@@ -227,7 +227,7 @@
                             <archive>
                                 <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
                                 <manifestEntries>
-                                    <Specification-Title>Jakarta Bean Validation</Specification-Title>
+                                    <Specification-Title>Jakarta Validation</Specification-Title>
                                     <Specification-Version>3.0</Specification-Version>
```
yrodiere, line 231: Ah, so it's 3 after all?

engine/src/main/java/org/hibernate/validator/overview.html
```diff
@@ -10,7 +10,7 @@
 </head>
 <body>
 <p>This is the Hibernate Validator API documentation. Hibernate Validator is the reference implementation of
- https://projects.eclipse.org/projects/ee4j.bean-validation[Jakarta Bean Validation 2.0].
+    https://projects.eclipse.org/projects/ee4j.validation[Jakarta Validation 3.1].
```
yrodiere, line 13: ... so it's 3.1 after all?

documentation/src/main/asciidoc/ch01.asciidoc
yrodiere: I didn't expect the "Jakarta Bean Validation" => "Jakarta Validation" change to happen in this PR, but okay :) Maybe a different issue though?

Also... There seem to be inconsistencies in the version of Jakarta Validation, as well as in the Java version (why talk about Java 11, don't we require 17?).

documentation/src/main/asciidoc/ch02.asciidoc
marko-bekhta: yeah I was a bit puzzled about how to change this 😄 
I mean the `ElementType.TYPE_USE` is available since Java Bean Validation 2.0 (the spec that was not yet Jakarta). But since we are talking about `Jakarta Bean Validation 2.0` here .. which was a rebrand-to-jakarta-only thing... should we change the names of "older" specs or is the shorter name only applicable to 3.1 and forward? 😖😖😖 🤔 😃

engine/src/main/java/org/hibernate/validator/overview.html
marko-bekhta: yes ! 🙈 🙈 🙈 🙈 3.1
I should probably go through it once more to see where the versions are incorrect, as I mostly was looking at the name only in files that are not docs.

documentation/src/main/asciidoc/ch02.asciidoc
yrodiere: No you're right, I didn't pay enough attention. Please ignore :)

engine/pom.xml
marko-bekhta: I've used the `build-helper-maven-plugin` to parse the jakarta validation version and use that where possible, so we wouldn't need to track where we must change the version 🙈

engine/pom.xml
```diff
@@ -260,7 +260,7 @@
                         <Bundle-SymbolicName>${hibernate-validator.module-name}</Bundle-SymbolicName>
                         <Import-Package>
                             jakarta.persistence.*;version="[3.0.0,4.0.0)";resolution:=optional,
-                            jakarta.validation.*;version="[3.0.0,4.0.0)",
+                            jakarta.validation.*;version="[${version.jakarta.validation-api},4.0.0)",
```
yrodiere, line 263: If you parse versions... may as well take advantage of `nextMajorVersion`. See https://www.mojohaus.org/build-helper-maven-plugin/parse-version-mojo.html

```suggestion
                            jakarta.validation.*;version="[${version.jakarta.validation-api},${parsed-version.jakarta.validation-api.nextMajorVersion}.0.0)",
```

performance/pom.xml
```diff
@@ -22,6 +22,7 @@
     <properties>
         <!-- Disable enforcer as we are using banned dependencies for old versions -->
         <enforcer.skip>true</enforcer.skip>
+        <add.bv2.sources.skip>true</add.bv2.sources.skip>
```
yrodiere, line 25: Maybe a comment would help here... ? Not sure what's the purpose of this.

README.md
yrodiere: Bump :) PR description is okay now, but not the backport, nor the Jira issue :see_no_evil:

yrodiere (approved):

engine/pom.xml
marko-bekhta: oh nice 👍🏻

performance/pom.xml
marko-bekhta: Yeah will add something. 
The thing is that the helper plugin was used to add the sources, and since I've used the same plugin to parse the version in the root pom, now adding the sources happens all the time 🙈 unless we skip it 

And thinking more about it ... we probably should just clean up the perf module. See all those other profiles, they are for `javax` packages so none of them really work as perf test in the main/java are for `jakarta`...
sooooooooooo we probably should have to "optional" source directories one for javax one for jakarta with the same tests in them and then include the dir depending on the spec we are targeting ....

marko-bekhta: I've created a https://hibernate.atlassian.net/browse/HV-2010 so that I stop adding more and more to this PR 🙈 
and a https://hibernate.atlassian.net/browse/HV-2009 for jakarta rename (part of this PR)

PR #1365, opened 2024-06-21
HV-1999 new CNPJ alphanumeric format

detinho:
https://hibernate.atlassian.net/browse/HV-1999

Added an option to allow for gradual adoption.

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraints/Mod11Check.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/Mod11Check.java
@@ -122,6 +122,12 @@ public @interface Mod11Check {
 	ProcessingDirection processingDirection() default ProcessingDirection.RIGHT_TO_LEFT;
 
 	/**
+	 * @return Whether upper case letters characters in the validated input should be allowed ({@code true}) or result in a
+	 * validation error ({@code false}).
+	 */
+	boolean alphanumeric() default false;
+
+	/**
 	 * Defines several {@code @Mod11Check} annotations on the same element.
 	 */
 	@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
--- a/engine/src/main/java/org/hibernate/validator/constraints/br/CNPJ.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/br/CNPJ.java
@@ -31,7 +31,7 @@ import org.hibernate.validator.constraints.br.CNPJ.List;
  *
  * @author George Gastaldi
  */
-@Pattern(regexp = "([0-9]{2}[.]?[0-9]{3}[.]?[0-9]{3}[/]?[0-9]{4}[-]?[0-9]{2})")
+@Pattern(regexp = "([0-9A-Z]{2}[.]?[0-9A-Z]{3}[.]?[0-9A-Z]{3}[/]?[0-9A-Z]{4}[-]?[0-9]{2})")
 @ReportAsSingleViolation
 @Documented
 @Constraint(validatedBy = { })
@@ -44,6 +44,8 @@ public @interface CNPJ {
 	Class<?>[] groups() default { };
 
 	Class<? extends Payload>[] payload() default { };
+
+	boolean alphanumeric() default false;
 
 	/**
 	 * Defines several {@code @CNPJ} annotations on the same element.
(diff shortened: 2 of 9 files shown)


engine/src/main/java/org/hibernate/validator/constraints/br/CNPJ.java
```diff
@@ -31,7 +31,7 @@
  *
  * @author George Gastaldi
  */
-@Pattern(regexp = "([0-9]{2}[.]?[0-9]{3}[.]?[0-9]{3}[/]?[0-9]{4}[-]?[0-9]{2})")
+@Pattern(regexp = "([0-9A-Z]{2}[.]?[0-9A-Z]{3}[.]?[0-9A-Z]{3}[/]?[0-9A-Z]{4}[-]?[0-9]{2})")
```
marko-bekhta, line 34: Since this pattern now depends on the CNPJ's version/format, I'd suggest moving it into the CNPJ validator itself and picking the pattern that suits the format.
Doing so would also mean that `@ReportAsSingleViolation`  can be removed.

engine/src/main/java/org/hibernate/validator/constraints/Mod11Check.java
```diff
@@ -121,6 +121,12 @@
 	 */
 	ProcessingDirection processingDirection() default ProcessingDirection.RIGHT_TO_LEFT;
 
+	/**
+	 * @return Whether upper case letters characters in the validated input should be allowed ({@code true}) or result in a
+	 * validation error ({@code false}).
+	 */
+	boolean alphanumeric() default false;
```
marko-bekhta, line 128: I'd suggest limit the changes to CNPJ constraint only

engine/src/main/java/org/hibernate/validator/constraints/br/CNPJ.java
```diff
@@ -45,6 +45,8 @@
 
 	Class<? extends Payload>[] payload() default { };
 
+	boolean alphanumeric() default false;
```
marko-bekhta, line 48: Wouldn't an enum here be more future-proof option? something like 
```suggestion
	FormatRevision formatRevision() default FormatRevision.V1;
```
```java
enum FormatRevision {
    // Obviously, the constant names have to be more meaningful 
    //  and connected to the versions of current and new formats.
    V1, V2; 
}
```
This way, if another version is published, it can be added without any new parameters on the constraint annotation...

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ModCheckBase.java
```diff
@@ -79,11 +82,22 @@ public boolean isValid(final CharSequence value, final ConstraintValidatorContex
...
+	protected void initialize(int startIndex, int endIndex, int checkDigitIndex, boolean ignoreNonValidCharacters) {
 		this.startIndex = startIndex;
 		this.endIndex = endIndex;
 		this.checkDigitIndex = checkDigitIndex;
-		this.ignoreNonDigitCharacters = ignoreNonDigitCharacters;
+		this.ignoreNonValidCharacters = ignoreNonValidCharacters;
+		this.alphanumeric = false;
+
+		this.validateOptions();
+	}
```
marko-bekhta, line 93: maybe just call 
```java
initialize(startIndex, endIndex, checkDigitIndex, ignoreNonValidCharacters, false)
```

and also if you'll end up still needing to modify the `ModCheckBase` consider using this overloaded version of initialize in places where `alphanumeric` is not needed so that less classes are modified.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ModCheckBase.java
```diff
@@ -98,11 +112,21 @@ protected void initialize(int startIndex, int endIndex, int checkDigitIndex, boo
 	 * @throws NumberFormatException in case character is not a digit
 	 */
 	protected int extractDigit(char value) throws NumberFormatException {
-		if ( Character.isDigit( value ) ) {
-			return Character.digit( value, DEC_RADIX );
+		if ( alphanumeric ) {
+			if ( (value >= '0' && value <= '9' ) || (value >= 'A' && value <= 'Z') ) {
+				 return value - BASE_CHAR_INDEX;
```
marko-bekhta, line 117: This way of converting seems somewhat specific to CNPJ, I'd suggest creating an extended version of `Mod11CheckValidator` in `CNPJValidator` and overriding `extractDigit` in it.

marko-bekhta (commented):
Hey 👋🏻 
Thanks for your pull request. I've added some comments inline.

marko-bekhta:
Thanks for the work and for bringing up the fact that there's an upcoming change in the format. I've addressed the review suggestions and aligned things more closely to how HV does things on top of your initial patch in this PR https://github.com/hibernate/hibernate-validator/pull/1386 .

I am closing this one. Thanks again!

PR #1370, opened 2024-07-02
HV-2001 Generate module-info for published artifacts

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-2001

This is the first iteration to generate module-info for published artifacts. It is based on https://github.com/hibernate/hibernate-validator/pull/1364 as it has some useful cleanups that make module-info generation cleaner (hence, opening only as a draft for now).

Only the last commit in this PR is really relevant to module-info.

I added specific plugin configurations to each module as it seems that each of them has specific needs and there's not much in common...

## test utils module

This one seems pretty straightforward. I've dropped the dependency on testng, as it was there just for a few assert calls that we can replace with the assertj ones. The constraint assert that this util jar offers is based on assertj, so it makes sense to keep assertj 😃.

There is a simple test  (`test-utils`).

## engine (actual validator)

The plugin couldn't pick up which services are loaded via `ServiceLoader` so I've listed those explicitly.
all seems to work fine and look ok until we get to CDI stuff... CDI needs reflection access to `ValidatorImpl`, `ValidatorFactoryImpl` and all constraint validator implementations... so that means we have to open the packages that contain those classes .... but open to CDI impl, which is ... ? so I just have them opened to all (at least for now).

Going back to the CDI stuff... I wonder whether we should move the classes around, maybe to some SPI packages for the things that are potentially instantiated through the CDI?

There are tests for the case with EL (`simple`) and without EL dependency (`no-el`).

## CDI

see the comments to the engine section just above. Another problem is that Weld is using the previous version of JBoss Logging. This leads to the test failure since Weld cannot access the logger correctly in the module test.

I've moved some classes around as CDI needed reflection access to some of the classes in this module and they lived in an internal package. 

The test is currently disabled (`cdi`), but I've tested it locally using the previous version of the logging jar.

## Annotaion processor

This one seems pretty straightforward as it doesn't really have any dependencies, and the AP is added as provided to the module info by the plugin automatically. 

With that, I'm not sure what would be a good way to test it, just run a compile task and pass the AP through the `annotationProcessorPaths` ?

----------------------

----------------------


cdi/pom.xml
```diff
@@ -225,6 +234,35 @@
...
+                        </goals>
+                        <configuration>
+                            <module>
+                                <moduleInfo>
+                                    <name>${java.module.name}</name>
+                                    <exports>
+                                        !org.hibernate.validator.*.internal;
+                                        !org.hibernate.validator.*.internal.*;
+                                        *;
+                                    </exports>
```
yrodiere, line 255: Can't you have this part in common (parent POM) while overriding only `<opens>` in modules where it makes sense?

annotation-processor/pom.xml
```diff
@@ -142,6 +141,34 @@
...
+                                    <exports>
+                                        !org.hibernate.validator.ap.internal;
+                                        !org.hibernate.validator.ap.internal.*;
+                                        *;
+                                    </exports>
+                                </moduleInfo>
+                            </module>
+                            <jdepsExtraArgs>
+                                --multi-release=${maven.compiler.release}
+                            </jdepsExtraArgs>
```
yrodiere, line 167: Is this absolutely necessary?

If so, please maybe add a comment explaining why, like in Hibernate Search?

```

                                <!--
                                    Running the jdeps tool will fail without this option, if there are any multi-release jars
                                    on the path. And since some modules might have such jars, e.g. util-common has Jandex,
                                    it's safer to add this additional option:
                                -->
```

cdi/src/main/java/org/hibernate/validator/cdi/ValidationExtension.java
```diff
@@ -51,8 +51,8 @@
 import org.hibernate.validator.cdi.internal.ValidationProviderHelper;
 import org.hibernate.validator.cdi.internal.ValidatorBean;
 import org.hibernate.validator.cdi.internal.ValidatorFactoryBean;
-import org.hibernate.validator.cdi.internal.interceptor.ValidationEnabledAnnotatedType;
-import org.hibernate.validator.cdi.internal.interceptor.ValidationInterceptor;
+import org.hibernate.validator.cdi.interceptor.spi.internal.ValidationEnabledAnnotatedType;
```
yrodiere, line 54: `.spi.internal`, really? :thinking: 

Are you trying to confuse automatic tooling? :)

cdi/src/main/java/org/hibernate/validator/cdi/interceptor/spi/ValidationInterceptor.java
```diff
@@ -4,14 +4,16 @@
  * License: Apache License, Version 2.0
  * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
  */
-package org.hibernate.validator.cdi.internal.interceptor;
+package org.hibernate.validator.cdi.interceptor.spi;
```
yrodiere, line 7: Any reason this cannot remain in an internal package *and* have access to reflection?

annotation-processor/pom.xml
```diff
@@ -23,7 +23,7 @@
         <!-- This is a publicly distributed module that should be published: -->
         <deploy.skip>false</deploy.skip>
         <hibernate-validator-parent.path>..</hibernate-validator-parent.path>
-        <automatic.module.name>org.hibernate.validator.annotationprocessor</automatic.module.name>
+        <java.module.name>org.hibernate.validator.annotationprocessor</java.module.name>
```
yrodiere, line 26: I'm really not sure it makes sense to put an annotation processor in the module path but... :shrug:

engine/pom.xml
```diff
@@ -322,6 +322,51 @@
...
+                                        <!-- jboss annotations are getting pulled even though they are for compilation only -->
+                                        !jboss.logging.annotations;
+                                        *;
+                                    </requires>
+                                    <exports>
+                                        org.hibernate.validator.*.logging to org.jboss.logging, org.hibernate.validator.cdi;
+                                        org.hibernate.validator.internal to org.hibernate.validator.cdi;
+                                        org.hibernate.validator.internal.* to org.hibernate.validator.cdi;
+                                        org.hibernate.validator.*.internal.* to org.hibernate.validator.cdi;
+                                        *;
```
yrodiere, line 349: Might as well export `org.hibernate.validator.*`, no? It's not like we don't trust our own modules...

engine/pom.xml
```diff
@@ -322,6 +322,51 @@
...
+                                    </exports>
+                                    <opens>
+                                        <!-- We open constraints for reflection as these can be instantiated by a CDI container -->
+                                        org.hibernate.validator.internal.constraintvalidators.bv;
+                                        org.hibernate.validator.internal.constraintvalidators.bv.*;
+                                        org.hibernate.validator.internal.constraintvalidators.hv;
+                                        org.hibernate.validator.internal.constraintvalidators.hv.*;
+                                        <!-- As both Validator and ValidatorFactory have to be accessible by CDI ... -->
+                                        org.hibernate.validator.internal.engine;
```
yrodiere, line 358: You mean Weld accesses the Validator/ValidatorFactory impl? Why... ? How?

java/cdi/src/main/java/module-info.java
```diff
@@ -0,0 +1,23 @@
+/*
```
yrodiere, line 1: The directory structure you picked is *very* confusing. I expect people seeing a `java` directory will assume all java code is in there, but that's not true here.

I'd suggest `integrationtest/java/modules` instead, but that might require moving around some other integration tests.

yrodiere (commented):
Hey, thanks. My comments below.

> Going back to the CDI stuff... I wonder whether we should move the classes around, maybe to some SPI packages for the things that are potentially instantiated through the CDI?

Maybe, though I'd love to know what those things are and how the CDI container manages to guess the names of the relevant classes. Are those hardcoded somewhere, or... ?

annotation-processor/pom.xml
marko-bekhta: ok I've double-checked it and this was "forced" by the fact AP being multi-release jar... but we don't need that anymore. I've opened https://github.com/hibernate/hibernate-validator/pull/1380 for that

cdi/src/main/java/org/hibernate/validator/cdi/interceptor/spi/ValidationInterceptor.java
marko-bekhta: well ... I mean ... 🙈 my thinking was that we don't really want to open or export something with `.internal` .. that's  until I've seen the problem with `ValidatorImpl` that it has to be opened 😖. Couldn't apply such logic there 😄

annotation-processor/pom.xml
marko-bekhta: I'm even less sure than that 😄. But as it had an automatic module name...

engine/pom.xml
marko-bekhta: I think if we export `org.hibernate.validator.*` to `org.hibernate.validator.cdi` then we'll be exporting only to that module and all the public packages won't be exported to any other module (even public packages)

engine/pom.xml
marko-bekhta:
```
Caused by: org.jboss.weld.exceptions.WeldException: WELD-001524: Unable to load proxy class for bean ValidatorFactoryBean [id=org.hibernate.validator.cdi.internal.ValidatorFactoryBean_default] with class class org.hibernate.validator.internal.engine.ValidatorFactoryImpl
        at org.jboss.weld.bean.proxy.ProxyFactory.getProxyClass(ProxyFactory.java:409)
        at org.jboss.weld.bean.proxy.ProxyFactory.run(ProxyFactory.java:360)
        at org.jboss.weld.bean.proxy.ProxyFactory.create(ProxyFactory.java:352)
        at org.jboss.weld.bean.proxy.ClientProxyFactory.create(ClientProxyFactory.java:83)
        at org.jboss.weld.bean.proxy.ClientProxyProvider.createClientProxy(ClientProxyProvider.java:205)
        at org.jboss.weld.bean.proxy.ClientProxyProvider.createClientProxy(ClientProxyProvider.java:195)
        at org.jboss.weld.bean.proxy.ClientProxyProvider$CreateClientProxy.apply(ClientProxyProvider.java:52)
        at org.jboss.weld.bean.proxy.ClientProxyProvider$CreateClientProxy.apply(ClientProxyProvider.java:48)
        at org.jboss.weld.util.cache.ReentrantMapBackedComputingCache.lambda$new$0(ReentrantMapBackedComputingCache.java:55)
        at org.jboss.weld.util.LazyValueHolder$1.computeValue(LazyValueHolder.java:32)
        at org.jboss.weld.util.LazyValueHolder.get(LazyValueHolder.java:46)
        at org.jboss.weld.util.cache.ReentrantMapBackedComputingCache.getValue(ReentrantMapBackedComputingCache.java:72)
        at org.jboss.weld.util.cache.ReentrantMapBackedComputingCache.getCastValue(ReentrantMapBackedComputingCache.java:78)
        at org.jboss.weld.bean.proxy.ClientProxyProvider.getClientProxy(ClientProxyProvider.java:229)
        at org.jboss.weld.manager.BeanManagerImpl.getReference(BeanManagerImpl.java:678)
        at org.jboss.weld.manager.BeanManagerImpl.getReference(BeanManagerImpl.java:666)
        at org.jboss.weld.manager.BeanManagerImpl.getInjectableReference(BeanManagerImpl.java:784)
        at org.jboss.weld.injection.FieldInjectionPoint.inject(FieldInjectionPoint.java:92)
        at org.jboss.weld.util.Beans.injectBoundFields(Beans.java:345)
        at org.jboss.weld.util.Beans.injectFieldsAndInitializers(Beans.java:356)
        at org.jboss.weld.injection.producer.ResourceInjector$1.proceed(ResourceInjector.java:69)
        at org.jboss.weld.injection.InjectionContextImpl.run(InjectionContextImpl.java:48)
        at org.jboss.weld.injection.producer.ResourceInjector.inject(ResourceInjector.java:71)
        at org.jboss.weld.injection.producer.BasicInjectionTarget.inject(BasicInjectionTarget.java:117)
        at org.jboss.arquillian.testenricher.cdi.CDIInjectionEnricher.injectNonContextualInstance(CDIInjectionEnricher.java:127)
        at org.jboss.arquillian.testenricher.cdi.CDIInjectionEnricher.injectClass(CDIInjectionEnricher.java:110)
        ... 68 more
Caused by: java.lang.RuntimeException: java.lang.IllegalAccessException: module org.hibernate.validator does not open org.hibernate.validator.internal.engine to unnamed module @6b81ce95
        at org.jboss.weld.bean.proxy.util.WeldDefaultProxyServices.defineWithMethodLookup(WeldDefaultProxyServices.java:151)
        at org.jboss.weld.bean.proxy.util.WeldDefaultProxyServices.defineClass(WeldDefaultProxyServices.java:60)
        at org.jboss.weld.bean.proxy.ProxyFactory.toClass(ProxyFactory.java:940)
        at org.jboss.weld.bean.proxy.ProxyFactory.createProxyClass(ProxyFactory.java:495)
        at org.jboss.weld.bean.proxy.ProxyFactory.getProxyClass(ProxyFactory.java:401)
        ... 93 more
Caused by: java.lang.IllegalAccessException: module org.hibernate.validator does not open org.hibernate.validator.internal.engine to unnamed module @6b81ce95
        at java.base/java.lang.invoke.MethodHandles.privateLookupIn(MethodHandles.java:287)
        at org.jboss.weld.bean.proxy.util.WeldDefaultProxyServices.defineWithMethodLookup(WeldDefaultProxyServices.java:148)
        ... 97 more
```

java/cdi/src/main/java/module-info.java
marko-bekhta: I've done that and moved the WF tests to the same test directory. I wasn't sure about the TCK tests as that's a strange beast 🧟 😃  so I kept those in the root

engine/pom.xml
yrodiere: It seems to me that the only reason CDI would need to access `ValidatorFactoryImpl` would be to instantiate it. But we instantiate the class ourselves in `org.hibernate.validator.cdi.internal.ValidatorFactoryBean#create`, so I think it's only being instantiated is for CDI proxies.

I don't know Weld very well, but in Arc (Quarkus's CDI impl) the CDI proxy would typically extend `ValidatorFactory`, not `ValidatorFactoryImpl`.

I see that here, we essentially expose the `ValidatorFactoryImpl` type in CDI:

https://github.com/hibernate/hibernate-validator/blob/f5715b41eee327a4442c072a2f513ac438314971/cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidatorFactoryBean.java#L70-L74

IMO we shouldn't. The bean types should only include public types, so `org.hibernate.validator.HibernateValidatorFactory` and superinterfaces. Maybe `PredefinedScopeHibernateValidatorFactory` and superinterfaces, when relevant. But definitely not `ValidatorFactoryImpl`.

Maybe if you the exposed types of the CDI bean, the problem will be different?

For constraint validators, however, I'm not sure it's the same problem... I find it weird that Weld tries to enable constructor injection on these, though; we don't rely on CDI anywhere in our constraint validators, do we?

marko-bekhta: > Maybe if you the exposed types of the CDI bean, the problem will be different?

Thanks for the suggestion, I'll give that a try! 😃 

> For constraint validators, however, I'm not sure it's the same problem... I find it weird that Weld tries to enable constructor injection on these, though; we don't rely on CDI anywhere in our constraint validators, do we?

hmm, right so we do not rely on CDI in any built-in constraints. There is this option for custom constraints if someone decides to do that (https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#example-constraint-validator-injected-bean) sooooooo maybe we update the factory here:

https://github.com/hibernate/hibernate-validator/blob/125feaa30e91fd149d29111bec5c3179b9f6feaf/cdi/src/main/java/org/hibernate/validator/cdi/internal/InjectingConstraintValidatorFactory.java#L40-L44

And if it is a built-in constraint, we return an instance and do not go through CDI ?

yrodiere: > And if it is a built-in constraint, we return an instance and do not go through CDI ?

That looks like a plan. Now how you'll detect it's a built-in constraint is going to be interesting... Please remember some demos use the package name `org.hibernate.validator.demo` or something like that... Ideally you'd re-use some pre-existing code that's designed to detect built-in classes, if that exists.

cdi/src/main/java/org/hibernate/validator/cdi/spi/InjectingConstraintValidatorFactory.java
```diff
@@ -39,6 +45,14 @@ public InjectingConstraintValidatorFactory(BeanManager beanManager) {
...
 	@Override
 	public <T extends ConstraintValidator<?, ?>> T getInstance(Class<T> key) {
+		// we can do a simple check like :
+		//		if ( key.getPackageName().startsWith( "org.hibernate.validator.internal.constraintvalidators.bv" )
+		//				|| key.getPackageName().startsWith( "org.hibernate.validator.internal.constraintvalidators.hv" ) ) {
+		//			return super.getInstance( key );
+		//		}
+		if ( builtInConstraintValidators.contains( key ) ) {
+			return super.getInstance( key );
+		}
```
marko-bekhta, line 55: There wasn't exactly a way to get the built-in validators as is soooo that's what I've done here.

- add a util in main module (engine) to provide a list of built-in constraint validator classes based on the ConstraintHelper that handles all constraint-validator related things there.
- passed the set of these classes to the factory and add a check to bypass CDI if a built-in constraint validator is requested. 

I've left that commented-out code as that's what I've started with to see if the approach of bypassing CDI will work and if there won't be any other unexpected errors. But then also, I wonder if we keep the check for those packages, do we even need the set? I mean, what are the odds of someone using that package? 😃

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidatorFactoryBean.java
```diff
@@ -252,7 +255,18 @@ private ConstraintValidatorFactory createConstraintValidatorFactory(Configuratio
 
 		if ( constraintValidatorFactoryFqcn == null ) {
 			// use default
-			return createInstance( InjectingConstraintValidatorFactory.class );
+			InjectingConstraintValidatorFactory validatorFactory = createInstance( InjectingConstraintValidatorFactory.class );
+
+			Set<Class<? extends ConstraintValidator<?, ?>>> builtInConstraintValidators;
+			if ( config instanceof PredefinedScopeConfigurationImpl ) {
```
marko-bekhta, line 266: BTW, the predefined validator factory and related classes are for Quarkus, and it does not use this CDI extension so IDK if we should even consider this case here ...

cdi/src/main/java/org/hibernate/validator/cdi/spi/InjectingConstraintValidatorFactory.java
yrodiere: > add a util in main module (engine) to provide a list of built-in constraint validator classes based on the ConstraintHelper that handles all constraint-validator related things there.

Makes sense, but:

1. Are you positive classes of built-in constraint validators were always being loaded before?
2. Maybe you can add some test OR a JQAssistant rule to be sure the util is up to date?


> what are the odds of someone using that package? 😃

In correct code? 0 chance.

In real code? Well... :)

At the very least, let's say it's possible in demos. We may not want to care, though.

yrodiere: > > add a util in main module (engine) to provide a list of built-in constraint validator classes based on the ConstraintHelper that handles all constraint-validator related things there.
> 
> Makes sense, but:
> 
>     1. Are you positive classes of built-in constraint validators were always being loaded before?
> 
>     2. Maybe you can add some test OR a JQAssistant rule to be sure the util is up to date?

Ah, you're computing the list at runtime... Then I guess item 2 is irrelevant.

cdi/src/main/java/org/hibernate/validator/cdi/spi/InjectingConstraintValidatorFactory.java
```diff
@@ -39,6 +45,14 @@ public InjectingConstraintValidatorFactory(BeanManager beanManager) {
 
 	@Override
 	public <T extends ConstraintValidator<?, ?>> T getInstance(Class<T> key) {
+		// we can do a simple check like :
+		//		if ( key.getPackageName().startsWith( "org.hibernate.validator.internal.constraintvalidators.bv" )
+		//				|| key.getPackageName().startsWith( "org.hibernate.validator.internal.constraintvalidators.hv" ) ) {
+		//			return super.getInstance( key );
+		//		}
```
yrodiere, line 52: Yeah so all things considered, this looks like a better solution. I'd use a simple `startsWith("org.hibernate.validator.internal.constraintvalidators.")` though.
And you'll need some test to check that all built-in constraint validators indeed are in that package...

cdi/src/main/java/org/hibernate/validator/cdi/spi/InjectingConstraintValidatorFactory.java
marko-bekhta: yeah.. the downside I can think of, for this approach is that we will be creating the ConstraintHelper 2 times, once for CDI and once for the ValidatorFactory itself ...

marko-bekhta:
@yrodiere can you think of anything else we can/should do here? Or should we try merging this and then formatting/sorting in and attempting to do a release after that? 🙈 😃

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidatorBean.java
```diff
@@ -44,6 +46,11 @@ public ValidatorBean(BeanManager beanManager, Bean<?> validatorFactoryBean,
 		this.types = Collections.unmodifiableSet(
 				CollectionHelper.<Type>newHashSet(
 						ClassHierarchyHelper.getHierarchy( validationProviderHelper.getValidatorBeanClass() )
+								.stream()
+								// We do not include impl class of a validator if it comes from Hibernate Validator.
+								// In this case we only want to share the interfaces:
+								.filter( c -> !( validationProviderHelper.isHibernateValidator() && c.equals( validationProviderHelper.getValidatorBeanClass() ) ) )
```
yrodiere, line 52: This seems... fragile?

If we add some abstract class above `ValidatorImpl`, it will probably no longer work.

And with `ValidatorProviderHelper.forDefaultProvider`, it will definitely not work -- though I admit I have no idea what's the purpose of that thing.

Shouldn't you adjust how `ValidationProviderHelper` is built instead? E.g. `forHibernateValidator` would pass `HibernateValidatorFactory.class`/`HibernateValidator.class` to the constructor, while `forDefaultProvider` would do that if `ValidatorFactory` implements `HibernateValidatorFactory`, and default to some other behavior if not (maybe `ValidatorFactory.class`/`Validator.class`?).

cdi/src/test/java/org/hibernate/validator/cdi/internal/util/BuiltInConstraintValidatorUtilsTest.java
```diff
@@ -0,0 +1,44 @@
...
+			Class<? extends Annotation> constraint = (Class<? extends Annotation>) Class.forName( builtinConstraint );
+			for ( ConstraintValidatorDescriptor<? extends Annotation> descriptor : helper.getAllValidatorDescriptors( constraint ) ) {
+				assertThat( isBuiltInConstraintValidator( descriptor.getValidatorClass() ) ).isTrue();
+			}
+		}
+	}
+
+	@Test
+	public void testSomeRandomClasses() {
+		assertThat( isBuiltInConstraintValidator( Random.class ) ).isFalse();
```
yrodiere, line 39: Being literal here, eh? :grin:

engine/pom.xml
```diff
@@ -59,11 +60,11 @@
             <groupId>org.glassfish.expressly</groupId>
             <artifactId>expressly</artifactId>
             <scope>provided</scope>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>org.jboss.logging</groupId>
             <artifactId>jboss-logging-annotations</artifactId>
-            <scope>provided</scope>
```
yrodiere, line 66: FWIW the dependency is provided in `hibernate-search-engine`...

integrationtest/java/modules/cdi/pom.xml
```diff
@@ -24,13 +24,14 @@
...
         <!--
-           Caused by: java.lang.IllegalAccessError: superinterface check failed:
-           class org.hibernate.validator.internal.engine.ValidatorFactoryImpl$Proxy$_$$_WeldClientProxy
-           (in module org.hibernate.validator) cannot access class org.jboss.weld.interceptor.util.proxy.TargetInstanceProxy
-           (in unnamed module @0x6d763516) because module org.hibernate.validator does not read unnamed module @0x6d763516
+          If we do not read all unnamed for jboss logging it'll lead to:
+            Caused by: java.lang.IllegalArgumentException: This library does not have private access to interface org.jboss.weld.logging.VersionLogger
+                at org.jboss.logging@3.6.0.Final/org.jboss.logging.Logger.getMessageLogger(Logger.java:2569)
+                at org.jboss.logging@3.6.0.Final/org.jboss.logging.Logger.getMessageLogger(Logger.java:2549)
+                at org.jboss.weld.logging.VersionLogger.<clinit>(VersionLogger.java:35)
```
yrodiere, line 31: Shouldn't Weld be in the module path in your test?

yrodiere (approved):
Thanks. Added a few comments, but feel free to merge when you think it's ready.

cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidatorBean.java
marko-bekhta: hmm, if we pass just the interface, then we might lose some interfaces:
```java
ValidatorImpl implements Validator, ExecutableValidator
```
On the other hand, the `ValidatorFactoryImpl` extends `AutoCloseable`, and since we are taking the class hierarchy, we add `AutoCloseable` to the types, so that means we allow something like ...
```
@Inject
@HibernateValidator
AutoClosable validatorFactory;
```

is possible? 🤔 😃 

Maybe... if it is an HV factory/validator, we filter out any internal classes/interfaces by looking at the package, and otherwise, we keep everything. There's this test that checks if someone can inject their own custom factory/validator, and in that case, the implementation classes are injected, that's why I think if it is not "ours", we just keep it all and leave it to the user to deal with.

https://github.com/hibernate/hibernate-validator/blob/e5a569e122330feb37677a5a33d6322948a819b1/cdi/src/test/java/org/hibernate/validator/test/cdi/internal/injection/InjectionOfCustomProviderTest.java#L51-L73

yrodiere: > if it is an HV factory/validator, we filter out any internal classes/interfaces by looking at the package

Okay, but perhaps hardcode that list (with a test to check it stays up to date).

> otherwise, we keep everything. There's this test that checks if someone can inject their own custom factory/validator, and in that case, the implementation classes are injected, that's why I think if it is not "ours", we just keep it all and leave it to the user to deal with.

Makes sense.

integrationtest/java/modules/cdi/pom.xml
marko-bekhta: this particular exception is something from jboss-logging itself, there's a check for modules and since current version of Weld is on an older JBoss logging things fail.

As to where Weld should be .... IDK 🫨 😃. My understanding was that we should only have the `jakarta.cdi` as required in the module definition, and then Weld is just added as a test dependency... no?

yrodiere: > this particular exception is something from jboss-logging itself, there's a check for modules and since current version of Weld is on an older JBoss logging things fail.

ok :shrug:

PR #1399, opened 2024-07-31
HV-2018 Keep the interface if it is the hierarchy root

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-2018

I guess we never anticipated that the validation constraints will be only defined on an interface and that interface is passed to the `validateValue()`. 
What currently happens is that `beanMetaData.getClassHierarchy()` is empty and once we want to validate things we never get into this loop:

https://github.com/hibernate/hibernate-validator/blob/bb6129938a096d0928f64b7d34eb844a2d682419/engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorImpl.java#L458-L462

Which means we actually do not evaluate any constraints... 

----------------------

----------------------

### DIFF
--- a/cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidationProviderHelper.java
+++ b/cdi/src/main/java/org/hibernate/validator/cdi/internal/ValidationProviderHelper.java
@@ -13,7 +13,6 @@ import java.lang.reflect.Type;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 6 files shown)


marko-bekhta:
@gsmet I'll let you take a look at this one once you have time. I think the patch makes sense, but maybe I'm missing something.. and since the filtering part should only happen once when the metadata is created it shouldn't cause issues because of those additional allocations to create filter instances.

PR #1421, opened 2024-08-12
[HV-2027] Upgrade jsoup to 1.15.4 in hibernate-validator 6.0 branch

gaol:
Issue: https://hibernate.atlassian.net/browse/HV-2027

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,7 +131,7 @@
             See http://search.maven.org/#search|gav|1|g%3A"org.wildfly"%20AND%20a%3A"wildfly-parent"
         -->
         <classmate.version>1.3.4</classmate.version>
-        <jsoup.version>1.8.3</jsoup.version>
+        <jsoup.version>1.15.4</jsoup.version>
         <joda-time.version>2.9.7</joda-time.version>
         <slf4j.version>1.7.22</slf4j.version>
         <javax.persistence-api.version>2.2</javax.persistence-api.version>
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/SafeHtmlValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/SafeHtmlValidator.java
@@ -17,7 +17,7 @@ import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Node;
 import org.jsoup.parser.Parser;
 import org.jsoup.safety.Cleaner;
-import org.jsoup.safety.Whitelist;
+import org.jsoup.safety.Safelist;
 
 /**
  * Validate that the string does not contain malicious code.
@@ -31,7 +31,7 @@ import org.jsoup.safety.Whitelist;
  */
 @Deprecated
 public class SafeHtmlValidator implements ConstraintValidator<SafeHtml, CharSequence> {
-	private Whitelist whitelist;
+	private Safelist safelist;
 
 	private String baseURI;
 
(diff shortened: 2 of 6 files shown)


marko-bekhta (commented):
Hi,

Thanks for the patch. Maybe you could take a look at https://github.com/hibernate/hibernate-validator/blob/85798436a55dcff97c323397308005b8032582b9/engine/src/test/java/org/hibernate/validator/test/internal/constraintvalidators/hv/SafeHtmlValidatorTest.java#L37 and try adding a test there, to confirm that this update actually resolved the problem?

Also, 6.0 is an EOL version... would you need a release from us with this patch included ?

pom.xml
```diff
@@ -131,7 +131,7 @@
             See http://search.maven.org/#search|gav|1|g%3A"org.wildfly"%20AND%20a%3A"wildfly-parent"
         -->
         <classmate.version>1.3.4</classmate.version>
-        <jsoup.version>1.8.3</jsoup.version>
+        <jsoup.version>1.15.4</jsoup.version>
```
yrodiere, line 134: Note that upgrading from JSoup 1.8 to 1.15 means some JSoup features will change incompatibly.

Applications that happen to use JSoup directly, or through other dependencies than Hibernate Validator, and rely on those features, will break.

Doing this in a micro release is generally frowned upon.

If we absolutely need this to fix a CVE, so be it, but +1 on Marko's request to add a test to make sure this actually fixes the CVE.

gaol: Thanks for the reviews. @marko-bekhta, @yrodiere 

We are using a combination of jsoup:1.14.2 and hibernate-validator:6.0.23.Final.

PST scanned the jsoup as one of the shipment to mark our project as CVE affected, the jsoup:1.15.4 version is so far the clean version, which is suited for us. But the upgrade will fail because hibernate-validator:6.0.23.Final uses the [Whitelist which got removed and replaced by Safelist](https://github.com/jhy/jsoup/commit/2cc2399728d32f7277597901d8b6fcce619c0f11) in jsoup:1.15.1+, that is the reason to drive this upgrade of jsoup in hibernate-validator 6.0 branch, and replace the Whitelist with Safelist.

 Which means we need to address the security flaw which does not need to go through hibernate-validator. An example is [CVE-2022-36033](https://github.com/jhy/jsoup/security/advisories/GHSA-gp7f-rwcx-9369), sets `SafeList.preserveRelativeLinks` to `false` can be a workaround, but it is always `false` [from hibernate-validator point of view](https://github.com/hibernate/hibernate-validator/pull/1421/files#diff-0fc91a42b895d381cc1ce8295a171fa38e3578efe4ecde222fdbb77432b3426dR42).

So yes, it would be great if we can do the upgrade in hibernate-validator:6.0 branch and a new release for it. :) 

For the tests, I don't see much we can add to SafehtmlValidatorTest, because so far I think it is pure jsoup issue and fix. :handshake:

marko-bekhta: hey, mm ok, if it's purely a JSoup problem and nothing to test in HV, then we should be ok. 👍🏻  

I'll merge this one and tag it for you.  As per Yoann's comment on the Jira it usually was enough to just have a tag.  So I won't be looking into doing a release for 6.0 at the moment, let's see if the tag will be enough. It would be better if we don't introduce braking change in a micro. I'll monitor that ticket for any further developments.
And thanks for the patch once more! 🎉

marko-bekhta (approved):

pom.xml
marko-bekhta: https://github.com/hibernate/hibernate-validator/releases/tag/6.0.23.SP1 here's the tag

marko-bekhta: Actually, we've noticed that there were a few commits since the 6.0.23 release in the 6.0 branch. We will branch out from the release point and add this upgrade, so only this change is included.

marko-bekhta: and the build for it passed, see here https://ci.hibernate.org/blue/organizations/jenkins/hibernate-validator/detail/6.0.23.SP/2/pipeline

gaol: Thank you very much on the tag, SP tag works for me :+1:

marko-bekhta: Ok great! Thanks for letting us know.🙂

PR #1460, opened 2024-09-25
HV-2051 Use spotless maven plugin instead of formatter/impsorter/license check plugins

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-2051

This one is only to evaluate the option and not actually to apply the change ...
I thought that it would be better to test this on the Validator as it uses all 3 plugins and has a smaller codebase.
Best to look by commit (if looked at all 😅).

- Not all things that we are checking with checkstyle have an alternative (or at least I couldn't find one). I kept those rules in the checkstyle file so we can review them. I'll add a comment on that file so it's easier to find.
- I couldn't split the license header check from formatting; hence, the files with different licenses are not formatted... (I suspect that having multiple executions would do the trick, but I didn't want to go that route as I suppose the idea was to simplify the build and have fewer executions....)
- import sorting removes the comments. This means that the tags we have in the docs tests to include in the documentation file have to be excluded from spotless, resulting in those imports not being sorted (not a big deal, but a bit annoying 😃)
- While trying to figure out the patterns for XML headers I've noticed that the plugin does not always fail when there's a problem in the config. I  had a bad regex that was capturing the license header itself resulting in infinite update of the same file (at least  that is what I've seen while debugging). This resulted in the build just passing successfully and the files not affected (I'd rather it failed with a build error .... )

(this work is not complete: the maven properties and builds should also be updated to make it all work and I haven't done that here yet...)

----------------------

----------------------


marko-bekhta:
@yrodiere I'll let you have a look at this one (probably whenever you have nothing interesting to look into 😄 )

yrodiere:
> * I couldn't split the license header check from formatting; hence, the files with different licenses are not formatted... (I suspect that having multiple executions would do the trick, but I didn't want to go that route as I suppose the idea was to simplify the build and have fewer executions....)
> [...]
> * While trying to figure out the patterns for XML headers I've noticed that the plugin does not always fail when there's a problem in the config. I  had a bad regex that was capturing the license header itself resulting in infinite update of the same file (at least  that is what I've seen while debugging). This resulted in the build just passing successfully and the files not affected (I'd rather it failed with a build error .... )

Looks like we'd be better off keeping license header checks in checkstyle?

> import sorting removes the comments. This means that the tags we have in the docs tests to include in the documentation file have to be excluded from spotless, resulting in those imports not being sorted (not a big deal, but a bit annoying 😃)

Annoying indeed... though it's quite rare that we include comments in docs code snippets, at least for Hibernate Search, so I suppose we can live with it.
Do you know if there's a bug report / feature request in Spotless regarding this?

> Not all things that we are checking with checkstyle have an alternative (or at least I couldn't find one). I kept those rules in the checkstyle file so we can review them. I'll add a comment on that file so it's easier to find.

Will have a look, but if we need checkstyle for license header checks anyway, I suppose it's no big deal.

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter02/containerelement/custom/CarTest.java
```diff
@@ -1,3 +1,9 @@
+/*
+ * Hibernate Validator, declare and validate application constraints
+ *
+ * License: Apache License, Version 2.0
+ * See the license.txt file in the root directory or <http://www.apache.org/licenses/LICENSE-2.0>.
```
yrodiere, line 5: Can you create an issue to ultimately switch to SPDX headers, like in Search/ORM?

distribution/pom.xml
```diff
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
yrodiere, line 1: Oh, so it's not limited to code, it also gets applied to poms. Nice.

pom.xml
```diff
@@ -1416,58 +1304,84 @@
...
-                                <goal>${goal.impsort-maven-plugin}</goal>
-                            </goals>
-                            <phase>process-sources</phase>
-                        </execution>
-                    </executions>
-                </plugin>
-                <plugin>
-                    <groupId>net.revelc.code.formatter</groupId>
-                    <artifactId>formatter-maven-plugin</artifactId>
-                    <version>${version.formatter-maven-plugin}</version>
```
yrodiere, line 1445: So we *can* replace both of these plugins with just spotless? That's already a nice improvement, no?

pom.xml
```diff
@@ -1416,58 +1304,84 @@
...
+                                <exclude>**/ConcurrentReferenceHashMap.java</exclude>
+                            </excludes>
+                            <indent>
+                                <tabs>true</tabs>
+                                <spacesPerTab>4</spacesPerTab>
+                            </indent>
+
+                            <importOrder>
+                                <!-- you can use an empty string for all the imports you didn't specify explicitly, '|' to join group without blank line, and '\#` prefix for static imports. -->
+                                <order>\#,java,javax,jakarta,org.openjdk.jmh,org.hibernate,org.hibernate.testing,org.hibernate.test,org.junit,org.jboss,com,javafx.beans,</order>
```
yrodiere, line 1353: Can this be imported from a file? People will need a file specifying import orders in order to set up formatting in their IDE.

Unless spotless gets executed automatically in the IDE, but I suspect it only is when compiling...

pom.xml
```diff
@@ -1416,58 +1304,84 @@
...
+                            <eclipse>
+                                <!-- optional version of Eclipse Formatter -->
+                                <!-- <version>4.26</version> -->
+                                <file>hibernate_validator_style.xml</file>
+                            </eclipse>
+                            <removeUnusedImports>
+                                <!-- default one fails to process the classes -->
+                                <engine>cleanthat-javaparser-unnecessaryimport</engine>
+                            </removeUnusedImports>
+                            <licenseHeader>
```
yrodiere, line 1366: Does this apply the header automatically, on top of just checking?

pom.xml
```diff
@@ -1416,58 +1304,84 @@
                     </executions>
                 </plugin>
                 <plugin>
-                    <groupId>net.revelc.code</groupId>
-                    <artifactId>impsort-maven-plugin</artifactId>
-                    <version>${version.impsort-maven-plugin}</version>
+                    <groupId>com.diffplug.spotless</groupId>
+                    <artifactId>spotless-maven-plugin</artifactId>
```
yrodiere, line 1308: We probably need two execution mode: one for dev (auto formats) and one for CI (checks only, no auto format)

pom.xml
marko-bekhta: yes there's an option to pass a file:
``` XML
<order>java|javax,org,com,com.diffplug,,\#com.diffplug,\#</order>  <!-- or use <file>${project.basedir}/eclipse.importorder</file> -->
```

pom.xml
marko-bekhta: it's like with those other plugins, if `spotless:apply` is executed then the headers are added, and if it's `spotless:check` then it fails the build if the header is missing...

distribution/pom.xml
marko-bekhta: There's also this "pom" section that can be configured: https://github.com/diffplug/spotless/tree/main/plugin-maven#maven-pom where it can keep the pom "formatted" too, but I didn't want to try it as we have some cases where the profile has to go last (those reporting modules etc....) so I thought better leave it as is 😃

yrodiere: Ah, so this one was manual. Gotcha.

You're right, it's probably safer to skip poms, I didn't think of that.

documentation/src/test/java/org/hibernate/validator/referenceguide/chapter02/containerelement/custom/CarTest.java
marko-bekhta: https://hibernate.atlassian.net/browse/HV-2050

PR #1493, opened 2024-11-12
Use OSGi framework utils to get a classloader (aaaand not depend on the EL implementation in engine)

marko-bekhta:
I am opening this as a draft just to share.

That's what I ended up with to make that work .... and since we don't have a dependency on OSGi in the engine (and we wouldn't want that one, right ?! 😄) I ended up with some reflection... which is essentially doing this:

```java
	ClassLoader classLoaderForExpressionFactory() {
		for ( Bundle bundle : FrameworkUtil.getBundle( ResourceBundleMessageInterpolator.class ).getBundleContext().getBundles() ) {
			try {
				return bundle.loadClass( "com.sun.el.ExpressionFactoryImpl" ).getClassLoader();
			}
			catch (ClassNotFoundException e) {
				// ignore
			}
		}
		return null;
	}
```

----------------------

----------------------

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -53,11 +53,6 @@
         Provided dependencies
         -->
         <dependency>
-            <groupId>org.glassfish</groupId>
-            <artifactId>jakarta.el</artifactId>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
             <groupId>org.jboss.logging</groupId>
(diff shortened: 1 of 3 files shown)


yrodiere (commented):
Looks like a lot of work / complexity for what was only supposed to be a workaround, but... Assuming tests pass here, and your merge this in `main` only (where we can't run tests on OSGi :x), fine by me.

PR #1494, opened 2024-11-14
HV-2020 Use OSGi framework utils to get a classloader

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-2020

Use OSGi framework utils to get a classloader, and do not rely on a dependency to expressly.

Based on the change from https://github.com/hibernate/hibernate-validator/pull/1493 that targets 7.0 where we still could run OSGi tests

----------------------

----------------------

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -75,12 +75,6 @@
             <optional>true</optional>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


marko-bekhta:
well... we won't get the coverage here since we do not have the OSGi tests that this change is for ...

PR #1495, opened 2024-11-18
Upgrade to hibernate-jenkins-pipeline-helpers 1.18 -- and configure the version on CI instead of in Jenkinsfile

yrodiere:
Upgrade to 1.18 to avoid buggy notifications.
See https://hibernate.zulipchat.com/#narrow/channel/132096-hibernate-user/topic/Jenkins.20Notifications
Full changeset: https://github.com/hibernate/hibernate-jenkins-pipeline-helpers/compare/1.17..1.18

Don't configure the version in Jenkinsfile so that it's easier to upgrade next time.
See https://ci.hibernate.org/manage/configure#global-untrusted-pipeline-libraries to change the (default) version globally for all projects.

----------------------

----------------------

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,8 @@ import groovy.transform.Field
 /*
  * See https://github.com/hibernate/hibernate-jenkins-pipeline-helpers
  */
-@Library('hibernate-jenkins-pipeline-helpers@1.17')
+@Library('hibernate-jenkins-pipeline-helpers@1.17') _
+
 import org.hibernate.jenkins.pipeline.helpers.job.JobHelper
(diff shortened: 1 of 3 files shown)


marko-bekhta (approved):
This looks nice 🎉 🥳  less version updates to think about 😄 

Thinking of whether this can be a problem in the future,  I could only come up with this case: 
If we introduce a breaking change in the helper lib, then we'd need to go back to all (maintained) branches (where the version is not specified) to adjust the Jenkins file (since it doesn't point to a particular version anymore)

marko-bekhta:
let's keep that in mind 😄 
Thanks!

PR #1587, opened 2025-03-13
HV-2004 Add constraint initialization payload

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-2004

I didn't want to have some hardcoded cache for patterns within the pattern constraint validator impl itself. Hence I was thinking that leveraging something similar to what we have for validation with payload could work. 
initialization payload would be available to any constraints. As for patterns and their caching, limiting it to the predefined scope makes sense as we have all constraints initialized and can clear out the cache. 

----------------------

----------------------

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
+++ b/engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
@@ -56,4 +56,18 @@ public interface HibernateConstraintValidatorInitializationContext {
 	 */
 	@Incubating
 	Duration getTemporalValidationTolerance();
+
+	/**
+	 * Returns an instance of the specified type or {@code null} if the current constraint initialization context does not
+	 * contain an instance of such type.
+	 *
+	 * @param type the type of payload to retrieve
+	 * @return an instance of the specified type or {@code null} if the current constraint initialization context does not
+	 * contain an instance of such type
+	 *
+	 * @since 9.0.0
+	 * @see org.hibernate.validator.HibernateValidatorConfiguration#addConstraintValidatorInitializationPayload(Object)
+	 */
+	@Incubating
+	<C> C getConstraintValidatorInitializationPayload(Class<C> type);
 }
--- a/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/PatternValidator.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/PatternValidator.java
@@ -8,11 +8,14 @@ import java.lang.invoke.MethodHandles;
 import java.util.regex.Matcher;
 import java.util.regex.PatternSyntaxException;
 
-import jakarta.validation.ConstraintValidator;
 import jakarta.validation.ConstraintValidatorContext;
 import jakarta.validation.constraints.Pattern;
+import jakarta.validation.metadata.ConstraintDescriptor;
 
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidator;
 import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorContext;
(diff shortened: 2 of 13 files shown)


engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateConstraintValidatorInitializationContext.java
```diff
@@ -56,4 +56,18 @@ public interface HibernateConstraintValidatorInitializationContext {
...
+	 *
+	 * @param type the type of payload to retrieve
+	 * @return an instance of the specified type or {@code null} if the current constraint initialization context does not
+	 * contain an instance of such type
+	 *
+	 * @since 9.1.0
+	 * @see org.hibernate.validator.HibernateValidatorConfiguration#addConstraintValidatorInitializationPayload(Object)
+	 */
+	@Incubating
+	<C> C getConstraintValidatorInitializationPayload(Class<C> type);
```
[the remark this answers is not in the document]
marko-bekhta: yeah ... naming is hard 😔 🙂
Yes, it's closer to shared context. The other thought that I had about this: if we were always running the validator in the CDI context and with the predefined scope (i.e. all metadata is built at boot and not dynamically at runtime), I'd think of `PatternConstraintInitializer` as a bean that we inject into the `PatternValidator`, and the scope of such beans would be the init phase of the validator factory... but we aren't always running in CDI 🙈

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/PatternValidator.java
```diff
@@ -8,35 +8,40 @@
...
+		Pattern parameters = constraintDescriptor.getAnnotation();
 		Pattern.Flag[] flags = parameters.flags();
 		int intFlag = 0;
 		for ( Pattern.Flag flag : flags ) {
 			intFlag = intFlag | flag.getValue();
 		}
 
 		try {
-			pattern = java.util.regex.Pattern.compile( parameters.regexp(), intFlag );
+			pattern = initializationContext.getConstraintValidatorInitializationSharedService( PatternConstraintInitializer.class )
```
marko-bekhta, line 43: I thought ... that shared service might be a good fit for the name 🙂 🫣

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateValidatorFactoryObserver.java
```diff
@@ -0,0 +1,43 @@
...
+	default void factoryClosing(HibernateValidatorFactory factory) {
+	}
+
+	/**
+	 * A callback invoked after the factory has been successfully shut down.
+	 *
+	 * @param factory The closed factory.
+	 */
+	default void factoryClosed(HibernateValidatorFactory factory) {
+	}
```
marko-bekhta, line 42: and since clearing the cache for the pattern "shared service" explicitly felt more like a "hack" I've added this observer to react to HV factory lifecycle events and used it to clear the cache instead...

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/PatternValidator.java
yrodiere: I would say that a service is generally stateless, but YMMV :)

If `cache` really rubs you the wrong way, maybe `sharedData`?

By the way, why `getConstraintValidatorInitialization*`? This is a method in `initializationContext`, so `get*` would be enough, no?

engine/src/main/java/org/hibernate/validator/constraintvalidation/HibernateValidatorFactoryObserver.java
marko-bekhta: Hmm, I'll have to see how it is done for the predefined scope... I don't remember if the init context there is "short lived" or if it stays available for the whole lifetime of the factory... it should stay available for the regular case, as for it, the metadata is built lazily on the first validation attempt...

PR #1635, opened 2025-05-28
HV-1591 Log warning when `@Valid` is used improperly

DavideD:
Fix https://hibernate.atlassian.net/browse/HV-1591

I've updated the log4j configuration because changing it wasn't having any effect.
Not sure if there's a way to test that's actually logging.

I can add a test that shows the log (if you need it), but it's not really a test.

----------------------

----------------------

### DIFF
--- a/annotation-processor/src/test/java/org/hibernate/validator/ap/testutil/CompilerTestHelper.java
+++ b/annotation-processor/src/test/java/org/hibernate/validator/ap/testutil/CompilerTestHelper.java
@@ -10,8 +10,8 @@ import static org.testng.Assert.fail;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.List;
@@ -57,7 +57,7 @@ public class CompilerTestHelper {
 
 		private final String name;
 
-		private Library(String name) {
+		Library(String name) {
 			this.name = name;
 		}
 
@@ -82,7 +82,7 @@ public class CompilerTestHelper {
 		PROCESSOR_OUT_DIR = new File( TARGET_DIR, "processor-generated-test-classes" );
 		if ( !PROCESSOR_OUT_DIR.exists() ) {
 			if ( !PROCESSOR_OUT_DIR.mkdirs() ) {
-				fail( "Unable to create test output directory " + PROCESSOR_OUT_DIR.toString() );
+				fail( "Unable to create test output directory " + PROCESSOR_OUT_DIR );
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


marko-bekhta (commented):
Hey @DavideD, thanks for taking a look at this one!

HV has three ways to define the constraints: annotations, XML and programmatic API, so if we'd go with this, we'd need to do it in a place where we combine all the sources.... 

_But_ ... I realised that I've misled you by pointing to this JIRA while actually thinking of something slightly different 😖 🙈 

The problem is that people keep using the legacy approach for validating container elements, i.e. write this:

```java
@Valid List<MyPojo> list; // expecting HV to check that the list elements are `@Valid`.
```
and we want them to finally switch to the more appropriate way of:
```java
List<@Valid MyPojo> list;
```
particularly because we'd eventually want to get rid of that legacy approach (see https://github.com/jakartaee/validation/issues/266)

and eventually something like 
```java
@Valid // (1)
MyContainer<@Valid /*(2)*/ MyElement> container;
```
would actually be doing what it's intended to: `(1)` will apply cascading validation to the container itself, while `(2)` will apply it to the container elements.

Hence, we'd want to log a warning for `@Valid List<MyPojo> list` instead. 

All the "cascading magic" happens in here:

https://github.com/hibernate/hibernate-validator/blob/d0ecda58ddb991c2ef4eb64b66dda9de56f81576/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaDataBuilder.java#L210-L263

and then a bit more here at "runtime":

https://github.com/hibernate/hibernate-validator/blob/d0ecda58ddb991c2ef4eb64b66dda9de56f81576/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/PotentiallyContainerCascadingMetaData.java#L89-L102

These would be the primary places to add the new warning. 

HV also has this annotation processor that checks some constraint definition rules at compile time ( e.g. https://github.com/hibernate/hibernate-validator/blob/3f8dc0abec5d81e62761003cacbb288a24ab7c59/annotation-processor/src/main/java/org/hibernate/validator/ap/internal/checks/MixDirectAndListAnnotationCheck.java#L24). We could add a new check there that would look for `@Valid` at built-in containers and report warnings. But you can put the annotation processor into a different ticket/PR if you'd like.

Some other possible "test cases":
```java
@Valid List<MyBean> list;
@Valid Object = List<MyBean> objectButActuallyList;
@Valid Map<String, MyBean> map;

@Valid List<@Valid MyBean> list; // this will also get a warning in that cascading metadata builder. 
```
btw, you could probably try logging the context in that new warning as done here: https://github.com/hibernate/hibernate-validator/blob/3f8dc0abec5d81e62761003cacbb288a24ab7c59/engine/src/main/java/org/hibernate/validator/internal/metadata/aggregated/CascadingMetaDataBuilder.java#L195 so it'll be easier for the user to find what HV is complaining about. 


--------------------------

> Not sure if there's a way to test that's actually logging.

There's this `ListAppender` in test utils that can be used to collect logged messages. It's not too widely used 😃 but here's a test with it:

https://github.com/hibernate/hibernate-validator/blob/3f8dc0abec5d81e62761003cacbb288a24ab7c59/engine/src/test/java/org/hibernate/validator/test/internal/engine/messageinterpolation/ParameterMessageInterpolatorTest.java#L35

DavideD:
Thanks @marko-bekhta, I will update the pr

PR #1648, opened 2025-06-12
HV-2116 Do not recreate PersistenceUtil in the built-in JPATraversableResolver

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-2116

[Please describe here what your change is about]

----------------------

----------------------

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/resolver/JPATraversableResolver.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/resolver/JPATraversableResolver.java
@@ -8,6 +8,7 @@ import java.lang.annotation.ElementType;
 import java.lang.invoke.MethodHandles;
 
 import jakarta.persistence.Persistence;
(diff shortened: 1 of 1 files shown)


gsmet:
Did you discuss this with the Hibernate ORM team? Because I think it's a bit weird to create a new instance each time in `Persistence`?

And the fact that we missed this is a good example of why this can be misleading.

gsmet:
Note that it looks like something that should be merged and backported. I was just wondering if it could impact other projects out there.

marko-bekhta:
No, I was only looking into that other Quarkus issue related to the resolver and noticed this, so I pushed the change to be sure I wouldn't forget about it. Thanks for the suggestion, I'll start a topic on Zulip once I finish with the other HV "backport" 😃 🙈

gsmet:
Yeah, let's get it in though and backported because it allocates useless things in a relatively hot path.

gsmet:
Good catch :)

PR #1672, opened 2025-08-06
HV-2127 Make Path implementation not rely on Lists (or other collections)

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-2127

----------------------

----------------------

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
@@ -287,6 +287,9 @@ public class NodeImpl
 	}
 
 	public static NodeImpl setTypeParameter(NodeImpl node, Class<?> containerClass, Integer typeArgumentIndex) {
+		if ( node.typeArgumentIndex != null && node.typeArgumentIndex.equals( typeArgumentIndex ) && node.containerClass == containerClass ) {
+			return node;
+		}
 		return new NodeImpl(
 				node.name,
 				node.parent,
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
@@ -8,9 +8,9 @@ import java.io.Serializable;
 import java.lang.invoke.MethodHandles;
 import java.util.Arrays;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 
 import jakarta.validation.ElementKind;
 import jakarta.validation.Path;
@@ -44,9 +44,11 @@ public class NodeImpl
 	private static final Log LOG = LoggerFactory.make( MethodHandles.lookup() );
 
 	static final NodeImpl ROOT_NODE;
+
 	static {
 		ROOT_NODE = NodeImpl.createBeanNode( null );
 		ROOT_NODE.valueSet = true;
+		ROOT_NODE.nodes = new NodeImpl[] { ROOT_NODE };
 		ROOT_NODE.hashCode();
 	}
 
@@ -81,6 +83,7 @@ public class NodeImpl
 
 	private int hashCode = -1;
 	private String asString;
+	private NodeImpl[] nodes;
 
 	private NodeImpl(
 			String name, NodeImpl parent, boolean isIterable, Integer index, Object key, ElementKind kind, Class<?>[] parameterTypes,
@@ -568,17 +571,16 @@ public class NodeImpl
 
 	@Override
 	public Iterator<Path.Node> iterator() {
-		// TODO: keep the initialized list so next iterator calls can reuse it?
-		if ( parent == null ) {
-			return List.of( (Path.Node) this ).iterator();
+		if ( nodes == null ) {
+			nodes = new NodeImpl[size - 1];
+			NodeImpl curr = this;
+			while ( curr.parent != null ) {
+				nodes[curr.size - 2] = curr;
+				curr = curr.parent;
+			}
 		}
-		List<Path.Node> result = new LinkedList<>();
-		NodeImpl curr = this;
-		while ( !curr.isRootPath() ) {
-			result.add( 0, curr );
-			curr = curr.parent;
-		}
-		return result.iterator();
+
+		return new NodeIterator( nodes );
 	}
 
 	boolean isSubPathOf(NodeImpl other) {
@@ -648,5 +650,27 @@ public class NodeImpl
 		}
 
 		return curr == null && otherCurr == null;
+	}
+
+	private static class NodeIterator implements Iterator<Path.Node> {
+		private final NodeImpl[] array;
+		private int index;
+
+		public NodeIterator(NodeImpl[] array) {
+			this.array = array;
+		}
+
+		@Override
+		public boolean hasNext() {
+			return index < array.length;
+		}
+
+		@Override
+		public Path.Node next() {
+			if ( index < array.length ) {
+				return array[index++];
+			}
+			throw new NoSuchElementException();
+		}
 	}
 }
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
@@ -502,6 +502,16 @@ public class NodeImpl
 			return false;
 		}
 		NodeImpl other = (NodeImpl) obj;
+		return samePath( other );
+	}
+
+	boolean sameNode(NodeImpl other) {
+		if ( this == other ) {
+			return true;
+		}
+		if ( other == null ) {
+			return false;
+		}
 		if ( hashCode != -1 && other.hashCode != -1 && hashCode != other.hashCode ) {
 			return false;
 		}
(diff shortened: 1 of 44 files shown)


marko-bekhta:
and some numbers:

```
main:

Benchmark                                                                                                          Mode  Cnt      Score    Error   Units
o.h.v.p.cascaded.CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems        thrpt   20   9210.215 ± 54.555   ops/s
o.h.v.p.simple.SimpleValidation.testSimpleBeanValidation                                                          thrpt   20  17446.229 ± 95.293  ops/ms

Benchmark                                                                                                          Mode  Cnt      Score    Error   Units
o.h.v.p.cascaded.CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems        thrpt   20   9447.360 ± 28.172   ops/s
o.h.v.p.simple.SimpleValidation.testSimpleBeanValidation                                                          thrpt   20  15886.786 ± 39.997  ops/ms


change:
Benchmark                                                                                                          Mode  Cnt      Score    Error   Units
o.h.v.p.cascaded.CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems        thrpt   20  13766.536 ± 43.834   ops/s
o.h.v.p.simple.SimpleValidation.testSimpleBeanValidation                                                          thrpt   20  26797.112 ± 80.112  ops/ms

Benchmark                                                                                                          Mode  Cnt      Score    Error   Units
o.h.v.p.cascaded.CascadedWithLotsOfItemsAndMoreConstraintsValidation.testCascadedValidationWithLotsOfItems        thrpt   20  14434.296 ± 67.744   ops/s
o.h.v.p.simple.SimpleValidation.testSimpleBeanValidation                                                          thrpt   20  23922.016 ± 89.195  ops/ms
```

marko-bekhta:
> > One "downside" of this is that we do not have an easy way to iterate the path from the "root" as we only have the leaf node.
> 
> But do we need to? Or does the user need to?

I hope not, and I hope others would just use the string representation of the path ... but it may be that some lib goes through the path to construct the "violation" .. I'm thinking about something like `org.zalando:problem` (https://github.com/zalando/problem) I'll take a look at it a bit later to see what's happening there 🙈

marko-bekhta:
> > > One "downside" of this is that we do not have an easy way to iterate the path from the "root" as we only have the leaf node.
> > 
> > 
> > But do we need to? Or does the user need to?
> 
> I hope not, and I hope others would just use the string representation of the path ... but it may be that some lib goes through the path to construct the "violation" .. I'm thinking about something like `org.zalando:problem` (https://github.com/zalando/problem) I'll take a look at it a bit later to see what's happening there 🙈

ok... I had a look at Quarkus and there are a few places where the path iterator is used to construct the response... e.g.:

https://github.com/quarkusio/quarkus/blob/ebee0a4b66d1b4dba9c51165e78d22fbf3b90d3e/extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/jaxrs/ResteasyReactiveViolationExceptionMapper.java#L59-L69

or here:

https://github.com/quarkusio/quarkus/blob/ebee0a4b66d1b4dba9c51165e78d22fbf3b90d3e/extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/interceptor/AbstractMethodValidationInterceptor.java#L155-L161

I could add some HV-specific methods to address these use cases... would Quarkus be able to leverage those?

gsmet:
> I could add some HV-specific methods to address these use cases... would Quarkus be able to leverage those?

I think I misunderstood you previously: I don't think you can break the contract of `Path`. It extends `Iterable` so `iterator()` should work.

Now the question is: can you somehow materialize the `Path` when we get a constraint violation? Because having a constraint violation is somehow a corner case. Most of the time, you shouldn't have one.

engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
```diff
@@ -568,17 +571,16 @@ boolean isRootPath() {
...
+			nodes = new NodeImpl[size - 1];
+			NodeImpl curr = this;
+			while ( curr.parent != null ) {
+				nodes[curr.size - 2] = curr;
+				curr = curr.parent;
+			}
 		}
-		return result.iterator();
+
+		return new NodeIterator( nodes );
```
marko-bekhta: > > I could add some HV-specific methods to address these use cases... would Quarkus be able to leverage those?
>
> I think I misunderstood you previously: I don't think you can break the contract of Path. It extends Iterable so iterator() should work.
>
> Now the question is: can you somehow materialize the Path when we get a constraint violation? Because having a constraint violation is somehow a corner case. Most of the time, you shouldn't have one.

oh, it's still iterable, so we are all ok from that perspective... My concern is more that it would be less performant to get the iterator the first time, as we need to get from the leaf to the root and collect all the nodes in some collection (or array in this case), and then "iterate" back.

I'm thinking more like in that case of getting the leaf node -- we get an iterator and go to the "last node" but the thing is that the path has the leaf for us right from the start so if we have some:

```java
interface HibernateValidatorPath extends jakarta.validation.Path {

  Node leaf();

}
```

that would be "faster" 🙂

engine/src/main/java/org/hibernate/validator/internal/engine/path/NodeImpl.java
```diff
@@ -525,4 +564,113 @@ else if ( !Arrays.equals( parameterTypes, other.parameterTypes ) ) {
...
+		if ( nodes == null ) {
+			nodes = new NodeImpl[size - 1];
+			NodeImpl curr = this;
+			while ( curr.parent != null ) {
+				nodes[curr.size - 2] = curr;
+				curr = curr.parent;
+			}
+		}
+
+		return new NodeIterator( nodes );
```
gsmet, line 583: AFAICS, this is not thread safe, which is a problem IMHO.

I think it would be safer to publish a specific full initialized object when we push a `Path` out of the engine. My understanding is that it's only when we build a `ConstraintViolation`?

Or is it more than that?

marko-bekhta: yes, we only need this for the `ConstraintViolation`.... 

yeah I didn't want to synchronize that block .. 
ok let me try something else instead ...

engine/src/main/java/org/hibernate/validator/internal/engine/path/PublicPath.java
```diff
@@ -0,0 +1,55 @@
...
+package org.hibernate.validator.internal.engine.path;
+
+import java.io.Serial;
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.Iterator;
+
+import jakarta.validation.Path;
+
+final class PublicPath implements Path, Serializable {
```
gsmet, line 14: Maybe `MaterializedPath` would be better? But unsure, it's just a suggestion.

engine/src/main/java/org/hibernate/validator/internal/engine/path/PublicPath.java
```diff
@@ -0,0 +1,55 @@
...
+	private static final long serialVersionUID = -8906501301223202169L;
+
+	private final NodeImpl leafNode;
+	private final NodeImpl[] nodes;
+	private final int hashCode;
+
+	PublicPath(ModifiablePath path) {
+		this.leafNode = path.getLeafNode();
+		this.nodes = NodeImpl.constructPath( this.leafNode );
+		this.hashCode = path.hashCode();
```
gsmet, line 26: I think I would avoid doing that here, and let it be done later.

Except if we are using it on the HV side or if it's a common use case amongst consumers?

engine/src/main/java/org/hibernate/validator/internal/engine/validationcontext/AbstractValidationContext.java
```diff
@@ -247,7 +247,7 @@ public void addConstraintFailure(
 				constraintViolationCreationContext.getExpressionVariables()
 		);
 		// at this point we make a copy of the path to avoid side effects
-		Path path = PathImpl.createCopy( constraintViolationCreationContext.getPath() );
+		Path path = ModifiablePath.createPublicCopy( constraintViolationCreationContext.getPath() );
```
gsmet, line 250: Then you could go with `materialize()` here

gsmet (commented):
I added a few comments.

FWIW, we could have a specific benchmark with only things that are valid and measure the improvements there.

Then have one where we have failures to measure that things have not gone too badly for this case.

gsmet:
@marko-bekhta were you able to re-run the benchmarks with this new code?

marko-bekhta:
yes. it seems still better than what we currently have 🙂 here's what I saved from Friday:

```
main:
Benchmark                                                     Mode  Cnt      Score    Error   Units
SimpleSingleElementValidation.invalidObjectValidation        thrpt   20   6474.390 ± 31.799  ops/ms
SimpleSingleElementValidation.validObjectValidation          thrpt   20  17840.565 ± 51.801  ops/ms

this patch:
Benchmark                                                     Mode  Cnt      Score     Error   Units
SimpleSingleElementValidation.invalidObjectValidation        thrpt   20   8527.579 ±  40.861  ops/ms
SimpleSingleElementValidation.validObjectValidation          thrpt   20  33352.222 ± 402.655  ops/ms
```
or you mean just the few latest commits with the `MaterializedPath` ?

(btw ^ this is from a "new benchmark" I've added with one simple entity being either valid or invalid. I noticed that some of our other benchmarks that use random can have a quite different results from one run to the other) so I wanted to have something more "stable")


> it seems still better than what we currently have

which does make sense since we do not create/copy the lists on each operation, but just create one array for the failing constraint.

gsmet:
> (btw ^ this is from a "new benchmark" I've added with one simple entity being either valid or invalid. I noticed that some of our other benchmarks that use random can have a quite different results from one run to the other) so I wanted to have something more "stable")

Yeah, it makes perfect sense.

Looks all good to me if you want to merge it.

marko-bekhta:
Thanks for having a look at this one, Guillaume! 🙏🏻 🙇🏻 🙂 
yeah.. I should probably merge it before switching to the bean tracking PR so that there would be less problems with rebasing... I'll have a look at what Sonar suggests to fix and then merge 🙂

gsmet:
Woot! Congrats on this change, it's awesome!

marko-bekhta:
thanks! Processed bean tracking is coming next 😉🤞🏻

PR #1707, opened 2025-09-15
HV-2139 Introduce an IndexedPath as an extension of the org.hibernate.validator.path.Path

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-2139

----------------------

----------------------

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/internal/engine/path/MaterializedPath.java
+++ b/engine/src/main/java/org/hibernate/validator/internal/engine/path/MaterializedPath.java
@@ -6,15 +6,20 @@ package org.hibernate.validator.internal.engine.path;
 
 import java.io.Serial;
 import java.io.Serializable;
+import java.lang.invoke.MethodHandles;
 import java.util.Iterator;
 
-import org.hibernate.validator.path.Path;
+import org.hibernate.validator.internal.util.logging.Log;
+import org.hibernate.validator.internal.util.logging.LoggerFactory;
+import org.hibernate.validator.path.RandomAccessPath;
 
 
-final class MaterializedPath implements Path, Serializable {
+final class MaterializedPath implements RandomAccessPath, Serializable {
+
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/hibernate/validator/path/IndexedPath.java
```diff
@@ -0,0 +1,34 @@
...
+import org.hibernate.validator.Incubating;
+
+/**
+ * An extended representation of the validation path, provides Hibernate Validator specific functionality.
+ * Represents a path with access to the nodes by their index.
+ *
+ * @since 9.1
+ */
+@Incubating
+public interface IndexedPath extends Path {
```
marko-bekhta, line 16: @yrodiere do you think `IndexedPath` would be a good name for a path that allows access to the nodes by an index? (or could someone think that this path is to be stored in a Lucene index 😁 )

yrodiere: I'm not sure why you didn't just pick `HibernatePath`, similar to what is being done for every other Jakarta Validation interface that we extend in Hibernate Validator?

If this really needs to be about the access by index, maybe `RandomAccessPath`, similar to `RandomAccessList`? Or maybe try to find some inspiration at https://openjdk.org/jeps/431 ?

marko-bekhta: > I'm not sure why you didn't just pick HibernatePath

right ... well we sort of have two in HV -- one using a link approach and the other one backed by an array ... 
so `ArrayPath` `LinkedPath` as with lists ? 🙈

yrodiere: Do you need two?

Serious question, as performance wise I've been talked `ArrayList` often outperforms `LinkedList` anyway, so...

yrodiere (approved):
Makes sense to me. Thanks.

PR #1938, opened 2026-04-02
HV-2202 Do not recreate HibernateConstraintValidatorContext if possible

marko-bekhta:
https://hibernate.atlassian.net/browse/HV-2202

```
Benchmark                                                              Mode  Cnt      Score     Error   Units
SimpleSingleElementValidation.invalidObjectNoInterpolationValidation  thrpt   10  26834.348 ± 155.542  ops/ms
SimpleSingleElementValidation.invalidObjectValidation                 thrpt   10  13511.785 ±  49.245  ops/ms
SimpleSingleElementValidation.validObjectValidation                   thrpt   10  53209.841 ±  71.306  ops/ms

Benchmark                                                                    Mode  Cnt      Score      Error   Units
SimpleSingleElementValidation.invalidObjectNoInterpolationValidation        thrpt   10  32649.943 ±  394.400  ops/ms
SimpleSingleElementValidation.invalidObjectValidation                       thrpt   10  14809.263 ±   31.111  ops/ms
SimpleSingleElementValidation.validObjectValidation                         thrpt   10  82862.102 ± 1110.725  ops/ms
```

this mostly helps for the cases where there are no violations as we create much-much less objects ...

----------------------

----------------------


marko-bekhta:
(if anyone want's to go through the changes ... 🙈 I'll leave it open till next week 🙂)

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ComposingConstraintTree.java
```diff
@@ -64,15 +64,11 @@ private <U extends Annotation> ConstraintTree<U> createConstraintTree(Constraint
...
-		List<ConstraintValidatorContextImpl> violatedConstraintValidatorContexts = new ArrayList<>( 5 );
-		validateConstraints( validationContext, valueContext, violatedConstraintValidatorContexts );
-		if ( !violatedConstraintValidatorContexts.isEmpty() ) {
-			for ( ConstraintValidatorContextImpl constraintValidatorContext : violatedConstraintValidatorContexts ) {
-				for ( ConstraintViolationCreationContext constraintViolationCreationContext : constraintValidatorContext.getConstraintViolationCreationContexts() ) {
-					validationContext.addConstraintFailure(
-							valueContext, constraintViolationCreationContext, constraintValidatorContext.getConstraintDescriptor()
-					);
-				}
+		List<ConstraintViolationCreationContext> constraintViolationCreationContexts = new ArrayList<>( 5 );
```
yrodiere, line 67: Unrelated to your changes, but it seems curious to me that we assume there will be violations (init size = 5). I'd expect an empty list to be much more common?

Then again I suspect this sort of micro-optimization does not bring much in practice...

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/HibernateConstraintValidatorReusableContext.java
```diff
@@ -184,6 +184,29 @@ private ConstraintViolationCreationContext getDefaultConstraintViolationCreation
...
 
+	@Override
+	public List<String> getMethodParameterNames() {
+		if ( ContextKind.CROSS_PARAMETER.equals( contextKind ) ) {
+			return parameterNames;
+		}
+		throw LOG.getUnexpectedConstraintValidatorContextCall();
+	}
+
+	public void supplyConstraintViolationCreationContexts(Collection<ConstraintViolationCreationContext> contexts) {
```
yrodiere: This feels more natural to me, but I'm biased towards Search, so... up to you :)

```suggestion
	public void contributeConstraintViolationCreationContexts(Collection<ConstraintViolationCreationContext> contexts) {
```

engine/src/test/resources/log4j2.properties
```diff
@@ -15,11 +15,11 @@ appender.console.layout.pattern = %d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
 logger.parametermessageinterpolator.name = org.hibernate.validator.messageinterpolation.ParameterMessageInterpolator
 logger.parametermessageinterpolator.level = info
 
-logger.constraintvalidatorcontextimpl.name = org.hibernate.validator.internal.engine.constraintvalidation.ConstraintValidatorContextImpl
-logger.constraintvalidatorcontextimpl.level = info
+logger.hibernateconstraintvalidatorreusablecontext.name = org.hibernate.validator.internal.engine.constraintvalidation.HibernateConstraintValidatorReusableContext
+logger.hibernateconstraintvalidatorreusablecontext.level = info
```
yrodiere: Is this something that could impact users, i.e. have we recommended using this specific logger somewhere, down to the class name?

If so, you might want to consider keeping the same name for `ConstraintValidatorContextImpl`. As long as you only have one implementation, the name doesn't matter much anyway.

If this is just about internal/debug logs, on the other hand... Please ignore me :)

engine/src/main/java/org/hibernate/validator/internal/util/logging/Log.java
```diff
@@ -962,4 +962,7 @@ ConstraintDefinitionException getConstraintValidatorDefinitionConstraintMismatch
 
 	@Message(id = 273, value = "Index %1$d is out of bounds for the path of length %2$d.")
 	IndexOutOfBoundsException pathIndexOutOfBounds(int index, int length);
+
+	@Message(id = 274, value = "Unsupported operation for the current constrain validator kind.")
```
yrodiere:
```suggestion
	@Message(id = 274, value = "Unsupported operation for the current constraint validator kind.")
```

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/HibernateConstraintValidatorReusableContext.java
```diff
@@ -83,7 +78,7 @@ public final String getDefaultConstraintMessageTemplate() {
 
 	@Override
 	public HibernateConstraintViolationBuilder buildConstraintViolationWithTemplate(String messageTemplate) {
-		return new ConstraintViolationBuilderImpl(
+		return new HibernateConstraintValidatorReusableContext.ConstraintViolationBuilderImpl(
```
yrodiere: This kind of change is sprinkled throughout the class, but I don't think it's necessary -- probably an artefact of IDE refactoring?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/HibernateConstraintValidatorReusableContext.java
```diff
@@ -426,4 +462,49 @@ private void addLeafNode() {
 			}
 		}
 	}
+
+	public HibernateConstraintValidatorReusableContext asRegularContext(
```
yrodiere: While I understand it's convenient, returning `this` in an `asSomethingElse` method could cause confusion: is the return a new instance or the same one?

returning void and renaming to `resetAsRegularContext` or something like that might be wiser.

yrodiere (approved):
LGTM! I added a few comments below, mostly so I could claim I did review, nothing critical :)

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/ComposingConstraintTree.java
marko-bekhta: oh yeah, this one is another thing I'd wish we could eliminate ... (the whole list in general but I don't have a good idea for now) 🙂 🫣 

soo why it's 5 ? 🙂 It's been a while, but I think this comes from some earlier optimisations where we wanted to reduce the memory consumption, and 5 was somewhat of a magic value where you most likely won't need to resize the list (if there are composing constraints, this list can grow) but still won't take much if it's a simple constraint.

engine/src/test/resources/log4j2.properties
marko-bekhta: hmmm ...
the `logger.hibernateconstraintvalidatorreusablecontext.name` part I think is just a log4j config side, user can name it anything they wish... as for the class ... I cannot find it anywhere else (like in the docs or public javadocs) sooo we probably should be ok, but then yeah ... it was useful to have a different name while refectoring and keeping all the versions around, but now with one remaining might as well keep the initial name 🙂

yrodiere:
> I'll leave it open till next week 🙂

Are you waiting for me? Because I wasn't planning to do anything :grimacing:

marko-bekhta:
ah thanks for the reminder ! 🙂 
I just got carried away with other things 🙈 🙂 merging now 🤞🏻

PR #2000, opened 2026-06-08
HV-2220 Add @Port constraint

koentsje:
Add a new `@Port` constraint that validates whether a numeric value is a valid port number (0-65535 inclusive).

This is a composed constraint using `@Min(0)` and `@Max(65535)` with `@ReportAsSingleViolation`, so it supports all numeric types out of the box without a dedicated validator class.

Null values are considered valid, consistent with Bean Validation conventions.

----------------------

----------------------

### DIFF
--- a/engine/src/main/java/org/hibernate/validator/constraints/Port.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/Port.java
@@ -1,0 +1,65 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.validator.constraints;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.CONSTRUCTOR;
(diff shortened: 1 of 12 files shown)


engine/src/main/java/org/hibernate/validator/constraints/Port.java
```diff
@@ -0,0 +1,65 @@
...
+ * @since 9.2
+ */
+@Documented
+@Constraint(validatedBy = { })
+@SupportedValidationTarget(ValidationTarget.ANNOTATED_ELEMENT)
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
+@Retention(RUNTIME)
+@Repeatable(List.class)
+@Min(0)
+@Max(65535)
```
marko-bekhta, line 46: Maybe let's have explicit validators instead (so that we don't need to build `ComposingConstraintTree` at runtime)? Usually, going with `SimpleConstraintTree` is better for performance.

marko-bekhta (commented):
Thanks! Looks good overall! 🙂 🎉

PR #2010, opened 2026-06-20
HV-1852 Add enum validation warning regression test

xfocus3:
https://hibernate.atlassian.net/browse/HV-1852

This adds regression coverage for validating enum constants whose constructors have the implicit enum parameters. The test validates an enum constant and captures Hibernate Validator WARN logs to ensure the old `HV000254: Missing parameter metadata` warning is not emitted.

Validation:
- `./mvnw -pl engine -am -Dtest=org.hibernate.validator.test.internal.properties.javabean.JavaBeanExecutableTest test -DskipITs -DdisableDistributionBuild=true -Dformat.skip=true -Dsurefire.failIfNoSpecifiedTests=false`
- `git diff --check`

----------------------

----------------------

### DIFF
--- a/engine/src/test/java/org/hibernate/validator/test/internal/properties/javabean/JavaBeanExecutableTest.java
+++ b/engine/src/test/java/org/hibernate/validator/test/internal/properties/javabean/JavaBeanExecutableTest.java
@@ -10,9 +10,17 @@ import java.lang.reflect.ParameterizedType;
 import java.util.List;
 import java.util.Map;
 
+import jakarta.validation.Validation;
+import jakarta.validation.ValidatorFactory;
+
 import org.hibernate.validator.internal.properties.javabean.JavaBeanConstructor;
 import org.hibernate.validator.testutil.TestForIssue;
+import org.hibernate.validator.testutils.ListAppender;
 
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.core.Logger;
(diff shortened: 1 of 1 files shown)


marko-bekhta (commented):
hey 👋🏻 🙂 
thanks for the PR! A  couple of things ...
- the tests are compiled with `-parameters` so it wouldn't be able to catch that scenario ...
- I think in HV-1927, where the change in how things are resolved, that particular message 254 was also gone, so the test won't be able to catch anything anymore ?

maybe I am missing something ?

xfocus3:
Thanks, you were right to point that out. The original version only covered the regular test-compiled enum.

I updated the PR with an additional regression that compiles the reproducer enum during the test without `-parameters`, loads it, and validates the enum constant before asserting that no `HV000254` / missing-parameter-metadata warning is emitted.

Verification:

```bash
./mvnw -DdisableDistributionBuild=true -pl engine -am -Dtest=JavaBeanExecutableTest -Dsurefire.failIfNoSpecifiedTests=false test
```

This runs the full `JavaBeanExecutableTest` class successfully in both the default and Byte Buddy test executions.

marko-bekhta:
hey, thanks again for the effort 🙂 
since that area responsible for producing the warning at a time, was reworked and we are not returning to previously used message keys, the test won't really catch any regression ... I am going to close the related issue (and the PR). Thanks!

PR #2032, opened 2026-07-02
Password related validation: HV-2232 @NotCompromised / HV-2233 @PasswordStrength / HV-2234 @PasswordPolicy / HV-2235 Use BeanResolver / BeanProvider 

marko-bekhta:
Hey @yrodiere 🙂 👋🏻 

I thiiiink this turned out to be pretty ok. Going through all the changes would take ages 🙈 but if you'd have the time to build the docs and look at those -- I'd like to hear your opinion 🙂 (or look at asciidoc, but then you'd need to go back and forth to the test classes with code examples 🙈)

https://hibernate.atlassian.net/browse/HV-2232
https://hibernate.atlassian.net/browse/HV-2233
https://hibernate.atlassian.net/browse/HV-2234
https://hibernate.atlassian.net/browse/HV-2235

----------------------

----------------------


engine/src/test/java/org/hibernate/validator/spi/password/PasswordStrengthScore.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.hibernate.validator.spi.password;
+
+import org.hibernate.validator.Incubating;
+
+/**
+ * Named constants for common password strength score thresholds.
+ * <p>
+ * These values align with the 0–4 scoring scale used by libraries
+ * such as zxcvbn and nbvcxz.
```
yrodiere, line 13: You say 0-4, but the implementation is 1-4.

Also, is this scale somewhat standard? If not, I suspect it would be just as well to leave the definition of constants to password strength service implementors...

documentation/src/main/asciidoc/reference/_ch09.adoc
```diff
@@ -525,6 +525,268 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/CustomScri
...
 ====
 
+[[section-password-validation-services]]
+==== Configuring password validation services
+
+The `@PasswordStrength` and `@NotCompromised` constraints delegate to user-provided validation services
+that must be registered at bootstrap time. The two service interfaces are:
+
+* `PasswordStrengthEstimator` -- estimates password strength on a 0–4 scale (see `PasswordStrengthScore`)
+* `CompromisedPasswordChecker` -- checks whether a password has appeared in a known data breach
```
yrodiere, line 646: I would recommend that you had a big fat warning somewhere that "you should never send passwords to external services without first transforming them the make sure they can't be used to log in (hashing, ...)".

Not that it should be our problem, but you just know we'll get an automated security report about that.

documentation/src/main/asciidoc/reference/_ch09.adoc
```diff
@@ -525,6 +525,268 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/CustomScri
...
+
+====
+[source, java, indent=0]
+----
+include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/PasswordPolicyTest.java[tags=serviceRules]
+----
+====
+
+See <<section-constraint-definition-resolver>> for how to register the resolver, and <<section-getting-started-cdi>>
+for CDI integration details.
```
yrodiere: I think this is going too far into "how to structure an application using CDI in 2026" territory. Almost none of this is specific to Hibernate Validator, so I would leave it out. At best I would document that custom rules are possible and show one that implements the rule inline (without relying on other components).

Unless I misunderstood what was going on here?

documentation/src/main/asciidoc/reference/_ch09.adoc
```diff
@@ -525,6 +525,268 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/CustomScri
...
+
+[[section-constraint-definition-resolver]]
+===== Custom policy definition resolver
+
+By default, `PasswordPolicyDefinition` classes are instantiated via their no-arg constructor.
+For dependency injection or custom initialization, register a `PasswordPolicyDefinitionResolver`
+as a validation service:
```
yrodiere: Shouldn't this be a higher-level concept with broader applications (to every other component that can be resolved)? Thinking of how we do things in Search with the `BeanResolver`, or in ORM with the `BeanContainer`.

documentation/src/main/asciidoc/reference/_ch09.adoc
```diff
@@ -525,6 +525,268 @@ include::{sourcedir}/org/hibernate/validator/referenceguide/chapter09/CustomScri
...
+====
+
+[[section-password-policy-class-level]]
+===== Class-level validation with `AbstractPasswordPolicyValidator`
+
+When a password policy rule needs access to other bean properties -- for example, to reject passwords
+that contain the username -- `@PasswordPolicy` can be placed on the class instead of the password field.
+
+Hibernate Validator does not ship a class-level validator; you provide your own by extending
+`AbstractPasswordPolicyValidator<T>`. This base class resolves the policy definition, builds
```
yrodiere, line 811: Again this is nice but it looks like you're implementing far-reaching features (composition of validators, class-level validator) in a very specific validator.

Do you intend to pull this out of here someday and allow such things for any validator? I could see a "field a must be lower than field b" validator...

yrodiere (commented):
Looks very nice! I don't have much to say, except maybe that some parts of this feel like they're going too far (for such a feature) or not far enough (failing to extend concepts to other validators).

engine/src/test/java/org/hibernate/validator/spi/password/PasswordStrengthScore.java
marko-bekhta: ah right .. it's based on those two libs. They can return 0-4 as the score, 0 felt like -- if you need 0 password strength, you might as well just not place the constraint 🙈 

probably better to just say that 

> These values align with the scoring scale used by libraries such as zxcvbn and nbvcxz.

I didn't find any "standards"  for this and also from various registration forms I've seen the scale varies ... 

Maybe we should move this to the docs instead ... my main intention here really was to show -- use a well-defined list of constants for your pwd strength scores instead of just putting random ints in that annotation attribute.

I was not convinced about including the libs as dependencies (even as optional), but at the same time, as I couldn't find other Java-based impls for the strength, I thought it might be nice to "ship" this for the user ?

documentation/src/main/asciidoc/reference/_ch09.adoc
marko-bekhta: yeah maybe it's not a reference-grade example ... I mainly wanted to say that by default we use a no-arg constructor to create custom rule instances, and if someone needs something more complex and passes some parameters to the rule, they need to implement their own rule resolver ... but maybe CDI is a step too far to show it .. wanted a "realistic" example 🙈

documentation/src/main/asciidoc/reference/_ch09.adoc
marko-bekhta: That's actually a nice idea... thanks 🙂 ! that way maybe we'd be able to get rid of the "validation services" (I've "borrowed" the idea from ORM, should've looked closer (at Search) !!! )

documentation/src/main/asciidoc/reference/_ch09.adoc
marko-bekhta: > I could see a "field a must be lower than field b" validator...

yeah that's something users want and ask for from time to time... problem is when constraint goes on a field it doesn't see the bean it belongs to ... but definetly something to consider if we can solve it in a nice (and ideally typesafe) manner ... 

With the password policies, my thinking was that we can provide a validator on a field level, but if someone needs to do a username/password combination, they are required to fully implement that on their own, and some of the classes are internal, so they won't be able to reuse things ... so we keep supporting the field level validation for passwords, and if user needs to go at class level they implement their own thing ? 
(our own field level implementation stays at a field level and cannot reach other fields of the containing bean)

don't know if that explanation helped at all 🙈 🥲

engine/src/test/java/org/hibernate/validator/spi/password/PasswordStrengthScore.java
yrodiere: > Maybe we should move this to the docs instead ... my main intention here really was to show -- use a well-defined list of constants for your pwd strength scores instead of just putting random ints in that annotation attribute.

In principle, I agree.

In practice, I feel like that's none of our business :)

> I was not convinced about including the libs as dependencies (even as optional), but at the same time, as I couldn't find other Java-based impls for the strength, I thought it might be nice to "ship" this for the user ?

Ugh, I didn't notice that. Why do you need that dependency exactly, if the implementation is up to the user?

documentation/src/main/asciidoc/reference/_ch09.adoc
yrodiere: You can do a blog post separately from this, you know :) In fact it would be a very good idea, to showcase what's new after the next release.

engine/src/test/java/org/hibernate/validator/spi/password/PasswordStrengthScore.java
marko-bekhta: > In practice, I feel like that's none of our business :)

🙈 🙂 

> Ugh, I didn't notice that. Why do you need that dependency exactly, if the implementation is up to the user?

oh we don't need it -- we have it in tests only, just for the sake of having some realistic test running, but yes it's all up to the user to do it.  When I was saying that I was not convinced about including the libs -- I meant I was not convinced it is a good idea to include them, so I didn't 🙈

marko-bekhta:
hey @yrodiere 🙂 👋🏻 

coming back to this one ... the bean resolver stuff worked out fine more or less ... there are a few tricky parts here 🫣 
- a lot of components that potentially could go through a bean resolver are needed early (getter/setter strategy, bean normalizer ...) so they can't be resolved lazily and keeping the bean holder for them doesn't make sense.
- lifecycle of most (all ?) components is really the for the lifetime of the factory itself.
so with that in mind I've decided that we should state it clearly that those basic components won't be trying to go to the bean provider and will just be resolved eagerly (constructor/reflection/instance). Laziness would only be available in the constraint validators. That way users would be able to pass some beans as e.g. password strength estimators or other similar components.

yrodiere:
Okay, well. Where should I look? Need feedback on something specific? :)

cdi/src/main/java/org/hibernate/validator/cdi/internal/CdiBeanProvider.java
```diff
@@ -0,0 +1,74 @@
...
+		@Override
+		public T get() {
+			return handle.get();
+		}
+
+		@Override
+		public void close() {
+			if ( Dependent.class.equals( handle.getBean().getScope() ) ) {
+				handle.destroy();
+			}
```
marko-bekhta, line 71: it seems Arc's instance handlers better handle destroying part .. Here in the tests, if I call destroy on an application-scoped one, things gets ugly when the container tries closing the application-scoped beans 🙂

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/password/PasswordStrengthValidatorForCharSequence.java
```diff
@@ -0,0 +1,71 @@
...
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidator;
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorContext;
+import org.hibernate.validator.constraintvalidation.HibernateConstraintValidatorInitializationContext;
+import org.hibernate.validator.spi.password.PasswordStrengthEstimator;
+import org.hibernate.validator.spi.password.PasswordStrengthResult;
+
+public class PasswordStrengthValidatorForCharSequence implements HibernateConstraintValidator<PasswordStrength, CharSequence> {
+
+	private int minScore;
+	private BeanHolder<PasswordStrengthEstimator> estimatorHolder;
```
marko-bekhta, line 25: since we might build the metadata early, some cdi beans might not be there yet, so we have a holder and then we "get" in the `isValid`

engine/src/main/java/org/hibernate/validator/internal/engine/ValidatorFactoryConfigurationHelper.java
```diff
@@ -201,32 +215,40 @@ static boolean determineFailFastOnPropertyViolation(AbstractConfigurationImpl<?>
...
-			}
-			catch (Exception e) {
-				throw LOG.getUnableToInstantiateScriptEvaluatorFactoryClassException( scriptEvaluatorFactoryFqcn, e );
+		if ( propertyKey != null ) {
+			String prop = properties.get( propertyKey );
+			if ( prop != null ) {
+				return beanResolver.resolve( beanType, prop, BeanRetrieval.CONSTRUCTOR ).get();
 			}
 		}
+		return beanResolver.resolve( beanType, defaultName, BeanRetrieval.BUILTIN ).get();
```
marko-bekhta, line 243: but then for "factory-level components" we basically don't allow CDI beans, it's either constructor or a built-in one.

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/PredefinedScopeConstraintValidatorManagerImpl.java
```diff
@@ -21,6 +24,8 @@
  */
 public class PredefinedScopeConstraintValidatorManagerImpl extends AbstractConstraintValidatorManagerImpl {
 
+	private final Set<HibernateConstraintValidator<?, ?>> createdValidators = ConcurrentHashMap.newKeySet();
```
marko-bekhta, line 27: a bit unfortunate ... but we need to track the validators somehow so we can "close" them and they can release any bean holders they've obtained...

marko-bekhta (commented):
thanks 🙂 🙏🏻  added a few inline comments to show what I meant about factory-level and constraint-validator beans

cdi/src/main/java/org/hibernate/validator/cdi/internal/CdiBeanProvider.java
yrodiere: This is... odd. I'm pretty sure we have tests with application-scoped beans in Hibernate Search, and releasing these beans does not throw? So there must be a solution for that in Hibernate ORM already

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/password/PasswordStrengthValidatorForCharSequence.java
yrodiere: Interesting, I wonder where you got this idea of a bean holder!

More seriously, long term I wonder if we want to:

1. Introduce separate APIs for bean access in each project (though sharing some concepts)
2. Introduce a common API for bean access that each project depends on
3. Just... use a subset of CDI? Possibly with adapters for other DI systems.

E.g. we might need `BeanHolder` and such as an "agnostic" representation of things in our own code, but a user implementing a validator would likely want to use a field like `@Inject Estimator` or `@Inject Instance<Estimator>` or `@Autowired Estimator` instead.

There's a secondary question hiding in there: shouldn't the validator itself be retrieved as a CDI bean, so that it can be injected?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/PredefinedScopeConstraintValidatorManagerImpl.java
yrodiere: I'm surprised you didn't store the validators somewhere already... did you just create them every time you needed them?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/password/PasswordStrengthValidatorForCharSequence.java
marko-bekhta: > Interesting, I wonder where you got this idea of a bean holder!

I don't remember if I mentioned it already, but it was highly inspired by this other project I've been working on 😉🙂 

> a user implementing a validator would likely want to use a field like @Inject Estimator or @Inject Instance<Estimator> or @Autowired Estimator instead.

that ... they can actually do today: https://docs.hibernate.org/stable/validator/reference/en-US/html_single/#example-constraint-validator-injected-bean

🙈 🙂 but it does require CDI 🙂 so the way I was thinking of this was -- we have the bean provider that can work with other sources than DI ones and we use that to implement our own validators so that way user can configure and use them in any env, with or without DI. When a user implements their own validator .. they either do the same or just use injection directly ?

engine/src/main/java/org/hibernate/validator/internal/engine/constraintvalidation/PredefinedScopeConstraintValidatorManagerImpl.java
marko-bekhta: yeah it's a predefined scope, so constraints just go directly into constraint tree nodes and live there for the factory lifetime. with non-predefined factory there already was a map as things are more lazy there ..

mkurz:
any progress? thanks!

PR #2033, opened 2026-07-03
HV-2226 DateTimeFormat Constraint

SeanOkafor:
https://hibernate.atlassian.net/browse/HV-2226

This PR introduces a new @DateFormat constraint that validates date strings against a specified format pattern.                                                                                                                                                                                                                                                                                                                                                          
Features:                                                                                                                                                                                                                             

- Validates date strings using SimpleDateFormat patterns (e.g., "yyyy-MM-dd", "dd/MM/yyyy")                                                                                                                                

- Supports both strict and lenient parsing modes via lenient parameter                                                                                                                                                               

- Strict mode (default): Rejects invalid dates like "2026-02-30"                                                                                                                                                                     

- lenient mode: Accepts invalid dates by rolling them over (e.g., Feb 30 → Mar 2) 


                                                                                                                                                   

Follows Bean Validation spec: null and empty strings are handled appropriately
----------------------

----------------------

### DIFF
--- a/annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/ConstraintHelper.java
+++ b/annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/ConstraintHelper.java
@@ -289,6 +289,7 @@ public class ConstraintHelper {
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.BITCOIN_ADDRESS, CharSequence.class );
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.CODE_POINT_LENGTH, CharSequence.class );
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.CURRENCY, JavaMoneyTypes.MONETARY_AMOUNT );
+		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.DATE_TIME_FORMAT, CharSequence.class );
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.DURATION_MAX, Duration.class );
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.DURATION_MIN, Duration.class );
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.EMAIL, CharSequence.class );
--- a/annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/TypeNames.java
+++ b/annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/TypeNames.java
@@ -68,6 +68,7 @@ public class TypeNames {
 
 		public static final String CODE_POINT_LENGTH = ORG_HIBERNATE_VALIDATOR_CONSTRAINTS + ".CodePointLength";
 		public static final String CURRENCY = ORG_HIBERNATE_VALIDATOR_CONSTRAINTS + ".Currency";
+		public static final String DATE_TIME_FORMAT = ORG_HIBERNATE_VALIDATOR_CONSTRAINTS + ".DateTimeFormat";
 		public static final String EMAIL = ORG_HIBERNATE_VALIDATOR_CONSTRAINTS + ".Email";
 		public static final String IP_ADDRESS = ORG_HIBERNATE_VALIDATOR_CONSTRAINTS + ".IpAddress";
 		public static final String ISBN = ORG_HIBERNATE_VALIDATOR_CONSTRAINTS + ".ISBN";
--- a/documentation/src/main/asciidoc/reference/_ch02.adoc
+++ b/documentation/src/main/asciidoc/reference/_ch02.adoc
@@ -694,6 +694,10 @@ The default is `ANY`, which means both IPv4 and IPv6 addresses are considered va
 	Supported data types::: `CharSequence`
 	Hibernate metadata impact::: None
 
+`@DateTimeFormat`:: Checks that the annotated character sequence is valid for the date and/or time format. The `pattern` parameter specifies the date/time format pattern (e.g., "yyyy-MM-dd", "dd/MM/yyyy HH:mm:ss") following `java.time.format.DateTimeFormatter` conventions. The `locale` parameter accepts one or more locale strings (e.g., `"en-US"`, `{"en-US", "fr-FR"}`), validation succeeds if the value parses with any of them; defaults to `"ROOT"` which works for numeric patterns but not text-based ones like month names. The `lenient` parameter determines date validation behaviour: when `false` (default), invalid dates like February 31st are rejected; when `true`, invalid dates are accepted and rolled over to subsequent months (e.g., February 30th becomes March 2nd). `null` values are always considered valid.
+	Supported data types::: `CharSequence`
+	Hibernate metadata impact::: None
+
 `@Length(min=, max=)`:: Validates that the annotated character sequence is between `min` and `max` included
 	Supported data types::: `CharSequence`
 	Hibernate metadata impact::: Column length will be set to max
--- a/engine/src/main/java/org/hibernate/validator/cfg/defs/DateTimeFormatDef.java
+++ b/engine/src/main/java/org/hibernate/validator/cfg/defs/DateTimeFormatDef.java
@@ -1,0 +1,48 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.validator.cfg.defs;
+
+
+import java.util.Locale;
+
+import org.hibernate.validator.cfg.ConstraintDef;
+import org.hibernate.validator.constraints.DateTimeFormat;
+
+/**
+ * An {@link org.hibernate.validator.constraints.DateTimeFormat} constraint definition.
+ * @author Sean Okafor
+ * @since 9.2
+ */
+public class DateTimeFormatDef extends ConstraintDef<DateTimeFormatDef, DateTimeFormat> {
+
+	public DateTimeFormatDef() {
+		super( DateTimeFormat.class );
+	}
+
+	public DateTimeFormatDef pattern(String pattern) {
+		addParameter( "pattern", pattern );
+		return this;
+	}
+
+	public DateTimeFormatDef locale(String... locale) {
+		addParameter( "locale", locale );
+		return this;
+	}
+
+	public DateTimeFormatDef locale(Locale... locale) {
+		final String[] locales = new String[locale.length];
+		for ( int i = 0; i < locales.length; i++ ) {
+			locales[i] = locale[i].toLanguageTag();
+		}
+		addParameter( "locale", locales );
+		return this;
+	}
+
+	public DateTimeFormatDef lenient(boolean lenient) {
+		addParameter( "lenient", lenient );
+		return this;
+	}
+
+}
--- a/engine/src/main/java/org/hibernate/validator/constraints/DateTimeFormat.java
+++ b/engine/src/main/java/org/hibernate/validator/constraints/DateTimeFormat.java
@@ -1,0 +1,92 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright Red Hat Inc. and Hibernate Authors
+ */
+package org.hibernate.validator.constraints;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.CONSTRUCTOR;
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import jakarta.validation.Constraint;
+import jakarta.validation.Payload;
+
+import org.hibernate.validator.Incubating;
+
+/**
+ * Validates that the annotated character sequence represents a valid date and/or time
+ * according to the specified {@link #pattern() format}.
+ * <p>
+ * The format pattern follows the same syntax as {@link java.time.format.DateTimeFormatter}.
+ * For example, {@code "yyyy-MM-dd"} for dates like {@code "2026-07-06"},
+ * or {@code "yyyy-MM-dd'T'HH:mm:ss"} for date-times like {@code "2026-07-06T11:00:00"}.
+ * <p>
+ * The supported type is {@code CharSequence}. {@code null} is considered valid.
+ *
+ * @author Sean Okafor
+ * @since 9.2
+ *
+ * @see java.time.format.DateTimeFormatter
+ */
+@Documented
+@Constraint(validatedBy = { })
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
+@Retention(RUNTIME)
+@Repeatable(DateTimeFormat.List.class)
+@Incubating
+public @interface DateTimeFormat {
+
+	String message() default "{org.hibernate.validator.constraints.DateTimeFormat.message}";
+
+	Class<?>[] groups() default { };
+
+	Class<? extends Payload>[] payload() default { };
+
+	/**
+	 * Must be a valid pattern for {@link java.time.format.DateTimeFormatter}.
+	 */
+	String pattern();
+
+	/**
+	 * @return the locale(s) to use for parsing text-based date/time elements (e.g., month names, day names).
+	 * When multiple locales are specified, validation succeeds if parsing succeeds with any of them (OR logic).
+	 * Defaults to {@code "ROOT"}, which works for numeric patterns but not for text-based patterns
+	 * like {@code "dd MMMM yyyy"}.
+	 * <p>
+	 * Examples:
+	 * <ul>
+	 *   <li>{@code locale = "en-US"} - Accept only English month/day names</li>
+	 *   <li>{@code locale = {"en-US", "fr-FR", "de-DE"}} - Accept English, French, or German</li>
+	 * </ul>
+	 */
+	String[] locale() default { "ROOT" };
+
+	/**
+	 * @return whether to be lenient when parsing dates.
+	 * When false (default), invalid dates like February 31st are rejected
+	 * using {@link java.time.format.ResolverStyle#STRICT}.
+	 * When true, invalid dates are accepted and rolled over to subsequent months
+	 * (e.g., February 30th becomes March 2nd) using {@link java.time.format.ResolverStyle#LENIENT}.
+	 */
+	boolean lenient() default false;
+
+	/**
+	 * Defines several {@code @DateTimeFormat} annotations on the same element.
+	 */
+	@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
+	@Retention(RUNTIME)
+	@Documented
+	@interface List {
+		DateTimeFormat[] value();
+	}
+
+}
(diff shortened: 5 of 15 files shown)


annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/ConstraintHelper.java
```diff
@@ -287,6 +287,7 @@ public ConstraintHelper(Types typeUtils, AnnotationApiHelper annotationApiHelper
 
 		//register HV-specific constraints
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.BITCOIN_ADDRESS, CharSequence.class );
+		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.DATE_TIME_FORMAT, CharSequence.class );
```
marko-bekhta: let's move it after 
```
		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.CURRENCY, JavaMoneyTypes.MONETARY_AMOUNT );
```

to keep it alphabetical ?

annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/TypeNames.java
```diff
@@ -66,6 +66,8 @@ public static class HibernateValidatorTypes {
 
 		public static final String BITCOIN_ADDRESS = ORG_HIBERNATE_VALIDATOR_CONSTRAINTS + ".BitcoinAddress";
 
+		public static final String DATE_TIME_FORMAT = ORG_HIBERNATE_VALIDATOR_CONSTRAINTS + ".DateTimeFormat";
```
marko-bekhta: same here about ordering (just to keep things consistent)

documentation/src/main/asciidoc/reference/_ch02.adoc
```diff
@@ -694,6 +694,10 @@ The default is `ANY`, which means both IPv4 and IPv6 addresses are considered va
 	Supported data types::: `CharSequence`
 	Hibernate metadata impact::: None
 
+`@DateTimeFormat`:: Checks that the annotated character sequence is valid for the date and/or time format. The `pattern` parameter specifies the date/time format pattern (e.g., "yyyy-MM-dd", "dd/MM/yyyy HH:mm:ss") following `java.time.format.DateTimeFormatter` conventions. The `lenient` parameter determines date validation behavior: when `false` (default), invalid dates like February 31st are adjusted to the nearest valid date (February 28th or 29th); when `true`, invalid dates are accepted and rolled over to subsequent months (e.g., February 30th becomes March 2nd). See https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html for complete formatting rules.
```
marko-bekhta: `lenient` -- based on this doc, we can have invalid dates, they would just either "truncate" or "roll over" ?
that doesn't seem ok, does it ? I'd expect that we either have a strict/lenient attribute that would either disallow or allow the "wrong" dates, whether it's truncating or rolling over doesn't make a difference, does it (in terms of validation)?

engine/src/main/java/org/hibernate/validator/constraints/DateTimeFormat.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Checks that the date and/or time is formatted correctly according to the specified pattern.
+ * When {@code lenient=false} (default), invalid dates are adjusted to the nearest valid date
+ * (e.g., February 31st becomes February 28th/29th). When {@code lenient=true}, invalid dates
+ * are rolled over to subsequent months (e.g., February 30th becomes March 2nd).
+ * Pattern examples: "yyyy-MM-dd", "dd/MM/yyyy HH:mm:ss", "uuuu-MM-dd'T'HH:mm:ss"
+ * See <a href="https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html">DateTimeFormatter</a>
+ * for complete formatting rules.
+ * @author Sean Okafor
+ * @since 9.1
```
marko-bekhta:
```suggestion
 * @since 9.2
```

9.1 is already final, so this new constraint would go into 9.2 🙂

engine/src/main/java/org/hibernate/validator/constraints/DateTimeFormat.java
```diff
@@ -0,0 +1,75 @@
...
+@Incubating
+public @interface DateTimeFormat {
+
+	String message() default "{org.hibernate.validator.constraints.DateTimeFormat.message}";
+
+	Class<?>[] groups() default { };
+
+	Class<? extends Payload>[] payload() default { };
+
+	/**                                                                                                                                                                                                                                  │
```
marko-bekhta:
```suggestion
	/**
```

engine/src/main/java/org/hibernate/validator/constraints/DateTimeFormat.java
```diff
@@ -0,0 +1,75 @@
...
+
+import org.hibernate.validator.Incubating;
+
+/**
+ * Checks that the date and/or time is formatted correctly according to the specified pattern.
+ * When {@code lenient=false} (default), invalid dates are adjusted to the nearest valid date
+ * (e.g., February 31st becomes February 28th/29th). When {@code lenient=true}, invalid dates
+ * are rolled over to subsequent months (e.g., February 30th becomes March 2nd).
+ * Pattern examples: "yyyy-MM-dd", "dd/MM/yyyy HH:mm:ss", "uuuu-MM-dd'T'HH:mm:ss"
+ * See <a href="https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html">DateTimeFormatter</a>
```
marko-bekhta:
```suggestion
 * @see DateTimeFormatter
```
This would render a link without needing to have a dedicated url for a particular JDK

engine/src/main/java/org/hibernate/validator/constraints/DateTimeFormat.java
```diff
@@ -0,0 +1,75 @@
...
+ * for complete formatting rules.
```
marko-bekhta: Let's try to make it more consistent with the way other constraints are defined ? It usually contains 
- what the constraint verifies
- what types does it accept
- null is valid (even it is somewhat implicit, but best to note that down)

I'd go for something like

```java
  /**
   * Validates that the annotated character sequence represents a valid date and/or time
   * according to the specified {@link #pattern() format}.
   * <p>
   * The format pattern follows the same syntax as {@link java.time.format.DateTimeFormatter}.
   * For example, {@code "yyyy-MM-dd"} for dates like {@code "2026-07-06"},
   * or {@code "yyyy-MM-dd'T'HH:mm:ss"} for date-times like {@code "2026-07-06T11:00:00"}.
   * <p>
   * The supported type is {@code CharSequence}. {@code null} is considered valid.
   *
   * @author Sean Okafor
   * @since 9.2
   *
   * @see java.time.format.DateTimeFormatter
   */
```

engine/src/main/java/org/hibernate/validator/constraints/DateTimeFormat.java
```diff
@@ -0,0 +1,75 @@
...
+	String pattern();
+
+	/**
+	 * @return whether to be lenient when parsing dates.
+	 * When false (default), invalid dates like February 31st are adjusted to the nearest valid date
+	 * (e.g., February 28th or 29th for leap years) using {@link java.time.format.ResolverStyle#SMART}.
+	 * When true, invalid dates are accepted and rolled over to subsequent months
+	 * (e.g., February 30th becomes March 2nd) using {@link java.time.format.ResolverStyle#LENIENT}.
+	 */
+	boolean lenient() default false;
```
marko-bekhta, line 80: since it's a constraint, I think we should either accept or fail "invalid" dates instead of configuring how they roll over ?
I'd say `lenient == false` and `!samePattern.format(parsedDate).equals(initialValue)` -- fail, `lenient == true` -- just parse and if no errors return as true ?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/DateTimeFormatValidator.java
```diff
@@ -0,0 +1,53 @@
...
+public class DateTimeFormatValidator implements ConstraintValidator<DateTimeFormat, CharSequence> {
+
+	private String pattern;
+	private boolean lenient;
+
+	@Override
+	public void initialize(DateTimeFormat constraintAnnotation) {
+		this.pattern = constraintAnnotation.pattern();
+		this.lenient = constraintAnnotation.lenient();
+	}
```
marko-bekhta: now that we switched to `DateTimeFormatter` we can actually create it just once here ?


```suggestion
	private DateTimeFormatter formatter;

	@Override
	public void initialize(DateTimeFormat constraintAnnotation) {
		this.formatter = DateTimeFormatter.ofPattern( constraintAnnotation.pattern(), Locale.ROOT )
					.withResolverStyle( constraintAnnotation.lenient() ? ResolverStyle.LENIENT : ResolverStyle.SMART );
	}
```

wdyt ?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/DateTimeFormatValidator.java
```diff
@@ -0,0 +1,53 @@
...
+
+	@Override
+	public boolean isValid(CharSequence value, ConstraintValidatorContext context) {
+		if ( value == null ) {
+			return true;
+		}
+		if ( value.isEmpty() ) {
+			return false;
+		}
```
marko-bekhta, line 72: if we pass an empty string to the formatter -- what happens ?

if we decide to keep the check -- we should mention that in the javadoc, alongside the "null is considered valid".

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/DateTimeFormatValidator.java
```diff
@@ -0,0 +1,53 @@
...
+
+		try {
+			DateTimeFormatter formatter = DateTimeFormatter.ofPattern( pattern, Locale.ROOT )
```
marko-bekhta: I wonder about that `Locale.ROOT` here ... especially when you'd have patterns where months/days are represented as words ? that would be locale-specific, no ? Maybe the locale should go on the constraint as an attribute ? (and moreover, if we end up doing so -- we probably want to accept multiple locales, and then do an "OR" check?)

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
```diff
@@ -41,6 +41,7 @@ org.hibernate.validator.constraints.ScriptAssert.message            = script exp
 org.hibernate.validator.constraints.UniqueElements.message          = must only contain unique elements
 org.hibernate.validator.constraints.URL.message                     = must be a valid URL
 org.hibernate.validator.constraints.UUID.message                     = must be a valid UUID
+org.hibernate.validator.constraints.DateTimeFormat.message            = Invalid Formatting
```
marko-bekhta:
```suggestion
org.hibernate.validator.constraints.DateTimeFormat.message            =  must be a valid date/time representation matching the format '${pattern}'
```
let's try to make it more consistent with other messages ?

marko-bekhta (commented):
hey thanks @SeanOkafor 👍🏻 🙂 

good progress on this one! I've added some ideas in the comments inline. 
let's figure out those lenient and local attributes, otherwise looks good!

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/DateTimeFormatValidator.java
SeanOkafor: only reason i had the locale set to root as with SimpleDateFormat it was necessary to specify, it would set the locale to root device which was giving me errors, ill test to see if it is still necessary

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/DateTimeFormatValidator.java
SeanOkafor: empty strings in the datetimeformatter throws and exception, so my check makes sense i think

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
```diff
@@ -41,6 +41,7 @@ org.hibernate.validator.constraints.ScriptAssert.message            = script exp
 org.hibernate.validator.constraints.UniqueElements.message          = must only contain unique elements
 org.hibernate.validator.constraints.URL.message                     = must be a valid URL
 org.hibernate.validator.constraints.UUID.message                     = must be a valid UUID
+org.hibernate.validator.constraints.DateTimeFormat.message            = must be a valid date/time representation matching the format '${pattern}'
```
marko-bekhta:
```suggestion
org.hibernate.validator.constraints.DateTimeFormat.message            = must be a valid date/time representation matching the format '{pattern}'
```

ah, so for simple arguments we can probably do without relying on EL interpolation.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/DateTimeFormatValidator.java
```diff
@@ -0,0 +1,78 @@
...
+		String[] parts = localeString.split( "_", 3 );
+
+		Locale.Builder builder = new Locale.Builder();
+		return switch ( parts.length ) {
+			case 1 -> builder.setLanguage( parts[0] ).build();
+			case 2 -> builder.setLanguage( parts[0] ).setRegion( parts[1] ).build();
+			case 3 -> builder.setLanguage( parts[0] ).setRegion( parts[1] ).setVariant( parts[2] ).build();
+			default -> throw new IllegalArgumentException( "Invalid locale format: " + localeString );
+		};
+	}
```
marko-bekhta: We've been using `Locale.forLanguageTag` in other places; maybe let's stick to that?
(we can specify in which format the locale string should be on the locale attribute of the constraint)

marko-bekhta (commented):
We are getting closer 🙂👍🏻 thanks !

annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/TypeNames.java
```diff
@@ -66,8 +66,10 @@ public static class HibernateValidatorTypes {
 
 		public static final String BITCOIN_ADDRESS = ORG_HIBERNATE_VALIDATOR_CONSTRAINTS + ".BitcoinAddress";
 
+
```
marko-bekhta:
```suggestion
```

engine/src/main/java/org/hibernate/validator/cfg/defs/DateTimeFormatDef.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * An {@link org.hibernate.validator.constraints.DateTimeFormat} constraint definition.
+ *
+ * @author Sean Okafor
+ * @since 9.2
+ */
+
+
+public class DateTimeFormatDef extends ConstraintDef<DateTimeFormatDef, DateTimeFormat> {
```
marko-bekhta, line 18:
```suggestion
 */
public class DateTimeFormatDef extends ConstraintDef<DateTimeFormatDef, DateTimeFormat> {
```

engine/src/main/java/org/hibernate/validator/cfg/defs/DateTimeFormatDef.java
```diff
@@ -0,0 +1,40 @@
...
+	public DateTimeFormatDef pattern(String pattern) {
+		addParameter( "pattern", pattern );
+		return this;
+	}
+
+	public DateTimeFormatDef locale(String... locale) {
+		addParameter( "locale", locale );
+		return this;
+	}
+
```
marko-bekhta, line 33: let's have an overload for locales ?

```java
	public DateTimeFormatDef locale(Locale... locale) {
		final String[] locales = new String[locale.length];
		for ( int i = 0; i < locales.length; i++ ) {
			locales[i] = locale[i].toLanguageTag();
		}
		addParameter( "locale", locales );
		return this;
	}
```

engine/src/main/java/org/hibernate/validator/internal/metadata/core/ConstraintHelper.java
```diff
@@ -331,44 +333,7 @@
...
-import org.hibernate.validator.internal.constraintvalidators.hv.PortValidatorForFloat;
-import org.hibernate.validator.internal.constraintvalidators.hv.PortValidatorForInteger;
-import org.hibernate.validator.internal.constraintvalidators.hv.PortValidatorForLong;
-import org.hibernate.validator.internal.constraintvalidators.hv.PortValidatorForNumber;
-import org.hibernate.validator.internal.constraintvalidators.hv.PortValidatorForShort;
-import org.hibernate.validator.internal.constraintvalidators.hv.ScriptAssertValidator;
-import org.hibernate.validator.internal.constraintvalidators.hv.URLValidator;
-import org.hibernate.validator.internal.constraintvalidators.hv.UUIDValidator;
-import org.hibernate.validator.internal.constraintvalidators.hv.UniqueElementsValidator;
+import org.hibernate.validator.internal.constraintvalidators.hv.*;
```
marko-bekhta: we've talked about this one elsewhere 🙂

engine/src/main/java/org/hibernate/validator/constraints/DateTimeFormat.java
```diff
@@ -0,0 +1,93 @@
...
+	/**
+	 * @return the locale(s) to use for parsing text-based date/time elements (e.g., month names, day names).
+	 * When multiple locales are specified, validation succeeds if parsing succeeds with any of them (OR logic).
+	 * Defaults to {@code "ROOT"}, which works for numeric patterns but not for text-based patterns
+	 * like {@code "dd MMMM yyyy"}.
+	 * <p>
+	 * Examples:
+	 * <ul>
+	 *   <li>{@code locale = "en_US"} - Accept only English month/day names</li>
+	 *   <li>{@code locale = {"en_US", "fr_FR", "de_DE"}} - Accept English, French, or German</li>
```
marko-bekhta:
```suggestion
	 *   <li>{@code locale = "en-US"} - Accept only English month/day names</li>
	 *   <li>{@code locale = {"en-US", "fr-FR", "de-DE"}} - Accept English, French, or German</li>
```

let's use the format that language tags are working with out of the box in our examples ?

marko-bekhta (commented):
thanks 🙂

annotation-processor/src/main/java/org/hibernate/validator/ap/internal/util/ConstraintHelper.java
```diff
@@ -289,6 +289,7 @@ public ConstraintHelper(Types typeUtils, AnnotationApiHelper annotationApiHelper
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.BITCOIN_ADDRESS, CharSequence.class );
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.CODE_POINT_LENGTH, CharSequence.class );
 		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.CURRENCY, JavaMoneyTypes.MONETARY_AMOUNT );
+		registerAllowedTypesForBuiltInConstraint( HibernateValidatorTypes.DATE_TIME_FORMAT, CharSequence.class );
```
marko-bekhta, line 292: oh, seems this class got quite re-formatted 🫣 
Let's revert those changes and just keep the ^ 🙂 ( I think we just need this one line added here, right?)

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/DateTimeFormatValidator.java
```diff
@@ -0,0 +1,145 @@
...
+		}
+
+		public DateTimeFormatter[] of(String pattern, boolean lenient, String[] localeStrings) {
+			Locale[] locales = parseLocales( localeStrings );
+
+			return cache.computeIfAbsent(
+					new FormatterKey( pattern, lenient, locales ), // The cache key
+					key -> buildFormatters( key.pattern(), key.lenient(), key.locales() ) // How to create it if missing
+			);
+		}
```
marko-bekhta: since that experiment with `DateTimeFormatterBuilder` and optional formatters did not work out... maybe let's still cache single formatters ?

```suggestion
		public DateTimeFormatter of(String pattern, boolean lenient, Locale locale) {
			return cache.computeIfAbsent(
					new FormatterKey( pattern, lenient, locale ), // The cache key
					key -> DateTimeFormatter.ofPattern(...)... // How to create it if missing
			);
		}
```
that way you can keep the cache key as a simple record without a need to override hash/equals + if there would be constraints with the same pattern but maybe a different set of locales you'd get more "cache hits"

marko-bekhta (commented):
hey 🙂 

thanks for experimenting with the format builder! we are getting closer 🤞🏻 🙂

marko-bekhta (approved):
thanks 🙂 !

PR #2034, opened 2026-07-05
HV-2230 Add Argentine and Paraguayan tax ID constraints

ggomarighetti:
### Resume

This PR adds country-specific Hibernate Validator constraints for validating Argentine and Paraguayan tax identifiers. It introduces `@CUIL` and `@CUIT` for Argentina, and `@RUC` for Paraguay, including their validator implementations, built-in constraint registration, programmatic constraint definitions, default validation messages, reference guide documentation, and focused test coverage. The Paraguayan RUC validation follows the official DNIT modulo 11 routine, including support for alphanumeric base values.

### Evidence

> No response

### Reference

https://hibernate.atlassian.net/browse/HV-2230

----------------------

----------------------


engine/src/test/java/org/hibernate/validator/test/constraints/annotations/hv/ar/CUITValidatorTest.java
```diff
@@ -0,0 +1,69 @@
...
+		assertNoViolations( validator.validate( new Person( "20-12345678-6" ) ) );
+		assertNoViolations( validator.validate( new Person( "23-12345678-5" ) ) );
+		assertNoViolations( validator.validate( new Person( "24-12345678-1" ) ) );
+		assertNoViolations( validator.validate( new Person( "27-12345678-0" ) ) );
+		assertNoViolations( validator.validate( new Person( "30-12345678-1" ) ) );
+		assertNoViolations( validator.validate( new Person( "33-12345678-0" ) ) );
+		assertNoViolations( validator.validate( new Person( "34-12345678-7" ) ) );
+	}
+
+	@Test
```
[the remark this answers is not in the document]
ggomarighetti: Comment noted. Changes promoted.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ar/AbstractArgentineTaxIdValidator.java
```diff
@@ -0,0 +1,68 @@
...
+		for ( int i = 0; i < WEIGHTS.length; i++ ) {
+			sum += Character.digit( taxId.charAt( i ), 10 ) * WEIGHTS[i];
+		}
+
+		int checkDigit = 11 - ( sum % 11 );
+		if ( checkDigit == 11 ) {
+			checkDigit = 0;
+		}
+		else if ( checkDigit == 10 ) {
+			return false;
```
martin2844, line 52: Small thought here: when the computed check digit comes out to `10`, AFIP's actual rule isn't to reject the number outright — it reassigns the prefix to `23` and fixes the check digit to `9` (originally male/`20`) or `4` (originally female/`27`) for exactly this case. So a legitimately-issued number like `23-00000040-9` (base sum → remainder 1 → raw digit 10) would currently get rejected here.

Reference for the substitution rule: https://es.wikipedia.org/wiki/Clave_%C3%9Anica_de_Identificaci%C3%B3n_Tributaria, and an independent implementation encoding the same 9/4 substitution: https://github.com/brielov/cuit/blob/master/src/index.ts

Could be worth accepting the check digit here when it's `10` and the prefix is `23`/`24` with a trailing `9` or `4` — no pressure either way, just flagging in case it wasn't considered. Might also be a good spot for a test case, since none of the current CUIL/CUIT tests exercise this remainder=1 path.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ar/CUILValidator.java
```diff
@@ -0,0 +1,20 @@
...
+package org.hibernate.validator.internal.constraintvalidators.hv.ar;
+
+import org.hibernate.validator.constraints.ar.CUIL;
+
+/**
+ * Validator for {@link CUIL}.
+ */
+public class CUILValidator extends AbstractArgentineTaxIdValidator<CUIL> {
+
+	private static final String[] PREFIXES = { "20", "23", "24", "27" };
```
martin2844: Minor note: some sources (e.g. https://es.wikipedia.org/wiki/Clave_%C3%9Anica_de_Identificaci%C3%B3n_Tributaria) list `25` and `26` alongside `23`/`24` as additional contingency prefixes for individuals (used once `23`/`24` are also exhausted by DNI collisions), though other references only mention `20/23/24/27` like this list does — sources aren't fully consistent here, so just flagging in case it's worth double-checking against an authoritative AFIP source. Same prefix set is used in `CUITValidator`, so it'd apply there too if it turns out to be missing anything.

martin2844 (commented):
Nice work adding these — the CUIL/CUIT/RUC wiring (BuiltinConstraint, ConstraintHelper, message properties, test suites, docs) all looks solid and consistent with the existing constraints. I hand-traced the RUC checksum against the official DNIT reference implementation and it matches. Two small things worth a look before merge, more "food for thought" than blockers — happy to be told I'm wrong on either.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ar/CUILValidator.java
ggomarighetti: Comment noted, fix promoted.

formal docs ref: 
- https://www.afip.gob.ar/InformacionFinancieradeSujetosnoResidentes/documentos/Estructuras-TIN-por-Pais.pdf
- https://www.oecd.org/content/dam/oecd/en/topics/policy-issue-focus/aeoi/tin_argentina.pdf

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ar/AbstractArgentineTaxIdValidator.java
ggomarighetti: Point reviewed, no change pursued.

I rechecked the special `10` case, and the linked implementation appears to apply that rule while generating a CUIT from DNI/type, not while validating an already assigned CUIT. The final reassigned value, e.g. the `23...9` / `23...4` variant, still passes the regular modulo-11 validation. Since this constraint validates final identifiers rather than generating them, I’d keep rejecting final values whose own check digit calculation results in `10` unless we find an official ARCA/AFIP validation rule saying otherwise.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ar/AbstractArgentineTaxIdValidator.java
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		return checkDigit == Character.digit( taxId.charAt( taxId.length() - 1 ), 10 );
+	}
+
+	protected abstract String[] getValidPrefixes();
```
marko-bekhta:
```suggestion
	protected abstract Set<String> getValidPrefixes();
```
no need to convert it each time, right ?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/py/RUCValidator.java
```diff
@@ -0,0 +1,84 @@
...
+
+import org.hibernate.validator.constraints.py.RUC;
+
+/**
+ * Validator for {@link RUC}.
+ */
+public class RUCValidator implements ConstraintValidator<RUC, CharSequence> {
+
+	private static final Pattern UNFORMATTED_PATTERN = Pattern.compile( "[0-9A-Za-z]+[0-9]" );
+	private static final Pattern FORMATTED_PATTERN = Pattern.compile( "[0-9A-Za-z]+-[0-9]" );
```
marko-bekhta: I think these shouldn't be unbound patterns? (need some length limits there ...)

engine/src/main/java/org/hibernate/validator/cfg/defs/ar/CUILDef.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+package org.hibernate.validator.cfg.defs.ar;
+
+import org.hibernate.validator.cfg.ConstraintDef;
+import org.hibernate.validator.constraints.ar.CUIL;
+
+/**
+ * A {@link CUIL} constraint definition.
+ */
+public class CUILDef extends ConstraintDef<CUILDef, CUIL> {
```
marko-bekhta, line 17: Let's have @Incubating on the new constraints (on public API classes)

engine/src/main/java/org/hibernate/validator/constraints/ar/CUIL.java
```diff
@@ -0,0 +1,50 @@
...
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import jakarta.validation.Constraint;
+import jakarta.validation.Payload;
+import jakarta.validation.ReportAsSingleViolation;
+
+/**
+ * Validates that a {@code CharSequence} is an Argentine CUIL (Unique Labor Identification Code).
```
marko-bekhta, line 27:
```suggestion
 * Validates that a {@code CharSequence} is an Argentine CUIL (Unique Labor Identification Code).
 * @since 9.2
```

let's have a since tag (and on other public types)

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/py/RUCValidator.java
```diff
@@ -0,0 +1,84 @@
...
+			char character = upperCaseNumber.charAt( i );
+			if ( Character.isDigit( character ) ) {
+				normalizedNumber.append( character );
+			}
+			else {
+				normalizedNumber.append( (int) character );
+			}
+		}
+		return normalizedNumber.toString();
+	}
```
marko-bekhta: can we drop the method and instead do the convertion in the checkdigit calculation ? Something along the lines of 
```
if digit -- just proceed as is
if char -- cast to int (would be a 2-digit number) -- just %10 /10 to get the digits and process them that way ?
```

marko-bekhta (commented):
hey thanks for the patch 🙂 👋🏻 

could you also add an update to the annotation processor (https://in.relation.to/2018/01/04/adding-new-constraint-to-engine/#step-3-add-annotation-processor-support)

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ar/AbstractArgentineTaxIdValidator.java
ggomarighetti: Comment noted. Fix promoted.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/py/RUCValidator.java
ggomarighetti: Comment noted. Fix promoted.

engine/src/main/java/org/hibernate/validator/cfg/defs/ar/CUILDef.java
ggomarighetti: Comment noted. Fix promoted.

engine/src/main/java/org/hibernate/validator/constraints/ar/CUIL.java
ggomarighetti: Comment noted. Fix promoted.

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/py/RUCValidator.java
ggomarighetti: Comment noted. Fix promoted.

ggomarighetti:
Hey @marko-bekhta, thanks for the very quick review! I pushed the updates addressing your comments, including the annotation processor support and the requested adjustments around the new constraints and RUC validation. I’ll wait for your feedback.

marko-bekhta (approved):
thanks! looks nice 🙂

ggomarighetti:
Hi @marko-bekhta ,

Based on Sonar’s quality gate, I saw that they were basic errors and easy to fix, along with a testing branch that was also fixed. I pushed the commits to bring it up to 100%.

marko-bekhta (approved):
thanks 🙂 !

PR #2052, opened 2026-07-13
HV-2236 Added Contains Constraint.

SeanOkafor:
https://hibernate.atlassian.net/browse/HV-2236

Adds a new @Contains constraint that validates a CharSequence contains specified substrings. Supports multiple values with minRequired to control AND/OR semantics (defaults to all must match), and an optional ignoreCase flag for case-insensitive comparison.

----------------------

----------------------


engine/src/main/java/org/hibernate/validator/constraints/Contains.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @author Sean Okafor
+ * @since 9.2
+ */
+@Documented
+@Constraint(validatedBy = { })
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
+@Retention(RUNTIME)
+@Repeatable(List.class)
+public @interface Contains {
```
marko-bekhta, line 44:
```suggestion
@Incubating
public @interface Contains {
```

let's add new constraints as incubating, especially if we are adding some attributes to the annotation

engine/src/main/java/org/hibernate/validator/cfg/defs/ContainsDef.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.hibernate.validator.cfg.defs;
+
+import org.hibernate.validator.cfg.ConstraintDef;
+import org.hibernate.validator.constraints.Contains;
+
+/**
+ * A {@link Contains} constraint definition.
+ * @author Sean Okafor
```
marko-bekhta, line 14:
```suggestion
 * @author Sean Okafor
 *  @since 9.2
```

engine/src/main/java/org/hibernate/validator/constraints/Contains.java
```diff
@@ -0,0 +1,63 @@
...
+	String[] value();
+
+	int minRequired() default -1;
+
+	boolean ignoreCase() default false;
```
marko-bekhta, line 68: Let's have some javadoc explaining the constraint attributes ?

engine/src/main/java/org/hibernate/validator/constraints/Contains.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+@Documented
+@Constraint(validatedBy = { })
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
+@Retention(RUNTIME)
+@Repeatable(List.class)
+public @interface Contains {
+	String[] value();
+
+	int minRequired() default -1;
```
marko-bekhta: how about we add a 
```
int MATCH_ALL = -1;
```
somewhere after the  
```
public @interface List {
		Contains[] value();
	}
```

and then use it here as a default ?
```suggestion
	int minRequired() default MATCH_ALL;
```

engine/src/main/java/org/hibernate/validator/constraints/Contains.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.hibernate.validator.constraints.Contains.List;
+
+/**
+ * Validates that the annotated character sequence contains all (or a minimum number of) the specified substrings.
+ * <p>
+ * By default, all specified values must be present (AND semantics). Use {@code minRequired} to require only
+ * a minimum number of matches (e.g., {@code minRequired=1} for OR semantics).
+ * <p>
+ * When {@code ignoreCase} is set to {@code true}, the comparison is case-insensitive.
```
marko-bekhta, line 32:
```suggestion
 * When {@code ignoreCase} is set to {@code true}, the comparison is case-insensitive.
 * <p>
 * {@code null} values are considered valid.
```

let's be clear that `null`s are valid

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ContainsValidator.java
```diff
@@ -0,0 +1,63 @@
...
+		return matchCount >= minRequired;
+	}
+
+	private void validateParameters() {
+		if ( minRequired < 0 ) {
+			throw new IllegalArgumentException( "The minRequired parameter cannot be negative." );
+		}
+		if ( minRequired > values.length ) {
+			throw new IllegalArgumentException( "The minRequired parameter cannot exceed the number of values." );
+		}
```
marko-bekhta, line 75: we do not throw exceptions directly ... we should use the `throw LOG.someNewMethodOnThatLogInterface`

see for example 

https://github.com/hibernate/hibernate-validator/blob/3f8dc0abec5d81e62761003cacbb288a24ab7c59/engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/bv/size/SizeValidatorForCharSequence.java#L55-L65

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ContainsValidator.java
```diff
@@ -0,0 +1,63 @@
...
+	private int minRequired;
+	private boolean ignoreCase;
+
+	@Override
+	public void initialize(Contains parameters) {
+		this.values = parameters.value();
+		this.ignoreCase = parameters.ignoreCase();
+
+		int min = parameters.minRequired();
+		this.minRequired = ( min == -1 ) ? values.length : min;
```
marko-bekhta:
```suggestion
		this.minRequired = ( min == Contains.MATCH_ALL ) ? values.length : min;
```
that is if we'll go with that constant from the other comment

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/ContainsValidator.java
```diff
@@ -0,0 +1,63 @@
...
+
+	@Override
+	public boolean isValid(CharSequence value, ConstraintValidatorContext constraintValidatorContext) {
+		if ( value == null ) {
+			return true;
+		}
+		String str = ignoreCase ? value.toString().toLowerCase( Locale.ROOT ) : value.toString();
+		int matchCount = 0;
+		for ( String v : values ) {
+			String target = ignoreCase ? v.toLowerCase( Locale.ROOT ) : v;
```
marko-bekhta: let's do this in the `initialize` ? 
we already know there if we need to ignore the case or not so we can prepare the correct substrings ahead of time ?

engine/src/main/java/org/hibernate/validator/constraints/Contains.java
marko-bekhta, line 32: > the comparison is case-insensitive

I think you were telling me that `toLowerCase( Locale.ROOT )` won't necessarly work for all languages, so maybe let's be explicit here that if user asks for case-insensitve it's actually using `toLowerCase( Locale.ROOT )` ? wdyt ?

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
```diff
@@ -28,6 +28,7 @@ org.hibernate.validator.constraints.IpAddress.message               = invalid IP
 org.hibernate.validator.constraints.ISBN.message                    = invalid ISBN
 org.hibernate.validator.constraints.Length.message                  = length must be between {min} and {max}
 org.hibernate.validator.constraints.CodePointLength.message         = length must be between {min} and {max}
+org.hibernate.validator.constraints.Contains.message               = must contain {value}
```
marko-bekhta:
```suggestion
org.hibernate.validator.constraints.Contains.message                = must contain {value}
```

just so it's more aligned 🙂

engine/src/main/java/org/hibernate/validator/cfg/defs/ContainsDef.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class ContainsDef extends ConstraintDef<ContainsDef, Contains> {
```
marko-bekhta, line 18:
```suggestion
@Incubating
public class ContainsDef extends ConstraintDef<ContainsDef, Contains> {
```
oh and yeah let's mark this one incubating as well...

engine/src/main/resources/org/hibernate/validator/ValidationMessages.properties
marko-bekhta: btw did you check it correctly prints the array of strings?

marko-bekhta (commented):
Hey, thanks!
looks good overall, I've added a few suggestions, and I think we are missing a test extending the `AbstractConstrainedTest` ?

engine/src/main/java/org/hibernate/validator/constraints/Contains.java
SeanOkafor: what do you mean it already is explained no?

engine/src/main/java/org/hibernate/validator/cfg/defs/ContainsDef.java
marko-bekhta: looks like incubating is still missing ?

marko-bekhta (commented):
> I think we are missing a test extending the `AbstractConstrainedTest`

or have I missed it ?  🫣 

oh and if you send the update could you also adjust the commit message ? so that it's more about adding the constraint in general rather than just annotation ?

thanks!

engine/src/main/java/org/hibernate/validator/cfg/defs/CurrencyDef.java
```diff
@@ -12,6 +13,7 @@
  *
  * @author Gunnar Morling
  */
+@Incubating
```
marko-bekhta:
```suggestion
```

I think you've intended it to be on `ContainsDef` ? 🫣

marko-bekhta (approved):
thanks! 🙂

PR #2096, opened 2026-08-25
HV-2246 Adding @IBAN Constraint

dreab8:
https://hibernate.atlassian.net/browse/HV-2246

----------------------

----------------------


engine/src/main/java/org/hibernate/validator/cfg/defs/IBANDef.java
```diff
@@ -0,0 +1,21 @@
...
+import org.hibernate.validator.cfg.ConstraintDef;
+import org.hibernate.validator.constraints.IBAN;
+
+/**
+ * An {@link IBAN} constraint definition.
+ *
+ * @author Andrea Boriero
+ * @since 9.2
+ */
+public class IBANDef extends ConstraintDef<IBANDef, IBAN> {
```
marko-bekhta, line 18: We usually add `@Incubating` to new constraints (so that we have more room to adjust one if needed)

engine/src/main/java/org/hibernate/validator/constraints/IBAN.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @author Andrea Boriero
+ * @since 9.2
+ */
+@Documented
+@Constraint(validatedBy = { })
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER, TYPE_USE })
+@Retention(RUNTIME)
+@Repeatable(List.class)
+public @interface IBAN {
```
marko-bekhta, line 47: and same here about `@Incubating`

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/IBANValidator.java
```diff
@@ -0,0 +1,171 @@
...
+		int length = iban.length();
+		int mod = 0;
+		for ( int i = 0; i < length; i++ ) {
+			// Start with the BBAN (chars after the first four), then wrap around to the country code and check digits.
+			char c = iban.charAt( ( i + 4 ) % length );
+			if ( c >= '0' && c <= '9' ) {
+				mod = ( mod * 10 + ( c - '0' ) ) % MODULUS;
+			}
+			else {
+				mod = ( mod * 100 + ( Character.toUpperCase( c ) - 'A' + 10 ) ) % MODULUS;
```
marko-bekhta, line 93: I wonder if lowercase would even be valid ?

marko-bekhta (commented):
Thanks!
I've added a few comments inline

engine/src/main/java/org/hibernate/validator/constraints/IBAN.java
```diff
@@ -0,0 +1,63 @@
...
+
+	String message() default "{org.hibernate.validator.constraints.IBAN.message}";
+
+	Class<?>[] groups() default { };
+
+	Class<? extends Payload>[] payload() default { };
+
```
marko-bekhta, line 54: Let's also have an attribute to disallow lowercase characters by default, but let the user flip the flag to allow them.

marko-bekhta (approved):
Thanks!

PR #2103, opened 2026-09-03
HV-2249 Add @BIC constraint

dreab8:
https://hibernate.atlassian.net/browse/HV-2249

----------------------

----------------------


engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/BICValidator.java
```diff
@@ -0,0 +1,418 @@
...
+		codes.add( "VI" );
+		codes.add( "VN" );
+		codes.add( "VU" );
+		codes.add( "WF" );
+		codes.add( "WS" );
+		codes.add( "YE" );
+		codes.add( "YT" );
+		codes.add( "ZA" );
+		codes.add( "ZM" );
+		codes.add( "ZW" );
```
marko-bekhta: there's a `Locale.getISOCountries( Locale.IsoCountryCode.PART1_ALPHA2)`

let's maybe use that and append it with XK ?

engine/src/main/java/org/hibernate/validator/internal/constraintvalidators/hv/BICValidator.java
```diff
@@ -0,0 +1,418 @@
...
+	public void initialize(BIC parameters) {
+		this.allowLowercase = parameters.allowLowercase();
+
+		// Normalize country codes to uppercase for case-insensitive matching
+		String[] countryCodes = parameters.countryCodes();
+		if ( countryCodes.length > 0 ) {
+			this.allowedCountryCodes = new HashSet<>( countryCodes.length );
+			for ( String code : countryCodes ) {
+				this.allowedCountryCodes.add( code.toUpperCase( Locale.ROOT ) );
+			}
```
marko-bekhta, line 49: I guess we should check that it's actually a country code, and fail fast if it's not ? (I think we are throwing `IllegalArgumeException`s in such cases ...

